PR #77, opened 2016-02-23
Remove memoization, ~100x speed improvement.

olafurpg:
Instead of memoization we dequeue all states when we enter a new
statement *after* a newline. After spaces, we are likely inside a single
line block which has long-distance effects.

JMH Benchmarks
==============
Benchmark                  Mode  Cnt     Score   Error    Units
BaseLinker.scalafmt        avgt   10    65.884 ± 2.074    ms/op
Division.scalafmt          avgt   10   134.656 ± 3.048    ms/op
SourceMapWriter.scalafmt   avgt   10    31.851 ± 0.989    ms/op
OptimizerCore.scalafmt     avgt   10  1045.331 ± 236.486  ms/op
OptimizerCore.scalariform  avgt   10   230.654 ± 8.470    ms/op

NOTE. OptimizerCore was unformattable before.

Scala.js results
================
Formatter timed out: 10
org.scalafmt.Error$FormatterChangedAST: 1
java.lang.ClassCastException: 11
Success: 891
org.scalafmt.Error$CantFormatFile: 7
Total: 920
+------------+--------+--------------+---------+---------+---------+----------+
|         Max|     Min|           Sum|     Mean|       Q1|       Q2|        Q3|
+------------+--------+--------------+---------+---------+---------+----------+
|9.958,026 ms|1,782 ms|328.591,625 ms|368,79 ms|31,157 ms|82,589 ms|241,443 ms|
+------------+--------+--------------+---------+---------+---------+----------+

Othe non-signifant changes:
* Manual tests support urls to urls instead of local files
* Removed optimal at optimization
* Removed members from Split and State classes.
* Upgrade scalameta

### DIFF
--- a/benchmarks/src/main/scala/org/scalafmt/benchmarks/FormatBenchmark.scala
+++ b/benchmarks/src/main/scala/org/scalafmt/benchmarks/FormatBenchmark.scala
@@ -68,7 +68,7 @@ abstract class FormatBenchmark(path: String*) {
   }
 
   // No need to run same benchmark again and again.
-  //  @Benchmark
+    @Benchmark
   def scalariform(): String = {
     ScalaFormatter.format(code, scalariformPreferences)
   }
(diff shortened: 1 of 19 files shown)


olafurpg:
This commit does affect the state visits a lot.

Some fewer states here
screen shot 2016-02-23 at 14 30 35

Some more states here:
screen shot 2016-02-23 at 14 32 21

PR #206, opened 2016-05-04
Fixes the check for URLs vs files

propensive:


propensive:
Tested locally, and appears to fix the problem.

olafurpg:
Whoop whoop, you are the first contributor to scalafmt! 🎉

PR #229, opened 2016-05-12
Use unmanagedSourceDirectories for sourceDirectories

fthomas:
`sourceDirectories in hasScalafmt` is currently initialized with
`List(scalaSource.value)` which means that scalafmt only looks into
one source directory per project. But a typical Scala.js cross
project has at least two source directories js/ and shared/ or
jvm/ and shared/. Sometimes there are also multiple source
directories for different Scala versions.

This change initializes `sourceDirectories in hasScalafmt` with
`unmanagedSourceDirectories` which contains all source directories
with manually created sources. In my tests
`unmanagedSourceDirectories` always includes `scalaSource`.

Here are for example both settings for typelevel/cats:

> coreJVM/scalaSource
[info] cats/core/.jvm/src/main/scala

> coreJVM/unmanagedSourceDirectories
[info] List(
  cats/core/.jvm/src/main/scala-2.11,
  cats/core/.jvm/src/main/scala,
  cats/core/.jvm/src/main/java,
  cats/core/src/main/scala)

And since cats has no sources in cats/core/.jvm/src/main/scala,
scalafmt won't format anything.

### DIFF
--- a/readme/Readme.scalatex
+++ b/readme/Readme.scalatex
@@ -415,16 +415,7 @@
        scalafmt update.
 
      @sect{The SBT plugin doesn't format anything}
-       @ul
-         @li
-           Run @code("reload"), @code("clean") and then try again.
-         @li
-           If that doesn't work. Try adding this to your build.sbt:
-
-       @hl.scala
-         sourceDirectories in (Compile,
-                               org.scalafmt.sbt.ScalaFmtPlugin.autoImport.hasScalafmt) :=
-           (sourceDirectories in Compile).value
+       Run @code("reload"), @code("clean") and then try again.
 
   @sect{Changelog}
     @sect{0.2.3}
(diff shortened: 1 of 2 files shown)


fthomas:
I should mention that I've tested this with refined and it allowed to get rid of this [workaround](https://github.com/fthomas/refined/blob/fdebe97575bbefdd66bb8a5f2344b0571fc7295e/build.sbt#L337-L340).

olafurpg:
From the sbt documentation:

> sbt collects sources from unmanagedSourceDirectories, which by default consists of scalaSource and javaSource.

Nice. Thanks a lot @fthomas 😄

fthomas:
For the sake of completeness, this is how `unmanagedSourceDirectories` is [defined by sbt by default](https://github.com/sbt/sbt/blob/bcd36599357eac5163be9c9d836fd4fdf068624b/main/src/main/scala/sbt/Defaults.scala#L209):
```scala
unmanagedSourceDirectories := makeCrossSources(
  scalaSource.value, javaSource.value, scalaBinaryVersion.value, crossPaths.value)
```

PR #245, opened 2016-05-16
Drop %% in addSbtPlugin

dwijnand:
Sbt plugins aren't cross built, see:

  * https://github.com/sbt/sbt/blob/v0.13.11/main/src/main/scala/sbt/Defaults.scala#L1948-L1953
  * https://github.com/sbt/sbt/blob/v0.13.11/main/src/main/scala/sbt/Defaults.scala#L899-L900

crossVersion is disabled

### DIFF
--- a/readme/Readme.scalatex
+++ b/readme/Readme.scalatex
@@ -71,7 +71,7 @@
 
     @sect{SBT}
       @hl.scala
-        addSbtPlugin("com.geirsson" %% "sbt-scalafmt" % "@org.scalafmt.Versions.stable")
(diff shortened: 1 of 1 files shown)


olafurpg:
Cool, I didn't know that. Thanks!

sjrd:
> Sbt plugins aren't cross built

Well ... actually, they are, but with a more complicated scheme than just `%%`. This scheme is taken care of by `addSbtPlugin`, which by the way disregards `%%` vs `%` because it overrides it.

dwijnand:
That's more precise, you're right.

PR #246, opened 2016-05-17
fallback to config file in home dir

maximn:
https://github.com/olafurpg/scalafmt/issues/239

This will fallback to `.scalafmt` config file in home directory in case of no such file in project directory.

* Been able to compile & run the plugin but wasn't able to test it.

### DIFF
--- a/intellij/src/org/scalafmt/intellij/ScalafmtAction.scala
+++ b/intellij/src/org/scalafmt/intellij/ScalafmtAction.scala
@@ -4,6 +4,8 @@
  */
 package org.scalafmt.intellij
 
+import java.io.File
+
 import scala.collection.mutable
 import scala.meta.parsers.ParseException
 
@@ -66,11 +68,9 @@ class ScalafmtAction extends AnAction {
   }
 
   private def getConfigFileInPath(path: String) = {
-    val file = FileOps.getFile(path, ".scalafmt")
-    if (file.isFile)
-      Some(file.getAbsolutePath)
-    else
-      None
+    Option(FileOps.getFile(path, ".scalafmt")).collect {
+      case file: File if file.isFile => file.getAbsolutePath
+    }
   }
 
   private val homeDir = System.getProperty("user.home")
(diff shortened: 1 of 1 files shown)


intellij/src/org/scalafmt/intellij/ScalafmtAction.scala
```diff
@@ -65,19 +65,29 @@ class ScalafmtAction extends AnAction {
...
+
   private def getStyle(event: AnActionEvent): ScalafmtStyle = {
     val customStyle = for {
       project <- Option(event.getData(CommonDataKeys.PROJECT))
-      configFile = FileOps.getFile(project.getBasePath, ".scalafmt")
-          if configFile.isFile
-      style <- StyleCache.getStyleForFile(configFile.getAbsolutePath)
+      localConfig = getConfigFileInPath(project.getBasePath)
+      globalConfig = getConfigFileInPath(homeDir)
+      configFile <- localConfig orElse globalConfig
```
olafurpg: I try to avoid infix application for non-symbolic operators: `localConfig.orElse(globalConfig)`

intellij/src/org/scalafmt/intellij/ScalafmtAction.scala
```diff
@@ -65,19 +65,29 @@ class ScalafmtAction extends AnAction {
     }
   }
 
+  private def getConfigFileInPath(path: String) = {
+    val file = FileOps.getFile(path, ".scalafmt")
```
olafurpg: `Option(FileOps.getFile(...)).filter(_.isFile)`

olafurpg:
Awesome, LGTM 👍 I don't have unit tests for the IntelliJ plugin, manual testing will do for now.

olafurpg:
Thanks a lot for this 😄 I've updated to the IntelliJ plugin to v0.2.5 with this change.

maximn:
Awesome! I've tested the updated published version and seems to work.
Couldn't find the usage readme in github so please also update there.

And something else, in the usage readme you're writing "See ScalafmtConfig scaladoc.", I think you mean `ScalafmtStyle` instead.

olafurpg:
I've updated the docs on the IntelliJ installation and also renamed ScalafmtConfig. FWIW, the docs are [here](https://github.com/olafurpg/scalafmt/blob/master/readme/Readme.scalatex).

Thanks!

PR #298, opened 2016-05-31
Space after triple equals

anna-zzz:
Little thingy that always bothers me with Scalatest matchers :)
What do you think?

### DIFF
--- a/core/src/main/scala/org/scalafmt/ScalafmtStyle.scala
+++ b/core/src/main/scala/org/scalafmt/ScalafmtStyle.scala
@@ -40,6 +40,7 @@ import sourcecode.Text
   *                                  scalafmt will fit as many parent constructors
   *                                  on a single line. If false, each parent
   *                                  constructor gets its own line.
+  * @param spaceAfterTripleEquals If true, formats ===( as === (
   * @param alignByArrowEnumeratorGenerator If true, aligns by <- in for comprehensions.
   * @param alignByIfWhileOpenParen If true, aligns by ( in if/while/for. If false,
   *                                indents by [[continuationIndentCallSite]].
@@ -59,6 +60,7 @@ case class ScalafmtStyle(
     spacesInImportCurlyBrackets: Boolean,
     allowNewlineBeforeColonInMassiveReturnTypes: Boolean,
     binPackParentConstructors: Boolean,
+    spaceAfterTripleEquals: Boolean,
     alignByArrowEnumeratorGenerator: Boolean,
     alignByIfWhileOpenParen: Boolean
 ) {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -282,6 +282,7 @@ class Router(formatOps: FormatOps) {
           } =>
         val modification: Modification = leftOwner match {
           case _: Mod => Space
+          case t: Term if style.spaceBeforeTripleEquals && t.tokens.map(_.code) == Seq("===") => Space
```
olafurpg: `t: Term.Name`

core/src/main/scala/org/scalafmt/ScalafmtStyle.scala
```diff
@@ -59,6 +59,7 @@ case class ScalafmtStyle(
     spacesInImportCurlyBrackets: Boolean,
     allowNewlineBeforeColonInMassiveReturnTypes: Boolean,
     binPackParentConstructors: Boolean,
+    spaceBeforeTripleEquals: Boolean,
```
olafurpg: I suppose `spaceAfterTripleEquals` is more accurate :)

olafurpg:
The first PR that touches `Router` 🎉 Thanks for doing this.

At first I thought there was a bug in scalafmt since I expected the `must === (2)` to be an infix application and arguments to infix applications always get a space. Turns out it's just a regular method call that's passed as an argument to must. I personally don't like how ScalaTest has made this a convention since it makes infix applications even more confusing. I'm happy to include this as a flag since I suspect many users see this differently than me.

core/src/main/scala/org/scalafmt/ScalafmtStyle.scala
anna-zzz: Oops :)

anna-zzz:
Fantastic!
So I also need to update readme I suppose? Anything else except your comments?

olafurpg:
Right, I forgot, for now it's sufficient to only document this flag in the `ScalafmtStyle` docstring. The cli flags are automatically included in the readme which appears on the website. The plan is add clearer configuration examples to the website soon.

olafurpg:
LGTM, no need to wait for travis since the previous fix passed CI.

olafurpg:
Thanks a lot for this, it's the first non-trivial contribution to scalafmt :) FYI, I don't expect to publish 0.2.6 until in the later part of June as I have to complete on my thesis writing first.

olafurpg:
@anna-zzz Just wanna draw your attention to #303 if you're itching to use the new flag.

PR #302, opened 2016-06-01
gitignore and clean repos.tar.gz

anna-zzz:


sjrd:
A better fix would be to put that file inside `target/`.

.gitignore
```diff
@@ -1,6 +1,8 @@
 *.class
 *.log
 
+repos.tar.gz
```
olafurpg: 👍

build.sbt
```diff
@@ -49,6 +49,7 @@ lazy val commonSettings = Seq(
   assemblyJarName in assembly := "scalafmt.jar",
   ScoverageSbtPlugin.ScoverageKeys.coverageExcludedPackages :=
       ".*Debug;.*ScalaFmtLogger;.*Versions",
+  cleanFiles <+= baseDirectory(_ / "repos.tar.gz"),
```
olafurpg: Is this necessary? I would prefer not to re-download it on every sbt clean.

olafurpg:
Good catch, seems like I ignore `*.gz` in my `$HOME/.gitignore`.

The file is not inside `target/` because I want to keep it between `sbt clean`s. Is that a terrible idea? As long as I don't change the contents of `repos.tar.gz` I think it will be fine.

sjrd:
> Is that a terrible idea?

No, it's reasonable.

build.sbt
anna-zzz: Nope, but it's a PR open for discussion :) Easier to add a line and submit PR than to ask. Will remove.

olafurpg:
LGTM, thanks! :)

PR #314, opened 2016-06-04
Correct all other mentions of "paretheses"

i-am-the-slime:
I didn't realise the mistake was repeated.


olafurpg:
Silly me, thanks! :)

Issue #316, opened 2016-06-07
More robust/standard configuration


olafurpg:
Following discussions on Gitter: https://gitter.im/olafurpg/scalafmt?at=5755e80492fc7c915f574b06

The current configuration is built on scopt and is quite hacky: https://olafurpg.github.io/scalafmt/#Configuration
It would be great to use a more standard syntax.

Requirements:
* Scala.js compatible
* supports comments
* not verbose
* allows multiple configurations on same line, like `a=1;b=2`. See #315

Considered options:
* [shocon](https://github.com/unicredit/shocon) seems like a perfect fit. @andreaTP, are there any docs/non-SNAPSHOT releases?
* typesafe config: no Scala.js support.
* http://oncue.github.io/knobs/: can't see Scala.js support.
* YAML: Don't think there's any Scala.js implementation.

andreaTP:
shocon is right now used in akka-js as a "drop in" replacement of typesafe config.
It is developed by @evacchi, it works, but it is not complete.
If you want to give a try use
```
"eu.unicredit" %%% "shocon" % "0.0.2-SNAPSHOT"
```
from 
```
resolvers += "sonatype-snapshots" at "https://oss.sonatype.org/content/repositories/snapshots"
```
We will be happy to support!

djspiewak:
As a counter-point…  Hocon is an incredibly annoying configuration format.  :-)  It's better than JSON, but I've just had universally massively negative experiences with it.  YAML is very lightweight, and its only negative aspects are a) hard to write a fully-compliant parser, and b) association with Rails.

Verizon uses YAML internally quite extensively, and it appears that we primarily use the following dependency:

```sbt
libraryDependencies += "org.yaml" %  "snakeyaml" % "1.16"
```

Not the prettiest API to work with, but it does work.

Why does scalafmt need to be Scala.js compatible?

olafurpg:
> Why does scalafmt need to be Scala.js compatible?

I want make a browser extension that formats Scala code on Github with scalafmt. If #315 gets implemented then then scalafmt-core will depend on the parser for the configuration language.

Does YAML support multiple statements on the same line? Like `a=1;b=2`.

djspiewak:
> Does YAML support multiple statements on the same line? Like a=1;b=2.

Yes.  It's not common style, but it is allowed.  The following are equivalent:

```yaml
foo:
  a: 1
  b: 2

# and this

foo: { a: 1, b: 2 }
```

[...]

PR #359, opened 2016-07-05
Created a scalafmt shell script to bootstrap usage

hntd187:
I wrote this to help me use the different versions of scalafmt easier.

Basically it has all the same uses, but adds a --v as the first argument to the script. If such an arg exists it will use that version number and if it doesn't have that version it will download it. Otherwise it will default to the latest version of scalafmt as determined by a call to the github API. This also creates a "jars" folder in the current directory where it stays, but I'm sure it'd be easy to change that to something like a ~/.scalafmt folder for more global use.

Some example usage
```
./scalafmt --v 0.2.9 --build-info
Using version: v0.2.9
build commit: 5b8483afad
build time: Thu Jun 30 04:55:10 EDT 2016
```

Other than that, all normal usage is the same.
```./scalafmt -v```

### CODE REVIEW

**Comment by olafurpg** [comment]
This looks like a great feature, thanks for putting it together. I will need more time to try it out before merging.

**Comment by hntd187** [comment]
I'm also more than open to harsh criticism of my bash scripting if anyone would like to comment. It may or may not be terrible code.

**Comment by olafurpg** [comment]
Sorry for the slow response, I've been traveling in West Africa with EDGE internet connection for 3 weeks. 

Writing a cross-platform command line script is no easy task and I am by no means an expert on the topic. I really like the idea of having a bootstrap script so I am happy to iterate a bit on this.

When I ran the current script on zsh on my Mac I got some errors:
```
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
(23) Failed writing body
You don't have jars folder, creating it...
No scalafmt jar for version , downloading it...
From: https://github.com/olafurpg/scalafmt/releases/download//scalafmt.tar.gz
######################################################################## 100.0%
tar: Unrecognized archive format
tar: cli/target/scala-2.11/scalafmt.jar: Not found in archive
tar: Error exit delayed from previous errors.
Error: Invalid or corrupt jarfile /Users/ollie/dev/scala/scalafmt/bin/jars/scalafmt-.jar
```


- A good start is to enable `set -e` so the script fails on non-zero exit code.
- The `command` command is a good way to test dependencies, see how it's used in publish.sh: https://github.com/olafurpg/scalafmt/blob/e4851c01ae3f323edcfa7e2fad632639333ab5fb/bin/publish.sh#L11 You can go fancy with arrays to avoid the boilerplate, but I don't mind a little boilerplate as long as it's readable.
- Where do the jars get downloaded? I personally prefer to default to `$HOME/.scalafmt-bin/0.2.11/...` but allow users to provide a `--dir /path/` flag to override. Unfortunately, we can't use `$HOME/.scalafmt` without changing the IntelliJ plugin since it will try to read configuration options from that file (this will probably change with #316 into .scalafmt.yml). 
- we can't rely on any bash specific features like `BASH_SOURCE` since users (like myself) use alternative shells like zsh or fish. `pwd` should do the trick, although I don't think we'll need it
- the `-P` flag doesn't work on mac's default grep installation. It's easiest probably to rewrite the regex to grep's default regex engine.


WDYT? :)

hntd187:
I agree with everything you've mentioned. I wrote this originally to assist with my own usage, but obviously to make it more full featured we'd need to tweak some of code a bit. I honestly have never used zsh, I use the standard bash that comes with Linux. I'll also have to revert the [[ ]] style testing as I think that is specific to bash as well. I'll do some research and read a bit on it and make those changes. Do you think just making it posix compatible will ensure its function?

olafurpg:
> Do you think just making it posix compatible will ensure its function?

I am not sure here, especially what's needed to make it work on Windows. I suggest we keep the original `scalafmt` script as it is and add a new script `scalafmt-bootstrap` (or something more meaningful) that users can optionally link to as `scalafmt` if they choose so. That way we don't need to guarantee compatibility with all OSes but still make it useful on Linux/OSX with bash/zsh.

> I'll also have to revert the [[ ]] style testing as I think that is specific to bash as well.

That won't be necessary I think, at least it works fine in zsh.

hntd187:
Also, how did you run this script? If should be ```chmod +x scalafmt``` and then ran from the command line, that way it still runs in bash even if your shell is zsh. It'll I believe blow up if you run is via sh

olafurpg:
Aaah, it doesn't matter at all if I use zsh because it of the `#!/bin/bash` pragma. Never mind my previous comment then, I just run the script normally.

hntd187:
Hey @olafurpg I realized I updated this script as per your suggestions awhile back but never checked it in. This I guess was prompted by the fact I downloaded your script to bootstrap my migration to hocon and it was broken trying to download 0.4.2-SNAPSHOT, but my script worked perfectly fine :+1:

olafurpg:
@hntd187 I tried out the script and it works great. I would love to include this in the project.

I noticed that the startup penalty is ~1s (with my current wifi) to check for the latest version. What do you think about making that opt-in by running for example `scalafmt --upgrade`? In the end, you upgrade a lot less often than you run scalafmt.


>  it was broken trying to download 0.4.2-SNAPSHOT,

I used coursier to create that bootstrap script and was to impatient wait for maven to sync. I guess I shouldn't do that again. Lesson learned.

olafurpg:
Also, it would be nice to include a comment at the top of the script what a short explanation what the script does, that it accepts a `--v` flag and that it stores files in `$HOME/.scalafmt-bin`.

BTW. Since configuration is just moved to `.scalafmt.conf`, it will soon be OK to store the files in `$HOME/.scalafmt`. I'm fine with keeping it in `$HOME/.scalafmt-bin` for now though.

hntd187:
@olafurpg Yea, the opt in for the upgrade seems reasonable, I think I will write a file to .scalafmt-bin/version that has the current latest version and when you upgrade it will check. The github API seems kinda slow. I can also add some comments in about what options people have.

hntd187:
I updated this as per your suggestions @olafurpg I also added bats tests to check functionality https://github.com/sstephenson/bats

olafurpg:
Amazing work @hntd187 I had not heard of bats, very cool. Thanks a lot!

olafurpg:
@hntd187 btw, I just realised that a package manager (like my homebrew script) might rely on the old `bin/scalafmt` script so I renamed this script to `bin/scalafmt_auto`. I will update the docs to promote this script.

hntd187:
@olafurpg makes sense with homebrew, and if I don't pop up on issues, feel free to ping me if anyone has portability issues or bugs in the script, I'll do my best to maintain the script going forward for you, and if you'd like to see more functionality on it too we can discuss something like that too, more like a version manager esq thing because scalafmt moves really fast in versions so it might not be a horrible idea

olafurpg:
Thank you @hntd187. I appreciate it.

One feature I have thought of adding to the `scalafmt` script is support for nailgun. We could add nailgun as a dependency to the `cli` project and make sure that the nailgun server injects the correct working directory + input/output stream. The challenge is start/stop/restart the nailgun server from bash. I'd be interested to know if you have any ideas on how to approach this.

hntd187:
@olafurpg I think it's certainly possible to do that, you wanted say a constantly running nailgun JVM for scalafmt? I guess the start-up for it is pretty expensive? What other aspects of that do you think makes it an attractive option?

olafurpg:
> constantly running nailgun JVM for scalafmt

Yes, exactly. This could be an opt-in feature, for example by creating the file `~/.scalafmt-bin/nailgun`. We might want to restart the server every 100th invocation or so.

> I guess the start-up for it is pretty expensive?

The startup time for the Nailgun server is typically <1s. You do it so rarely that I doesn't really matter.

> What other aspects of that do you think makes it an attractive option?

My experiments show that scalafmt becomes ~2-3x faster once the JVM is hot (usually after 1-2 invocations). I think this significant speedup comes from the fact that the JVM is able to JIT and you don't pay the 500ms JVM startup penalty.

The fast startup time would make it easier to integrate scalafmt with text editors like vim/emacs/atom. Each editor plugins just calls the `scalafmt` script via the console.

I'll try to put something together so that the scalafmt cli can be used with nailgun. Then maybe you could get a better idea of how it might work smoothly with the bash script.

olafurpg:
BTW, one other thing I think would greatly completent the bash script is shrinking the fat jar with proguard: https://github.com/olafurpg/scalafmt/issues/466 I think the binary could easily become <10mb.

olafurpg:
The docs now promote this script as the recommended way to install scalafmt on Linux  and OSX: https://olafurpg.github.io/scalafmt/#Linux/OSX

Issue #180, opened 2016-04-18
Don't insert space before colon 


densh:
Diff:
```scala
-    def map[T: reflect.ClassTag](f: Node => T): Seq[T] = {
+    def map[T : reflect.ClassTag](f: Node => T): Seq[T] = {
```

I'd like not to have colon inserted after `T`.

Using: 0.2.1

olafurpg:
Interesting comment from @dwijnand on gitter:

> when it's one type I agree with Denys I like `T: ClassTag` too, but when it's more than one I prefer `T : ClassTag : Functor`.

sjrd:
@dwijnand's comment makes sense. I have also seen (and used) a space being used if there is a lower bound and/or upper bound on the type.

olafurpg:
Looking at cats, they seem to [consistently put no space before colon](https://github.com/typelevel/cats/blob/master/kernel-laws/src/test/scala/cats/kernel/laws/LawTests.scala#L25):

```scala
implicit def orderLaws[A: Eq: Arbitrary] = OrderLaws[A]
```

densh:
@olafurpg that looks great.

PR #369, opened 2016-07-25
Add spaceBeforeContextBoundColon option

triggerNZ:
Context bound constraints are often written with spaces surrounding the `:` , for example:

```scala
def f[T : Numeric](t: T) = ???
```
, which is consistent with other type constraint usages such as `[T <: AnyRef]`.

https://github.com/olafurpg/scalafmt/issues/180 dealt with this and explicitly dropped spaces before, but I think an option to have it (off by default) is still useful

This PR adds the `spaceBeforeContextBoundColon ` option to allow inserting a space before the `:`.

### DIFF
--- a/cli/src/main/scala/org/scalafmt/cli/Cli.scala
+++ b/cli/src/main/scala/org/scalafmt/cli/Cli.scala
@@ -267,6 +267,12 @@ object Cli {
               |        It's best to play around with scala.meta in a console to
               |        understand which regexp you should use for the owner.
               |""".stripMargin
+
+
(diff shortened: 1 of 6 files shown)


olafurpg:
Looks great, I'm happy to add this flag. There's a missing test case for `method[F[_]: TypeClass]`, this was a bug even before #180. I've consolidated the three matches on colons in the Router and fixed that test case in fd716dc932c91206da575c6816e9785996851ec9. Could you merge that commit and squash into one commit?

triggerNZ:
Done, thanks for reviewing :)

olafurpg:
LGTM, thanks for contributing 😄

PR #396, opened 2016-08-07
update sbt-scoverage from 1.0.1 to 1.3.5

NicolasRouquette:
While scalafmt 0.2.13 isn't released yet, I'm doing local builds with a source-level SBT project dependency; this led to finding that scalafmt's dependency on sbt-scoverage is out-of-date.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,4 @@
-import scoverage.ScoverageSbtPlugin.ScoverageKeys.coverageHighlighting
+import scoverage.ScoverageKeys.coverageHighlighting
 
 lazy val buildSettings = Seq(
   organization := "com.geirsson",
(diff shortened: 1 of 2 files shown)


olafurpg:
I tried to upgrade scoverage to 1.3.5 earlier this week but it introduced flakyness in the tests, the same error I see in the travis log here. Been meaning to file a bug report on scoverage but haven't gotten to it yet.

NicolasRouquette:
It's really surprising that just upgrading sbt-scoverage makes a functional difference in the behavior of the unit tests. I'm curious to learn what's going on.

olafurpg:
scoverage is a compiler plugin that injects a lot of instrumentation at every code statement so it's actually not so surprising that it makes a functional difference. In fact, I'm impressed that scoverage works so well :)

olafurpg:
I'm gonna close this since I don't know when I'll get around to try the latest scoverage release.

PR #401, opened 2016-08-13
0.3.0 rc1

### LINKED ISSUE #400: Make default continuation indent 2 spaces instead of 4?
Why?

- It seems to be most common in many coding styles. For example, the intellij default formatting settings.
- it frees up horizontal space
- It means all bumps in scope add 2 space indentation.

olafurpg:
This commit should clearly by several commits. The changes are:

- No more superfluous indent settings. Indentation is relaxed
  by default just like scalariform.
- More agressive newline penalty to avoid code like this:

    function("foobar bananas" +
        "continued")

  Instead, this is preferred:

    function(
      "foobar bananas" +
        "continued")

  We could make this configurable, but it's not for now.

- The default continuation indent at call-site is now 2 spaces
  instead of 4. Closes #400.


olafurpg:
I ran the tests on a 48 core 64gb ram machine, all passes. Hoping to integrate the server once I get a public IP on the machine.

PR #402, opened 2016-08-13
Format sbt files, fixes #378

### LINKED ISSUE #378: Format SBT files
The latest version of scala.meta supports parsing SBT files. Once #214 is done, it will be easy to add a flag to scalafmt to indicate that the file should be formatted as an SBT file. The relevant line is here: https://github.com/olafurpg/scalafmt/blob/c1a7233d9edbe2600a60bce55c64ccd18f707d42/core/src/main/scala/org/scalafmt/ScalaFmt.scala#L30

olafurpg:


olafurpg:
CI passed on the monster machine.

PR #407, opened 2016-08-17
Make SBT plugin pick up .sbt files

### LINKED ISSUE #405: Format *.sbt files via SBT plugin
The plugin uses `sourceDirectories` to find which *.scala files to format, see https://github.com/olafurpg/scalafmt/blob/b56564d56a7960bcbbc0c152452e29e9014f3a91/scalafmtSbt/src/main/scala/org/scalafmt/sbt/HasScalaFmt.scala#L47
All that's needed is to include *.sbt files in that list.

I am happy to help anyone fix this issue, I don't think it should be so hard. I am less motivated to do this myself since I don't use the SBT plugin.

mdedetrich:
This addresses to fix #405. Note that this is a "cheap" solution, the better solution would be to read the `scalaFmtConfig` file, read the `formatSbtFiles` config option and then adjust the filter depending on whether we need to read SBT files or not.

I can add this functionality if asked

### DIFF
--- a/scalafmtSbt/src/main/scala/org/scalafmt/sbt/ScalaFmtPlugin.scala
+++ b/scalafmtSbt/src/main/scala/org/scalafmt/sbt/ScalaFmtPlugin.scala
@@ -92,7 +92,7 @@ object ScalaFmtPlugin extends AutoPlugin {
   def configScalafmtSettings: Seq[Setting[_]] =
     List(
       (sourceDirectories in hasScalafmt) := unmanagedSourceDirectories.value,
-      includeFilter in Global in hasScalafmt := "*.scala",
+      includeFilter in Global in hasScalafmt := "*.scala" || "*.sbt",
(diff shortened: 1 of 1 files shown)


olafurpg:
Cool!

I just added a `formatSbtFiles: SettingKey[Boolean]` to the plugin, see https://github.com/olafurpg/scalafmt/commit/e4834fdbfe58988f49077d084565b613ee027d44#diff-2901d773691784fbc97da6272e8f5225R50
Then `formatSbtFiles.value` can be used to decide whether to include *.sbt files or not. I think that makes more sense than reading it from the configuration file.

I also added a check to the scripted test to see that the file contents are actually formatted, could uncomment the test in the following line just so we can automatically test that it actually works? https://github.com/olafurpg/scalafmt/commit/e4834fdbfe58988f49077d084565b613ee027d44#diff-9a3523e02d8ef9c8b85cb5d931beb000R32

PR #437, opened 2016-09-11
Exclude files with a CLI option

stefanobaghino:
Adds a `--exclude` flag to the CLI utility to ignore certain files.


olafurpg:
Awesome, LGTM. Thanks 😄

Issue #438, opened 2016-09-11
Add an option to force a newline before the parameters of a lambda


stefanobaghino:
It may be interesting for some to have the possibility to always have a newline between the open brace token and the parameters of an anonymous function, like the following:

    something.map {
      n =>
        // do awesome stuff with `n`
    }

stefanobaghino:
Fixed by #439 :smiley:

PR #439, opened 2016-09-11
Option to put a newline to be put before the parameters of a lambda

stefanobaghino:
Addresses #438

### DIFF
--- a/core/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/core/src/main/scala/org/scalafmt/internal/Router.scala
@@ -173,10 +173,17 @@ class Router(formatOps: FormatOps) {
         val skipSingleLineBlock =
           startsLambda || newlines > 0
 
+        val newlineBeforeLambdaParameters =
+          formatOps.style.newlineBeforeLambdaParameters
+
         Seq(
           Split(Space, 0, ignoreIf = skipSingleLineBlock)
             .withOptimalToken(close, killOnFail = true)
             .withPolicy(SingleLineBlock(close)),
+          Split(nl, 0, ignoreIf = !newlineBeforeLambdaParameters)
+            .withOptimalToken(lambdaArrow)
+            .withIndent(2, close, Right)
+            .withPolicy(lambdaPolicy),
           Split(Space, 0, ignoreIf = !startsLambda)
             .withOptimalToken(lambdaArrow)
             .withIndent(lambdaIndent, close, Right)
--- a/core/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/core/src/main/scala/org/scalafmt/internal/Router.scala
@@ -173,18 +173,14 @@ class Router(formatOps: FormatOps) {
         val skipSingleLineBlock =
           startsLambda || newlines > 0
 
-        val newlineBeforeLambdaParameters =
+        val lambdaParamsNl =
           formatOps.style.newlineBeforeLambdaParameters
 
         Seq(
-          Split(Space, 0, ignoreIf = skipSingleLineBlock)
+          Split(Space, 0, ignoreIf = lambdaParamsNl || skipSingleLineBlock)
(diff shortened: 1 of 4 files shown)


stefanobaghino:
`[error] (*:update) sbt.ResolveException: unresolved dependency: pl.project13.scala#sbt-jmh;0.2.6: not found`

:anguished:

stefanobaghino:
My bad, looks like I had to rebase.

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -173,10 +173,17 @@ class Router(formatOps: FormatOps) {
...
           startsLambda || newlines > 0
 
+        val newlineBeforeLambdaParameters =
+          formatOps.style.newlineBeforeLambdaParameters
+
         Seq(
           Split(Space, 0, ignoreIf = skipSingleLineBlock)
             .withOptimalToken(close, killOnFail = true)
             .withPolicy(SingleLineBlock(close)),
+          Split(nl, 0, ignoreIf = !newlineBeforeLambdaParameters)
```
olafurpg: I don't think we need a new branch here, it is probably enough to do something similar to what @melrief does in https://github.com/olafurpg/scalafmt/pull/440/files#diff-ded7f78a4a0d490270763a26fa35f6b7R828

Basically, we want to ignore the Space on line 180 if `newlines > 0`.

core/src/test/resources/newlineBeforeLambdaParams/LambdaNewline.stat
```diff
@@ -0,0 +1,12 @@
...
+lst.map { x =>
+      println(x)
+  x + 1
+}
+>>>
+lst.map {
+  x =>
+    println(x)
+    x + 1
+}
```
olafurpg: What happens for this?

```
lst.map { x => x }
```

stefanobaghino: In this case in particular I'd like one-liners to be preserved, unless explicitly indented... but I'm struggling with it.

stefanobaghino:
Ok, I think this can be satisfactory as a start, let me know what you think.

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -173,11 +173,14 @@ class Router(formatOps: FormatOps) {
         val skipSingleLineBlock =
           startsLambda || newlines > 0
 
+        val lambdaParamsNl =
+          formatOps.style.newlineBeforeLambdaParameters
```
olafurpg: `&& newlines > 0` should do the trick to support single line lambdas

olafurpg: Also, you can use `style` instead of `formatOps.style`.

stefanobaghino: Thanks! I'll try it out.

stefanobaghino:
Ok, I've ditched the one-liner preservation thing completely since it's not what's currently done anyway.

olafurpg:
LGTM, thanks 😄

olafurpg:
@stefanobaghino FYI, I renamed the flag to `newlines.alwaysBeforeCurlyBraceLambdaParams` to be consisted with a group of other `newlines.*` options.

stefanobaghino:
@olafurpg Great, thanks for accepting the PR, can't wait to try out the new version! :tada:

core/src/test/resources/newlineBeforeLambdaParams/LambdaNewline.stat
fmv1992: hey, I think its an overkill to open an issue for this

but how can I avoid

```
def f(x) = {
    x => x
}
```

becoming

```
def f(x) = {
    x =>
      x
}
```

?

Issue #356, opened 2016-07-02
Preserve newlines between vals and their body


fthomas:
Original:
```scala
val x: Int =
  1
```

Scalafmt:
```scala
val x: Int = 1
```

It would be nice if the input is not reformatted. This behaviour is already implemented for `def`s, i.e. if `x` is a `def`, the newline is kept. I think it is more regular if `val`s and `def`s are treated the same in this regard.

Using:
* 0.2.10
* --style default

psttf:
Is it possible to apply the same behavior to `<-` and `=` in `for` expressions?

PR #445, opened 2016-09-17
#356 Code with windows line endings after formatting still has windows endings

mmatloka:
Related to: https://github.com/olafurpg/scalafmt/issues/365#issuecomment-243346157

### DIFF
--- a/core/src/main/scala/org/scalafmt/Scalafmt.scala
+++ b/core/src/main/scala/org/scalafmt/Scalafmt.scala
@@ -8,6 +8,7 @@ import scala.util.control.NonFatal
 import scala.meta.Input.stringToInput
 
 import org.scalafmt.Error.Incomplete
+import org.scalafmt.ScalafmtStyle.{PreserveLineEndings, WindowsLineEndings}
 
 object Scalafmt {
 
@@ -48,11 +49,13 @@ object Scalafmt {
         val search = new BestFirstSearch(formatOps, range, formatWriter)
         val partial = search.getBestPath
         val formattedString = formatWriter.mkString(partial.splits)
-        val correctedFormattedString = if (isWindows) {
-          formattedString.replaceAll(UnixLineEnding, WindowsLineEnding)
-        } else {
-          formattedString
-        }
+        val correctedFormattedString =
+          if ((style.lineEndings == PreserveLineEndings && isWindows) ||
+              style.lineEndings == WindowsLineEndings) {
+            formattedString.replaceAll(UnixLineEnding, WindowsLineEnding)
+          } else {
+            formattedString
+          }
         if (partial.reachedEOF) {
           FormatResult.Success(correctedFormattedString)
         } else {
--- a/core/src/main/scala/org/scalafmt/Scalafmt.scala
+++ b/core/src/main/scala/org/scalafmt/Scalafmt.scala
@@ -11,6 +11,9 @@ import org.scalafmt.Error.Incomplete
 
 object Scalafmt {
 
+  private val WindowsLineEnding = "\r\n"
+  private val UnixLineEnding = "\n"
+
   /**
     * Format Scala code using scalafmt.
     *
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/scalafmt/Scalafmt.scala
```diff
@@ -39,15 +48,23 @@ object Scalafmt {
...
         }
       }
     } catch {
       // TODO(olafur) add more fine grained errors.
       case NonFatal(e) => FormatResult.Failure(e)
     }
   }
+
+  private[this] def containsWindowsLineEndings(code: String): Boolean =
+    code.contains("\r\n")
```
olafurpg: `WindowsLineEnding`?

core/src/test/scala/org/scalafmt/LineEndingsTest.scala
```diff
@@ -0,0 +1,22 @@
+package org.scalafmt
+
+import org.scalafmt.util.DiffAssertions
+import org.scalatest.FunSuite
+
+class LineEndingsTest extends FunSuite with DiffAssertions {
+  val javadocStyle = ScalafmtStyle.default.copy(scalaDocs = false)
```
olafurpg: Why not default style?

mmatloka:
Corrected, squashed, pushed :)

olafurpg:
One last thought, should we have a flag to disable this new behavior?

olafurpg:
Whoops, I was fiddling with automatic publishing from travis (only possible from master branch) and just remembered that I disabled the tests suite to get faster experiments. Could you rebase on master and re-sync?

sjrd:
> One last thought, should we have a flag to disable this new behavior?

Yes please, so that `scalafmtTest` can assert that all my files have Linux line endings.

olafurpg:
@sjrd how about a flag like `--lineEndings unix/windows/preserve` where preserve will not change any line endings, unix enforces \n and windows enforces \r\n.

This option could be a sealed abstract class `LineEnding` and go into ScalafmtStyle. @mmatloka Here is an example on how to parse a custom type in the Cli: https://github.com/olafurpg/scalafmt/blob/892298eb38eec9590bef732478a7f2d8c0eadec8/cli/src/main/scala/org/scalafmt/cli/Cli.scala#L87

mmatloka:
@olafurpg as `preserve` you mean behaviour implemented in this PR or current master behaviour?

olafurpg:
@mmatloka Hmmm. Excellent question. I think preserve should be the behavior implemented in this PR. The current master branch behavior is buggy.

olafurpg:
@mmatloka For the readme, it should be `@b{preserve}` instead of `@b(preserve)`. With the parentheses it assumes the argument is a variable name.

readme/Configuration.scalatex
```diff
@@ -293,3 +293,14 @@
       function(
         argument1,
         argument2)
+  @sect{--lineEndings}
+    Default: @b(preserve)
```
olafurpg: `@b{preserve}` will do the trick ;)

mmatloka:
I have added the flag. Included in tests and docs, however I was not able to generate readme from scalatex locally, I don't know why yet (``` found   : scalatex.site.Section.Proxy
[error]  required: scalatags.Text.all.Frag``` in not touched areas), so I have not checked how does "look like" the added configuration section. Please feel free to adjust the docs descriptions.

readme/Configuration.scalatex
mmatloka: thanks :)

olafurpg:
LGTM. Thanks a lot for this contribution 😄  I expect to release very soon, the wifi is pretty bad at home so I may wait until tomorrow when I go to work.

mmatloka:
Thanks :)

olafurpg:
I am afraid the 0.3.2 release is taking longer than I expected. Is that OK? I have some exciting features in the pipeline that are almost complete and I want to bundle everything in a 0.4.

PR #447, opened 2016-09-17
Semantics preserving rewrites before formatting.

### LINKED ISSUE #304: Enable sorting imports within braces
Original:
```scala
import foo.{Zilch, bar, Random, sand}
```

```scala
import foo.{bar, sand, Random, Zilch}
```

----

While scalafmt clearly cannot reorder general import statements due to hierarchical import semantics, those semantics do not apply within braces as shown, so reordering is valid.

I can see two valid rules here:

- Alphabetical, with lower-case grouped
- Strictly alphabetical

I like the first option, and this seems to be relatively uncontroversial, but I can imagine some people prefer the second.  The first option is demonstrated above.  The second option would yield the following:

```scala
import foo.{bar, Random, sand, Zilch}
```

### LINKED ISSUE #432: feature request: remove unneeded `{ }`
I'd love to see code like:

```scala
def foo: Int = {
  List(1, 2, 3).sum
}
```
be rewritten into
```scala
def foo: Int = List(1, 2, 3).sum
```

So, I want an option to only keep `= { }` if they are required (not a single expression).

### LINKED ISSUE #446: Rewrite rule: add final to case class
Context: https://twitter.com/lambdista/status/769112213072277504

```scala
case class Foo(i: Int)
```
```scala
final case class Foo(i: Int)
```

cc/ @lambdista

olafurpg:
This commit adds a new feature to scalafmt so that the input is
rewritten before formatting.

Rewrite rules
-------------
 The commit implements these rewrite rules:

- Sort imports within curly braces, fixes #304
- Strip away curlies for single statement blocks in defs, fixes #432

See this diff on Kafka from running sort imports and strip away curlies:
https://github.com/olafurpg/scala-repos/pull/7/files

Performance
-----------
This commit does not implement any specific optimizations, the main
concern is to have a proof-of-concept. I haven't benchmarked this implementation yet.
My wild guess is that this new features adds 1-10ms overhead to the
total running time depending on the size of the source file.

Caveats
-------

There are some funky corner cases where the current implementation may
perform differently from user expectation. See "caveats" sections in
tests in org.scalafmt.rewrite package.

Configuration
-------------
```
rewrite.rules = [SortImports, RedundantBraces]
```

### DIFF
--- a/.scalafmt
+++ b/.scalafmt
@@ -1,4 +1,3 @@
 --maxColumn 80
 --alignTokens %;Infix,%%;Infix
 --assumeStandardLibraryStripMargin true
-
--- a/build.sbt
+++ b/build.sbt
@@ -164,9 +164,18 @@ lazy val macros = project
     )
   )
 
+lazy val cliJvmOptions = Seq(
+  "-Xss4m"
+)
+
 lazy val cli = project
   .settings(allSettings)
   .settings(
+    packSettings,
+    packMain := Map("scalafmt_pack" -> "org.scalafmt.cli.Cli"),
+    packJvmOpts := Map(
+      "scalafmt_pack" -> cliJvmOptions
+    ),
     moduleName := "scalafmt-cli",
     mainClass in assembly := Some("org.scalafmt.cli.Cli"),
     libraryDependencies ++= Seq(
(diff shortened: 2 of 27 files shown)


olafurpg:
After [some feedback](https://twitter.com/olafurpg/status/777273182919294976) I've decided to drop the "add final to case class" rewrite.

Issue #424, opened 2016-09-07
--spacesInImportCurlyBraces true will add spaces everywhere


schmitch:
This template isn't a strict requirement to open issues, but please try to provide as much information as possible.

**Version**: 0.3.1
**Integration**: intellij, sbt, cli
**Configuration**: --spacesInImportCurlyBraces true

```scala
import play.api.libs.json.{JsObject, JsString, Json}

object Hase {

    def name(xs: OtherObject): String = {
        s"Hallo ${xs.name}"
    }
    
}
```

Scalafmt:
```scala
import play.api.libs.json.{ JsObject, JsString, Json }

object Hase {

    def name(xs: OtherObject): String = {
        s"Hallo ${ xs.name }"
    }
    
}
```

I would like the input to look like this:
```scala
import play.api.libs.json.{ JsObject, JsString, Json }

object Hase {

    def name(xs: OtherObject): String = {
        s"Hallo ${xs.name}"
    }
    
}
```

olafurpg:
My guess is that this is caused by a small bug in spacesInImportCurlyBraces. I don't personally use this flag so it would be great if someone who does use the flag is willing to take a shot at this.

From my comment on gitter:
> the tutorial explains how to trace that space https://github.com/olafurpg/scalafmt#tutorial-add-spacesinparentheses-flag
> that'll pinpoint a line number in Router.scala where a Space is returned instead of a NoSplit

olafurpg:
Fixed in #454

PR #454, opened 2016-09-20
Distinction between Import braces and string interpolation braces #424

hasumedic:
This PR is a proposed fix for #424.

An important note would be that the proposed solution splits the "left brace" case into two to help distinguish between an _import_ brace and a _string interpolation_ one. The existing code contemplated both cases within the same block, creating multiple "inner" boolean checks to distinguish between the two cases. Keeping them separate allows for more fine tuning in the future and IMHO, simpler code to reason about due to less branching logic.

I'm not quite sure if this is the right way to solve it, please let me know (tests in particular!)

### DIFF
--- a/core/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/core/src/main/scala/org/scalafmt/internal/Router.scala
@@ -109,31 +109,36 @@ class Router(formatOps: FormatOps) {
         Seq(
           Split(NoSplit, 0)
         )
+      // Import left brace
       case FormatToken(open @ LeftBrace(), _, _)
-          if parents(leftOwner).exists(_.is[Import]) ||
-            leftOwner.is[Term.Interpolate] =>
-        val isInterpolate = leftOwner.is[Term.Interpolate]
-        val policy =
-          if (isInterpolate) NoPolicy
-          else SingleLineBlock(matchingParentheses(hash(open)))
+          if parents(leftOwner).exists(_.is[Import]) =>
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -108,31 +108,36 @@ class Router(formatOps: FormatOps) {
...
+      // Interpolated string left brace
+      case FormatToken(open @ LeftBrace(), _, _)
+          if leftOwner.is[Term.Interpolate] =>
+        Seq(
+          Split(NoSplit, 0)
+        )
       case FormatToken(_, close @ RightBrace(), _)
           if parents(rightOwner).exists(_.is[Import]) ||
             rightOwner.is[Term.Interpolate] =>
+        val isImport = parents(rightOwner).exists(_.is[Import])
```
olafurpg: Could you do change into `isInterpolate = righOwner.is[...]`? The `parents` method is really inefficient.

olafurpg:
Hey @hasumedic, any update? I would love to include this fix in 0.4.3, which I hope to release this coming week :)

hasumedic:
Sorry @olafurpg, had a bit rough couple of weeks. Will try to get this done for 0.4.3 ;)

olafurpg:
@hasumedic No stress, take your time.

olafurpg:
LGMT 👍 Thanks a lot @hasumedic, I'm sure this fix will make many users happy!

hasumedic:
👍

PR #483, opened 2016-10-01
New rewrite rule for replacing parens with braces in for expressions

caoilte:
- Only applied on expressions with more than one generator
- Also replaces semi-colons with newlines
- Works on ForYield and For

As suggested in https://github.com/olafurpg/scalafmt/issues/470


olafurpg:
Here's the diff from running the rewrite on Term.For:
https://github.com/olafurpg/scala-repos/pull/9/commits/d3f8da790b51f8fc34f4bae5d2c863bd3d3d885e

I'm OK with rewriting Term.For by default. If someone would like more control we can later introduce a flag.

core/src/main/scala/org/scalafmt/rewrite/ReplaceForParensWithBraces.scala
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    builder.result()
+  }
+
+  override def rewrite(code: Tree, ctx: RewriteCtx): Seq[Patch] = {
+    val builder = Seq.newBuilder[Patch]
+
+    code.collect {
+      case fy: Term.ForYield if fy.enums.size > 1 =>
```
olafurpg: How about `fy.enums.count(_.is[Enumerator.Generator]) > 1`?

I think this case should maybe not be rewritten.
```diff
-    for (sup <- direct if !res.contains(sup)) res ++= sup.getSupers
+    for {
+      sup <- direct if !res.contains(sup)
+    } res ++= sup.getSupers
```

caoilte: Oops! That was me misunderstanding the shape of the ADT! Will put that fix in with a test.

core/src/main/scala/org/scalafmt/rewrite/ReplaceForParensWithBraces.scala
```diff
@@ -0,0 +1,92 @@
...
+  *   } yield (a, b)
+  *
+  */
+object ReplaceForParensWithBraces extends Rewrite {
+
+  def findForParens(forTokens: Tokens, ctx: RewriteCtx): Option[(Token, Token)] = {
+    import ctx.tokenTraverser._
+
+    @tailrec
+    def nextValidLeftParen(token: Token):Option[Token] = {
```
caoilte: That's a great suggestion. The only change I'd like to make is on the return type. I think `Option[Token]` would be more idiomatic.

``` scala
def find(start: Token)(f: Token => Boolean): Option[Token]
def reverseFind(start: Token)(f: Token => Boolean): Option[Token]
```

core/src/main/scala/org/scalafmt/rewrite/ReplaceForParensWithBraces.scala
```diff
@@ -0,0 +1,92 @@
...
+  *
+  * becomes,
+  *
+  *   for {
+  *     a <- as
+  *     b <- bs if b > 2
+  *   } yield (a, b)
+  *
+  */
+object ReplaceForParensWithBraces extends Rewrite {
```
caoilte: I very much agree!

core/src/main/scala/org/scalafmt/rewrite/ReplaceForParensWithBraces.scala
olafurpg: Right, of course, I meant `Token`!

core/src/main/scala/org/scalafmt/rewrite/PreferCurlyFors.scala
```diff
@@ -22,42 +21,29 @@ import scala.meta.{Tree, _}
...
-
-    forEnumerators.flatMap { enumerator =>
-      enumerator.tokens.headOption.flatMap(prevSemiColon).toList
-    }
+    for {
+      enumerator <- forEnumerators
+      token <- enumerator.tokens.headOption.toIterable
+      semicolon <- reverseFind(token)(_.isNot[Whitespace]).filter(_.is[Token.Semicolon]).toIterable
+    } yield {
+      println(semicolon)
```
olafurpg: Remove

core/src/test/resources/rewrite/PreferCurlyForYields.stat
```diff
@@ -26,7 +26,19 @@ object a {
     for (a <- as) yield a
   }
 }
-<<< replaces parens with braces and removes semi-colons on multi line multi generator for yield expression
+<<< doesn't change single generator for-yield expression with guard that uses parens
+object a {
+  def x(as: List[Int]): List[Int] = {
```
olafurpg: Remove

core/src/test/resources/rewrite/PreferCurlyFors.stat
```diff
@@ -34,6 +34,18 @@ object a {
     }
   }
 }
+<<< doesn't change single generator for expression with guard that uses parens
+object a {
+  def x(as: List[Int]): List[Int] = {
```
olafurpg: Remove

core/src/test/resources/rewrite/PreferCurlyFors.stat
```diff
@@ -34,6 +34,18 @@ object a {
...
+    for (a <- as if a < 5) yield a
+  }
+}
+>>>
+object a {
+  def x(as: List[Int]): List[Int] = {
```
olafurpg: Remove

core/src/main/scala/org/scalafmt/rewrite/PreferCurlyFors.scala
olafurpg: Remove println + redundant enclosing curly braces.

core/src/test/resources/rewrite/PreferCurlyForYields.stat
```diff
@@ -0,0 +1,74 @@
+rewrite.rules = [PreferCurlyFors]
+<<< replaces parens with braces and removes semi-colons on single line multi generator for-yield expression
+object a {
+  def x(as: List[Int], bs: List[Int]): List[(Int, Int)] = {
```
olafurpg: Could you remove this def in all of the tests? I think it's enough to only test

```scala
object a {
  for (...) yield x
}
```

olafurpg:
I'm slightly confused by this new review system. There's only two comments on removing defs and a println statement. Then I believe we can go ahead and merge :D

core/src/test/resources/rewrite/PreferCurlyForYields.stat
caoilte: Will do

core/src/main/scala/org/scalafmt/rewrite/PreferCurlyFors.scala
caoilte: *face palm*

caoilte:
Pushed. I think.

olafurpg:
Excellent, thank you for putting this together!

I compared the output from the new version with the older ReplaceCurlyXXX implemenatation and the only diffs were related to for comprehensions with a single generator and an `if` guard. We may later choose to add a configuration option that rewrites those as well. However, I prefer we start conservatively.

LGTM 🎉

caoilte:
Woo! That was fun. Thanks for guiding me through it. I'll think I'll look for another issue to work on this week!

olafurpg:
>  I'll think I'll look for another issue to work on this week

Awesome! Don't hesitate to ask me if there's anything.

PR #489, opened 2016-10-03
Produce clearer error messages for config typos in tests.

### LINKED ISSUE #488: Confusing FormatTest error message when stat file contains invalid config
I saw a unit test error message while working on [470](https://github.com/olafurpg/scalafmt/issues/470) that was hard to understand. 

It turns out the error message is being thrown away just like in [478](https://github.com/olafurpg/scalafmt/issues/478). 

The specific problem is we don't handle the error here,

https://github.com/olafurpg/scalafmt/blob/7c765ac28bf0ad2f8f7c7e902635a07554dda8bd/core/src/test/scala/org/scalafmt/util/HasTests.scala#L90

and so the code continues until

https://github.com/olafurpg/scalafmt/blob/7c765ac28bf0ad2f8f7c7e902635a07554dda8bd/core/src/test/scala/org/scalafmt/util/HasTests.scala#L195

at which point all context about the error (except which directory it is in - which is very confusing) has been lost.

-----

To reproduce this error take an existing stat file and break the config, eg I had ```rewrite.rules = [ReplaceForParensWithBraces]``` (which no longer exists).

A simple `case Left(e)` on line 90 would handle this case. I'd be happy to submit a small PR with the change if that all sounds okay in principle.

olafurpg:
Fixes #488. Example error message:

```
[error] Could not run test org.scalafmt.FormatTests:
org.scalafmt.Error$UnableToFindStyle: Unable to find style for file
rewrite/RedundantBraces.stat. metaconfig.ConfigErrors: Errors:
java.lang.IllegalArgumentException: Unknown input 'RedndantBraces'.
Expected one of RedundantBraces, SortImports
```


caoilte:
Ah yes. Much better than my suggestion! ... As long as you never expect people to add config for directories fringed in spec2style I guess.

caoilte:
Should be found not fringed (phone fail)

olafurpg:
> As long as you never expect people to add config for directories fringed in spec2style I guess.

Correct. I opened #490 to remind me about that.

PR #491, opened 2016-10-03
Port macros to inline meta

### LINKED ISSUE #467: Migrate scala.reflect macros to scala.meta inline/meta macros
The CLI uses two scala.reflect macros to get the compile time git hash and timestamp. That information is very useful to have when debugging user issues. See macro implementations:
https://github.com/olafurpg/scalafmt/blob/master/macros/src/main/scala/org/scalafmt/macros/Macros.scala#L10-L26  

The [`@ConfigReader`](https://github.com/olafurpg/scalafmt/blob/master/metaconfig/src/main/scala/metaconfig/ConfigReader.scala) macro uses the simpler, dotty-ready, more amazing [inline/meta](https://github.com/scala/scala.github.com/pull/567) macro system based on scala.meta.

It would be great to migrate the `gitCommit` and `buildTimeMs` scala.reflect macros to inline/meta macro annotations. It could work like this:

```
object Macros {
  @BuildTime
  def buildTimeMs: Long = ???

  @GitCommit
  def gitCommit: Option[String] = ???
}
```

where the macro replaces the `???` part with the actual implementation. We can put these new macros into the `metaconfig` project in the `org.scalafmt.util` package.

cb372:
* Rewrite the "build time" and "git commit" macros as inline/meta
* Delete the unneeded `macros` project

Fixes #467


metaconfig/src/main/scala/org/scalafmt/util/BuildTime.scala
```diff
@@ -0,0 +1,18 @@
+package org.scalafmt.util
+
+import scala.annotation.compileTimeOnly
+import scala.meta._
+
+@compileTimeOnly("Gets expanded away")
```
olafurpg: This message appears when users forget to add the paradise compiler plugin. A clearer error might be:

"@BuildTime not expanded. Have you enabled the scala.meta paradise plugin?" 

Same for GitCommit.

metaconfig/src/main/scala/org/scalafmt/util/BuildTime.scala
```diff
@@ -0,0 +1,18 @@
...
+class BuildTime extends scala.annotation.StaticAnnotation {
+
+  inline def apply(defn: Any): Any = meta {
+    defn match {
+      case defn @ Defn.Val(mods, pats, optType, rhs) if pats.size == 1 =>
```
olafurpg: I think this might be clearer with a quasiquote, something  in this direction:

```scala
case q"val $name = ???" => q"val $name = ${System.currentTimeMillis()}"
```

metaconfig/src/main/scala/org/scalafmt/util/BuildTime.scala
cb372: Oh, so that's what that annotation is for. I'll admit I didn't really understand it.

Fixed.

metaconfig/src/main/scala/org/scalafmt/util/BuildTime.scala
cb372: I originally tried to wrote it with quasiquotes, but I was tripped up by the fact that you have to explicitly type the val name as `${name: Pat.Var.Term}` (otherwise it gets typed as `Pat`, which you can't use when you come to construct the new tree).

Got it working the second time around. I agree that this is more readable.

cb372:
@olafurpg Thanks for the review. I've made some fixes and rebased.

olafurpg:
LGTM 👍

We're still in the very early days of scala.meta macros. In this particular case, macro annotations are a bit awkward because of the `= ???` part. Def macros are on the roadmap to fix that :) 

Thanks!

PR #517, opened 2016-10-12
Add AvoidInfix rewrite rule.

olafurpg:
The AvoidInfix rule rewrites unconventional infix operators into regular term
application. For example

``` diff
-          promise success "done"
+          promise.success("done")
```

The configuration is like this

```
rewrite.rules = [AvoidInfix]
rewrite.neverInfix.includeFilters = ["\\w+"]
rewrite.neverInfix.excludeFilters = ["until", "to"]
```

The defaults exclude symbolic operators and regulars like `until` and `should` (from scalatest).

This implementation ignores nested infix operators, for example

``` scala
an[Exception] should be thrownBy foo
```

Diff from running AvoidInfix on spark and kafka:
  https://github.com/olafurpg/scala-repos/pull/11/files

### DIFF
--- a/core/src/main/scala/org/scalafmt/config/Pattern.scala
+++ b/core/src/main/scala/org/scalafmt/config/Pattern.scala
@@ -17,7 +17,7 @@ case class Pattern(
 
 object Pattern {
   val neverInfix = Pattern(
-    Seq("\\w+"),
+    Seq("[\\w\\d_]+"),
     Seq(
       "until",
       "to",
@@ -25,6 +25,7 @@ object Pattern {
       "eq",
       "ne",
       "should.*",
+      "contain.*",
       "be",
       "synchronized",
       "have"
--- a/core/src/main/scala/org/scalafmt/config/Pattern.scala
+++ b/core/src/main/scala/org/scalafmt/config/Pattern.scala
@@ -1,0 +1,33 @@
+package org.scalafmt.config
+
+import scala.util.matching.Regex
+
+import metaconfig.ConfigReader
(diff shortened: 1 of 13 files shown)


olafurpg:
@bdarfler Did you have something like this in mind?

bdarfler:
Um waoh! This is awesome!

core/src/main/scala/org/scalafmt/config/Pattern.scala
```diff
@@ -0,0 +1,33 @@
...
+    excludeFilters: Seq[String]
+) {
+  def toMatcher: Matcher =
+    new Matcher(RegexOps.mkRegexp(includeFilters),
+                RegexOps.mkRegexp(excludeFilters))
+}
+
+object Pattern {
+  val neverInfix = Pattern(
+    Seq("\\w+"),
```
sjrd: I suggest `"[\\w\\d_]+"`.

olafurpg: Good catch! Updated.

olafurpg:
FWIW, I'm waiting with this PR until we have a more robust `Patch` in scala.meta/scalafix (it's a wip, probably won't be ready until january) that can compose patches when possible and gracefully handle patch conflicts.

dwijnand:
`org.scalafmt.rewrite.EmbraceInfix`

olafurpg:
I'm gonna close this PR for now since it's been open for several months. I still really want to get this rewrite in!

PR #522, opened 2016-10-14
Use typelevel/scala in SBT plugin, fixes #485.

### LINKED ISSUE #485: Cross-version issue with Scala 2.12.0-RC1
```
[info] Updating {file:/Users/heiko/projects/scala/new-in-scala-212/}new-in-scala-212...
[warn] Binary version (2.11) for dependency org.scala-lang#scala-library;2.11.8
[warn]  in de.heikoseeberger#new-in-scala-212_2.12.0-RC1;f023808b8f857706a6c3d4d3a633496683cf471b-SNAPSHOT differs from Scala binary version in project (2.12.0-RC1).
[info] Resolving com.github.scopt#scopt_2.11;3.3.0 ...
[info] Done updating.
[error] Modules were resolved with conflicting cross-version suffixes in {file:/Users/heiko/projects/scala/new-in-scala-212/}new-in-scala-212:
[error]    org.scala-lang.modules:scala-xml _2.11, _2.12.0-RC1
[trace] Stack trace suppressed: run last *:update for the full output.
[error] (*:update) Conflicting cross-version suffixes in: org.scala-lang.modules:scala-xml
[error] Total time: 0 s, completed Oct 2, 2016 3:39:12 PM
```

olafurpg:
- SBT plugins must build against 2.10
- scalafmt is 2.11 only because scala.meta makes heavy usage of macro
  annotations.
- To solve this problem so far, the scalafmt SBT plugin classloads itself with
  scala-library 2.11.8 in a hidden configuration (see
    http://www.scala-sbt.org/0.13/docs/Library-Management.html#ivy-configurations)
- This solution is problematic because in 2.12 SBT builds the scala-library:2.11.8
  dependency is evicted in favor of 2.12, causing issues like #485. Why? It appears SBT special handles
  org.scala-lang:scala-library for some reason.
- This commit changes the SBT plugin to use
org.typelevel:scala-library instead of org.scala-lang:scala-library.

This solution may not work on a project that use
org.typeleve:scala-library:2.12.  Hopefully everyone will have switched to SBT 1.0
by then.

### DIFF
--- a/core/src/main/scala/org/scalafmt/Versions.scala
+++ b/core/src/main/scala/org/scalafmt/Versions.scala
@@ -5,7 +5,7 @@ package org.scalafmt
   */
 object Versions {
   // Nightly, used in CLI, build.sbt, etc.
-  val nightly = "0.4.6"
+  val nightly = "0.4.7-SNAPSHOT"
   // Stable, used in official user docs.
   val stable = "0.4.6"
   val scala = "2.11.8"
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=0.13.11
+sbt.version=0.13.12
--- a/scalafmtSbt/src/main/scala/org/scalafmt/sbt/ScalaFmtPlugin.scala
+++ b/scalafmtSbt/src/main/scala/org/scalafmt/sbt/ScalaFmtPlugin.scala
@@ -88,10 +88,23 @@ object ScalaFmtPlugin extends AutoPlugin {
     List(
       ivyConfigurations += config("scalafmt").hide,
       libraryDependencies ++= Seq(
-        // scala-library needs to be explicitly added to fix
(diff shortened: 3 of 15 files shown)


milessabin:
Can't you use macro-paradise for 2.10.x and [macro-compat](https://github.com/milessabin/macro-compat)? Mixing multiple versions of Scala libraries seems like a recipe for disaster.

olafurpg:
I agree that would be an ideal solution. Unfortunately, attempts to port scala.meta (which scalafmt depends on) to 2.10 have so far been unsuccessful, see https://github.com/scalameta/scalameta/issues/295.

> Mixing multiple versions of Scala libraries seems like a recipe for disaster.

I agree. I would love to find a better solution to this problem.

olafurpg:
I checked the update report from using org.typelevel:scala-library and there are indeed two versions of scala-library on the classpath. Thanks for the help @milessabin  : )

PR #534, opened 2016-10-20
Format on save and update the notification system

joan38:
Solving https://github.com/olafurpg/scalafmt/issues/530

settings

But also moving to the better notification system:
balloon

event log

Cheers


intellij/resources/META-INF/plugin.xml
```diff
@@ -21,7 +21,7 @@
   </change-notes>
 
   <!-- please see http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/build_number_ranges.html for description -->
-  <idea-version since-build="141.0"/>
+  <idea-version since-build="162"/>
```
olafurpg: Any idea how many people still use <2016.2? Unfortunately, I can't tell from the download numbers on plugins.jetbrains.com

olafurpg:
PS. I appreciate you cleaned up the project structure, long overdue.

intellij/resources/META-INF/plugin.xml
joan38: Actually I though it was the version it's been built with.
I will revert that then.

olafurpg:
I can't promise when the next release will be, but for the impatient here's a zip file for manual installation: https://github.com/olafurpg/scalafmt/releases/download/v0.4.7/intellij-4.7-autosave.zip

rcavalcanti:
I just tested the zip and indeed works very well.

However, it seems that it enable the format on save behavior to all projects. It's a IDE settings instead of project setting. 

I'm not using scalafmt all over the place, although I plan to do it, I can pretty much imagine that others won't be able to use scalafmt everywhere. 

Having to enable, disable the plugin on a per project base is very annoying and error prone.

olafurpg:
Great observation @rcavalcanti 

One thing we could do is enable format on save only if the setting is enabled globally AND there is a .scalafmt.conf file in the root directory of the current project. How would that work?

joan38:
Or we can have 2 checkboxes, one for project level and the other for global.
And we format if one of the 2 is enabled.

olafurpg:
That sounds good as well. Would you mind taking a swing at this @joan38 ?

joan38:
I can take a java.swing at it over the WE yeah.

lunaryorn:
@joan38 Thanks so much for this change… I feel that I'll be breaking our CI a lot less now 👏 🎆

Issue #349, opened 2016-06-28
CLI: non-zero exit code for formatting failure


olafurpg:
The behavior of the CLI should be
- warning is printed out
- exit code is non-zero
- no stack trace (less noise, easier for text editor integrations to consume and present a helpful message to users)

when
- formatting failed, and
- the CLI is formatting only a single file, for example via `--stdin` or `-f foo.scala` (not `-f <directory>`)

The current behavior is a bit annoying if you really care about knowing if something went wrong

``` scala
$ cat foo.scala
object a {
$ cat foo.scala | scalafmt --stdin
[warn] Error in stdin.scala: <input>:1: error: } expected but end of file found
object a {
          ^
// exit code 0
```

If we want to be fancy, we could reuse [rustfmt](https://github.com/rust-lang-nursery/rustfmt/blob/master/README.md#running) conventions for exit codes:

> If `rustfmt` successfully reformatted the code it will exit with `0` exit
> status. Exit status `1` signals some unexpected error, like an unknown option or
> a failure to read a file. Exit status `2` is returned if there are syntax errors
> in the input files. `rustfmt` can't format syntatically invalid code. Finally,
> exit status `3` is returned if there are some issues which can't be resolved
> automatically. For example, if you have a very long comment line `rustfmt`
> doesn't split it. Instead it prints a warning and exits with `3`.

One possible way to achieve this is to:
- set `runner.fatalWarnings = tre` when `inputMethods.length == 1`
- catch the exception somewhere higher,
- print the error message to `common.err`,
- return appropriate int exit code (for example, change return type to Int instead of Unit here: https://github.com/olafurpg/scalafmt/blob/7ea32fd4141bfcd17acc9bb3cffaa48fc5af8e03/cli/src/main/scala/org/scalafmt/cli/Cli.scala#L47)
- `System.exit(returnCode)` somewhere at the end of the universe, where the tests don't run

siddarth:
`--test` allegedly does this, but i don't think it works?

olafurpg:
I updated the description to better reflect the current goals of this issue. `--test` is meant to run in CI, which is a bit different from what I had in mind for this ticket.

PR #540, opened 2016-10-21
cli: Exit with non-zero status if in testing mode.

siddarth:
Fixes https://github.com/olafurpg/scalafmt/issues/349

Did it the (admittedly lazy) way. Happy to capture all failures and raise them at the end if that's preferred.

### DIFF
--- a/cli/src/main/scala/org/scalafmt/cli/Cli.scala
+++ b/cli/src/main/scala/org/scalafmt/cli/Cli.scala
@@ -123,7 +123,7 @@ object Cli {
       case Formatted.Success(formatted) =>
         inputMethod.write(formatted, input, options)
       case Formatted.Failure(e) =>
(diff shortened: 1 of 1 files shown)


olafurpg:
`--test` is meant for running in CI to throw an error if scalafmt finds a mis-formatted file, #349 was meant for a different scenario.

The CLI has gone through a few changes since I opened #349, so I've now added more up-to-date details to the issue description. The issue is still very relevant. It may be more than what you had in mind, but I think it's totally doable and it would greatly improve quality-of-life for users of the CLI. WDYT? :)

siddarth:
@olafurpg: this seems super reasonable. I'll take a pass at this.

olafurpg:
Closing. Don't hesitate to re-open @siddarth if you find time/motivation to work on this! : )

Issue #399, opened 2016-08-10
Automate IntelliJ plugin update


olafurpg:
The publish step for the plugin is too manual so I don't bother for tiny releases, causing #398. The Jetbrains guys showed me at ScalaDays how they build and upload the Scala plugin automatically. It would be great to do something similar for scalafmt.

olafurpg:
Concretely, I want to make the `intellij` directory become part of the main SBT build by using this SBT plugin: https://github.com/dancingrobot84/sbt-idea-plugin

My current workflow for the intellij plugin is this:

- cp fat jar into intellij/lib
- open separate intellij project from `intellij` repo
- import `intellij.iml`
- Build > Prepare plugin for deployment
- Manually upload zip to plugins.jetbrains.com

I would like the workflow to be instead:
- `sbt publishPlugin`

Another great benefit to this change is that it becomes a lot easier for contributors to make changes to the intellij plugin, like in #544.

Any help from someone with SBT experience would be awesome.

rbellamy:
I haven't tried the `sbt publishPlugin` since I don't have the permissions to try that, and am a bit unclear how one would go about testing it without, well, actually publishing the plugin, but it shouldn't be a great stretch to go from PR#655 to publishing.

olafurpg:
Even though it seems I still need to manually upload the zip file to plugins.jetbrains.com, I'm happy to close this issue since most of the laborious steps needed to build the intellij plugin have now been automated thanks to #655. Thank you @rbellamy!

PR #544, opened 2016-10-22
Move Settings at project level

joan38:
I just moved the app setting at project level.
So if you want to set it globally you can set it for all your current projects in `Preferences...` and for all future projects in `File` > `Other settings` > `Default Settings...`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -109,10 +109,6 @@ lazy val core = project
   .settings(
     metaMacroSettings,
     moduleName := "scalafmt-core",
-    test in assembly := {
-      // TODO(olafur) This should be an integration test.
-      (runMain in Test).toTask(" org.scalafmt.FormatExperimentApp").value
-    },
     libraryDependencies ++= Seq(
       "com.lihaoyi"    %% "sourcecode"   % "0.1.2",
       "org.scalameta"  %% "scalameta"    % Deps.scalameta,
@@ -153,14 +149,6 @@ lazy val cli = project
     )
   )
   .dependsOn(core % "compile->compile;test->test")
-
-lazy val scalafmtIntellij = project
-  .settings(
-    allSettings,
-    noPublish,
-    moduleName := "scalafmt-intellij"
-  )
-  .enablePlugins(SbtIdeaPlugin)
 
 lazy val scalafmtSbt = project
   .settings(allSettings)
--- a/intellij/intellij.iml
+++ b/intellij/intellij.iml
@@ -23,5 +23,6 @@
     </orderEntry>
     <orderEntry type="library" name="lib" level="project" />
     <orderEntry type="library" name="scala-sdk-2.11.8" level="project" />
+    <orderEntry type="library" name="org.scala-lang.modules:scala-xml_2.11:1.0.6" level="project" />
   </component>
 </module>
--- a/intellij/resources/META-INF/plugin.xml
+++ b/intellij/resources/META-INF/plugin.xml
@@ -30,11 +30,11 @@
   -->
 
   <extensions defaultExtensionNs="com.intellij">
-    <applicationConfigurable groupId="tools"
+    <projectConfigurable groupId="tools"
                              displayName="Scalafmt"
                              id="preferences.Scalafmt"
                              instance="org.scalafmt.intellij.SettingsGui" />
-    <applicationService serviceImplementation="org.scalafmt.intellij.Settings"/>
+    <projectService serviceImplementation="org.scalafmt.intellij.Settings"/>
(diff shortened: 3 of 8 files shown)


olafurpg:
I tried this out. The settings menu looks great. However, it appears that format runs now randomly during editing, before I press save. Here is a gif where I press enter non-stop

save

I did not experience this with the previous PR. Any idea what might be the cause?

joan38:
I can't reproduce the issue.

On a separate point. I'm trying to resolve an issue when the formatter can't format it will display a ballon with this message:
```
Parse error: <input>:5: error: } expected but end of file found
```
Which is displayed like this:
screen shot 2016-10-23 at 14 01 53

Is order to resolve this issue, I need the lib `scala-xml`, do you know how I can add this lib? Do we use SBT for this project? I don't understand how the project `scalafmtIntellij` is working.

Cheers

joan38:
This new commit fixes the HTML (see below pic), I had to import `scala-xml.jar` manually in Idea to make it work so not sure how to do with SBT.
screen shot 2016-10-23 at 14 12 28

olafurpg:
> I don't understand how the project scalafmtIntellij is working.

I'm sorry about that, the scalafmtIntellij project is not used so it shouldn't even be part of the build. I don't know why it even got merged. I've been planning to move to https://github.com/dancingrobot84/sbt-idea-plugin for a while now but never got it working. For now, I just `sbt cli/assembly` and manually copy paste the fat jar to `intellij/lib` 

>  I had to import scala-xml.jar manually in Idea to make it work so not sure how to do with SBT.

Would it be possible to escape the angle brackets? That message should not be rendered as html. In either case, as long as the dependency is defined in `intellij.iml` then my idea project will pick it up.

> I can't reproduce the issue.

Let me give it another try.

joan38:
> Would it be possible to escape the angle brackets? That message should not be rendered as html. In either case, as long as the dependency is defined in intellij.iml then my idea project will pick it up.
Is it ok then now with my last commit?

olafurpg:
The project doesn't build because of the missing scala-xml library.

joan38:
Which line number is the best to add this dep in `build.sbt`?

olafurpg:
I'm investigating if we can use some package in com.intellij to escape xml strings, adding scala-xml seems overkill

olafurpg:
Okay, I couldn't quickly find any Xml escape method in com.intellij.xml. We can use scala-xml, I think the dependency needs to be listed in `intellij.iml`. It should automatically add to `intellij.iml` when you add the dependency in the project settings.

olafurpg:
Something like this


```xml
    <orderEntry type="library" name="org.scala-lang.modules:scala-xml_2.11:1.0.6" level="project" />
```

joan38:
But I don't see where the scalafmt lib is described there?

olafurpg:
Scalafmt is not described there, it's just an unmanaged dependency. I copy paste the fat jar in to `lib`. I admit this is terribly undocumented.

 #399 is the plan to move the intellij project into the main SBT build, I just haven't found the time get it done.

joan38:
Ok I can do what you mentioned, and that can be then moved at SBT level later.

joan38:
Done. Have you resolved your issue with random reformatting?

olafurpg:
That sounds like a good plan. Could you also remove `scalafmtIntellij` from build.sbt, to avoid further confusion?

olafurpg:
I am unable to reproduce the "saves randomly, not only on save" error once I build the plugin zip and install manually. So this PR is good to go as soon as scalafmtIntellij is out from build.sbt

joan38:
Sure, I will also do https://github.com/olafurpg/scalafmt/pull/548#discussion_r84596257

olafurpg:
Awesome! Thanks a lot :)

joan38:
No worries, I will have a look to https://github.com/olafurpg/scalafmt/issues/518 later next week.

joan38:
@olafurpg Did we updated the plugin zip file in the release with this change?

olafurpg:
@joan38 No, I had forgotten. I've updated the zip now: https://github.com/olafurpg/scalafmt/releases/download/v0.4.7/intellij-4.7-autosave.zip

I hope to release 0.4.8 soon, once I find the time to write up the changelog.

olafurpg:
Scalafmt is not described there, it's just an unmanaged dependency. I copy
paste the fat jar in to `lib`. I admit this is terribly undocumented, sorry
about that.  #399 is the plan to clean up the intellij project to make the
project a lot easier to work with, I just haven't found the time get it
done.

PR #548, opened 2016-10-23
Use constant instead of strings and remove useless statement in SBT

joan38:


build.sbt
```diff
@@ -110,7 +110,6 @@ lazy val core = project
     metaMacroSettings,
     moduleName := "scalafmt-core",
     test in assembly := {
-      (test in Test).value
```
olafurpg: The task runs, so the SBT warning is wrong, but I ever run root/assembly anymore so I'm OK with dropping this line.

olafurpg:
LGTM 👍 Thanks :)

build.sbt
joan38: Do you want to drop the whole setting then?
```
test in assembly := { ... }
```

olafurpg: Yes, that would be great.

PR #562, opened 2016-10-27
Update IDEA plugin

olafurpg:
Review @joan38


intellij/src/org/scalafmt/intellij/IdeaUtils.scala
```diff
@@ -26,7 +26,7 @@ object Utils {
       new Notification(PluginName,
                        PluginName,
                        Utility.escape(msg),
-                       NotificationType.INFORMATION))
+                       notificationType))
```
olafurpg: It happens to the best of us ^^

olafurpg:
Do you if there is a way to enable balloons by default?

PR #565, opened 2016-10-29
Add info about Arch Linux package

RatanRSur:


readme/Installation.scalatex
```diff
@@ -90,8 +90,13 @@
...
 
     @sect{Linux/OSX}
+    
+      @sect{Arch Linux}
+        @p
+          Install the @code{scalafmt} package from the Arch User Repository. Voting on the package will also increase the likelyhood that the package will be merged into the community repo.
+        
       @p
-        The simplest way to install the CLI in Linux and OSX is to run:
+        Otherwise, the simplest way to install the CLI in Linux and OSX is to run:
```
olafurpg: "To install the CLI in Linux or OSX, run:"

readme/Installation.scalatex
```diff
@@ -90,8 +90,13 @@
     There are several ways to install the CLI.
 
     @sect{Linux/OSX}
+    
+      @sect{Arch Linux}
```
olafurpg: It appears a scalatex section can't immediately start with another section. Could you move this subsection after the `curl` instructions?

RatanRSur:
I actually was in the process of making one as a contribution but someone else beat me by just a few days! I thought the next best thing to do would be to bring it to people's attention.

olafurpg:
LGTM, thanks! I still manually update the website so it may take a while before the change becomes live.

PR #567, opened 2016-10-31
Issue 457 long imports

mtomko:
This is a preliminary PR intended to fix https://github.com/olafurpg/scalafmt/issues/457


core/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -119,7 +119,7 @@ import org.scalafmt.util.ValidationOps
   * @param alignTokens Documented in scalafmt --help page.
   * @param spacesInImportCurlyBraces If true, formats `import a.b.{ c, d }`.
   *                                  If false, formats `import a.b.{c, d}`.
-  * @param binPackImportSelectors If true, breaks up import selectors if
+  * @param importSelectors If true, breaks up import selectors if
```
olafurpg: This section is outdated, maybe move it to `ImportSelectors`?

core/src/test/resources/unit/Import.source
```diff
@@ -82,6 +82,7 @@ import org.{
 <<< break #101 short 2
 import org.{Aaaa, Bbbbb, C, D, E, F, G, H, I, J, K, L}
 >>>
+
```
olafurpg: Is this needed?

core/src/test/resources/importSingleLine/ImportSingleLine.source
```diff
@@ -0,0 +1,87 @@
...
+import foo.bar
+
+import kaz.bar
+object a
+>>>
+import foo.bar
+
+import kaz.bar
+object a
+<<< One per line, with docstring
```
olafurpg: Are these tests necessary for the new flag? I would try to add only tests that check the functionality of the new flag.

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -116,13 +116,22 @@ class Router(formatOps: FormatOps) {
         val policy = SingleLineBlock(matchingParentheses(hash(open)))
         val close = matchingParentheses(hash(open))
         val newlineBeforeClosingCurly = newlineBeforeClosingCurlyPolicy(close)
-        val newlinePolicy =
-          if (style.binPackImportSelectors) newlineBeforeClosingCurly
-          else newlineBeforeClosingCurly.andThen(OneArgOneLineSplit(open))
+
+        val newlinePolicy = style.importSelectors match {
```
olafurpg: Awesome!

core/src/main/scala/org/scalafmt/config/ImportSelectors.scala
```diff
@@ -0,0 +1,17 @@
+package org.scalafmt.config
+
+/**
+  * ADT representing import selectors settings.
+  * TODO(mtomko) provide a brief description of each setting, with code examples of what each setting does
```
olafurpg: Please do : )

olafurpg:
LGTM 👍  Thanks @mtomko 😸

olafurpg:
Wait, I just realised that this is a breaking change because of the removal of the binPackImportSelectors flag. I've wanted to extend metaconfig to support alternative names for fields to more easily accommodate changes like this. Let me try and take a look at that this week :)

core/src/test/resources/importSingleLine/ImportSingleLine.source
```diff
@@ -0,0 +1,22 @@
...
+import org.{Aaaaaa, GlobalScope, Registration}
+<<< break #101 short 2
+import org.{Aaaa, Bbbbb, C, D, E, F, G, H, I, J, K, L}
+>>>
+import org.{Aaaa, Bbbbb, C, D, E, F, G, H, I, J, K, L}
+<<< undo the work of other import formatters
+import org.{
+  Aaaa,
+  Bbbbb,
+  C,
```
olafurpg: Can you add another test like this but with an inline comment `// foobar` somewhere inside the block? I suspect you need to pass `disallowInlineComments = false` to the SingleLinePolicy.

mtomko: Will do.

mtomko:
Good catch on the breaking change - I wasn't thinking of that. I don't think I got far enough into metaconfig to help without a steep learning curve, but I may have some spare time later in the week to take a look.

olafurpg:
Thanks for the offer. I was happy to implement this feature as it's been on my todo for a while now. Once #570 is merged, you can customize the `importSelectorsReader` reader like this

```scala
implicit importSelectorsReader: Reader[ImportsSelector] = Reader.instance[ImportsSelector] {
  case true => ??? // fill in
  case false => ???
  case els => ImportSelectors.reader.read(els)
}
```
Then you need to annotate the member in ScalafmtConfig like this `@metaconfig.ExtraName("binPackImportSelectors") importsSelector`.

You can keep the old tests to see it parses the old settings correctly.

mtomko:
Thanks, I'll update the import selectors reader and get this PR updated later today.

core/src/main/scala/org/scalafmt/config/ImportSelectors.scala
```diff
@@ -45,6 +47,20 @@ object ImportSelectors {
...
     ReaderUtil.oneOf[ImportSelectors](noBinPack, binPack, singleLine)
 
+  // This reader is backwards compatible with the old import selector
+  // configuration, which used the boolean flag binPackImportSelectors to
+  // decide between (what are now) the `binPack` and `noBinPack` strategies.
+  // It is defined here to keep the `Reader.instance[T} {}` lambda in a
+  // separate file from the @ConfigReader macro annotation; this is due to
+  // limitations in the current version of scalameta/paradise, but these will
+  // likely be fixed in the future, at which point this reader could be moved
+  // to ScalafmtConfig
```
olafurpg: Thanks for summarizing this.

core/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -174,7 +174,8 @@ case class ScalafmtConfig(
     newlines: Newlines = Newlines(),
     runner: ScalafmtRunner = ScalafmtRunner.default,
     // Settings which belong to no group
-    importSelectors: ImportSelectors = ImportSelectors.noBinPack,
+    @metaconfig.ExtraName("binPackImportSelectors") importSelectors: ImportSelectors =
+      ImportSelectors.noBinPack,
```
olafurpg: scalafmt is a bit awkward with annotations on parameters, just opened #572.

core/src/test/resources/importSingleLine/ImportSingleLine.source
```diff
@@ -20,3 +20,14 @@ import org.{
...
+<<< handle inline comments
+import org.{
+// what?
+  Aaaa,
+  Bbbbb,
+  C,
+  D
+}
+>>>
+import org.{// what?
```
olafurpg: I think this is totally fine, the user can move the comment elsewhere and re-format.

Issue #533, opened 2016-10-20
Scalafmt does not format sources in shared when using Scala.js crossproject


mdedetrich:
**Version**: 0.4.7
**Integration**: SBT 0.13.12
**Configuration**: 
```
style = defaultWithAlign
maxColumn = 120
```

There seems tobe an issue with scala.js crossprojects (i.e. projects based off https://github.com/scala-js/scalajs-cross-compile-example). If you are using a crossproject and have this in your standard `build.sbt`

```scala
scalafmtConfig in ThisBuild := Some(file(".scalafmt.conf"))
```

Then scalafmt won't actually format any sources in your `shared` folder. It will only format the files in the `jvm`/`js` folders

olafurpg:
Hmmm. This is annoying. I've kind of given up on figuring out how to use SBT settings to decide which files to format. It seems the SBT build landscape is too heterogeneous to make it work consistently well.

My strategy for 0.5 is to extract the project model from VCS and let users define which extra files to exclude/include in `.scalafmt.conf`. There's already git support in the 0.4.7 CLI and we are using that in scalameta/paradise, see [here](https://github.com/scalameta/paradise/blob/master/.drone.yml#L7). So far, it works great, but I would love to replace the bash script with something cleaner.

sjrd:
That sounds strange. scalafmt correctly grabs the list of source directories to process from `unmanagedSourceDirectories in Compile` ([here](https://github.com/olafurpg/scalafmt/blob/master/scalafmtSbt/src/main/scala/org/scalafmt/sbt/ScalaFmtPlugin.scala#L100)), and `CrossProject` precisely adds the shared sources in `unmanagedSourceDirectories in Compile` ([here](https://github.com/scala-js/scala-js/blob/master/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/cross/CrossProject.scala#L316)).

What does the following display in your build:

    > show projectJVM/compile:unmanagedSourceDirectories

and

    > show projectJVM/compile:hasScalafmt::sourceDirectories

?

mdedetrich:
I am actually thinking right now that it may have been an issue with sbt not correctly picking up a change `.scalafmt.conf` and hence not formatting the shared directory (it was a project which only has sources in `shared` and nowhere else) so it may be a false alarm, let me confirm this

mdedetrich:
Okay, false alarm, I did a `clean` in sbt and its now working. It appears that sbt didn't initially pick up a change `.scalafmt`

PR #578, opened 2016-11-05
Remove redundant braces in enumerator guard conditions

ShaneDelmore:
Implements rewrite requested in #533

### DIFF
--- a/core/src/main/scala/org/scalafmt/rewrite/RedundantParens.scala
+++ b/core/src/main/scala/org/scalafmt/rewrite/RedundantParens.scala
@@ -1,0 +1,31 @@
+package org.scalafmt.rewrite
+
+import scala.meta.Tree
+import scala.meta._
+import scala.meta.tokens.Token.{LeftParen, RightParen}
+
+object RedundantParens extends Rewrite {
+
+  def isWrappedInParens(tokens: Tokens): Boolean =
+    tokens.nonEmpty && tokens.head.is[LeftParen] && tokens.last.is[RightParen]
+
+  override def rewrite(code: Tree, ctx: RewriteCtx): Seq[Patch] = {
+    val builder = Seq.newBuilder[Patch]
+
+    code.collect {
+      case g: Enumerator.Guard =>
+        val tokens: Tokens = g.cond.tokens
+        if (isWrappedInParens(tokens)) {
+          val openParensCount = tokens.segmentLength(_.is[LeftParen], 0)
+          val closeParensCount = tokens.reverse.segmentLength(_.is[RightParen], 0)
+          val parensToRemoveCount = Math.min(openParensCount, closeParensCount)
+          val leftSegment: Tokens = tokens.slice(0, parensToRemoveCount)
+          val rightSegment: Tokens = tokens.slice(tokens.length - parensToRemoveCount, tokens.length)
+          builder += Patch(leftSegment.head, leftSegment.last, "")
+          builder += Patch(rightSegment.head, rightSegment.last, "")
+        }
+    }
+
+    builder.result()
+  }
+}
--- a/core/src/test/resources/rewrite/RedundantParens.stat
+++ b/core/src/test/resources/rewrite/RedundantParens.stat
@@ -1,0 +1,33 @@
+rewrite.rules = [RedundantParens]
+<<< basic
+object a {
+  def c(b: List[Int]): List[Int] =
+    for {
+      a <- b
+      if ((a))
+    } yield a
+}
+>>>
+object a {
+  def c(b: List[Int]): List[Int] =
+    for {
+      a <- b
+      if a
+    } yield a
+}
+<<< imbalanced sides
+object a {
+  def c(b: List[Int]): List[Int] =
+    for {
+      a <- b
+      if ((b ++ b).length >= 2)
+    } yield a
+}
+>>>
+object a {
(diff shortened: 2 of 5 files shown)


ShaneDelmore:
Are there settings or documentation that I should add with this PR?

olafurpg:
Thanks for this contribution! Quick answer about docs, please add a section under "rewrites" in Configuration . scalatex, same as for RedundantBraces. For configuration, you've already correctly added the rewrite in Rewrite.scala :)

I'll take a closer look later, on a trip in Zürich for the weekend.

core/src/main/scala/org/scalafmt/rewrite/RedundantParens.scala
```diff
@@ -0,0 +1,25 @@
...
+  def startsWithParen(cond: Term): Boolean =
+    cond.tokens.headOption.map(_.is[LeftParen]).getOrElse(false)
+
+  override def rewrite(code: Tree, ctx: RewriteCtx): Seq[Patch] = {
+    val builder = Seq.newBuilder[Patch]
+
+    code.collect {
+      case g: Enumerator.Guard if startsWithParen(g.cond) =>
+        val openParens = g.cond.tokens.takeWhile(_.is[LeftParen])
+        val closeParens = g.cond.tokens.takeRight(openParens.length)
```
ShaneDelmore: Completely missed that. I'll have to think about this but min(leadingParenCount, trailingParenCount) might be the correct number to remove.

ShaneDelmore: I will add a test of this scenario as well.

ShaneDelmore:
@olafurpg I wasn't sure if you had a preferred style of the solutions so I tried to fit in most with what I currently saw.  FWIW, I created an example of some different ways I might approach the problem if we are planning on adding many more rewrites, especially of wrap/unwrap type rewrites.  I think a few shared enrichments to Tokens might be fairly reusable.

https://gist.github.com/ShaneDelmore/50bee845755d8cf81946ee2fff3efe9e

olafurpg:
@ShaneDelmore That solution looks fine, it's not exactly how I would personally write it but it's totally readable. My only comment would be that the `excise` method makes it a bit unclear that you remove an equal number of parens on each side.

I agree that shared enrichment ops would be nice, feel free to put those somewhere in the `util` package in a `TokenPimps` object or some other name that makes sense. I'm personally a bigger fan of static methods on objects even though it a bit more verbose. Explicit object names make it easy to read code outside an IDE like for example on Github. But feel free to do what makes sense for you.

ShaneDelmore:
Should we just merge in this fix, and then I could create a second proposal pr with the utility functions and old examples converted to use them to see if you prefer that method or not?

core/src/test/resources/rewrite/RedundantParens.stat
```diff
@@ -0,0 +1,33 @@
...
+      a <- b
+      if ((b ++ b).length >= 2)
+    } yield a
+}
+>>>
+object a {
+  def c(b: List[Int]): List[Int] =
+    for {
+      a <- b
+      if (b ++ b).length >= 2
```
olafurpg: Awesome!

core/src/main/scala/org/scalafmt/rewrite/RedundantParens.scala
```diff
@@ -0,0 +1,31 @@
...
+
+    code.collect {
+      case g: Enumerator.Guard =>
+        val tokens: Tokens = g.cond.tokens
+        if (isWrappedInParens(tokens)) {
+          val openParensCount = tokens.segmentLength(_.is[LeftParen], 0)
+          val closeParensCount = tokens.reverse.segmentLength(_.is[RightParen], 0)
+          val parensToRemoveCount = Math.min(openParensCount, closeParensCount)
+          val leftSegment: Tokens = tokens.slice(0, parensToRemoveCount)
+          val rightSegment: Tokens = tokens.slice(tokens.length - parensToRemoveCount, tokens.length)
```
olafurpg: Isn't it possible to takeRight?

olafurpg:
I would be happy to merge a PR if you have ideas for re-usable components that would be useful around the codebase. We could also look into putting those components into `scalameta-contrib`

core/src/main/scala/org/scalafmt/rewrite/RedundantParens.scala
ShaneDelmore: It is possible to takeRight.  I wasn't sure if there was value in using operations like slice that returned Tokens instead of takeRight that returned IndexedSeq[Token].  It caught me by surprise a few times when I had to change my function signature as Tokens just turned into IndexedSeq[Token].  The alias is easy to lose.

ShaneDelmore: I also wasn't sure how critical it was to performance to avoid the usual collections operators, it looks like you spent some effort to make sure to create a builder and append to it, then return the result instead of collect.toVector.

olafurpg: Aaah, I see. We might be able to address this in scala.meta, I opened an issue https://github.com/scalameta/scalameta/issues/531

ShaneDelmore:
Great, I will make another pr demonstrating some ideas I had when I find some free time next.

PR #588, opened 2016-11-14
[WIP] Diff formatting

olafurpg:
This PR introduces a new feature called "diff formatting". With `scalafmt --diff`, you only format the lines that are touched by a diff. Examples

```shell
# via stdin
$ git diff -U0 master | scalafmt --diff --stdin
... # writes to files
$ svn diff --diff-cmd=diff -x-U0 | scalafmt --diff --stdin # untested, but should work
# via configuration
$ cat .scalafmt.conf
project.git = true
project.baseBranch = master
$ scalafmt --diff
... # writes to file
```

Use cases:
* Interactive IDE setting in a codebase that's not formatted with scalafmt. You could reformat a subsection you edited.
* (hypothetical) gradually introduce scalafmt into a project, or allow a subset of developers to use scalafmt in a codebase (caveat below)
* faster formatting (although nice, I wouldn't call it a selling feature unless for maybe huge source files).

Caveats:
* in a partially formatted codebase, scalafmt may still introduce unnecessarily large diffs. Example

```diff
// line limit    |
-  class Foo extends A with B with C
+  class Foo extends A with B with C with D
// after formatting
class Foo
    extends A
    with B
    with C
    with D
```
A blank enforcement of diff formatting might therefore be unwanted.

Challenges:
* need to guarantee that formatting a sequence of diffs will always produce the same output as formatting everything all the time.

Opening this PR to collect feedback. I would like to get this feature ready for 0.5.

TODOs:
- [x] handle indent for `Case`
- [x] handle indent for `Term.Lambda`
- [ ] setup test that goes through git repo history and formats diff on every commit. Assert the output is identical to formatting the entire project on every commit.
- [ ] select chains
- [ ] large config style lists

### DIFF
--- a/cli/src/main/scala/org/scalafmt/cli/Cli.scala
+++ b/cli/src/main/scala/org/scalafmt/cli/Cli.scala
@@ -189,14 +189,14 @@ object Cli {
       case Formatted.Success(formatted) =>
         inputMethod.write(formatted, input, options)
       case Formatted.Failure(e) =>
-        if (options.config.runner.fatalWarnings) {
-          throw e
-        } else {
-          logger.elem(options.config.runner.fatalWarnings)
-          options.common.err.println(
-            s"${LogLevel.warn} Error in ${inputMethod.filename}: $e"
-          )
-        }
+      if (options.config.runner.fatalWarnings) {
+        throw e
+      } else {
+        logger.elem(options.config.runner.fatalWarnings)
+        options.common.err.println(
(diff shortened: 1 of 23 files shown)


olafurpg:
This is blocked by https://github.com/olafurpg/scalafmt/issues/140

olafurpg:
0.5 made big progress on config argument lists, so this PR is somewhat closer to being ready. However, I'm a bit busy with other tasks at the moment so I'm gonna close this PR.

PR #594, opened 2016-11-18
Add indentYieldKeyword setting, fixes #592.

### LINKED ISSUE #592: Indent `yield` at same level as `for`
Scalafmt:
```scala
object Bar {
  for (x <- Nil)
    yield 2
}
```

```scala
object Bar {
  for (x <- Nil)
  yield 2
}
```

From gitter: https://gitter.im/olafurpg/scalafmt?at=582ec607e51081951edf151b

> for symmetry with if/else or try/catchs which seem to format like the former rather than the latter

olafurpg:


olafurpg:
Note, this PR breaks the expected behavior explained here https://github.com/olafurpg/scalafmt/issues/592#issuecomment-261499437

This needs more work

PR #602, opened 2016-12-01
Fix Linux installation command

felixmulder:
Was missing sudo in the `chmod` portion.


olafurpg:
In OSX, I think sudo is active for both commands. LGTM 👍 Thanks!

PR #615, opened 2016-12-18
Upgrade to scala.meta 1.3.

olafurpg:
1.3 introduced a new encoding for Type.Compound which complicated things
quite a bit. This commit tried to remap the new encoding into the old
encoding. The scala-repos diff is here: https://github.com/olafurpg/scala-repos/commit/c00c1bd6da91d2aec92f6d15933795f09d98ec2a

The diff comes from a quirky bug in
optimizer.recurseOnBlocks, see https://github.com/olafurpg/scalafmt/issues/614, which I think is a scalafmt bug and not scala.meta 1.3 bug.


olafurpg:
Broken cache on travis.

PR #619, opened 2016-12-19
fixing regex for latest version

sudev:
Not very familiar with this project, but I think this will fix cut issues where when tag name contains 
"-RC2". Current script is not considering -RC2 and is failing. 
Replaced with cut for simplicity.


olafurpg:
I just tried this out to upgrade to 0.5.0-RC2 and it worked great. Thanks! 👍

PR #622, opened 2016-12-19
Improve search state for reactjs style DSLs. Fixes #458

### LINKED ISSUE #458: Scalajs react element attributes
**Version**: 0.3.1
**Integration**: sbt
**Configuration**: `--style defaultWithAlign --maxColumn 100 --scalaDocs --continuationIndentCallSite 2`

``` scala
<.div(^.`class` := "col-xs-6",
  "text")
```

Scalafmt (**Failed to format file xxx. Cause: Search state exploded around line 1.**):

``` scala
<.div(^.`class` := "col-xs-6",
  "text")
```

``` scala
<.div(^.`class` := "col-xs-6", "text")
```

Looks like `^.`class`` cause this error.

olafurpg:
The trick is to handle symbolic operators differently for select chains. This eliminates a branch on every `<.` selection.

Before:
screen shot 2016-12-19 at 17 48 06

After:
screen shot 2016-12-19 at 17 47 49

### DIFF
--- a/core/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/core/src/main/scala/org/scalafmt/internal/Router.scala
@@ -876,14 +876,17 @@ class Router(formatOps: FormatOps) {
               case (_, lst) => Math.max(0, lst.length - 1)
             }.getOrElse(0)
           } else 0
-        Seq(
-          Split(NoSplit, 0, ignoreIf = ignoreNoSplit)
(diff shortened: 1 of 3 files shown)


olafurpg:
Awww shuck, this doesn't fix any "search state exploded" errors in the test corpus :( At least it might improve the life for folks using scalajs-react 😸

PR #623, opened 2016-12-19
Disable align.openParenCallSite in default

olafurpg:


olafurpg:
I'm gonna put this on hold for now, since I've already made quite significant changes to the default style today. The diff is here: https://github.com/olafurpg/scala-repos/commit/64008151fbd47671540863e0477d1afe103a1fbe

We may revisit this later.

Issue #451, opened 2016-09-19
Setup CI on Windows with Appveyor


olafurpg:
See #435. I don't have a windows machine so it's a pain for me to debug these.

olafurpg:
Fixed a long time ago.

PR #630, opened 2016-12-20
Improve builds in IntelliJ and on Windows

mpociecha:
- ScalaFile is moved to the separate, common module since in IntelliJ
the main scope can't depend on the test scope.
- Adopters.scala and Readme.scala moved to src/main so that the build
works in IntelliJ. In cmd it works as previously, unless I overlooked
something.
- The tarball with test sources is now downloaded directly in the
Scala code instead of using wget and tar. Therefore, now it works also
on Windows.

In general currently one can run compile in cmd (to generate sources)
and simply import the project to IntelliJ. At the beginning IDE will
inform that its output dirs for tests and main scope need to be split
but, once user confirmed the automatic fix, the build should work in
IntelliJ out of the box.

When one uses autocrlf = input, builds work also on Windows, but
I haven't fixed yet tests to have them all working on Windows.


mpociecha:
i hope you don't have anything against having it fully working also on Windows ;) (at some point)

I know that wget was at  least showing the progress and now there's nothing like that. If it does matter, I can change this to have a switch - to use the new way of downloading only on Windows.

mpociecha:
1. Likely they could, but does it make any sense? This ScalaFile is really used by various tests (unit + benchamarks) and I don't think `Adopters/Readme.scala` belong to the same category.
2. Thanks for pointing it out. I completely overlooked this.
3. If there's no logger configured, which I could use, then ok - I'll add simple printlns.
4. Any experience, I'm afraid.

olafurpg:
1. Aaah, sry, I thought you had moved the files to the `core` repo. This is just fine. Make sure that `readme/run` still works.
3. I'm OK with adding a logging framework in the tests, if it produces nice output when downloading. Otherwise println is fine.
4. Aah, OK, I may try to set it up once we get this PR merged :)

mpociecha:
Updated.

1. I was running it, there were no warnings/errors in cmd, and the generated documentation looks as always.

olafurpg:
Yay, I pulled down this PR and managed to build the project in IntelliJ! This should set the coast clear for #451.

Awesome work 👍 Thanks a lot for this contribution.

PR #631, opened 2016-12-20
Fix idempotency regression from #629

olafurpg:


olafurpg:
Yay, back down to 13 idempotency issues and 16 search state exploded. Tests now assert those numbers don't go up again.

PR #632, opened 2016-12-20
Setup AppVeyor

olafurpg:


olafurpg:
cc/ @mpociecha Do the errors in the Appvejor logs [here](https://ci.appveyor.com/project/olafurpg/scalafmt) look familiar? Running the tests fails on Windows. It seems to be an encoding issue,  I think I set uft8 correctly [here](https://github.com/olafurpg/scalafmt/pull/632/files#diff-180360612c6b8c4ed830919bbb4dd459R7).

mpociecha:
Yes, I'm also getting these errors:
```
org.scalafmt.Error$UnableToFindStyle: Unable to find style for file
core\src\test\resources\alig\AlignTokens.stat. com.typesafe.config.ConfigException$Parse: String: 1: Key '"80 columns
|"' may not be followed by token: end of file (if you intended '"80 columns
|"' to be part of a value, instead of a key, try adding double quotes around the whole value)
```

and others . As I wrote previously, I haven't fixed test failures on Windows. I looked at them briefly though and e.g. some of them disappear when one fixes line endings in a current and an expected text // Since e.g. there's the Linux line ending in one of them and there's used the File.separator - what means \r\n - when computing the second one. And so on, and so on.

PR #634, opened 2016-12-22
Add UnmaintainedScalafmtReformatOnCompilePlugin.

olafurpg:


sjrd:
IMO a better way to "offer" this would be to distribute it as a separate sbt plugin, in a separate GitHub repo. At least there's a reasonable chance that someone would actually become a maintainer, as they could maintain "the repo" rather than "one feature inside a complex repo".

olafurpg:
@sjrd You're probably right. My plan was to post it as a gist: https://gist.github.com/olafurpg/e045ef9d8a4273bae3e2ccf610636d66
I'll just do that instead.

PR #642, opened 2016-12-23
Fix path separators and line endings to make most of tests passing on Windows

mpociecha:
On Windows it's needed to set SBT_OPTS=-Dfile.encoding=UTF8

The changes are a bit tricky. One could easily make change where all compared line endings will be converted to the same form but, in fact, then we could miss cases where the tested code itself could change line endings in the unexpected way.

`sbt test` passed.
```
[info] Run completed in 29 seconds, 619 milliseconds.
[info] Total number of tests run: 798
[info] Suites: completed 14, aborted 0
[info] Tests: succeeded 798, failed 0, canceled 0, ignored 33, pending 0
[info] All tests passed.
```

Other than that:
- `scripted` still fails on Windows:
```
[info] java.lang.Exception: File: target\src_managed\Generated.scala
[info] Obtained output:
[info] object DontFormatMe { println(1) }
[info] Expected:
[info] object      DontFormatMe   {    println(1)    }
[info]
[info]  at $8bc643ac9d20807c5658$.assertContentsEqual(build.sbt:28)
```
- The result of `core/test:runMain org.scalafmt.FormatExperimentApp` on Windows:
```
Success: 9381

Total: 10698
+----------+----+---+----+--+--+----+----+----+
|Total time| Max|Min|Mean|Q2|Q3|90th|95th|99th|
+----------+----+---+----+--+--+----+----+----+
|    293829|7456|  0|  75|26|75| 177| 292| 693|
+----------+----+---+----+--+--+----+----+----+
[success] Total time: 309 s, completed 2016-12-23 16:19:58
```

### DIFF
--- a/bin/scalafmt.bat
+++ b/bin/scalafmt.bat
@@ -1,1 +1,1 @@
-java -jar PATH_TO/scalafmt.jar %*
+java -jar PATH_TO\scalafmt.jar %*
--- a/cli/src/test/scala/org/scalafmt/cli/CliOptionsTest.scala
+++ b/cli/src/test/scala/org/scalafmt/cli/CliOptionsTest.scala
@@ -8,7 +8,6 @@ import org.scalafmt.config.AlignToken
 import org.scalafmt.config.ImportSelectors
 import org.scalafmt.config.IndentOperator
 import org.scalafmt.config.ScalafmtConfig
-import org.scalafmt.util.LoggerOps._
 import org.scalatest.FunSuite
 
 class CliOptionsTest extends FunSuite {
--- a/cli/src/test/scala/org/scalafmt/cli/CliTest.scala
+++ b/cli/src/test/scala/org/scalafmt/cli/CliTest.scala
@@ -4,7 +4,6 @@ import java.io.ByteArrayInputStream
 import java.io.ByteArrayOutputStream
 import java.io.File
 import java.io.FileNotFoundException
-import java.io.InputStream
 import java.io.PrintStream
 import java.nio.charset.StandardCharsets
 import java.nio.file.Files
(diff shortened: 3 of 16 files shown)


olafurpg:
Wow, thank you for this contribution! 

- FidelityTests, I suspect these lines are the culprits: https://github.com/olafurpg/scalafmt/blob/6b82e05111c41e6ed51e89e5d8cfbbb8e0059803/core/src/test/scala/org/scalafmt/FidelityTest.scala#L34
- scripted tests, these lines: https://github.com/olafurpg/scalafmt/blob/6b82e05111c41e6ed51e89e5d8cfbbb8e0059803/scalafmtSbt/src/sbt-test/scalafmt-sbt/helloworld/build.sbt#L53
- I don't expect FormatExperimentApp to cause problems.
- FYI `StripMarginTest`/`ContinuationIndentTests`/... are using the old testing system, I'll probably migrate them soon to use the `.stat` tests in `core/src/test/resources`

olafurpg:
BTW. I am writing an updated tutorial as we speak :)

mpociecha:
Regardiing FidelityTests, only this one failed:
 .\benchmarks\src\resources\scala-js\`OptimizerCore.scala` *** FAILED *** (2 seconds, 358 milliseconds)
   org.scalafmt.Error$SearchStateExploded: Search state exploded around line 19591`

I pasted details in the description. :)

olafurpg:
That file should be excluded in the FidelityTests, but I suspect it's not filtered out because the filter is written as `/resources/`

mpociecha:
Ok, I get it. That's why there were more tests on Windows. ;)

mpociecha:
`FidelityTest` fixed. Regarding scripted, it's not enough. I can take a look but not now.

olafurpg:
Awesome! I am happy to merge this once the CI completes. We can fix the scripted tests later. I am currently doing christmas cleaning in the repo so this PR comes at the perfect time!

olafurpg:
LGTM 👍 Thanks!

PR #647, opened 2016-12-23
Fix links to Tutorial.md

mpociecha:
The name of the file starts with `T` so it didn't work with `t`. By the way I changed the strange wording. ;)


olafurpg:
This is why I prefer Scalatex, it catches bad links automatically! Thank you :)

PR #648, opened 2016-12-24
Make nested path exclusions working on Windows

mpociecha:
Now paths like target/foo are accepted also on Windows.
The existing unit test has been extended to verify such cases.
By the way this fixed the `scripted` task when running on Windows.

In fact it was possible to use \\ e.g. in paths passed via --exclude,
but it isn't very helpful and convenient in case of projects built on
various platforms.

### DIFF
--- a/core/src/main/scala/org/scalafmt/config/ProjectFiles.scala
+++ b/core/src/main/scala/org/scalafmt/config/ProjectFiles.scala
@@ -1,6 +1,7 @@
 package org.scalafmt.config
 
 import metaconfig.ConfigReader
+import org.scalafmt.util.OsSpecific
 
 @ConfigReader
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/scalafmt/config/ProjectFiles.scala
```diff
@@ -10,5 +11,6 @@ case class ProjectFiles(
     excludeFilters: Seq[String] = Nil
 ) {
   lazy val matcher: FilterMatcher =
-    FilterMatcher(includeFilters, excludeFilters)
+    FilterMatcher(includeFilters.map(OsSpecific.fixSeparatorsInPathPattern),
```
olafurpg: Awesome idea 👍

PR #649, opened 2016-12-25
Issue #607 Add param to don't insert new lines after annotations

### LINKED ISSUE #607: Add the parameter to keep annotations in the original lines
It would be helpful to have a parameter which simply allows to keep annotations in the original lines (relatively - obviously the number of the line could change). Only the indentation etc. would be fixed, if needed.

**Version**: 0.4.10
**Configuration**: the default one, but setting various params rather doesn't change anything

**Input == expected:**
```scala
  @annot @deprecated class B extends A {
    @annot override def foo = 1
  
    @annot
    override def bar = 1
  
    @annot @deprecated def foo2 = 1
  
    @annot
    @deprecated
    def bar2 = 1
  }
  
  @annot
  @deprecated class C
  
  @annot
  @deprecated
  class D
```

**Scalafmt:**
```scala
  @annot
  @deprecated
  class B extends A {
    @annot override def foo = 1
  
    @annot
    override def bar = 1
  
    @annot
    @deprecated def foo2 = 1
  
    @annot
    @deprecated
    def bar2 = 1
  }
  
  @annot
  @deprecated class C
  
  @annot
  @deprecated
  class D
```

mpociecha:
There's added `noNewlineAfterAnnotation` param to OptIn, `stat` test
and the additional check in CliOptionsTest.

The logic related to annotations is located in TreeOps. I changed
it to - when needed - use the alternative approach where we check
whether the line of the beginning of the annotation is the same as
the line of the end of previous one.

Fixes #607

### DIFF
--- a/cli/src/test/scala/org/scalafmt/cli/CliOptionsTest.scala
+++ b/cli/src/test/scala/org/scalafmt/cli/CliOptionsTest.scala
@@ -42,6 +42,7 @@ class CliOptionsTest extends FunSuite {
         |optIn: {
         |  configStyleArguments = true
         |  breakChainOnFirstMethodDot = true
+        |  noNewlineAfterAnnotation = true
         |}
         |maxColumn = 4000
         |poorMansTrailingCommasInConfigStyle = true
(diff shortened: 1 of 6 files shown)


mpociecha:
1. I hope it's not yet the spaghetti code.
2. It's very likely that I'm doing something really dumb. It's simply the first implementation that worked.

olafurpg:
BTW. I totally understand router.scala may be confusing, it's a weird pattern match that I've spent months polishing. It's become so complicated at this point that I'm seriously considering how I could solve this problem with an alternative design.

mpociecha:
Thanks. Can we just close this PR and merge your changes? :)

mpociecha:
@olafurpg Thanks. Can we just close this PR and merge your changes? :)

olafurpg:
Sure, we can do that. I opened #651. It also fixes the parameter annotation case. Note there's one caveat, you control all line breaks, even when you might expect scalafmt to break the line like here: https://github.com/olafurpg/scalafmt/pull/651/files#diff-9a648d9a9cbb16f8bd6ab6a8633abad3R29

PR #665, opened 2017-01-06
Remove extra newline from stdout

dguo-coursera:
I'm not sure if this is intended, but I noticed that the formatted code is written to stdout using println. This makes redirecting the stdout to a file incorrect as it adds an extra new line character.


olafurpg:
Good catch! Thank you for this contribution. I took the liberty to make the add the `size` assertions everywhere, which involved cleaning up a bit of the test infrastructure. Cheers,

olafurpg:
btw, I messed up the merge workflow for some reason, so it appears your changes got merged twice. Anyways, doesn't really matter, ci is happy ^_^ https://travis-ci.org/olafurpg/scalafmt/builds/189468091

Issue #660, opened 2016-12-31
Feature request: configurable spaces around infix types


Atry:
I created a compiler plugin to provide `##` syntax https://github.com/ThoughtWorksInc/implicit-dependent-type
I hope people feel that `Foo[Bar]##Baz` is similar to native syntax `Foo[Bar]#Baz`. However, there's currently no configuration to do that, and scalafmt always add spaces around `##`, i.e. `Foo[Bar] ## Baz`.

- **Version**: 0.5
- **Integration**: IntelliJ
- **Configuration**: default
## Unexpected formatting output

Obtained, scalafmt formats code like this:
```scala
val hlistForFoo: Generic[Foo] ## Repr = 1 :: "xxx" :: HNil
```

Expected, I would like the formatted output to look like this:
```scala
val hlistForFoo: Generic[Foo]##Repr = 1 :: "xxx" :: HNil
```

olafurpg:
As discussed on gitter, I am happy to add an option to configure the space around infix type parameters. I suggest to name the flag `spaces.neverAroundInfixTypes: Seq[String] = []`, which you could configure to be `["##"]`. 

If someone is motivated to take on this issue, the tutorial should contain all the necessary instructions to fix it, see https://github.com/olafurpg/scalafmt/blob/master/Tutorial.md

olafurpg:
This was fixed by @ysusuk! In 0.5.2 you will be able to configure
```
spaces.neverAroundInfixTypes = ["##"]
```

PR #669, opened 2017-01-08
#660: Feature request: configurable spaces around infix types

ysusuk:


core/src/main/scala/org/scalafmt/config/Spaces.scala
```diff
@@ -6,5 +6,6 @@ import metaconfig.ConfigReader
 case class Spaces(
     beforeContextBoundColon: Boolean = false,
     afterTripleEquals: Boolean = false,
-    inImportCurlyBraces: Boolean = false
+    inImportCurlyBraces: Boolean = false,
+    neverAroundInfixTypes: Seq[String] = Seq("##")
```
olafurpg: Could we keep the default at `Nil`? The `##` operator is only magical if you're using a specific compiler plugin.

olafurpg:
Two more things I forgot

- Could you add a test with `##` appearing in Term.ApplyInfix position? Example `val x: F[X] ## B = f##b` should format as `val x: F[X]##B = f ## b`. The option should only affect infix **type** operators.
- Could you document the new option in the `Spaces` docstring?

core/src/main/scala/org/scalafmt/config/Spaces.scala
ysusuk: Agree! I was expecting this ^_^ but was not sure what is better as default value.

core/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -474,10 +474,14 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
       token <- arg.tokens.lastOption
     } yield token).getOrElse(owner.tokens.last)
 
-    if (style.spaces.neverAroundInfixTypes.contains((op.value)))
-      Split(NoSplit, 0)
-    else
-      Split(modification, 0).withIndent(Num(indent), expire, ExpiresOn.Left)
+    owner.parent match {
```
ysusuk: not sure that such an construct clearly defines an intention.

ysusuk:
@olafurpg not sure what did you mean by `Spaces` docsring =(. I added param description in `ScalafmtConfig`

olafurpg:
I meant to create a docstring in spaces, but this is fine too. Need to move the scalafmtconfig docstrings to the nested values some, never got around to it after introducing the hierarchy. 

LGTM 👍 thanks!

PR #678, opened 2017-01-16
New ContinuationIndent: extendSite

pjrt:
This setting sets the indentation at extends/with site. By default is it four (4) (in order to not break current), but the Scala Style Guide (http://docs.scala-lang.org/style/declarations.html#classes) recommends two (2).


olafurpg:
Thanks!

PR #682, opened 2017-01-18
Adds a new style for function defs

pjrt:
This is a style that we are currently using and would like to keep using. It formats:

```scala
// maxColumn = 80
// this function is 98 chars long
def format_![T <: Tree](code: String, foo: Int)(f: A => B)(implicit ev: Parse[T], ev2: EC): String
```

to this:

```scala
def format_![T <: Tree](
    code: String,
    foo: Int
  )(f: A => B
  )(implicit ev: Parse[T],
    ev2: EC
  ): String
```

This format will only apply if the function would go over the maxColumn is single-lined. If it is under, it will be singled line.

### DIFF
--- a/core/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
+++ b/core/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
@@ -181,7 +181,21 @@ import org.scalafmt.util.ValidationOps
   *        x + 2
   *    }.filter(_ > 2)
   *  }}}
+  * @param verticalMultiline If true, reformat multi-line function definitions in
+  *  the following way
+  * {{{
+  *   def format(
+  *       code: String,
+  *       age: Int
+  *     )(implicit ev: Parser
+  *     ): String
+  * )
+  * }}}
   *
+  *  All parameters are on their on line indented by four (4), seperation between
+  *  parament groups are indented by two (2). ReturnType is on its own line at
+  *  then end. This will only trigger if the function is already multi-line
+  *  or if a single-line function goes over [[maxColumn]].
   */
 @ConfigReader
 case class ScalafmtConfig(
@@ -209,7 +223,7 @@ case class ScalafmtConfig(
     danglingParentheses: Boolean = false,
     poorMansTrailingCommasInConfigStyle: Boolean = false,
     bestEffortInDeeplyNestedCode: Boolean = false,
-    straightCurried: Boolean = false,
+    verticalMultiline: Boolean = false,
     project: ProjectFiles = ProjectFiles()
 ) {
 
--- a/core/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
+++ b/core/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
@@ -181,21 +181,24 @@ import org.scalafmt.util.ValidationOps
   *        x + 2
   *    }.filter(_ > 2)
   *  }}}
-  * @param verticalMultiline If true, reformat multi-line function definitions in
+  * @param verticalMultilineAtDefinitionSite If true, reformat multi-line function definitions in
   *  the following way
   * {{{
   *   def format(
   *       code: String,
   *       age: Int
-  *     )(implicit ev: Parser
+  *     )(implicit ev: Parser,
+  *       c: Context
   *     ): String
   * )
   * }}}
   *
   *  All parameters are on their on line indented by four (4), seperation between
   *  parament groups are indented by two (2). ReturnType is on its own line at
-  *  then end. This will only trigger if the function is already multi-line
-  *  or if a single-line function goes over [[maxColumn]].
+  *  then end. This will only trigger if the function would go over
+  *  [[maxColumn]]. If a multi-line funcion can fit in a single line, it will
+  *  make it so. Note that this setting ignores [[continuation.defnSite]],
+  *  [[binPack.defnSite]], and [[align.openParenDefnSite]].
   */
 @ConfigReader
 case class ScalafmtConfig(
@@ -223,7 +226,7 @@ case class ScalafmtConfig(
     danglingParentheses: Boolean = false,
     poorMansTrailingCommasInConfigStyle: Boolean = false,
     bestEffortInDeeplyNestedCode: Boolean = false,
-    verticalMultiline: Boolean = false,
+    verticalMultilineAtDefinitionSite: Boolean = false,
     project: ProjectFiles = ProjectFiles()
 ) {
 
--- a/core/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
+++ b/core/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
@@ -209,6 +209,7 @@ case class ScalafmtConfig(
     danglingParentheses: Boolean = false,
     poorMansTrailingCommasInConfigStyle: Boolean = false,
     bestEffortInDeeplyNestedCode: Boolean = false,
+    straightCurried: Boolean = false,
     project: ProjectFiles = ProjectFiles()
 ) {
 
--- a/core/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/core/src/main/scala/org/scalafmt/internal/Router.scala
@@ -410,6 +410,8 @@ class Router(formatOps: FormatOps) {
             )
         }
 
+      // Parameter opening for one parameter group. This format works
+      // on a group-by-group basis.
       case FormatToken(open @ LeftParen(), right, between)
           if style.straightCurried && isDefnSite(leftOwner) =>
         val close = matchingParentheses(hash(open))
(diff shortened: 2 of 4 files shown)


core/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -209,6 +223,7 @@ case class ScalafmtConfig(
     danglingParentheses: Boolean = false,
     poorMansTrailingCommasInConfigStyle: Boolean = false,
     bestEffortInDeeplyNestedCode: Boolean = false,
+    verticalMultiline: Boolean = false,
```
pjrt: Name pending. It doesn't sound right but I can't think of anything else.

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -409,6 +409,52 @@ class Router(formatOps: FormatOps) {
                 .withIndent(2, expire, Left)
             )
         }
+
+      // Parameter opening for one parameter group. This format works
+      // on a group-by-group basis.
```
pjrt: This doesn't sound right to me. I would like to have a holistic approach where I see the `def` as one thing, not as a bunch of `(`. I thought I could do so by using the `Decision` below, but for some reason a Decision for `(` (same as the top format case) just isn't picked up. 

Is there another way of doing this?

olafurpg: Unfortunately no. The design is intentionally low level, for better or worse. Policies are the way to achieve what you want, which is to enforce behavior in upcoming format tokens. The first open paren of a def is where you enable the policy for the entire def/class parameter list.

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -409,6 +409,52 @@ class Router(formatOps: FormatOps) {
...
+                Seq(Split(Newline, 0)
+                  .withIndent(indentSep, close, Right)
+                )
+              )
+          }
+          val configStyleLike =
+            oneLinePerArgPolicy.andThen(noNewLineOnParamSep)
+          // Find the very first leftParen in the `def`. This will tell us
+          // if we are in the first group or not.
+          val isFirstGroup =
```
olafurpg: I know this may feel hacky but it's totally idiomatic in the scalafmt code base.

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -409,6 +409,52 @@ class Router(formatOps: FormatOps) {
...
         }
+
+      // Parameter opening for one parameter group. This format works
+      // on a group-by-group basis.
+      case FormatToken(open @ LeftParen(), right, between)
+          if style.verticalMultiline && isDefnSite(leftOwner) =>
+        val close = matchingParentheses(hash(open))
+        val indentParam = Num(4)
+        val indentSep = Num(2)
+        if (isTuple(leftOwner)) {
```
olafurpg: I would be very careful with the single line policy

core/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
olafurpg: Naming is hard. I would love to use an adt for styles supported in parameter lists with possible values bin pack, default and vertical multi-line. Currently it's a boolean in BinPack.

I'm OK with making this a boolean in scalafmt config, but the  name needs to be more specific. I suggest verticalMultilineAtDefinitionSite

core/src/main/scala/org/scalafmt/internal/Router.scala
pjrt: This is actually not needed, left-over from when I was copy-and-pasting. Sorry for the noise.

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -409,6 +409,52 @@ class Router(formatOps: FormatOps) {
...
+          // Find the very first leftParen in the `def`. This will tell us
+          // if we are in the first group or not.
+          val isFirstGroup =
+            leftOwner.tokens.find {
+              case LeftParen() => true
+              case _ => false
+            }.filter(_ == open).isDefined
+          val mod = if (isFirstGroup) Newline else NoSplit
+          Seq(
+            Split(mod, 0)
```
pjrt: That would cause the split to happen partly. So it won't split on the first parameter group but it would on the second (if not splitting it causes it to go over the maxColumn). 

The test shows:

```scala
def format_![T <: Tree](code: String, foo: Int)(f: A => B
  )(implicit ev: Parse[T], ev2: EC): String
```

I need to somehow know, at the first `(` whether maxColumn has been violated or not and do one style or the other for the whole `def`.

olafurpg:
It's not possible to know the column width at this point. You only describe
the possible formatting options (including policies with expire tokens) and
their respective costs. The search will handle the max column.

core/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
pjrt: It is odd because this is a whole style, but only for `def` sites (as opposed to a "style" like `IntelliJ`). This "mini-style" will also override things like continuation.defnSite (I'll make sure to mention it in the doc).

pjrt:
Ok I managed to get it to do what I wanted (it seems so obvious now). I've renamed the setting and added docs and comments all over the place. 

The interface for setting this is a bit odd, we should consider doing something about how binPack and now "vertical" work with one another.

core/src/test/resources/test/VericalMultiline.stat
```diff
@@ -27,12 +27,15 @@ def format_![T <: Tree](code: String)(f: A => B)(implicit ev: Parse[T]): String
 
 <<< curried function mixed style 1
 def format_![T <: Tree](
-    code: String
-  )(implicit ev: Parse[T]): String = 1
+    code: String,
+    code2: String
+  )(implicit ev: Parse[T], ex: D): String = 1
```
olafurpg: I would have expected there to be a new line before ex. Are you sure this is correct?

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -412,49 +413,58 @@ class Router(formatOps: FormatOps) {
...
+                  .withIndent(indentParam, close2, Right)
+              )
+            )
         }
 
+        val policy =
+          Policy(oneLinePerArg.orElse(paramGroupSplitter), lastParen.end)
+
+        Seq(
+          Split(NoSplit, 0) // If it fits in one block, make it so
```
olafurpg: Split has an ignoreIf boolean field that you can set to true if this is not the first parameter list.

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -412,49 +413,58 @@ class Router(formatOps: FormatOps) {
...
-          val noNewLineOnParamSep: PartialFunction[Decision, Decision] = {
-            case Decision(t @ FormatToken(_, `close`, _), _) =>
-              Decision(t,
-                Seq(Split(Newline, 0)
-                  .withIndent(indentSep, close, Right)
-                )
-              )
+
+        @tailrec
+        def loop(token: Token): Option[Token] =
```
olafurpg: Can we move this to FormatOps with a clearer method name? Suggestion: lastClosingParam with return type Token.RightParen, no option since it can't be that by construction of the map. Exceptions are caught and dealt with higher up.

core/src/test/resources/test/VericalMultiline.stat
pjrt: It is, this is the "before" case. The after case is below. These tests might be redundant though.

olafurpg: Aha whoops! I didn't see that, on the phone -_-  good, then it makes sense!

core/src/main/scala/org/scalafmt/internal/Router.scala
pjrt: I removed that since I didn't need the whole "am I on the first block" thing. Now it is just "I am on a def, do either of these two policies: no split, single line or newline, vertical-style".

core/src/main/scala/org/scalafmt/internal/Router.scala
olafurpg: Nevermind, this is good.

core/src/main/scala/org/scalafmt/internal/Router.scala
olafurpg: Makes sense, this is clean!

olafurpg:
Thank you for this contribution, I'll try have another release out this week :)

core/src/main/scala/org/scalafmt/internal/Router.scala
pjrt: Oh, I was planning on moving it to FormatOps as you said tonight.

PR #700, opened 2017-01-26
Fix issue on verticalMultilineAtDefinitionSite

pjrt:
Fixes an issue where verticalMultilineAtDefinitionSite affected classes,
instead of just `def` sites.

### DIFF
--- a/core/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/core/src/main/scala/org/scalafmt/internal/Router.scala
@@ -414,7 +414,7 @@ class Router(formatOps: FormatOps) {
       // Parameter opening for one parameter group. This format works
       // on a group-by-group basis.
       case FormatToken(open @ LeftParen(), _, between)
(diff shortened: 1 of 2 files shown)


pjrt:
Yes, I was thinking that actually. Will do.

olafurpg:
I have been impressed with your contributions @pjrt. I am curious to know, would you be interested in becoming a scalafmt maintainer?

You would have merge rights and be listed in the readme/docs as such. With all the recent contributions and activity around the project, I could use some extra help.

If you don't have time or motivation, I totally understand. Even if you approve, please don't feel pressured to take on a responsibility unless you're personally motivated to work on said task. If you later change your mind then that's totally fine, too.

pjrt:
Sure! This all started as just trying to make a specific style we are using work with Scalafmt. But I've actually become intrigued by the project and your paper (which I haven't finished yet).

So yeah, I would love to be a contributor. I'll try to put aside some hours to fix the outstanding issues.

olafurpg:
Awesome! I'm happy to hear that 😸 

I'll update the docs/readme then.

Issue #690, opened 2017-01-23
Support check for "redundant enclosing brace" in string interpolation.


Synesso:
## Enhancement request

Prevent redundant enclosing block in string interpolation, as detected by IntelliJ's default code style. 

```scala
s"Hello ${"dlroW".reverse}"
```
is OK

But
```scala
val s = "World"
s"Hello ${s}"
```
is reformatted to
```
val s = "World"
s"Hello $s"
```

ysusuk:
does it make sense to wait until https://github.com/scalameta/scalameta/issues/251 will be implemented?

olafurpg:
No need to wait for that issue, scalafmt does not use quasi quotes. You can find the expression inside an interpolation string by matching against Interpolator.Part iirc

ysusuk:
Hi @olafurpg! shall it be another rewrite rule?

olafurpg:
I was thinking of bundling it together with RedundantBraces but disabled
behind a setting. We can enable the setting by default in the next minor
bump. Wdyt?

ysusuk:
i put in https://github.com/olafurpg/scalafmt/blob/master/core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala#L56 this cod

```scala
case Pat.Interpolate(x, y, z) =>
        println("!!! int pat")
      case Term.Interpolate(x, y, z) =>
        println("!!! int term")
```

olafurpg:
That's weird,

```scala
@ q""" 1 + q"a $${b}" """.collect {
    case Term.Interpolate(a, b, c) =>
      a -> b -> c
  }
res7: List[((Term.Name, collection.immutable.Seq[Lit]), collection.immutable.Seq[Term])] = List(((q, List("a ", "")), List(b)))
```

ysusuk:
looks like some strange behaviour in tests (e.g. https://github.com/olafurpg/scalafmt/issues/657)

changed

```scala
s"Hello ${c}"
```
to
```scala
object a {
  def b(c: String) = s"Hello ${c}"
}
```
and it works

olafurpg:
This issue was fixed by @ysusuk in #703

PR #703, opened 2017-01-29
#690: Support check for "redundant enclosing brace" in string interpo...

ysusuk:
…lation.

### DIFF
--- a/core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -46,8 +46,7 @@ object RedundantBraces extends Rewrite {
     bodyIsNotTooBig
   }
 
-  val ALPHA_NUMERIC_AND_UNDERSCORE = ((('a' to 'z') ++ ('A' to 'Z') ++ ('0' to '9')) :+ '_').toSet
-  def isAlphaNumericOrUnderscore(c: Char): Boolean = ALPHA_NUMERIC_AND_UNDERSCORE.contains(c)
+  def isIdentifierAtStart(value: String) = value.nonEmpty && (Character.isLetterOrDigit(value.head) || value.head == '_')
 
   override def rewrite(code: Tree, ctx: RewriteCtx): Seq[Patch] = {
     import ctx.tokenTraverser._
@@ -59,10 +58,8 @@ object RedundantBraces extends Rewrite {
           case (Lit(value: String), arg @ Term.Name(name)) =>
             val openBrace = prevToken(arg.tokens.head)
             val closeBrace = nextToken(arg.tokens.head)
-            (openBrace, closeBrace, value.headOption) match {
-              case (LeftBrace(), RightBrace(), Some(c)) if !isAlphaNumericOrUnderscore(c) =>
-                builder += Patch(openBrace, closeBrace, name)
-              case (LeftBrace(), RightBrace(), None) =>
+            (openBrace, closeBrace) match {
+              case (LeftBrace(), RightBrace()) if !isIdentifierAtStart(value) =>
                 builder += Patch(openBrace, closeBrace, name)
               case _ =>
             }
--- a/core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -55,11 +55,11 @@ object RedundantBraces extends Rewrite {
     code.collect {
       case t: Term.Interpolate if settings.stringInterpolation =>
         t.parts.tail.zip(t.args).collect {
-          case (Lit(value: String), arg @ Term.Name(name)) =>
+          case (Lit(value: String), arg @ Term.Name(name)) if !isIdentifierAtStart(value) =>
             val openBrace = prevToken(arg.tokens.head)
             val closeBrace = nextToken(arg.tokens.head)
             (openBrace, closeBrace) match {
-              case (LeftBrace(), RightBrace()) if !isIdentifierAtStart(value) =>
+              case (LeftBrace(), RightBrace()) =>
                 builder += Patch(openBrace, closeBrace, name)
               case _ =>
             }
--- a/core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -1,13 +1,10 @@
 package org.scalafmt.rewrite
 
-import scala.meta.Importee
 import scala.meta.Tree
 import scala.meta._
-import scala.meta.tokens.Token.Equals
 import scala.meta.tokens.Token.LF
 import scala.meta.tokens.Token.LeftBrace
 import scala.meta.tokens.Token.RightBrace
-import scala.meta.tokens.Token.Space
 
 import org.scalafmt.util.TreeOps._
 
@@ -51,8 +48,20 @@ object RedundantBraces extends Rewrite {
 
   override def rewrite(code: Tree, ctx: RewriteCtx): Seq[Patch] = {
     import ctx.tokenTraverser._
+    import ctx.style.rewrite.{redundantBraces => settings}
     val builder = Seq.newBuilder[Patch]
     code.collect {
+      case Term.Interpolate(x: Term.Name, lits: Seq[Lit], terms: Seq[Term]) if settings.stringInterpolation =>
+        terms.collect {
+          case t @ Term.Name(name) =>
+            val openBrace = prevToken(t.tokens.head)
+            val closeBrace = nextToken(t.tokens.head)
+            (openBrace, closeBrace) match {
+              case (LeftBrace(), RightBrace()) =>
+                builder += Patch(openBrace, closeBrace, name)
+              case _ =>
+            }
+        }
       case d: Defn.Def if isCandidate(d, ctx) =>
         val open = d.body.tokens.head
         val close = d.body.tokens.last
--- a/core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -46,6 +46,9 @@ object RedundantBraces extends Rewrite {
     bodyIsNotTooBig
   }
 
+  val ALPHA_NUMERIC_AND_UNDERSCORE = ((('a' to 'z') ++ ('A' to 'Z') ++ ('0' to '9')) :+ '_').toSet
+  def isAlphaNumericOrUnderscore(c: Char): Boolean = ALPHA_NUMERIC_AND_UNDERSCORE.contains(c)
+
   override def rewrite(code: Tree, ctx: RewriteCtx): Seq[Patch] = {
     import ctx.tokenTraverser._
     import ctx.style.rewrite.{redundantBraces => settings}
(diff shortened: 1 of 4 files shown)


olafurpg:
Thank you for this contribution! Great work.

I think there's a missing test case s"${foo}bar". Removing the curly brace will result in the ast to change.

ysusuk:
Thank you @olafurpg!

ysusuk:
can you suggest me a way to find out what is going after closing brace in `s"${foo}bar"`. i tried

```scala
val closeBrace = nextToken(t.tokens.head)
val afterCloseBrace = nextToken(closeBrace)
```

but it doesn't seems to work.

olafurpg:
Maybe something like this? 

```scala
@ val t = q""" q"$${foo}bar$$banana ok" """; t.parts.tail.zip(t.args)
t: Term.Interpolate = q"$foobar$banana ok"
res9_1: collection.immutable.Seq[(Lit, Term)] = List(("bar", foo), (" ok", banana))
```

olafurpg:
Note that there are Interpolation.Splice{End,Start} tokens that you may bump into
```scala
@ q""" q"$${foo}bar$$banana ok" """.tokens.map(x => x.getClass -> x.structure)
res12: collection.immutable.IndexedSeq[(Class[?0], String) forSome { type ?0 <: Token }] = Vector(
  (class scala.meta.tokens.Token$BOF,BOF [0..0)),
  (class scala.meta.tokens.Token$Interpolation$Id,q [0..1)),
  (class scala.meta.tokens.Token$Interpolation$Start," [1..2)),
  (class scala.meta.tokens.Token$Interpolation$Part, [2..2)),
  (class scala.meta.tokens.Token$Interpolation$SpliceStart,$ [2..3)),
  (class scala.meta.tokens.Token$Ident,foobar [3..9)),
  (class scala.meta.tokens.Token$Interpolation$SpliceEnd, [9..9)),
  (class scala.meta.tokens.Token$Interpolation$Part, [9..9)),
  (class scala.meta.tokens.Token$Interpolation$SpliceStart,$ [9..10)),
  (class scala.meta.tokens.Token$Ident,banana [10..16)),
  (class scala.meta.tokens.Token$Interpolation$SpliceEnd, [16..16)),
  (class scala.meta.tokens.Token$Interpolation$Part, ok [16..19)),
  (class scala.meta.tokens.Token$Interpolation$End," [19..20)),
  (class scala.meta.tokens.Token$EOF,EOF [20..20))
)
```

core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -51,8 +48,20 @@ object RedundantBraces extends Rewrite {
...
+    import ctx.style.rewrite.{redundantBraces => settings}
     val builder = Seq.newBuilder[Patch]
     code.collect {
+      case t: Term.Interpolate if settings.stringInterpolation =>
+        t.parts.tail.zip(t.args).collect {
+          case (Lit(value), arg @ Term.Name(name)) =>
+            val openBrace = prevToken(arg.tokens.head)
+            val closeBrace = nextToken(arg.tokens.head)
+            (openBrace, closeBrace, value) match {
+              case (LeftBrace(), RightBrace(), "") =>
```
olafurpg: Why `""`? Have you tested cases like?

```scala
q""
q"$a"
q" $a "
q"$a-"
q"${a}1"
q"${a}_2"
```

core/src/test/resources/rewrite/RedundantBraces.stat
```diff
@@ -159,3 +160,27 @@ object a {
...
+object a {
+  def b(foo: Int) = s"Hello ${foo}bar"
+}
+>>>
+object a {
+  def b(foo: Int) = s"Hello ${foo}bar"
+}
+<<< enclosing braces in str interpolation
+object a {
+  def b(d: Int) = s"Hello ${d.toString}"
```
olafurpg: I don't think the `def b(d: Int) = ` part is necessary. Maybe a block is sufficient
```
{ s"$a" }
```

olafurpg: Note that the test doesn't have to type check, only parse.

ysusuk: it doesn't work without `object`. max what i achieved is
```scala
<<< redundant enclosing braces in str interpolation
object a {
  s"Hello ${value}"
}
>>>
object a {
  s"Hello $value"
}
```

core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
ysusuk: Hi @olafurpg. I added all the test cases you mentioned. I'm just not sure that it does make big sense to test all the cases without braces! Since there is a check that args are inside braces! maybe one test here is enough?

olafurpg: Whoops, testing without braces is not so interesting I agree. What if those examples are expected outputs and the inputs have variables wrapped in braces? I still don't understand why it should only rewrite when value is ""

olafurpg: I suspect the current implementation misses some opportunities to rewrite

core/src/test/resources/rewrite/RedundantBraces.stat
```diff
@@ -194,35 +194,35 @@ object a {
...
-<<< spaces before and after arg no braces in str interpolation
+<<< spaces before and after arg in str interpolation
 object c {
-  s" $a "
+  s" ${a} "
 }
 >>>
 object c {
-  s" $a "
+  s" ${a} "
```
olafurpg: Is this expected output?

core/src/test/resources/rewrite/RedundantBraces.stat
```diff
@@ -159,3 +160,83 @@ object a {
...
+object c {
+  s" ${a} "
+}
+<<< special symbol after arg no braces in str interpolation
+object c {
+  s"${a}-"
+}
+>>>
+object c {
+  s"${a}-"
```
olafurpg: Same here, a can be unwrapped.

core/src/test/resources/rewrite/RedundantBraces.stat
ysusuk: it's a bug)

core/src/test/resources/rewrite/RedundantBraces.stat
ysusuk: and this one fails `s"${a} a${b}b"`

ysusuk: i was thinking about check like `value.isEmpty` or starts with \n, \r or anything like this! what else should it be? minus/plus... is there kind a list of them? or do you think there is a better approach?

olafurpg: You can look at the spec https://www.scala-lang.org/files/archive/spec/2.11/01-lexical-syntax.html#identifiers

Seems like this rewrite should not take place when the right hand part starts with underscore or an alphanumeric character. In other cases, you can safely remove the curly brace.

ysusuk:
Thanx a lot for support @olafurpg! i corrected logic and cleaned up tests!

core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -49,10 +46,27 @@ object RedundantBraces extends Rewrite {
     bodyIsNotTooBig
   }
 
+  val ALPHA_NUMERIC_AND_UNDERSCORE = ((('a' to 'z') ++ ('A' to 'Z') ++ ('0' to '9')) :+ '_').toSet
```
olafurpg: What about variables like `æ1`?

```
res0: Boolean = true
@ Character.isLetterOrDigit('-')
res1: Boolean = false
@ Character.isLetterOrDigit('æ')
res3: Boolean = true
```

core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -49,10 +46,27 @@ object RedundantBraces extends Rewrite {
...
+  def isAlphaNumericOrUnderscore(c: Char): Boolean = ALPHA_NUMERIC_AND_UNDERSCORE.contains(c)
+
   override def rewrite(code: Tree, ctx: RewriteCtx): Seq[Patch] = {
     import ctx.tokenTraverser._
+    import ctx.style.rewrite.{redundantBraces => settings}
     val builder = Seq.newBuilder[Patch]
     code.collect {
+      case t: Term.Interpolate if settings.stringInterpolation =>
+        t.parts.tail.zip(t.args).collect {
+          case (Lit(value: String), arg @ Term.Name(name)) =>
```
olafurpg:
```scala
def isIdentifierStart(value: String) = value.isEmpty || Character.isLetterOrDigit(value.head) || value.head == '_'
...
case xxx if !isIdentifierStart(value) =>
```

core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -49,10 +46,27 @@ object RedundantBraces extends Rewrite {
...
+            val openBrace = prevToken(arg.tokens.head)
+            val closeBrace = nextToken(arg.tokens.head)
+            (openBrace, closeBrace, value.headOption) match {
```
olafurpg: remove `, value.headOption`

core/src/test/resources/rewrite/RedundantBraces.stat
```diff
@@ -159,3 +160,59 @@ object a {
...
+object a {
+  s"Hello ${foo}5"
+}
+>>>
+object a {
+  s"Hello ${foo}5"
+}
+<<< not redundant enclosing braces in str interpolation (right part starts with underscore)
+object a {
+  s"Hello ${foo}_bar"
```
olafurpg: Maybe also test `s"${foo}æ"`

core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
ysusuk: is more like this

```scala
def isIdentifierStart(value: String) = value.nonEmpty && (Character.isLetterOrDigit(value.head) || value.head == '_')
```
when value is empty we don't want to skip patch

core/src/test/resources/rewrite/RedundantBraces.stat
ysusuk: added, thank you!

core/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -59,10 +58,8 @@ object RedundantBraces extends Rewrite {
           case (Lit(value: String), arg @ Term.Name(name)) =>
             val openBrace = prevToken(arg.tokens.head)
             val closeBrace = nextToken(arg.tokens.head)
-            (openBrace, closeBrace, value.headOption) match {
-              case (LeftBrace(), RightBrace(), Some(c)) if !isAlphaNumericOrUnderscore(c) =>
-                builder += Patch(openBrace, closeBrace, name)
-              case (LeftBrace(), RightBrace(), None) =>
+            (openBrace, closeBrace) match {
+              case (LeftBrace(), RightBrace()) if !isIdentifierAtStart(value) =>
```
olafurpg: can we move the `!isIdentifierAtStart(value)` guard to the `case Lit(value...` line?

core/src/test/resources/rewrite/RedundantBraces.stat
```diff
@@ -192,6 +192,14 @@ object a {
 object a {
   s"Hello ${foo}_bar"
 }
+<<< not redundant enclosing braces in str interpolation (right part starts with special symbol)
```
olafurpg: Can we move all of the "in str interpolation" tests into a separate file named `RedundantBracesStrInterpolation.stat`? Also maybe remove the "redundant braces" part since it's implicit for that test suite, I prefer to keep test names as short as possible.

olafurpg:
Yay \o/ thanks for going through all the comments.

I will try to get a release out tomorrow

PR #711, opened 2017-02-03
Another fix for `verticalMultilineAtDefinitionSite`

pjrt:
Yet another bug fix for `verticalMultilineAtDefinitionSite`. I suspect this should be the last one though. 

Fixes an issue where `verticalMultilineAtDefinitionSite` would incorrectly split defaulted parameters that called a method.

ie:
```
def someFunc(
   code: String = Defaults.getCode(a)(b)
  ): RT
```

### DIFF
--- a/core/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/core/src/main/scala/org/scalafmt/internal/Router.scala
@@ -435,7 +435,8 @@ class Router(formatOps: FormatOps) {
         val oneLinePerArg = OneArgOneLineSplit(open).f
         val paramGroupSplitter: PartialFunction[Decision, Decision] = {
           // Indent seperators `)(` by `indentSep`
-          case Decision(t @ FormatToken(_, rp @ RightParen(), _), _) =>
+          case Decision(t @ FormatToken(_, rp @ RightParen(), _), _)
+            if rp == lastParen || next(t).right.is[LeftParen] =>
             Decision(t,
                      Seq(
                        Split(Newline, 0)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -435,7 +435,8 @@ class Router(formatOps: FormatOps) {
         val oneLinePerArg = OneArgOneLineSplit(open).f
         val paramGroupSplitter: PartialFunction[Decision, Decision] = {
           // Indent seperators `)(` by `indentSep`
-          case Decision(t @ FormatToken(_, rp @ RightParen(), _), _) =>
+          case Decision(t @ FormatToken(_, rp @ RightParen(), _), _)
+            if rp == lastParen || !isCallSite(owners(rp)) =>
```
olafurpg: I think owners (rp) == leftOwner should also work. For pathological cases, you could for example define a new method inside the default parameter value block.

pjrt: Good point, added a test and fixed it.

core/src/test/resources/test/VerticalMultiline.stat
```diff
@@ -53,6 +53,51 @@ def format_![T <: Tree](
...
+  createdAt: Instant = Default.getInstant(a)(b))
+  (implicit ev: Parse[T],
+  ev2: EC)
+  : String
+>>>
+def format_![T <: Tree](
+    name: Name,
+    code: String = Defaults.code,
+    updatedAt: Instant = Instant.now(),
+    user: User = new User { def someFunc(): RT = () },
```
olafurpg: I have btw seen crazy code like this! iirc it was somewhere in twitter/summingbird ^_^

pjrt: I can gladly say I've never seen this; wasn't even sure it would parse :\

Issue #706, opened 2017-02-01
Migrate to scalafix patches


olafurpg:
Scalafix recently added new awesome patches that support high-level tree operations and low-level token operations. SEe re's the relevant section https://github.com/scalacenter/scalafix/blob/master/core/src/main/scala/scalafix/util/Patch.scala#L33-L44

Benefits include:
- Easier to solve conflicts (some patches compose)
- Faster (TokenPatches are grouped by single token)

It would be nice to use these in scalafmt, this would involve refactoring the existing rewrites to use the new patches.

NOTE. It might be worth considering putting the patches in scalameta/contrib  and depend on them from here https://github.com/scalameta/scalameta/pull/624

ysusuk:
Hi @olafurpg! will be glad to contribute on this. what do you think should be done? Replacing org.scalafmt.rewrite.Patch with scalafix.util.Patch (and its childs)

olafurpg:
After more thinking I think that patches should belong in scalameta/contrib, a new module that just got merged yesterday. Otherwise, we end up duplicating code + bugs! I think we'll try to use the scalafix patch implementation as a base for the scalameta/contrib patch. I think it's best I do the migration to scalameta/contrib myself, since the scalafix patch implementation is still experimental and not thoroughly tested yet.

Instead, I think we can go ahead with https://github.com/olafurpg/scalafmt/issues/707 straight away and get immediate performance benefits. Would you be interested in that? Before doing the refactoring, it would be really useful to benchmark the current patch implementation when multiple rewrites are configured.

olafurpg:
Fixed in #712

PR #712, opened 2017-02-03
#706: Migrate to scalafix patches

ysusuk:


ysusuk:
benchmark shows better results as well

core/src/main/scala/org/scalafmt/rewrite/ExpandImportSelectors.scala
```diff
@@ -20,18 +22,24 @@ object ExpandImportSelectors extends Rewrite {
                 }
             }.flatten
             val patchStr = expandedImport.mkString("\n")
-            Patch(`import`.parent.get.tokens.head,
-                  `import`.parent.get.tokens.last,
-                  patchStr)
+
+            //TODO move up this side effect
+            `import`.parent.get.tokens.tail.foreach { tok =>
```
olafurpg: It would be nice to add helper utilities for common operations like this. For example, `Patch.remove(Tree/Tokens)`

We can leave that for a separate PR!

olafurpg:
Whoops, the benchmark screenshots didn't appear at first. Now I see them, sweet 😎

core/src/main/scala/org/scalafmt/rewrite/ExpandImportSelectors.scala
ysusuk: agree!

PR #714, opened 2017-02-03
Fix #710: Extra space added for unary operations

### LINKED ISSUE #710: Unnecessary space for unary operator + _
Seems like the fix for #697  bug was too aggressive, it adds spaces when it's not necessary

```diff
-    assert(!_evaluatorSet, "Evaluator can be set at most once")
+    assert(! _evaluatorSet, "Evaluator can be set at most once")
```

pjrt:


core/src/main/scala/org/scalafmt/util/TokenOps.scala
```diff
@@ -221,7 +221,11 @@ object TokenOps {
   )
 
   def isSymbolicIdent(tok: Token): Boolean = tok match {
```
pjrt: The name of this function is confusing (or maybe I'm missing something). `isSymbolicIdent`? As in "is this a symbolic identifier"? Shouldn't this be `isOperator`?

pjrt: Oh wait, this could be an identifier for a class or type too. Nvm I guess.

olafurpg: Exactly, an identifier can be anything from a method, class, etc.

Issue #709, opened 2017-02-02
Feature request: AvoidInfix rewrite


olafurpg:
```diff
-          promise success "done"
+          promise.success("done")
```

I made progress on this rewrite in #517.

It got stuck because I was missing Patch.Add{Left,Right}, we can add those to the scalafmt patch impl

ysusuk:
cool, you made a lot on AvoidInfix! do you think i can reuse most of the staff?

olafurpg:
Yes, we can reuse most of that. I never merged it because the rewrite didn't handle the most tricky cases from scalatest where whitelisted infix operators were mixed together with blacklisted operators. This is really common in scalatest matcher code.

I would recommend you run the rewrite against projects such as akka that use scalatest matchers and look carefully at the diffs. For test data, you can clone this repo https://github.com/olafurpg/scala-repos It contains a ton of projects already formatted with scalafmt (branch name == version). I just pushed to branch 0.5.5

olafurpg:
btw, note that there exists a FilterMatcher now so there's no need to add the one from the old PR

ysusuk:
there is not such a branch (0.5.5) the highest is 0.5.1

olafurpg:
Whoops, I maybe forgot to push. I'll do it when I get back to my desk.

ysusuk:
not a problem) i'm having a hard time finding discrepencies %)

olafurpg:
Just pushed to 0.5.5 https://github.com/olafurpg/scala-repos/tree/0.5.5

One challenge with this rewrite is to be guaranteed that it doesn't change the semantic of the program. One way to do this is to check the normalized ast is unchanged.

You can normalise a tree to remove all infix applications like this

```scala
@   q"1 + 1 + 2 should be ( 1) ".transform {
      case Term.ApplyInfix(lhs, op, targs, args) =>
       if (targs.nonEmpty) q"$lhs.$op[..$targs](..$args)"
       else q"$lhs.$op(..$args)"
    }
res7: Tree = 1.+(1).+(2).should(be(1))
@ res7.structure == q"1.+(1).+(2).should(be(1))".structure
res8: Boolean = true
```

ysusuk:
hey @olafurpg! can you pls explain as well what did you mean by missing a  Patch.Add{Left,Right}

ysusuk:
in a WordSpec, should

```scala
class StackSpec extends WordSpec {
  "A Stack" when {
    "empty" should {
[...]

PR #715, opened 2017-02-05
#709: Feature request: AvoidInfix rewrite

ysusuk:
- this is initial commit
- works with 2+ infix applications
- tested on akka repo

### DIFF
--- a/core/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala
+++ b/core/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala
@@ -16,8 +16,7 @@ case object AvoidInfix extends Rewrite {
   override def rewrite(code: Tree, ctx: RewriteCtx): Seq[Patch] = {
     val matcher = ctx.style.rewrite.neverInfix.toMatcher
     code.collect {
-      case Term.ApplyInfix(lhs, op, targs, args)
-          if matcher.matches(op.value) =>
+      case Term.ApplyInfix(lhs, op, _, args) if matcher.matches(op.value) =>
         val fstOpToken = op.tokens.head
 
         val selectorToBeAdded = Seq(
@@ -27,7 +26,7 @@ case object AvoidInfix extends Rewrite {
         val fstArgsToken = args.head.tokens.head
         val lastArgsToken = args.last.tokens.last
         val isSingleArg = args.size == 1
-        val parensToBeAdded =
+        val selectorParensToBeAdded =
           if (isSingleArg
               && fstArgsToken.isNot[LeftParen]
               && fstArgsToken.isNot[LeftBrace])
@@ -35,6 +34,14 @@ case object AvoidInfix extends Rewrite {
                 TokenPatch.Add(lastArgsToken, "", ")", true))
           else
             Nil
+
+        val lhsParensToBeAdded = lhs match {
+          case Term.ApplyInfix(lhs1, op1, _, _)
+              if !matcher.matches(op1.value) =>
+            Seq(TokenPatch.Add(lhs.tokens.head, "(", "", true),
+                TokenPatch.Add(lhs.tokens.last, "", ")", true))
+          case _ => Nil
+        }
 
         val toBeRemoved = ctx.tokenTraverser
           .filter(fstOpToken, fstArgsToken)(_.is[LF])
@@ -47,7 +54,7 @@ case object AvoidInfix extends Rewrite {
         if (hasInlineComment)
           Nil
         else
-          selectorToBeAdded ++ parensToBeAdded ++ toBeRemoved
+          lhsParensToBeAdded ++ selectorToBeAdded ++ selectorParensToBeAdded ++ toBeRemoved
     }.flatten
   }
 }
--- a/core/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala
+++ b/core/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala
@@ -37,7 +37,8 @@ case object AvoidInfix extends Rewrite {
 
         val lhsParensToBeAdded = lhs match {
           case Term.ApplyInfix(lhs1, op1, _, _)
-              if !matcher.matches(op1.value) =>
+              if !matcher.matches(op1.value)
+                && lhs.tokens.head.isNot[LeftParen] =>
             Seq(TokenPatch.AddLeft(lhs.tokens.head, "(", keepTok = true),
                 TokenPatch.AddRight(lhs.tokens.last, ")", keepTok = true))
           case _ => Nil
--- a/core/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala
+++ b/core/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala
@@ -20,7 +20,7 @@ case object AvoidInfix extends Rewrite {
         val fstOpToken = op.tokens.head
 
         val selectorToBeAdded = Seq(
-          TokenPatch.Add(fstOpToken, ".", "", true)
+          TokenPatch.AddLeft(fstOpToken, ".", keepTok = true)
         )
 
         val fstArgsToken = args.head.tokens.head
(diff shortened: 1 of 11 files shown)


ysusuk:
[akka.diff.zip](https://github.com/olafurpg/scalafmt/files/753352/akka.diff.zip)

core/src/test/resources/rewrite/AvoidInfix.stat
```diff
@@ -0,0 +1,102 @@
...
+_.compareTo(_)
+<<< akka.ActivatorDist 1
+PathFinder(dir) descendantsExcept ("*", gitignoreFileFilter) filter (_.isFile)
+>>>
+PathFinder(dir).descendantsExcept("*", gitignoreFileFilter).filter(_.isFile)
+<<< akka.ActivatorDist 2
+filteredPathFinder pair Path.rebase(dir, activatorDistDirectory) map {
+  case (source, target) =>
+    s.log.info(s"copying: $source -> $target")
+    println("blah")
```
olafurpg: Are these two lines necessary for the test?

core/src/test/resources/rewrite/AvoidInfix.stat
```diff
@@ -0,0 +1,102 @@
...
+Iterator.continually(mq.dequeue) takeWhile (_ ne null) foreach System.err.println
+>>>
+Iterator
+  .continually(mq.dequeue)
+  .takeWhile(_ ne null)
+  .foreach(System.err.println)
+<<< SKIP type
+a map (_.asInstanceOf[AnyRef]): _*
+>>>
+a.map(_.asInstanceOf[AnyRef]): _*
```
olafurpg: Could we add a test like this?

foo + 1 bar 2

rewrites to 

(foo + 1).bar(2)

core/src/test/resources/rewrite/AvoidInfix.stat
```diff
@@ -0,0 +1,102 @@
...
+  case (source, target) =>
+    s.log.info(s"copying: $source -> $target")
+    println("blah")
+}
+<<< akka.typed.StubbedActorContext
+Iterator from 1 map (Helpers.base64(_))
+>>>
+Iterator.from(1).map(Helpers.base64(_))
+<<< akka.actor.dispatch.ActorModelSpec
+Iterator.continually(mq.dequeue) takeWhile (_ ne null) foreach System.err.println
```
olafurpg: Nice! :)

core/src/test/resources/rewrite/AvoidInfix.stat
```diff
@@ -0,0 +1,102 @@
...
+     // foobar
+}
+>>>
+an[SparkException] should be thrownBy {
+  // foobar
+}
+<<< nested infix
+a banana walked and orders a martini
+>>>
+a.banana(walked).and(orders).a(martini)
```
olafurpg: Great to see this work.

core/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  override def rewrite(code: Tree, ctx: RewriteCtx): Seq[Patch] = {
+    val matcher = ctx.style.rewrite.neverInfix.toMatcher
+    code.collect {
+      case Term.ApplyInfix(lhs, op, targs, args)
+          if matcher.matches(op.value) =>
+        val fstOpToken = op.tokens.head
+
+        val selectorToBeAdded = Seq(
+          TokenPatch.Add(fstOpToken, ".", "", true)
```
olafurpg: Could we use AddLeft/AddRight helpers to make this clearer to read? I think it might also help to name the remove parameter.

olafurpg:
Looked at Akka diff and found some problematic cases

```diff
-      terminated should be theSameInstanceAs Await.result(f, 10 seconds)
+      terminated should be.theSameInstanceAs(Await.result(f, 10 seconds)) // obtained
+      (terminated should be).theSameInstanceAs(Await.result(f, 10 seconds)) // expected
-      seq should have length 5
+      seq should have.length(5)
-          childProps2 should be theSameInstanceAs ActorCell.terminatedProps
+          childProps2 should be.theSameInstanceAs(ActorCell.terminatedProps)
-      1 to total foreach { _ ⇒
+      1 to total.foreach { _ ⇒
         perform(_ + 1)
       }


-        stay forMax (1.millis.dilated) replying Tick
+        stay.forMax(1.millis.dilated) replying Tick // obtained
+        stay.forMax(1.millis.dilated).replying(Tick) // expected
```

These were really nice :D 
```diff
                   System.err.println(
                     "Teammates left: " + team.size + " stopLatch: " +
                       stopLatch.getCount + " inhab:" + dispatcher.inhabitants)
-                  team.toArray sorted new Ordering[AnyRef] {
-                    def compare(l: AnyRef, r: AnyRef) = (l, r) match {
-                      case (ll: ActorCell, rr: ActorCell) ⇒
-                        ll.self.path compareTo rr.self.path
+                  team.toArray
+                    .sorted(new Ordering[AnyRef] {
+                      def compare(l: AnyRef, r: AnyRef) = (l, r) match {
+                        case (ll: ActorCell, rr: ActorCell) ⇒
+                          ll.self.path.compareTo(rr.self.path)
+                      }
+                    })
+                    .foreach {
+                      case cell: ActorCell ⇒
+                        System.err.println(
+                          " - " + cell.self.path + " " + cell.isTerminated +
+                            " " + cell.mailbox.currentStatus + " " +
+                            cell.mailbox.numberOfMessages + " " + cell.mailbox
+                            .systemDrain(SystemMessageList.LNil)
+                            .size)
                     }
-                  } foreach {
-                    case cell: ActorCell ⇒
-                      System.err.println(
-                        " - " + cell.self.path + " " + cell.isTerminated +
-                          " " + cell.mailbox.currentStatus + " " +
-                          cell.mailbox.numberOfMessages + " " + cell.mailbox
-                          .systemDrain(SystemMessageList.LNil)
-                          .size)
-                  }

-            future1 recover {
-              case e: ArithmeticException ⇒ 0
-            } map (_.toString)
+            future1
+              .recover {
+                case e: ArithmeticException ⇒ 0
+              }
+              .map(_.toString)

```

olafurpg:
Another interesting test case is build.sbt files. The sbt DSL overuses infix application, IMO. You might be able to find interesting test cases there, for example https://github.com/scalameta/scalameta/blob/master/build.sbt

core/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala
ysusuk: It is like this because of true (keepTok), by default i set it to false. False makes more sense for all rewrites except AvoidInfix. But i can introduce Token.AddLeft(token, keepTok)

core/src/test/resources/rewrite/AvoidInfix.stat
ysusuk: it was failing, because i was removing all the LFs in ApplyInfix.

olafurpg:
keepTok can stay false by default. I suggest to name the argument when it's
true, so the call site is like

AddLeft("foo", keepTok = true)

olafurpg:
Aha, good to know.

olafurpg:
Here's another problematic example I found
```diff
-            val future = actor1 ? "Hello" flatMap { case i: Int ⇒ actor2 ? i }
+            val future = actor1 ? "Hello".flatMap { case i: Int ⇒ actor2 ? i }

```

olafurpg:
You might find more interesting cases by running the rewrite against other projects like sbt, see https://github.com/olafurpg/scala-repos/pull/17 feel free to open PRs against the 0.5.5 branch.

ysusuk:
All mentioned test cases are covered!
missing thing:
- strict exclude of combinators

``` diff
-        stay forMax (1.millis.dilated) replying Tick
+        stay.forMax(1.millis.dilated) replying Tick // obtained
+        stay.forMax(1.millis.dilated).replying(Tick) // expected
```

doesn't work. because `replying` contains `in`

olafurpg:
Amazing! ^_^ That test suite  😍 

You can create a custom FilterMatcher (bypassing apply) with a strict regex wrapped between `^` and `$`. FilterMatcher is just a dumb container for two regexes.

ysusuk:
looking into sbt project =)

olafurpg:
btw, you can also add a `toStrict` method on `FilterMatcher` that returns a new FilterMatcher with regex.value wrapped in `^$`

olafurpg:
Here's the diff from running this rewrite compared to 0.5.5 (ignore the `! _` diffs) https://github.com/olafurpg/scala-repos/pull/18

olafurpg:
I found one non-problematic corner case

```diff
-        (0 to 1) map { _ =>
+        ((0 to 1)).map { _ =>
```
This is not at all a blocker. 

We also might want to make it configurable to skip the rewrite for cases like `a op b` or `_ op _`. Basically, when both sides are single tokens.

olafurpg:
Another case that might be nice to handle

```diff
-      assert((Top compare Undefined) < 0)
-      assert((Bottom compare Top) < 0)
-      assert((Undefined compare Undefined) == 0)
-      assert((Top compare Top) == 0)
-      assert((Bottom compare Bottom) == 0)
+      assert((Top.compare(Undefined)) < 0)
+      assert((Bottom.compare(Top)) < 0)
+      assert((Undefined.compare(Undefined)) == 0)
+      assert((Top.compare(Top)) == 0)
+      assert((Bottom.compare(Bottom)) == 0)
```
These original parentheses are now redundant so it'd be nice to remove them.

Having gone through a lot of diffs, I am really satisfied with the current implementation. I would like to include this PR it in the next release. What do you think?

olafurpg:
One last thing, could you please document this rewrite in Configuration.scalatex along with the other rewrites? A quick explanation for the configuration would also be good, same as RedundantBraces

ysusuk:
double paren case is fixed, i'm working on another one (didn't have much time today), when it's finished i think we are good to go.

ysusuk:
last push should fix most of the problems ^_^

ysusuk:
i will try to add docs today, not sure how much time will i have!

Issue #724, opened 2017-02-09
Update rewrite rules docs.


ysusuk:
Add documentation for missing rewrite rules:
- [x] AvoidInfix
- [x] ExpandImportSelectors

PR #725, opened 2017-02-09
 #724: Update rewrite rules docs

ysusuk:


readme/Configuration.scalatex
```diff
@@ -423,6 +423,27 @@
       @b{NEW}. Semantically preserving rewrites. To enable a
       rewrite rule, add it to the config like this @b{rewrite.rules = [SortImports]}.
 
+    @sect(Rewrite.rewrite2name(AvoidInfix))
+      @demoStyle(rewriteInfix)
+        future recover {
+          case e: ArithmeticException ⇒ 0
+        } map (_.toString)
+
```
olafurpg: Can we also document the include/exclude filters?

readme/Configuration.scalatex
```diff
@@ -423,6 +423,27 @@
       @b{NEW}. Semantically preserving rewrites. To enable a
```
olafurpg: While you're at it, can you remove "NEW"?

readme/Configuration.scalatex
```diff
@@ -423,6 +423,27 @@
...
+        import a.{
+            b,
+            c
+          }, h.{
+            k, l
+          }
+          import d.e.{f, g}
+          import a.{
+              foo => bar,
+              zzzz => _
```
olafurpg: What happens in this case?

```scala
import a.{b => c, d => _, _}
// this is OK
import a.{b => c, d => _, _}
// this is bug
import a.{b => c}
import a.{d => _}
import a. _
```

If it's a bug then we should open a ticket.

ysusuk: https://github.com/olafurpg/scalafmt/issues/726

olafurpg: Thanks! I bumped into this recently while implementing "organize imports" in scalafix. Didn't know renames and unimports needs to be bundled together with the wildcard.

Issue #726, opened 2017-02-09
unimport is not respected


ysusuk:
This template is a guideline, not a strict requirement.
- **Version**: 0.5
- **Integration**: IntelliJ
- **Configuration**:
```
maxColumn = 100
```

## Unexpected formatting output

Obtained, scalafmt formats code like this:
```scala
import a.{b => c}
import a.{d => _}
import a. _
```

Expected, I would like the formatted output to look like this:
```scala
import a.{b => c, d => _, _}
```

The code in question was originally formatted like this:
```scala
import a.{b => c, d => _, _}
```

PR #729, opened 2017-02-10
#726: unimport is not respected

ysusuk:


core/src/main/scala/org/scalafmt/rewrite/ExpandImportSelectors.scala
```diff
@@ -7,19 +7,38 @@ object ExpandImportSelectors extends Rewrite {
...
-                    else
-                      s"import $path.$importee"
-                }
+              case importer @ Importer(path, importees) =>
+                val hasRenames = importees.collect {
+                  case importee: Importee.Rename =>
+                    importee
+                }.nonEmpty
+
+                val hasUnimports = importees.collect {
```
olafurpg: Can this be done with importees.exists(X => X.is[foo] || X.is[bar])?

Issue #732, opened 2017-02-13
Use verticalMultilineAtDefinitionSite in default style


olafurpg:
Example of how the flag works: https://github.com/olafurpg/scalafmt/blob/5643d5d8948bf895bd24ee801fac758edf6c29af/core/src/test/resources/test/VerticalMultiline.stat

The default style doesn't handle curried parameter lists so well, IMO. I think `verticalMultilineAtDefinitionSite` is offers a more regular/simpler solution that scales for really long parameter lists.

Discussion points:

- Is this a good idea?
- Is there a better name for the flag?

pjrt:
Pasting what I said before here:

> ...the Scala style guide doesn't mention how vertical multiline methods should be defined. So `verticalMultilineAtDefinitionSite` is as official as anything else

Ofc I am biased since I made the flag so more input would be better.

jeantil:
alternate style suggestions
```
// curried function mixed style 3 ?
def format_![T <: Tree](code: String, foo: Int)
                                     (f: A => B, k: D)
                                     (implicit ev: Parse[T], ev2: EC): String
```
```
// curried function mixed style 4 
def format_![T <: Tree]
        (code: String, foo: Int)
        (f: A => B, k: D)
        (implicit ev: Parse[T], ev2: EC): String
```

olafurpg:
@jeantil style 4 has been discussed in several other issues (see https://github.com/olafurpg/scalafmt/issues/337), its problem is that the vertical alignment doesn't scale well with maxColumn.

Style 5 is interesting but I am afraid the leading commas has never been picked up by the Scala community. It is much more common in Haskell/Elm/Purescript circles.

The reason I like verticalMultilineAtDefinitionSite is that it looks similar to existing popular coding conventions in the Scala community AND it scales well with the introduction of new parameters, parameter lists, and really long parameter names.

pjrt:
I'm a Haskeller myself so I love leading commas. But that's just not how the Scala community rolls. I don't 100% understand `style 3`.

jeantil:
I realized style 3 was damaged by editing it in github non fixed-font web editor. I fixed it now

jeantil:
Basically style 4 goes to style 5 of one of the argument lists becomes to long. this gives you a compromise of compact yet readable code. 

[...]

Issue #730, opened 2017-02-11
Roadmap for v1.0


olafurpg:
Scalafmt v0.1 came out in March 2016. Here are some stats in January 2017

- 3 maintainers (including myself)
- 40 contributors
- ~10k downloads from Maven last month alone
- ~1k fat jar downloads from github releases last month alone
- IntelliJ plugin is at ~23k downloads in total, with >6k installs on the latest 0.5.x series
- 56 releases (!!)

It's been a really great year, and I'm super thankful for everyone who has been involved in the project so far. I have enjoyed it a LOT and I look forward to continue working on scalafmt.

The formatting output is fairly stable now (big changes like in v0.5.0 to the default style typically provide fallback config for old styles). I think now is a good time to start thinking about releasing v1.0.

I saw somewhere on the interwebs.

> If you want to go quickly, go alone. If you want to go far, go together

I admit it's a bit cheezy, but it does strike a chord with me.

These are organizational things I'd like to get done first

- [x] Move the project under the umbrella of [scala.meta](http://scalameta.org/). I've discussed about this with @xeno-by and we both believe this is a good idea
- [x] Publish under org.scalameta instead of com.geirsson
- [x] Move docs to scalameta.org/scalafmt
- [x] publish docs on merge into master
- [ ] Automatic releases on push tag
- [x] remove unsupported/dangerous options (e.g., bestEffortInDeeplyNestedFunctions, binpack.defnsite)
- [x] #875
I'd be interested to hear what people think.

pjrt:
Something that has been bugging me, and maybe 1.0 is the place to do this: do we want Scalafmt to be a fine-grain or an opinionated, holistic auto-formatting tool?

* The tool started fine-grain: things like how many spaces to indent stuff to and whether to put a new line here or there.
  * Advantage: You can pretty much make your own style
  * Disadvantage: Confusing UI since settings might conflict with one another
* Now we have some settings that throw all of that away and do a holistic approach, like the oddly named `verticalMultilineAtDefinitionSite`.
  * Advantage: It becomes clearer what you code will end up looking like and there are no conflict
  * Disadvantage: If you don't like it, well tough luck. 

[...]

PR #733, opened 2017-02-13
Add flag for newline before toplevel statements

Daxten:
Hi, thanks for the help in Gitter!
Here is my first try to add a  flag for adding newlines before toplevel statements

### DIFF
--- a/core/src/main/scala/org/scalafmt/config/Newlines.scala
+++ b/core/src/main/scala/org/scalafmt/config/Newlines.scala
@@ -30,5 +30,6 @@ case class Newlines(
     neverBeforeJsNative: Boolean = false,
     sometimesBeforeColonInMethodReturnType: Boolean = true,
     penalizeSingleSelectMultiArgList: Boolean = true,
-    alwaysBeforeCurlyBraceLambdaParams: Boolean = false
+    alwaysBeforeCurlyBraceLambdaParams: Boolean = false,
+    beforeTopLevelStatements: Boolean = false
 )
--- a/core/src/main/scala/org/scalafmt/config/Newlines.scala
+++ b/core/src/main/scala/org/scalafmt/config/Newlines.scala
@@ -31,5 +31,5 @@ case class Newlines(
     sometimesBeforeColonInMethodReturnType: Boolean = true,
     penalizeSingleSelectMultiArgList: Boolean = true,
     alwaysBeforeCurlyBraceLambdaParams: Boolean = false,
-    beforeTopLevelStatements: Boolean = false
+    alwaysBeforeTopLevelStatements: Boolean = false
 )
--- a/core/src/main/scala/org/scalafmt/util/TokenOps.scala
+++ b/core/src/main/scala/org/scalafmt/util/TokenOps.scala
@@ -58,7 +58,7 @@ object TokenOps {
       val newlines = newlinesBetween(tok.between)
       newlines > 1 || (isDocstring(tok.right) && !tok.left.is[Comment])
     } || {
-      style.newlines.beforeTopLevelStatements &&
+      style.newlines.alwaysBeforeTopLevelStatements &&
         tok.between.count(_.is[KwNew]) < 2 &&
         (tok.right.is[KwObject] || tok.right.is[KwDef] || tok.right.is[KwDef] || tok.right.is[KwCase] || tok.right.is[KwPackage])
     }
--- a/core/src/main/scala/org/scalafmt/util/TokenOps.scala
+++ b/core/src/main/scala/org/scalafmt/util/TokenOps.scala
@@ -8,6 +8,7 @@ import scala.meta.dialects.Scala211
 import scala.meta.tokens.Token
 import scala.meta.tokens.Token._
 
+import org.scalafmt.config.ScalafmtConfig
 import org.scalafmt.internal.Decision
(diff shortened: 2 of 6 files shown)


core/src/main/scala/org/scalafmt/config/Newlines.scala
```diff
@@ -30,5 +30,6 @@ case class Newlines(
     neverBeforeJsNative: Boolean = false,
     sometimesBeforeColonInMethodReturnType: Boolean = true,
     penalizeSingleSelectMultiArgList: Boolean = true,
-    alwaysBeforeCurlyBraceLambdaParams: Boolean = false
+    alwaysBeforeCurlyBraceLambdaParams: Boolean = false,
+    beforeTopLevelStatements: Boolean = false
```
olafurpg: `alwaysBeforeTopLevelStatements`?

pjrt:
I can see why we want this as a flag now (not to break everything) but I think this should eventually be the default, along with #732 .

Maybe for v1? #730

core/src/main/scala/org/scalafmt/config/Newlines.scala
pjrt: Could you add some doc for this variable?

core/src/main/scala/org/scalafmt/util/TokenOps.scala
```diff
@@ -52,10 +53,14 @@ object TokenOps {
...
 
-  def shouldGet2xNewlines(tok: FormatToken): Boolean = {
+  def shouldGet2xNewlines(tok: FormatToken, style: ScalafmtConfig): Boolean = {
     !isDocstring(tok.left) && {
       val newlines = newlinesBetween(tok.between)
       newlines > 1 || (isDocstring(tok.right) && !tok.left.is[Comment])
+    } || {
+      style.newlines.beforeTopLevelStatements &&
+        tok.between.count(_.is[KwNew]) < 2 &&
+        (tok.right.is[KwObject] || tok.right.is[KwDef] || tok.right.is[KwDef] || tok.right.is[KwCase] || tok.right.is[KwPackage])
```
olafurpg: FYI, this won't cover modifiers like `final class` or `sealed trait`. Depending on your use-case, you might want to support that. There's a `owners: Token => Tree` available in Router that you can use to get the tree parent of the token. Note that the owner of modifiers like `lazy` is the Mod tree while the owner of `class` for exampl is Defn.Class.

btw, KwDef appears twice.

pjrt: I would think modified top levels (`final`, `sealed`) would also need the newline. 

Also, maybe refactor that into a "isTopLevelStatment" function in FormatOps?

olafurpg: There is a `startsStatement: Set[Token]` iirc, if that was a Map[Token, Tree] then we could more easily support that

Daxten:
So, I've added a failing testcase and added everything which is needed (I think?) but I can't find a solution which works using owners, can someone give me a hand?

Daxten:
Thanks @olafurpg wasn't so hard after all!

core/src/main/scala/org/scalafmt/util/TokenOps.scala
```diff
@@ -52,13 +54,26 @@ object TokenOps {
...
     }
   }
 
+  def isTopLevelStatment(tok: Token, owners: Token => Tree): Boolean = {
+      tok.is[KwObject] ||
+        tok.is[KwDef] ||
+        tok.is[KwCase] ||
+        tok.is[KwPackage] ||
+        tok.is[KwClass] ||
+        owners(tok).parent.exists(_.is[Defn.Class]) ||
```
olafurpg: Can we combine these two lines? I would suggest to replace the Kw* matches with matches against the owner tree.

core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
```diff
@@ -55,9 +61,17 @@ package a {
 
       case class C1
 
+      sealed trait T1
+
+      final class FC1
```
olafurpg: These blank lines would be preserved with default

Daxten: Yes, that's correct, since this Test just tests if the new formatter doesn't add any new lines in this case

core/src/main/scala/org/scalafmt/util/TokenOps.scala
Daxten: the KwClass is actualy redundant yes, I removed that. Do you mean that also the rest should be changed?

readme/Configuration.scalatex
```diff
@@ -204,6 +204,35 @@
        *
        */
 
+  @sect{newlines.alwaysBeforeTopLevelStatements}
+    Default: default.newlines.alwaysBeforeTopLevelStatements
+
+    @h1.scala
```
olafurpg: This is failing the CI, I think it should be `hl.scala`.

olafurpg:
Thank you for this contribution @Daxten!

Daxten:
maybe add a second flag to disable it for objects ( / packages ) which only contain class (/ case / traits) definitions? Dunno if this gets too spaghetti

PR #737, opened 2017-02-14
Grouping of case [class|object] for newline.alwaysBeforeTopLevelStatements and fixed new test case

Daxten:
add failing test for match case and newline.alwaysBeforeTopLevelStatements,
fixed the failing test,
keep groups of case [class|object] with no body together

### DIFF
--- a/core/src/main/scala/org/scalafmt/util/TokenOps.scala
+++ b/core/src/main/scala/org/scalafmt/util/TokenOps.scala
@@ -77,13 +77,11 @@ object TokenOps {
   }
 
   def isTopLevelStatment(tok: Token, owner: Tree): Boolean = {
-      tok.is[KwObject] ||
-        tok.is[KwDef] ||
-        tok.is[KwPackage] ||
-        tok.is[KwClass] ||
-        tok.is[KwTrait] ||
-        owner.parent.exists(_.is[Defn.Class]) ||
-        owner.parent.exists(_.is[Defn.Trait])
+      tok.is[KwObject] || owner.parent.exists(_.is[Defn.Object]) ||
+        tok.is[KwClass] || owner.parent.exists(_.is[Defn.Class]) ||
+        tok.is[KwDef] || owner.parent.exists(_.is[Defn.Def]) ||
+        tok.is[KwTrait] || owner.parent.exists(_.is[Defn.Trait]) ||
+        tok.is[KwPackage] || (tok.is[KwProtected] && owner.parent.exists(_.is[Pkg]))
   }
 
   def isDocstring(token: Token): Boolean = {
--- a/core/src/main/scala/org/scalafmt/util/TokenOps.scala
+++ b/core/src/main/scala/org/scalafmt/util/TokenOps.scala
@@ -60,11 +60,9 @@ object TokenOps {
       val rightOwner = owners(tok.right)
 
       style.newlines.alwaysBeforeTopLevelStatements &&
-        tok.between.count(_.is[KwNew]) < 2 &&
         isTopLevelStatment(tok.right, rightOwner) &&
         !{
           // keep groups of case class/objects with no body in a block
-          
           def hasBody(tree: Tree) = tree.parent.map(_.children).getOrElse(Seq.empty).exists {
             case Template(_, _, _, Some(_)) => true
             case _ => false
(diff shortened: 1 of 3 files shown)


Daxten:
I think it will also fail for override def atm, will check that later

Daxten:
Okay I found some time right now, this should be complete now 👍

core/src/main/scala/org/scalafmt/util/TokenOps.scala
```diff
@@ -59,18 +56,32 @@ object TokenOps {
...
     } || {
+      val leftOwner = owners(tok.left)
+      val rightOwner = owners(tok.right)
+
       style.newlines.alwaysBeforeTopLevelStatements &&
-        tok.between.count(_.is[KwNew]) < 2 && isTopLevelStatment(tok.right, owners(tok.right))
+        isTopLevelStatment(tok.right, rightOwner) &&
+        !{
+          // keep groups of case class/objects with no body in a block
+          def hasBody(tree: Tree) = tree.parent.map(_.children).getOrElse(Seq.empty).exists {
```
pjrt: Why not move this to TreeOps too? It is very odd looking at this function inside a bool body :\

Daxten: wasn't sure if this is really a generic way to check for a body, it works in this case.. if you can confirm that it's generic enough I will move it

core/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -460,4 +460,24 @@ object TreeOps {
...
     defn.decltpe.exists(_.tokens.isEmpty)
+
+  def isCaseClass(tree: Tree): Boolean = {
+    tree.is[Defn.Class] && tree.children.exists(_.is[Mod.Case]) ||
+      tree.parent.exists(_.is[Defn.Class]) && tree.parent.exists(_.children.exists(_.is[Mod.Case]))
+  }
+
+  def isCaseObject(tree: Tree): Boolean = {
+    tree.is[Defn.Object] && tree.children.exists(_.is[Mod.Case]) ||
+      tree.parent.exists(_.is[Defn.Object]) && tree.parent.exists(_.children.exists(_.is[Mod.Case]))
```
olafurpg: Here is how I would implement this

```scala
def isCaseObject(tree: Tree): Boolean = tree match {
  case defn: Defn.Object => defn.mods.exists(_.is[Mod.Case])
  case _ => false
}
```

The current implementation would return true for a `Mod` or `Ctor.Primary`, which is just false.

If you have a `mod: Mod`, you can do `mod.parent.exists(isCaseObject)`.

Think of `Tree` as `Object` or `Any`, we should avoid working with it where possible.

core/src/main/scala/org/scalafmt/util/TokenOps.scala
```diff
@@ -59,18 +56,32 @@ object TokenOps {
...
+            case Template(_, _, _, Some(_)) => true
+            case _ => false
+          }
+
+          (TreeOps.isCaseClass(leftOwner) || TreeOps.isCaseObject(leftOwner)) &&
```
olafurpg: Is there a need to distinguish `case class` from `final class`? I think this rule could apply to all `Defn.{Object,Trait,Class}`

core/src/main/scala/org/scalafmt/util/TreeOps.scala
Daxten: true, maybe move it to a parameter, since in my limited experience I mostly want to know if a token is part of the definition?

```
  def isCaseClass(tree: Tree, matchOnParent: Boolean = false): Boolean = tree match {
    case defn: Defn.Class => defn.mods.exists(_.is[Mod.Case])
    case _ if matchOnParent => tree.parent.exists(isCaseClass(_))
    case _ => false
  }
```

core/src/main/scala/org/scalafmt/util/TokenOps.scala
Daxten: You are right, but we have at least to distingush between them since if we have 1 Block of traits and 1 Block of sealed traits we want to have newlines between the blocks
```
trait A
trait B
trait C

sealed trait ST1
sealed trait ST2
```

Issue #720, opened 2017-02-08
Feature request: scalafmt --pre-commit


olafurpg:
It would be nice if you could put in .git/hooks/pre-commit
```
scalafmt --pre-commit
```
and that would reformat changed files before each commit.

pjrt:
Shouldn't that be the job of a pre-commit script? What would that flag do?

olafurpg:
The pre commit hook would use this flag, eliminating the need for complex
bash scripts. It's possible to pipe the output of git diff name only into
--files but that doesn't handle project.excludeFilters and such.

pjrt:
> git diff name only into
--files but that doesn't handle project.excludeFilters and such.

Couldn't we just make `--files` handle excludeFilters then? If we can get that then we could just make a `.sh` script that people can copy over into their pre-commit hooks.

bjornj12:
@olafurpg My 2 cents on it is calling it --modified, i think that is what git calls it ?
with --files you also need to go through some extra work of filtering out non .scala files, working on this right now, can give better feedback soon.

This is a simple command i have gotten up to find all modified files and get it into scalafmt.

```
$ scalafmt -i -f $(git status --porcelain | cut -c 4- | tr '\n' ',')
```

Missing to exclude everything that is not .scala

For example is scalafmt erroring for me right now because in my modified is 

```
modified:   .scalafmt.conf
```

But @pjrt is kind of right though, that the pre-commit is really simple tbh, atleast for me as it is now, even though its slightly broken.

I think it would be more nice to have it as scalafmt flag, to give the user of scalamft the flexibility of where to use it, but a easy way to use it.

For example i am running a build tool called Pants, and i would love to later on to write a pants plugin that runs ``` scalafmt --modified ``` before compiling or something similar.

bjornj12:
Im sorry for late update, but i got stuck at doing something else, 
I ended up with a simple command layer to apply scalafmt on all modified files, using bash

```
scalafmt -i -f $(git status --porcelain | grep -e .scala$ | cut -c 4- | tr '\n' ',')
```

Its not hugely battle tested, so there might still be some bugs with it. but it atleast got me started using a pre-commit hook for now. Thought it could be usefull for someone.

pjrt:
@bjornj12 FYI: you can do `git diff --name-only` and it will give it the file path on its own (so no need to `cut` afterwards).

olafurpg:
Sry for the slow response, been swamped at work lately.

[...]

PR #742, opened 2017-02-22
Remove 'HEAD' from git diff, to include all uncommitted changes aswell

bjornj12:
helps to solve: #720

This should help scalafmt --diff to catch all uncommitted changes that differ from 'branch' along with committed changes.


olafurpg:
LGTM 👍 I swear the same thing failed in CI when I tried the first time, git complained that there was missing another argument. Glad to see it now works.

PR #750, opened 2017-02-26
fix isTopLevelStatement for edge-cases

Daxten:


olafurpg:
Could you include an accompanying test to test this fix?

Daxten:
ah sorry forgot to add that, I extended the original test case with some code which was failing for me

core/src/main/scala/org/scalafmt/util/TokenOps.scala
```diff
@@ -65,12 +65,11 @@ object TokenOps {
   }
 
   def isTopLevelStatment(tok: Token, owner: Tree): Boolean = {
-      tok.is[KwObject] ||
-        tok.is[KwDef] ||
-        tok.is[KwCase] ||
-        tok.is[KwPackage] ||
-        owner.parent.exists(_.is[Defn.Class]) ||
-        owner.parent.exists(_.is[Defn.Trait])
+    tok.is[KwObject] || owner.parent.exists(_.is[Defn.Object]) ||
```
olafurpg: Is the ` || ` part necessary? I think a match against `owner` might be sufficient.

Daxten: no the test fails without it:

```
package b {
       object a {
         val v1 = 1
         val v2 = 2
``` 

there should be whitespace between { and object in this case

olafurpg: Sure. LGTM 👍

olafurpg:
Thanks for fixing this!

olafurpg:
I'd love to have a system to publish on merge and all integrations (including intellij) could pickup the version number from `.scalafmt.conf`, that way you could enjoy this fix immediately ;)

Daxten:
at work we are using gitlab ci for stuff like that, but I bet it should also be possible with travis. Have you took a look at https://github.com/sbt/sbt-release ?

olafurpg:
Releasing artifacts to maven is easy, the challenge is that scalafmt has several integrations (homebrew, github releases, intellij) that require custom release steps. The IntelliJ plugin is the worst, I need to build the plugin via the IntelliJ ui and upload a zip via their web interface. There's an sbt plugin to automate it but I never got it working

olafurpg:
think it's just easier to use coursier to fetch maven artifacts and classload scalafmt, then it's not necessary to upload a new intellij plugin for every release. Would make it possible to use the intellij plugin with different scalafmt versions between proejcts too

olafurpg:
the sbt plugin already works like this, shouldn't be hard to do the same for intellij!

PR #760, opened 2017-03-03
Simplify build so `sbt test` "just works"

olafurpg:
- Don't require IDEA, it's slow and only relevant for intellij plugin development
- Add downloadIdea alias for intellij/updateIdea
- Automatically set file.encoding via .jvmopts
- Document gotchas in readme

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,7 +1,10 @@
 build:
   image: olafurpg/scalafix:0.0.1
   commands:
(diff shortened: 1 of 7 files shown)


olafurpg:
Yay, finally got this passing through travis-ci. @AnamikaD this change makes the `intellij/compile` error you bumped into go away.

AnamikaD:
@olafurpg Success!
screenshot from 2017-03-05 21-15-32

PR #761, opened 2017-03-03
Clarify config values & examples some

dwijnand:
Present the examples of the configurations in a more explicit, clarified
and format-consistent (oh the irony!) way.

As an example the comments "// ScalaDoc" and "// JavaDoc" weren't clear
if they were just giving an examples of ScalaDoc vs JavaDoc style, or if
they were the valid values to pass to "docstrings".

### DIFF
--- a/readme/Configuration.scalatex
+++ b/readme/Configuration.scalatex
@@ -195,11 +195,12 @@
     Default: @b(default.docstrings.toString)
 
     @hl.scala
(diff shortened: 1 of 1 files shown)


olafurpg:
DRY was the reason the full config path was excluded in the examples, this is purely for my personal convenience but I agree that this PR makes the docs clearer/better for users. Thanks!

PR #766, opened 2017-03-03
Generalise and structure the issue template

dwijnand:
Not all issues will be in the reduced form "I obtained ...; I would like ..".

So I generalised it a little, but kept the original template.

WDYT?


dwijnand:
> this looks similar to the sbt issue template

a strange coincidence

Issue #763, opened 2017-03-03
Varargs expansion shouldn't be split


dwijnand:
- **Version**: 0.5.7
- **Integration**: CLI
- **Configuration**:
```
    style = defaultWithAlign
maxColumn = 112
```

## Unexpected formatting output

Obtained, scalafmt formats code like this:
```scala
  assemblyMergeStrategy in assembly := {
    case PathList(ps @ _ *) if ps.last == "javax.inject.Named"      => MergeStrategy.first
    case PathList(ps @ _ *) if ps.last endsWith ".class"            => MergeStrategy.first
    case PathList(ps @ _ *) if ps.last endsWith "module.properties" => MergeStrategy.first
    case PathList(ps @ _ *) if ps.last == "MANIFEST.MF"             => MergeStrategy.rename
    case "LICENSE"                                                  => MergeStrategy.first
    case "NOTICE"                                                   => MergeStrategy.first
  }
```

Expected, I would like the formatted output to look like this:
```scala
  assemblyMergeStrategy in assembly := {
    case PathList(ps @ _*) if ps.last == "javax.inject.Named"      => MergeStrategy.first
    case PathList(ps @ _*) if ps.last endsWith ".class"            => MergeStrategy.first
    case PathList(ps @ _*) if ps.last endsWith "module.properties" => MergeStrategy.first
    case PathList(ps @ _*) if ps.last == "MANIFEST.MF"             => MergeStrategy.rename
    case "LICENSE"                                                 => MergeStrategy.first
    case "NOTICE"                                                  => MergeStrategy.first
  }
```

For clarity: I would like it to format `@ _*` not `@ _ *`.

olafurpg:
Thank you for reporting! This is probably an easy fix, a good place to start would be to follow the tutorial and track where that space comes from https://github.com/olafurpg/scalafmt/blob/b52acf0d78ab7d677b6eb807b18486e9acf4111f/Tutorial.md

I suggest we keep this fix behind a configuration flag named `spaces.beforeSeqWildcard`, which will be true for now but default to false on the next 0.x./1.0 release.

Fyi,

```scala
@ q"{case PathList(ps @ _ *) => MergeStrategy.first}".structure
res2: String = """
[...]

PR #772, opened 2017-03-03
#763 Fix Varargs expansion shouldn't be split

tgodzik:


dwijnand:
@olafurpg how do you feel about switch this to be the default behaviour?

If you need motivation, @eed3si9n has dug up that it's `_*` in the Scala Spec! https://www.scala-lang.org/files/archive/spec/2.12/08-pattern-matching.html#pattern-sequences

olafurpg:
@dwijnand I've opened https://github.com/scalameta/scalafmt/issues/890 to discuss changing a few fields in the default style to match the sbt settings.

Issue #774, opened 2017-03-03
Don't want a new line after enabling alwaysBeforeTopLevelStatements


d1egoaz:
- **Version**: 0.5.8
- **Integration**: shell
- **Configuration**:
```
newlines.alwaysBeforeTopLevelStatements = true
```

## Steps

I ran scalafmt like this:

```bash
scalafmt
```

## Problem

Scalafmt formats code like this:
```scala
  /** test */

  def foo(): Int = 0

  // bar

  def foo(): Int = 0
```

## Expectation

I would like the formatted output to look like this:
```scala
  /** test */
  def foo(): Int = 0

  // bar
  def foo(): Int = 0

```

olafurpg:
cc/ @Daxten

olafurpg:
This issue was fixed by @Daxten

diego-alvarez-hs:
tyvm!

PR #797, opened 2017-03-07
[WIP] fix for #774; no newline after comments

Daxten:


Daxten:
Should I make a new pull request and squash this, I think to remember that you squash these anyway?

PR #803, opened 2017-03-07
Fix #798.

### LINKED ISSUE #798: `::` is not correctly indented
## Steps

The following test fails:

```
<<< Indent `::` with tuples
val overrides =
      "akka.log-config-on-start" -> "on" ::
      "akka.cluster.split-brain-resolver.static-quorum.quorum-size" -> 1 ::
        "akka.remote.netty.tcp.hostname" -> akkaHost ::
        "akka.remote.netty.tcp.port" -> akkaPort ::
        Nil
>>>
  val overrides =
    "akka.log-config-on-start" -> "on" ::
      "akka.cluster.split-brain-resolver.static-quorum.quorum-size" -> 1 ::
      "akka.remote.netty.tcp.hostname" -> akkaHost ::
      "akka.remote.netty.tcp.port" -> akkaPort ::
      Nil
```

olafurpg:
infixSplit is used for both infix applications, where we want to handle
infix operators, and also for assignment, where we don't want to handle
right associative operators.

This commit is a poor fix/hack/workaround for this issue. I would love
to come with a better design/implementation to avoid this duct taping,
but I'm afraid I don't have so much time right now and I want to publish
a fix quickly.

### DIFF
--- a/core/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/core/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -476,6 +476,8 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
     }
     val isRightAssociative =
       style.indent.rightAssociativeInfixOperatorsLikeLeftAssociative &&
+        // NOTE. Silly workaround because we call infixSplit from assignment =, see #798
+        formatToken.left.syntax != "=" &&
         isRightAssociativeOperator(op.value)
     val expire = (for {
       arg <- {
--- a/core/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/core/src/main/scala/org/scalafmt/internal/Router.scala
@@ -888,7 +888,6 @@ class Router(formatOps: FormatOps) {
           insideBlock(formatToken, expire, _.isInstanceOf[LeftBrace])
         rhs match {
           case t: Term.ApplyInfix =>
-            val modification = newlines2Modification(between)
             Seq(
               infixSplit(t, formatToken)
             )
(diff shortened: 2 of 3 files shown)


olafurpg:
@pjrt The issue is that `infixSplits` is called from here https://github.com/olafurpg/scalafmt/blob/c53f03f7744bd04bef3e56fdcaea6294de48ec1e/core/src/main/scala/org/scalafmt/internal/Router.scala#L893, where `formatToken` is `= "..."` and not the `::` infix application. I'm not sure how to gracefully avoid the right-associativity handling when infixSplit is called from that case, any ideas/suggestions welcome.

Otherwise, I'm OK with just using this hack . I'm hoping to replace Router.scala this summer with the help of a GSoC student http://www.scala-lang.org/gsoc/2017.html#new-line-wrapping-algorithm-for-scalafmt

core/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -476,6 +476,8 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
     }
     val isRightAssociative =
       style.indent.rightAssociativeInfixOperatorsLikeLeftAssociative &&
+        // NOTE. Silly workaround because we call infixSplit from assignment =, see #798
+        formatToken.left.syntax != "=" &&
```
pjrt: So the bug happens when making a plain assignment? So the previous test passed because the "left" was never `=`.

pjrt:
I'm ok with this hack atm since this is a pretty large bug. Merging once CI is done.

olafurpg:
Sounds good, I'll merge and publish 0.6.2 then! :shipit:

olafurpg:
The update should hit maven in ~30 min

Issue #807, opened 2017-03-09
FetchError on Travis CI


hseeberger:
- **Version**: 0.6.2
- **Integration**: sbt

On Travis CI I get the following errors (5 times) for a project with 6 sub-projects:
```
org.scalafmt.bootstrap.FetchError: ArrayBuffer((Dependency(com.geirsson:scalafmt-cli_2.11,0.6.2,compile,Set(),Attributes(jar,),false,true),List(not found: /home/travis/.ivy2/local/com.geirsson/scalafmt-cli_2.11/0.6.2/ivys/ivy.xml, locked: /home/travis/.coursier/cache/v1/https/repo1.maven.org/maven2/com/geirsson/scalafmt-cli_2.11/0.6.2/scalafmt-cli_2.11-0.6.2.pom)))
	at org.scalafmt.bootstrap.ScalafmtBootstrap$.fromVersionUncached(Bootstrap.scala:71)
	at org.scalafmt.bootstrap.ScalafmtBootstrap$$anonfun$fromVersion$1.apply(Bootstrap.scala:45)
	at org.scalafmt.bootstrap.ScalafmtBootstrap$$anonfun$fromVersion$1.apply(Bootstrap.scala:45)
	at scala.collection.mutable.MapLike$class.getOrElseUpdate(MapLike.scala:189)
	at scala.collection.mutable.AbstractMap.getOrElseUpdate(Map.scala:91)
	at org.scalafmt.bootstrap.ScalafmtBootstrap$.fromVersion(Bootstrap.scala:45)
	at org.scalafmt.bootstrap.ScalafmtBootstrap$.main(Bootstrap.scala:94)
	at AutomateScalafmtPlugin$autoImport$$anonfun$automateScalafmtFor$1$$anonfun$apply$3.AutomateScalafmtPlugin$autoImport$$anonfun$$anonfun$$formattingHandler$1(AutomateScalafmtPlugin.scala:48)
```

On my local machine everything is fine.

olafurpg:
Are you using coursier for the build? What happens if you run "sbt scalafmt
--help" before compile?

hseeberger:
We are talking about this project: https://github.com/hseeberger/akka-http-json. The issue only occurs on Travis CI, not locally.

I don't use coursier, neither on my laptop nor on Travis CI.

I can run `scalafmt --help` locally in an sbt session which produces the following output, but I can't do that on Travis CI:
```
[...]

PR #817, opened 2017-03-11
Don't resolve from ivy2local in ScalafmtBotstrap.

olafurpg:
Addreses #807, I'm unable to reproduce the reported error so I can't say
if this commit fixes it or not. The reporter error indicates there's
some race condition with the default ivy resolver in ~/.ivy2/local,
I I'm hoping that this change avoids that error.

### DIFF
--- a/bootstrap/src/main/scala/org/scalafmt/bootstrap/Bootstrap.scala
+++ b/bootstrap/src/main/scala/org/scalafmt/bootstrap/Bootstrap.scala
@@ -53,10 +53,15 @@ object ScalafmtBootstrap {
         )
       )
 
-    val repositories = Seq(
-      Cache.ivy2Local,
-      MavenRepository("https://repo1.maven.org/maven2")
-    )
+    val repositories =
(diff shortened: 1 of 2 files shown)


d1egoaz:
yay it worked @olafurpg 💃 

as this commit hasn't being released, I've created for my test a local [ScalafmtBootstrap.scala](https://gist.github.com/d1egoaz/e7f8f09d479a6136f2cd1e608e30fb5d#file-scalafmtbootstrap-scala-L17) with the contents of this commit, but using fully qualified `org.scalafmt.bootstrap.` for some dependencies

Then I remove the `ScalafmtBootstrap` import on `AutomateScalafmtPlugin` in order to make it use my local version https://gist.github.com/d1egoaz/2ec81cefae169028a1a5d89ce33dc032

I just tested it on my failed CI (jenkins) job and test passed!

so, when @olafurpg release this commit, I think I can go back to the original `AutomateScalafmtPlugin` version provided by @hseeberger

olafurpg:
Awsome! Thanks for reporting @d1egoaz , I'll try to get the release out tonight :)

PR #818, opened 2017-03-12
Fix #780, add spaces in parens flag

### LINKED ISSUE #780: bring back spaces in parenthesis
Could we consider bringing back spaces in parentheses :)? https://github.com/olafurpg/scalafmt/issues/116

I personally find it much easier to parse (i.e. read) code with spaces inside of parenthesis. To the point that this is a necessary requirement for me to be able to use scalafmt. Other people at @xdotai like this style as well and we use it across our code base. I know we seem to be in the minority about this aspect, but we'd really love to migrate from scalariform to scalafmt and this is blocking us right now :).

olafurpg:
cc/ @cvogt

### DIFF
--- a/core/src/test/resources/spaces/InParentheses.stat
+++ b/core/src/test/resources/spaces/InParentheses.stat
@@ -1,0 +1,40 @@
+spaces.inParentheses = true
+maxColumn = 40
+<<< Spaces in parentheses
+function(a, b, c)
+>>>
+function( a, b, c )
+<<< no arg
+function()
+>>>
+function()
+<<< def
+def x(foo: Int) = 2
+>>>
+def x( foo: Int ) = 2
+<<< tuple
+def x = (1, 2)
+>>>
+def x = ( 1, 2 )
+<<< maxColumn (looks weird, but scalariform does the same)
+def x = (1, 2, aaaaaaaaaaaaaawwwwsssssssssswa)
+>>>
+def x =
+  ( 1,
+   2,
+   aaaaaaaaaaaaaawwwwsssssssssswa )
+<<< maxColumn apply
+def foo = bar(aaaaaaaaaaaaa, bbbbbbbbbbbbbb, cccccccccc)
+>>>
+def foo =
+  bar( aaaaaaaaaaaaa,
+      bbbbbbbbbbbbbb,
+      cccccccccc )
+<<< maxColumn apply 2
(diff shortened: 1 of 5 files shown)


core/src/test/resources/spaces/InParentheses.stat
```diff
@@ -0,0 +1,40 @@
...
+function()
+<<< def
+def x(foo: Int) = 2
+>>>
+def x( foo: Int ) = 2
+<<< tuple
+def x = (1, 2)
+>>>
+def x = ( 1, 2 )
+<<< maxColumn (looks weird, but scalariform does the same)
```
olafurpg: What do you think about these @cvogt?

cvogt: Interesting. Never seen that with scalariform.

Is it hard to let it produce this instead?
```
def x =
  ( 1,
    2,
    aaaaaaaaaaaaaawwwwsssssssssswa )
```

Are there any other flags I can set to let it produce either of:
```
def x = (
  1,
  2,
  aaaaaaaaaaaaaawwwwsssssssssswa )
```
```
def x = (
  1,
  2,
  aaaaaaaaaaaaaawwwwsssssssssswa
)
```

```
def x = (
  1, 2,
  aaaaaaaaaaaaaawwwwsssssssssswa
)
```
?

cvogt:
@olafurpg
I personally don't use spaces.inBrackets at the moment.

I believe some people like to distinguish call sites from definition sites for whitespace in parenthesis, but I personally like to have it in both, so the current solution works for me in this regard.

core/src/test/resources/spaces/InParentheses.stat
olafurpg: Seems like you want `style = IntelliJ`, that will produce these outputs

```scala
[info]   ===========
[info]   => Obtained
[info]   ===========
[info]   def foo =
[info]     bar(
[info]       aaaaaaaaaaaaa,
[info]       bbbbbbbbbbbbbb,
[info]       cccccccccc
[info]     )
[info]   ===========
[info]   => Obtained
[info]   ===========
[info]   def x =
[info]     (
[info]       1,
[info]       2,
[info]       aaaaaaaaaaaaaawwwwsssssssssswa
[info]     )

```

olafurpg:
I'll go ahead and merge this since it seems `style = IntelliJ` will handle the weird max column tests. You are the only user of this new option, so I won't bother with `inBrackets` or distinguishing def/call site

cvogt:
thx :D. I will check it out.

PR #825, opened 2017-03-12
Fix spaces.inParens for brackets

olafurpg:


olafurpg:
The appveyor test will be fixed in another PR, it's unrelated to this change

PR #828, opened 2017-03-13
Fix #816: Makes `verticalMultilineDefnSite` respect `continuationIndent.defnSite`.

### LINKED ISSUE #816: Make indentation in verticalMultilineAtDefinitionSite customisable
Perhaps it would make sense to allow indentation of parameters and separators to be customised in `verticalMultilineAtDefinitionSite`. Right now it is hardcoded to 4 and 2 spaces for params and sep, but some people (myself included) prefer 2 and 0 spaces, respectively. 

Current formatting:
```scala
def format(
    a: A
  )(implicit ctx: Context,
    ex: Executor
  ): B
```
What I would like it to be:
```scala
def format(
  a: A
)(implicit ctx: Context,
  ex: Executor
): B
```

If nobody is against it, I can try implementing it.

pjrt:
Makes `verticalMultilineDefnSite` respect `continuationIndent.defnSite`.

Parameters are now indented by `continuationIndent.defnSite`, while
parameter separators are indented by `continuationIndent.defnSite - 2`,
with `0` being the minimum.
    
`continuationIndent.defnSite` is defaulted to `4`, so this won't affect
how `verticalMultilineDefnSite` currently works.


pjrt:
Wow, not sure what happened, but that commit history is bad.

PR #831, opened 2017-03-13
Fix #618, add space after annotation

### LINKED ISSUE #618: Space removed after annotation - Want to keep it
When using [Play DI Guice](https://www.playframework.com/documentation/2.5.x/ScalaDependencyInjection#Declaring-runtime-DI-dependencies), I would like to keep the space after the `@Inject` annotation

**Version**: `scalafmt 0.5.0-RC2`
**Integration**: cli
**Configuration**: `style=defaultWithAlign`

```scala
class Foo @Inject() (conf: Configuration)
```

Scalafmt:
```scala
class Foo @Inject()(conf: Configuration)
```

```scala
class Foo @Inject() (conf: Configuration)
```

I think I can help with a PR if you could point me in the right direction/where to start.

And thank you for this amazing project 🍻 , I've been using it for a few weeks now, it has been really useful since I moved from IntelliJ to Emacs

AnamikaD:
The issue with the space between @inject() and ( is solved, though there are few failed cases

### DIFF
--- a/core/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/core/src/main/scala/org/scalafmt/internal/Router.scala
@@ -355,7 +355,8 @@ class Router(formatOps: FormatOps) {
           case _ => NoSplit
         }
         Seq(
-          Split(modification, 0)
+          Split(modification, 0),
+          Split(if (leftOwner.parent.exists(_.is[Mod])) Space else NoSplit, 0)
         )
       // Defn.{Object, Class, Trait}
       case tok @ FormatToken(KwObject() | KwClass() | KwTrait(), _, _) =>
--- a/core/src/test/resources/spaces/Hacking.stat
+++ b/core/src/test/resources/spaces/Hacking.stat
@@ -1,5 +1,5 @@
-maxColumn = 40
-<<< SKIP Spaces in paretheses
-function(a, b, c)
+maxColumn = 150
+<<< ONLY Spaces in after close braces
+class Foo @Inject()(conf: Configuration)
 >>>
-function( a, b, c )
+class Foo @Inject() ( conf: Configuration )
--- a/core/src/test/resources/spaces/Hacking.stat
+++ b/core/src/test/resources/spaces/Hacking.stat
@@ -1,5 +1,5 @@
 maxColumn = 150
-<<< ONLY Spaces in after close braces
+<<<Spaces in after close braces
 class Foo @Inject()(conf: Configuration)
 >>>
 class Foo @Inject() ( conf: Configuration )
--- a/core/src/test/resources/spaces/Hacking.stat
+++ b/core/src/test/resources/spaces/Hacking.stat
@@ -1,5 +1,4 @@
-maxColumn = 150
-<<<Spaces in after close braces
+<<<SKIP Spaces in after close braces
 class Foo @Inject()(conf: Configuration)
 >>>
 class Foo @Inject() ( conf: Configuration )


core/src/test/resources/spaces/Hacking.stat
```diff
@@ -1,5 +1,5 @@
-maxColumn = 40
-<<< SKIP Spaces in paretheses
-function(a, b, c)
+maxColumn = 150
```
olafurpg: I think we can remove this line here.

AnamikaD:
@olafurpg  instead of ONLY, SKIP is added. removed maxColumn

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -355,7 +355,8 @@ class Router(formatOps: FormatOps) {
           case _ => NoSplit
         }
         Seq(
-          Split(modification, 0)
+          Split(modification, 0),
```
olafurpg: The failing tests with "Search state exploded" are because there are two possible splits here. Each split creates a new branch that is explored by the search algorithm, similar to how [A*](https://en.wikipedia.org/wiki/A*_search_algorithm) works. 

You can remove the line below, and update instead `modification` , which is defined above.

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -355,7 +355,8 @@ class Router(formatOps: FormatOps) {
           case _ => NoSplit
```
olafurpg: Here we want to add a case that produces a `Space` that fixes issue #618. A good way to debug what's happening here is to print out `leftOwner` (which is a [scala.meta.Tree](http://scalameta.org/tutorial/#Part2-Trees))

```scala
case _ =>
  logger.elem(leftOwner.syntax, leftOwner.structure)
  NoSplit
```

core/src/test/resources/spaces/Hacking.stat
```diff
@@ -1,5 +1,4 @@
-maxColumn = 40
-<<< SKIP Spaces in paretheses
-function(a, b, c)
+<<<SKIP Spaces in after close braces
```
olafurpg: We want to run this test, even if it fails!

core/src/main/scala/org/scalafmt/internal/Router.scala
AnamikaD: I will create another PR when it gets fixed. This logic "leftOwner.parent.exists(_.is[Mod])" solves the issue #618.

core/src/test/resources/spaces/Hacking.stat
AnamikaD: so Hacking.stat looks like:
`maxColumn = 40
<<< SKIP Spaces in paretheses
function(a, b, c)
>>>
function( a, b, c )
<<<SKIP Spaces in after close braces
class Foo @Inject()(conf: Configuration)
>>>
class Foo @Inject() (conf: Configuration) `

core/src/main/scala/org/scalafmt/internal/Router.scala
AnamikaD: I define only this 
```
Split(if (leftOwner.parent.exists(_.is[Mod])) Space else NoSplit, 0)

File: core/src/test/resources/spaces/Hacking.stat (hunk 4 above)
Comment (olafurpg) [inline core/src/test/resources/spaces/Hacking.stat]: Note `<<<SKIP` should be `<<<`

File: core/src/main/scala/org/scalafmt/internal/Router.scala (hunk 3 above)
Comment (olafurpg) [inline core/src/main/scala/org/scalafmt/internal/Router.scala]: There's no need to open another PR, this is for fixing #618.

File: core/src/main/scala/org/scalafmt/internal/Router.scala [hunk 5]
```diff
@@ -355,7 +355,8 @@ class Router(formatOps: FormatOps) {
...
         }
         Seq(
-          Split(modification, 0)
+          Split(modification, 0),
+          Split(if (leftOwner.parent.exists(_.is[Mod])) Space else NoSplit, 0)
```
olafurpg: We should remove this line.

AnamikaD: Understood about the PR. Remove it and replace it with what!? Apologies for bugging you, soon will understand the code base :)

olafurpg: Replace it with another case [here](https://github.com/scalameta/scalafmt/pull/831#discussion_r105847958). We want to change the value of `modification`.

PR #838, opened 2017-03-15
Bootstrap should resolve with lock retry

stephennancekivell:
Bootstrap should resolve with lock retry, incase of concurrent dependency usage, as seen with concurrent scalafmt usage on travis-ci

As seen and discussed 
https://github.com/scalameta/scalafmt/issues/807
https://github.com/akka/alpakka/pull/216


hseeberger:
I think this is wrong, because we want to use scalafmt concurrently. Instead we should make the cache thread-safe: https://github.com/scalameta/scalafmt/blob/master/bootstrap/src/main/scala/org/scalafmt/bootstrap/Bootstrap.scala#L41 is a plain mutable map, we should use a thread-safe one here.

stephennancekivell:
Yea I also think its a bit budget. But its exposing a blocking methods `fromVersion` `main`.

The other way could be to use `Futures` and `Await.result`

We could change the cache to something like`mutable.Map.empty[String, Future[Either[Throwable, ScalafmtBootstrap]]]`

Im not sure if `fromVersionUncached` should really be public. using uncached sounds like you should be able to force your way back into the problem, unless some locking is added, lol. Which is the problem we're trying to fix.

hseeberger:
I don't think we need any of Future, Await, etc. All we need is make the cache thread-safe.

stephennancekivell:
Um, whys that.

I think its missing the cache because running `start.process.run(fetch).unsafePerformSync` takes seconds. Not because two threads are updating that map at the same time.

EG,
Thread A comes along, sees nothing in the cache, starts `run`. ~50ms later Thread B comes along also sees nothing in the map, also starts `run`. Some time later thread B crashes with the lock error. Thread B eventually finishes `run` then adds the result to the cache.

I agree a concurrent map would help make it more thread safe. But I dont think in practise we're hitting that problem. Maybe but i dont think so. Could be two causes.

Its also known as the thundering herd problem. Put nicely [here](http://spray.io/documentation/1.2.4/spray-caching/)

hseeberger:
Yeah, you are probably right. But thread-safe access to the cache is certainly needed, too.

bootstrap/src/main/scala/org/scalafmt/bootstrap/Bootstrap.scala
```diff
@@ -77,7 +77,7 @@ object ScalafmtBootstrap {
         logger = Some(logger)
       )
     )
-    val resolution = start.process.run(fetch).unsafePerformSync
```
olafurpg: I think we can solve this by using an `AtomicReference[Unit]` as a semaphore, you can only run the process with that reference. That should probably avoid the retry logic.

bootstrap/src/main/scala/org/scalafmt/bootstrap/Bootstrap.scala
```diff
@@ -77,7 +77,7 @@ object ScalafmtBootstrap {
         logger = Some(logger)
```
olafurpg: Can you please remove the `cache` file, since it didn't fix the issue. I prefer to reuse the default coursier cache.

olafurpg:
I also agree that we should use a concurrent map for the cache 👍

bootstrap/src/main/scala/org/scalafmt/bootstrap/Bootstrap.scala
olafurpg: Actually, it's probably enough to put a `synchronized` on the fromVersionUncached` method.

hseeberger: No, that doesn't guarantee that different threads see the same values for `cliCache`.

olafurpg: That's not too big of an issue IMO, it's OK to classload a few instances of scalafmt. sbt does a whole lot more of redundant classloading :D

hseeberger: But why not make it correct? It's easy, not?

olafurpg:
Absolutely, correct is better.

stephennancekivell:
Thanks for your comments guys. I've just put `synchronized` on those methods, to control single resolving and single access to the cache. Its much better.

Im not able to reproduce the problem in unit tests. Im not sure why.
```
  test("concurrent access to fromVersion doesnt crash with coursier lock errors") {
    val a = Future {
      ScalafmtBootstrap.fromVersion("0.6.3")
    }

    val b = Future {
      ScalafmtBootstrap.fromVersion("0.6.3")
    }

    val ab = for {
      aa <- a
      bb <- b
    } yield aa -> bb

    whenReady(ab) { abb =>
      abb._1.isRight === true
      abb._2.isRight === true
    }
  }
```

olafurpg:
Please run `./scalafmt` to fix the misformatted errors.

stephennancekivell:
I've been able to verify this fix locally, by publishing locally this branch/plugin and trying to reproduce in the alpakka project. I can see the error with 0.6.3 and its fixed with my 0.6.5-SNAPHSOT.

bootstrap/src/main/scala/org/scalafmt/bootstrap/Bootstrap.scala
```diff
@@ -41,11 +41,14 @@ object ScalafmtBootstrap {
   private val cliCache =
     mutable.Map.empty[String, Either[Throwable, ScalafmtBootstrap]]
 
-  def fromVersion(version: String): Either[Throwable, ScalafmtBootstrap] =
-    cliCache.getOrElseUpdate(version, fromVersionUncached(version))
+  def fromVersion(version: String): Either[Throwable, ScalafmtBootstrap] = {
+    synchronized {
```
olafurpg: I just noticed this line, do you think it's necessary? We already synchronize on `fromVersionUncached`.

stephennancekivell: Yea I think so.

Otherwise a second thread could pass these lines and find the cache empty while a first thread is busy running fromVersionUncached. Once the first has finished, the second will perform fromVersionUncached without knowing that its just been put in the cache.

So here its more of a performance optimisation than avoiding the coursier locking.

olafurpg: Sounds good 👍

PR #846, opened 2017-03-20
Switch to coursier 1.0.0-M15-5

alexarchambault:
Needed because of a few bin compat issues with the previous versions...

### DIFF
--- a/project/Deps.scala
+++ b/project/Deps.scala
@@ -2,5 +2,5 @@ object Deps {
   val scalameta = "1.6.0"
   val scalatest = "3.0.1"
   val scalariform = "0.1.8"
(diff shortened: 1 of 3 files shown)


olafurpg:
Thank you for the update! Sorry to see sbt-scalafmt caused problems, coursier + url classloading is so far the best workaround that I've found for running 2.11 libraries in sbt plugins. If https://github.com/sbt/sbt/issues/2879 got fixed or if sbt upgraded to 2.11/2.12, this would not be a problem.

alexarchambault:
@olafurpg Np, the binary compatibility was broken in coursier in the first place. Something along the lines of https://github.com/ensime/ensime-sbt/pull/312 could fix it too.

PR #854, opened 2017-03-23
Fix #736: Make verticalMultiline apply to classes too

### LINKED ISSUE #736: [Suggestion] Extend `verticalMultilineAtDefinitionSite` to classes
`verticalMultilineAtDefinitionSite` could be extended to apply to classes too. For example:

```scala
class SomeClass(f: Code, y: String)(implicit ex: Context, t: Task) extends Something with Else
```

```scala
class SomeClass(
    f: Code,
    y: String
  )(implicit ex: Context,
    t: Task)
  extends Something
  with Else
```

Whether that's a different flag or the same one is up to debate.

pjrt:
Makes `verticalMultilineAtDefinitionSite` apply to classes too. Fixes issue #736

### DIFF
--- a/core/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/core/src/main/scala/org/scalafmt/internal/Router.scala
@@ -415,7 +415,7 @@ class Router(formatOps: FormatOps) {
       // Parameter opening for one parameter group. This format works
       // on a group-by-group basis.
       case FormatToken(open @ LeftParen(), r, between)
-          if style.verticalMultilineAtDefinitionSite && isDefDef(leftOwner) =>
+          if style.verticalMultilineAtDefinitionSite && isDefnSite(leftOwner) =>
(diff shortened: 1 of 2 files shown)


pjrt:
Not sure I understand what failed in the build. It is complaining about finding a bad file?

olafurpg:
Hmm, I have no idea why the `push` CI passed and not the `pr` one.  I recommend we keep our WIP/PR branches on our private forks in order to keep this page clean https://github.com/scalameta/scalafmt/branches. 

I'll be working a bit on scalafmt today so I'll investigate if the CI continues to act weirdly. Hoping to port it to Scala.js now that meta is ported https://github.com/scalameta/scalameta/pull/761

olafurpg:
fwiw, ci passed after merge https://platform-ci.scala-lang.org/scalameta/scalafmt/60/1

PR #855, opened 2017-03-24
Fix appveyor

olafurpg:


olafurpg:
Aha, it seems the integration got broken for two reasons

1. sbt changed the uncompressed directory name in 0.13.13, causing `sbt` to be missing from the `%PATH%`
2. the appveyor integration was missing permissions in the scalameta organization

Should be fixed now ^^

PR #859, opened 2017-03-27
Fix #808: newlines after curly lambda

### LINKED ISSUE #808: Newline (at top) not preserved for lambda calls
- **Configuration**:
```
style = default
```

## Steps

Here's a test (that fails):

```
<<< Preserve newline in lambda call
def f = {
  something.call { x =>

    g(x)
  }
}
>>>
def f = {
  something.call { x =>

    g(x)
  }
}
```

## Expectation

I would expect it to preserve the line, similar to def's, the following test passes:

```
<<< Preserve newline in method def
def f = {

  something.call
}
>>>
def f = {

  something.call
}
```

pjrt:
Fixes issue #808 by introducing a new setting: `newlines.afterCurlyLambda`

This option can have one of three settings: never (default), always and
preserve.

* `never`(default) removes all lines. Before this commit, this is what was happening
* `always` inserts one line if there is non and removes extra lines if
    there are more. This means there is always one (1) line
* `preserve` is a combination of `never` and `always`. If the user does
    not have an extra line, there it will not add an extra line (like
    `never`). If there user has one or more lines, it will replaces it
    with one line (like `always`).

### DIFF
--- a/core/src/main/scala/org/scalafmt/config/Newlines.scala
+++ b/core/src/main/scala/org/scalafmt/config/Newlines.scala
@@ -1,6 +1,6 @@
 package org.scalafmt.config
 
-import metaconfig.ConfigReader
+import metaconfig.{ConfigReader, Reader}
 
 /**
   * @param penalizeSingleSelectMultiArgList
@@ -36,6 +36,24 @@ import metaconfig.ConfigReader
   *     n =>
   *       consume(n)
   *   }
+  * @param afterCurlyLambda
+  *   If `never` (default), it will remove any extra lines below curly lambdas
+  *   {{{
+  *   something.map { x =>
+  *
+  *     f(x)
+  *   }
+  *   }}}
+  *   will become
+  *   {{{
+  *   something.map { x =>
+  *     f(x)
+  *   }
+  *   }}}
+  *
+  *   If `always`, it will always add one empty line (opposite of `never`).
+  *   If `preserve`, and there isn't an empty line, it will keep it as it is.
+  *   If there is one or more empty lines, it will place a single empty line.
   */
 @ConfigReader
 case class Newlines(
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/scalafmt/config/Newlines.scala
```diff
@@ -44,5 +62,18 @@ case class Newlines(
...
+
+sealed abstract class NewlineCurlyLambda
+
+object NewlineCurlyLambda {
+
+  case object preserve extends NewlineCurlyLambda
+  case object always extends NewlineCurlyLambda
+  case object never extends NewlineCurlyLambda
+
+  implicit val newlineCurlyLambdaReader: Reader[NewlineCurlyLambda] =
```
pjrt: I noticed that for every other `Reader` we just define it in the companion object but don't make it implicit, but instead do so in ScalafmtConfig instead. This didn't work for me (due to a macro error I couldn't understand. But i think it wanted the implicit when creating `Newline()` in the config's default param). 

Putting the typeclass instance on the companion object fixes this, and it is proper typeclass useage anyways (avoid orphans).

pjrt:
Would be great if we could modify settings inside of a `.stat` file (currently, you can only modify them once at the top, right?). That's low priority though.

core/src/main/scala/org/scalafmt/config/Newlines.scala
```diff
@@ -36,6 +36,24 @@ import metaconfig.ConfigReader
...
+  *   }
+  *   }}}
+  *   will become
+  *   {{{
+  *   something.map { x =>
+  *     f(x)
+  *   }
+  *   }}}
+  *
+  *   If `always`, it will always add one empty line (opposite of `never`).
```
pjrt: Now that I think about it, I'm not sure `always` is the correct name for this. It isn't really "always" since the single-line split takes precedence. `force` might work, but it also isn't really "forced".

core/src/main/scala/org/scalafmt/config/Newlines.scala
olafurpg: Good point. I agree that absolute readers like this one should exist in the companion object. Only the `@ConfigReader` generated readers are not implicit and need to be inside `ScalafmtConfig`. 

> due to a macro error I couldn't understand.

Yeah, metaconfig/macros can be tricky. I badly wanted to avoid rolling my own config library/macros but I couldn't find anything that checked all boxes at the time, see https://github.com/scalameta/scalafmt/issues/316 I think I might be able to use circe path readers, but I didn't know about them at the time.

core/src/main/scala/org/scalafmt/config/Newlines.scala
olafurpg: Naming is hard...

How about `forceBlankLine`? I'm OK whichever route we go.

olafurpg:
Seems like the CI had some network error, I restarted the job, pls let me know if it misbehaves again. Me and @jvican are working on making the CI production-ready for the Scala platform modules.

core/src/main/scala/org/scalafmt/config/Newlines.scala
pjrt: That wouldn't be any different than just `force`. 

Part of me wants to leave it as `always` and rely on the fact that, as a general rule in Scalafmt, single-line is always tried before anything else. With that fact in mind `always` (and `force`) make sense.

PR #860, opened 2017-03-27
Fix #799 Allow ASCII sorting of imports

### LINKED ISSUE #799: Sorting of import group sorting doesn't match IntelliJ's sorting
When we have the following import statement:

`import sangria.schema.{BooleanType, fields, ObjectType, Field, interfaces}`

Scalafmt reorders the imports with lowecase-first 

`import sangria.schema.{fields, interfaces, BooleanType, Field, ObjectType}`

However the IntelliJ sorting is Uppercase first:

`import sangria.schema.{BooleanType, Field, ObjectType, fields, interfaces}`

Personally I find lowecase first fine since I "see" lowecase as "first", but this doesn't match the some-what universal standard that uppercase comes first (this coming from those character's ASCII codes, where uppercase chars are lower than lowercase ones).

pjrt:
Adds a new rewrite.rule: `AsciiSortImports`. This rules works exactly as
SortImports but sorts in the traditional ASCII form.

This is wanted since many editors (ie: IntelliJ) use the ASCII sorting
when sorting their imports. Having this option allows people to better
transition to Scalafmt from existing codebases and decreases tension
between Scalafmt and most editors (users can use a mix of Scalafmt and
their editors at will).

NOTE: This setting is incompatible with `SortImports`

### DIFF
--- a/core/src/main/scala/org/scalafmt/rewrite/Rewrite.scala
+++ b/core/src/main/scala/org/scalafmt/rewrite/Rewrite.scala
@@ -22,6 +22,7 @@ object Rewrite {
       RedundantBraces,
       RedundantParens,
       SortImports,
+      AsciiSortImports,
       PreferCurlyFors,
       ExpandImportSelectors,
       AvoidInfix
@@ -45,7 +46,8 @@ object Rewrite {
   val default: Seq[Rewrite] = name2rewrite.values.toSeq
 
   private def incompatibleRewrites: List[(Rewrite, Rewrite)] = List(
-    SortImports -> ExpandImportSelectors
+    SortImports -> ExpandImportSelectors,
+    SortImports -> AsciiSortImports
   )
 
   def validateRewrites(rewrites: Seq[Rewrite]): Seq[String] = {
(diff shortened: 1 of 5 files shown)


pjrt:
@olafurpg This one failed too, though it seems it failed for a different reason. 

```
Exception in thread "main" org.scalafmt.Error$MisformattedFile: /drone/src/github.com/scalameta/scalafmt/core/src/main/scala/org/scalafmt/rewrite/SortImports.scala is mis-formatted. To fix this, run ./scalafmt from the project root directory
4s
33
	at org.scalafmt.cli.InputMethod$FileContents.write(InputMethod.scala:47)
4s
34
	at org.scalafmt.cli.Cli$.org$scalafmt$cli$Cli$$handleFile(Cli.scala:144)
4s
35
	at org.scalafmt.cli.Cli$$anonfun$runFormat$1.apply(Cli.scala:217)
```

Are we expected to run `scalafmt` on the codebase ourselves? IIRC this was done automatically in CI.

olafurpg:
You're supposed to run ./scalafmt yourself before pushing. CI only checks
that you ran it.

Issue #863, opened 2017-03-28
Expand vertical-multiline to support type params


pjrt:
Currently, `verticalMultilineAtDefinitionSite` ignores type parameters.

To expand this, we will support 3 cases (splits).

* Single line

```case class SomeClass[N](name: String)```

* Single line types, multi-line params

```
case class SomeClass[N](
    someVeryLongName: String,
    someOtherVeryLongName: String
  )(implicit ex: Context)
```

* Full multi-line

```
case class SomeClass[
    N,
    T <: SomeVeryLongUpperBound: SomeTypeClass
  ](someVeryLongName: String,
    someOtherVeryLongName: String
  )(implicit ex: Context)
```

pjrt:
Implemented at: https://github.com/scalameta/scalafmt/commit/63bd7294e3d2ffac8e576f8e67eb46b62d0e08f2

AesaKamar:
@pjrt Is there a config setting to allow binpacking of type arguments up to a certain column width?

pjrt:
@AesaKamar Not really, vertical-multiline (best name ever) is specifically about forcing a multi-line pattern when the line limit is reached. Conditional bin-packing like that is what it is trying to avoid.

PR #864, opened 2017-03-29
Implement #863: vertical-multiline for type params

### LINKED ISSUE #862: Scalafmt does not indent type members when there are more than one
- **Version**: 0.6.6
- **Configuration**:
```
style = default
```

## Problem

```scala
final case class SomeCaseClass[N <: Node,
R <: SomeLongUpperBound: Ordering: Parser](projectOrdering: N => R)
```

## Expectation

```scala
final case class SomeCaseClass[N <: Node,
  R <: SomeLongUpperBound: Ordering: Parser](projectOrdering: N => R)
```

Or something similar to that.

## Notes

1. This issue happens for `def`s too, not just classes. 
2. Could/Should `verticalMultiline` be extended to type members?

pjrt:
Implements enhancement #863 by extending verticalMultilineAtDefinitionSite
to handle type parameters as well as data parameters.

The type parameter group behaves slightly different than the data
parameters. While data parameters are an all-or-nothing (either it all
fits in one line or it goes vertical), type parameters are allowed to go
in one line if they can fit regardless of whether the data parameters
fit too.

This PR does NOT fix #862. I plan on working on that one next but I found implementing this "fix" in vertical-multiline to be simpler. With the knowledge I gained here I can hopefully fix #862 now.


core/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -229,6 +229,25 @@ object TreeOps {
...
+  /**
+   * Whether the [[tree]] is a class, trait or def
+   *
+   * For classes this includes primary and secondary Ctors.
+   */
+  def isClassOrDef(tree: Tree): Boolean = tree match {
+    case _: Decl.Def | _: Defn.Def | _: Defn.Macro | _: Defn.Class |
+        _: Defn.Trait | _: Ctor.Secondary =>
+      true
+    case x: Ctor.Primary if x.parent.exists(_.isInstanceOf[Defn.Class]) =>
```
olafurpg: This seems like a reasonable workaround for the `owners` issue.

core/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -229,6 +229,25 @@ object TreeOps {
     case _ => None
   }
 
+  /**
+   * Whether the [[tree]] is a class, trait or def
+   *
+   * For classes this includes primary and secondary Ctors.
+   */
+  def isClassOrDef(tree: Tree): Boolean = tree match {
```
pjrt: I created this because I wasn't aware that `isDefnSite` includes things like type applications. 

I'm not sure whether that's correct (type applications are `definitions`?) but I opted for a more explicit function in order to not break old functionality (and since I might not have complete understanding on what a `defn` actually means).

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -420,45 +420,70 @@ class Router(formatOps: FormatOps) {
...
+        // DESNOTE(2017-03-28, pjrt) Classes and defs aren't the same.
+        // For defs, type params and data param have the same `owners`. However
+        // this is not the case for classes. Type params have the class itself
+        // as the owner, but data params have the Ctor as the owner. So a
+        // simple check isn't enough. Instead we check against the owner of the
+        // `lastParen`, which will be the same as the data param's owner.
+        val dataParamsOwner = owners(lastParen)
+        @inline def ownerCheck(rp: Token): Boolean = {
+          val owner = owners(rp)
+          owner == leftOwner || owner == dataParamsOwner
```
pjrt: The way this works is:

* When dealing with defs, then the `owners(rp)` will always equals to the leftOwner, for both type params and data params
* When dealing with classes though, this isn't the case. For classes `owners(rp)` will only equal `leftOwners` for type params. This means the second test will kick in for data params (who share the same owner as the `lastParen`).

pjrt: I could compare against for both the `owner(rp)` and the `owner(rp).parent`. That may work too.

core/src/test/resources/test/VerticalMultiline.stat
```diff
@@ -137,3 +140,59 @@ override def load(
...
+
+<<< class: if the type params fit in one line, make it so (single type param)
+final case class SomeCaseClass[N](projectOrdering: N => R, someOtherString: String)(implicit ex: Context)
+>>>
+final case class SomeCaseClass[N](
+    projectOrdering: N => R,
+    someOtherString: String
+  )(implicit ex: Context)
+
+<<< class: if the type params fit in one line, make it so (multiple type param)
```
pjrt: Add test for traits too

pjrt:
This style currently breaks if we have a modified Ctor. 

```scala
case class SomeClass[A] private (name: String)
```

core/src/main/scala/org/scalafmt/internal/Router.scala
pjrt: Well that didn't work (messes with local things like `def f(parser: Parser[R])`). It might work if I put more restrictions on it but I feel my current restriction is pretty solid (no relative checks. We only have one `lastParen`). 

Keeping as is.

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -420,67 +420,114 @@ class Router(formatOps: FormatOps) {
...
 
-        // If this is a class, we need to do some different things
+        // If this is a class/trait, we need to do some things differently.
         // We only care about the Primary Ctor since other Ctors will
         // be picked up by the `def` definition
-        val isCtor = leftOwner.is[meta.Ctor.Primary]
+        val isClassTrait =
+          leftOwner.is[meta.Ctor.Primary] ||
+          leftOwner.is[meta.Defn.Class] ||
+          leftOwner.is[meta.Defn.Trait]
```
pjrt: Since I have the holistic information about the defn (ie: I can ask things like "does this have any data params? Any type params?") I feel like this can all be cleaner (at the cost of, maybe, some code duplication).

core/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -108,6 +108,23 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
...
   }
 
+  object CtorModifier {
+    def is(tree: Tree): Boolean =
+      tree.is[meta.Mod.Private] || tree.is[meta.Mod.Protected]
+    def is(tok: Token): Boolean = tok.is[KwPrivate] || tok.is[KwProtected]
+    def unapply(tok: Token): Boolean = is(tok)
+  }
+
+  object LeftParenOrBracket {
```
olafurpg: Can we move this to https://github.com/scalameta/scalafmt/blob/235909ffdcbefe18457878a8c95c6c93ff288ed9/core/src/main/scala/org/scalafmt/util/TokenClasses.scala ?

core/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -108,6 +108,23 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
...
+    def is(tok: Token): Boolean = tok.is[LeftParen] || tok.is[LeftBracket]
+    def unapply(tok: Token): Boolean = is(tok)
+  }
+
+  object RightParenOrBracket {
```
olafurpg: Same, TokenClasses.

core/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -108,6 +108,23 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
     (packages.result(), imports.result(), arguments.toMap, optional)
   }
 
+  object CtorModifier {
```
olafurpg: Can we move this to https://github.com/scalameta/scalafmt/blob/master/core/src/main/scala/org/scalafmt/util/TreeClasses.scala ?
 
The `@classifier` macro annotation creates a `Classifiable` type-class instance, which makes it possible to use the `Token/Tree.is[T]` syntax. You can probably avoid the macro annotation by writing some boilerplate yourself, but I haven't looked at the expanded code. It's up to yourself.

The (undocumented) general rule where I keep utilities is

- FormatOps: anything that requires access to the fields in FormatOps, which is custom for each source file.
- Token/TreeOps: anything that is not require fields in FormatOps, agnostic to the source file being formatted.
- Token/TreeClasses: classifiers like this.

I should probably add this to the contributing docs.

core/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -165,7 +182,7 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
   @tailrec
   final def findFirst(start: FormatToken, end: Token)(
       f: FormatToken => Boolean): Option[FormatToken] = {
-    if (start.left.start < end.start) None
```
olafurpg: Whoah, good catch! Was this method used somewhere before? Logic bugs like this tend to pop up in dead code.

core/src/main/scala/org/scalafmt/internal/Policy.scala
```diff
@@ -16,6 +16,18 @@ case class Policy(
     noDequeue: Boolean = false,
     isSingleLine: Boolean = false)(implicit val line: sourcecode.Line) {
 
+  def merge(other: PartialFunction[Decision, Decision], newExpire: Int): Policy =
```
olafurpg: Nice 👍 I've written this too many times.

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -420,67 +420,121 @@ class Router(formatOps: FormatOps) {
...
       // Parameter opening for one parameter group. This format works
-      // on a group-by-group basis.
-      case FormatToken(open @ LeftParen(), r, between)
+      // on the WHOLE defnSite (via policies)
+      case FormatToken(open @ (LeftParen() | LeftBracket()), r, between)
           if style.verticalMultilineAtDefinitionSite &&
-            isDefnSite(leftOwner) =>
+            isClassOrDef(leftOwner) =>
+
         val close = matchingParentheses(hash(open))
```
olafurpg: Can we move the bulk of this logic outside of Router.scala into a utility file? See for example how `infixSplit` is used. I've noticed that incremental compile times in Router.scala are ~6-7 seconds now and it keeps growing.

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -420,67 +420,121 @@ class Router(formatOps: FormatOps) {
...
+          leftOwner.is[meta.Defn.Class] ||
+          leftOwner.is[meta.Defn.Trait]
+
+        // Since classes and defs aren't the same (see below), we need to
+        // create two (2) OneArgOneLineSplit when dealing with classes. One
+        // deals with the type params and the other with the data params.
+        val oneLinePerArg = {
+          val base = OneArgOneLineSplit(open)
+          if (mixedParams) {
+            val afterTypes = leftTok2tok(matchingParentheses(hash(open)))
```
olafurpg: Reminds me I should write a wrapper for `leftTok2tok(matchingParentheses(hash(open)))` 😄

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -420,67 +420,121 @@ class Router(formatOps: FormatOps) {
...
+        }
+
+        // DESNOTE(2017-03-28, pjrt) Classes and defs aren't the same.
+        // For defs, type params and data param have the same `owners`. However
+        // this is not the case for classes. Type params have the class itself
+        // as the owner, but data params have the Ctor as the owner, so a
+        // simple check isn't enough. Instead we check against the owner of the
+        // `lastParen` as well, which will be the same as the data param's
+        // owner.
+        val dataParamsOwner = owners(lastParen)
```
olafurpg: Nitpick, I typically use "value" instead of "data" to describe runtime values. At least in Scala, the convention seems to be to talk about types vs. values.

core/src/test/resources/test/VerticalMultiline.stat
```diff
@@ -137,3 +140,130 @@ override def load(
...
+<<< Work with modifiers (protected) for the constructor (full multi-line)
+final case class SomeVeryLongName[SomeLongTypeName <: UpperBounds, SomeOtherTypeName] protected(value: Long, name: String)(implicit ex: Executor)
+>>>
+final case class SomeVeryLongName[
+    SomeLongTypeName <: UpperBounds,
+    SomeOtherTypeName
+  ] protected (
+    value: Long,
+    name: String
+  )(implicit ex: Executor)
```
olafurpg: So I understand correctly, will the last closing parens never dangle? If so, why?

There seems to be missing a test case where the last param list of a class has >1 parameters.

core/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -229,6 +229,25 @@ object TreeOps {
...
+    case _: Decl.Def | _: Defn.Def | _: Defn.Macro | _: Defn.Class |
+        _: Defn.Trait | _: Ctor.Secondary =>
+      true
+    case x: Ctor.Primary if x.parent.exists(_.isInstanceOf[Defn.Class]) =>
+      true
+    case _ => false
+  }
+
+  /**
+   * Whether [[tree]] is a defintion site
```
olafurpg: I try structure docstrings like this:

"Returns X if Y(, else Z)?"

core/src/main/scala/org/scalafmt/util/TreeOps.scala
olafurpg: Type applications should probably not be included in isDefnSite, I opened https://github.com/scalameta/scalafmt/issues/867 to remind myself to investigate that and fix `isDefnSite`. For now, I think it's OK for you to create your own utility for vertical multiline.

`isClassOrDefn` is not so accurate here since this also includes `Defn.Trait`. A more appropriate name could be `isDefnSiteWithParams`, since it excludes `Defn.Val` which is a definition but has no parameters. It's up to you what name you prefer.

core/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -229,6 +229,25 @@ object TreeOps {
     case _ => None
   }
 
+  /**
+   * Whether the [[tree]] is a class, trait or def
```
olafurpg: See comment below.

core/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -420,67 +420,121 @@ class Router(formatOps: FormatOps) {
...
+        // for parameter groups.
+        val policy = oneLinePerArg.merge(paramGroupSplitter, lastParen.end)
 
         val firstIndent =
           if (r.is[RightParen]) // An empty param group
             indentSep
           else
             indentParam
 
+        val singleLineSplit =
```
olafurpg: Suggestion:
```scala
val singleLineExpire =
  if (isBracket) close // comment
  else lastParen // comment
val singleLineSplit = Split(NoSplit, 0)...
``

core/src/main/scala/org/scalafmt/internal/FormatOps.scala
pjrt: It wasn't used anywhere.

core/src/main/scala/org/scalafmt/internal/Router.scala
pjrt: Right! I kept thinking yesterday "what do you call non-tyle parameters?". For some reason (more than likely Haskell related) I ended up to data. 

But yeah, they are "value parameters". I'll change it when I get to my PC.

core/src/test/resources/test/VerticalMultiline.stat
pjrt: It should not dangle because it creates an ugly result when combined with `extends` and `with` (unless we do some manipulation with those, but then it would need to ignore other settings and make the overall style more complex). We do dangle for defs though.

olafurpg: Aha, makes sense. I just remembered now that IntelliJ style does the same, exclude dangling parentheses in defn site (although both `def` and `class`)

PR #881, opened 2017-04-08
Fix #880, filter diff files with canFormat

### LINKED ISSUE #880: scalafmt --diff should filter files through canFormat as standard use does
- **Version**: 0.6.8

Normal usage of scalafmt cli filters files to be formatted with canFormat as can be seen here,

https://github.com/scalameta/scalafmt/blob/master/cli/src/main/scala/org/scalafmt/cli/Cli.scala#L117

"scalafmt --diff" does not as can be seen here

https://github.com/scalameta/scalafmt/blob/master/cli/src/main/scala/org/scalafmt/cli/Cli.scala#L104

I propose a 1 line fix that adds a canFormat filter to getFilesFromDiff.

I can't see a practical way to test this (or a test for this feature currently) so it will be a very small PR ;-).

caoilte:
scalafmt already filters files to format with canFormat for standard invocation.
This brings the '--diff' option into line with this behaviour

### DIFF
--- a/cli/src/main/scala/org/scalafmt/cli/Cli.scala
+++ b/cli/src/main/scala/org/scalafmt/cli/Cli.scala
@@ -101,7 +101,10 @@ object Cli {
 
   private def getFilesFromDiff(options: CliOptions): Seq[AbsoluteFile] = {
     val branch = options.diff.get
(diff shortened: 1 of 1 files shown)


olafurpg:
Aha, that's what was wrong. Nice catch, thanks! 👍

PR #885, opened 2017-04-13
Upgrade dependencies

olafurpg:
This PR is the first step towards adding Scala.js support. 
The old version of metaconfig relied on a bit of runtime reflection, which was problematic in JS runtimes. The new metaconfig api is 1000x cleaner anyways to it's totally worth the brutal migration.


olafurpg:
cc/ @pjrt, fyi this PR reorganizes the reader instances for configuration to use implicits idiomatically. The new `@metaconfig.Recurse` helper also avoids the boilerplate `implicit val xReader: Reader[X] = x.reader`

pjrt:
Nice! :+1:

olafurpg:
Unfortunately, I just realised that scalameta 1.7 introduces a new dependency on protobuf and scalapb. I believe this dependency can be avoided with a `semantic-protobuf` module in scalameta (see https://github.com/scalameta/scalameta/issues/799), until then I'm not gonna worry about this new unnecessary dependency in scalafmt.

Issue #883, opened 2017-04-12
Align different tokens with different owners


i-am-the-slime:
This template is a guideline, not a strict requirement.
- **Version**: 0.6.3
- **Integration**: IntelliJ
- **Configuration**:

```hocon
style = defaultWithAlign
maxColumn = 120
bestEffortInDeeplyNestedCode = true
project.git = true
```

## Steps

I ran scalafmt like this:

```sbt scalafmt```

## Problem

Scalafmt formats code like this:
```scala
      val something = for {
        seven <- Option(7)
        eight = 8
        nine <- Option(9)
      } yield seven + eight + nine
```

## Expectation

I would like the formatted output to look like this:
```scala
      val something = for {
        seven <- Option(7)
        eight  = 8
        nine  <- Option(9)
      } yield seven + eight + nine
```

## Workaround
```scala
      // format: off
      val something = for {
        seven <- Option(7)
        eight  = 8
        nine  <- Option(9)
      } yield seven + eight + nine
      // format: on
```

olafurpg:
Thank you for reporting! I believe this should be possible to support using the same infrastructure as `align.mixedOwners`. The reason these don't align is because their "owners" (tree node types that are closest parents of the `<-` and `=` tokens) mismatch. See here https://github.com/scalameta/scalafmt/blob/abff74e65baa775521fa58c62ac099fe98d91208/core/src/main/scala/org/scalafmt/internal/FormatWriter.scala#L314 for how `=` is aligned across `def` and `val`. I suggest we do the same for this issue.

We can add a test for this here: https://github.com/scalameta/scalafmt/blob/abff74e65baa775521fa58c62ac099fe98d91208/core/src/test/resources/align/DefaultWithAlign.stat


Please let me know if you need more pointers.

olafurpg:
BTW, the tutorial contains some more info on how to hack on scalafmt https://github.com/scalameta/scalafmt/blob/abff74e65baa775521fa58c62ac099fe98d91208/Tutorial.md

but note that this issue does not require learning `Split` or editing `Router.scala`

PR #886, opened 2017-04-14
Start and stuck.

i-am-the-slime:
This does not work. It's only to demonstrate what I tried.

I started work on #883 but I am already stuck.
I figured out the right names to put in the `Map` but nothing is being formatted.
@olafurpg could you help me out?


olafurpg:
Thank you for opening a PR. I think you are on the right track, I need to checkout the code to see what's happening. I'm out of town over easter so I'll try to get back next week :)

olafurpg:
Superseded by #887

PR #887, opened 2017-04-14
Fix #883.

### LINKED ISSUE #883: Align different tokens with different owners

olafurpg:
This commit exposes new configuration options to control which
tree nodes and token kinds align together using `.productPrefix`
as keys.

### DIFF
--- a/core/src/main/scala/org/scalafmt/config/Align.scala
+++ b/core/src/main/scala/org/scalafmt/config/Align.scala
@@ -44,15 +44,43 @@ import metaconfig._
   * @param ifWhileOpenParen
   *   If true, aligns by ( in if/while/for. If false,
   *   indents by continuation indent at call site.
+  * @param tokenCategory
+  *   Customize which token kinds can align together. By default, only tokens with
+  *   the same `Token.productPrefix` align. To for example align = and <-,
+  *   set the values to:
+  *     Map("Equals" -> "Assign", "LeftArrow" -> "Assign")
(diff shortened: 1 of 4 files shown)


olafurpg:
@i-am-the-slime I realized it's possible to get rid of `.getClass.getName` by using `productPrefix` instead, this makes it OK to expose a configuration option for alignment of different token and tree types. I added a small example of how to do that in the `TrollAlignment.stat` file. The reason things are so strict by default is to avoid stuff like

```scala
val a           = b
for { a <- b; c = d } c
foobar(a        = d)
```
However, this commit makes it possible to achieve that output if you like.

olafurpg:
The Ci failure is unrelated to this change, I will fix that in a separate PR.

Issue #597, opened 2016-11-19
SBT plugin is looking for a new maintainer


olafurpg:
I have decided to remove the SBT plugin. Why? I don't personally use the SBT plugin, the plugin has a few [issues](https://github.com/olafurpg/scalafmt/labels/sbt) and maintaining the plugin in my free time has turned out to be an unrewarding experience. I have several times asked for help to maintain the plugin, but unfortunately with no luck so far.

It appears many people (and organizations) use the sbt plugin. I'm opening this issue to invite anyone who is interested to lend an extra hand to help bring the SBT plugin back. I would love to welcome you as a collaborator to the project, give you merge rights, and help bring you up to speed on making the SBT plugin a success.

To show your interest in maintaining the SBT plugin, you can do any of the following

- send a PR reverting 308f393ac71cbb290bc8166c9f82bed59b1f3a17 and 983b04372b76fd116c1b3767ac78012367412f44. This is probably the easiest option, but comes with the baggage of issues such as #485, which is caused by a [regression in sbt 0.13.12](https://github.com/sbt/sbt/issues/2786). The implementation is mostly borrowed from the scalariform sbt plugin but is slightly more complicated because scalafmt is 2.11 only and sbt plugins must compile against 2.10.
- send a PR implementing a new SBT plugin from scratch that uses the [`bootstrap` module](https://github.com/olafurpg/scalafmt/blob/master/bootstrap/src/main/scala/org/scalafmt/bootstrap/Bootstrap.scala). The bootstrap module can potentially fix issues such as #485 by using the coursier api as a replacement for SBT's `update` mechanism (the root cause of #485). Don't hesitate to ask for more details.
- send a PR creating a new SBT plugin that leverages the ["synthetic projects" feature in sbt 0.13.13](https://github.com/sbt/sbt/releases/tag/v0.13.13) to create an isolated classpath to invoke the scalafmt CLI
- send a PR implementing a new SBT plugin with whatever approach you like. However, please note that sbt plugins must compile against 2.10 and scalafmt is 2.11 only because of its dependency on scala.meta. Backporting scala.meta to 2.10 has so far been unsuccessful, https://github.com/scalameta/scalameta/issues/295

If you opt to implement a new SBT plugin, please include scripted tests like the previous plugin had.

### Migrating away from the SBT plugin

Use the CLI: https://olafurpg.github.io/scalafmt/#CLI

To enforce scalafmt in a project like was possible with the sbt plugin, my personal recommendation is to

[...]

PR #896, opened 2017-04-22
Migrate to sbt 1.0.0-M5 for sbt-scalafmt.

olafurpg:
- Remove sbt 0.13 plugin. It is not worth the maintenance cost.
  Users who want to stay on sbt 0.13 and want to use latest scalafmt
  versions can copy the single-file plugin
  https://github.com/scalameta/scalafmt/blob/master/scalafmtSbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
  and replace ScalafmtBootrap.main with
  ```scala
  val Right(scalafmt) = ScalafmtBootrap.fromVersion("<latest scalafmt version>")
  scalafmt.main(args)
  ```
- scripted tests are done with a dummy scalafmtSbtTest project because
  I couldn't figure out how to run scripted tests on the sbt 1.0 plugin
  because the scalafmt build is still on 0.13.15


sjrd:
That's a harsh move. Very few people can afford to migrate to sbt 1 as long as it's not a final release. I also don't understand the maintenance cost argument: the last commit that touched the sbt plugin is dated from December.

I understand the good feeling that removing code provides, but I believe it is way too early for this move.

olafurpg:
> maintenance

See 
https://github.com/scalameta/scalafmt/issues?utf8=%E2%9C%93&q=is%3Aissue%20label%3Asbt%20
and
https://github.com/scalameta/scalafmt/commits/master/bootstrap/src/main/scala/org/scalafmt/bootstrap/Bootstrap.scala
All of those issues, of which some are still open, are related to hacks around the 2.10 requirement in sbt 0.13.

Those who are stuck on 0.13 can continue to use future scalafmt versions by staying on sbt-scalafmt 0.6.6 and writing this (untested) command

```scala
commands += Command.args("scalafmt") {
  case (state, args) =>
    val Right(scalafmt) org.scalafmt.bootstrap.ScalafmtBootstrap.fromVersion("<latest version>")
    scalafmt.main("--non-interactive" +: args.toArray)
    state
}
```
 
I'll add a section in the docs explaining this workaround.

I've tried really hard and put a lot of hours into making sbt-scalafmt work on 0.13. When I called for help in https://github.com/scalameta/scalafmt/issues/597 no-one stepped up to help. Given that I personally don't even use the sbt plugin, my patience for supporting sbt 0.13 is running out now that sbt 1.0 is on the way.

olafurpg:
I've added an example repo on how to use the latest scalafmt versions with sbt 0.13 here: https://github.com/olafurpg/sbt-scalafmt-0.13/blob/master/build.sbt

I've updated this PR to add a "sbt 0.13" section in the docs explaining this workaround.

olafurpg:
Given that sbt 0.13 users can continue to use the latest scalafmt version with scalafmt-bootstrap, see https://github.com/olafurpg/sbt-scalafmt-0.13, I'm gonna go ahead and merge this PR. I hope that this change motivates people to upgrade to sbt 1.0 as soon as possible.

PS. I think it's amazing that this new plugin for sbt 1.0 is only 22 lines of code. In fact, the build configuration needed to run the scripted tests contains code than the plugin implementation 😄

olafurpg:
The appveyor setup has been broken for a while, it's unrelated to this PR.

Issue #893, opened 2017-04-20
Add NL before multi-line vals


dwijnand:
This template is a guideline, not a strict requirement.
- **Version**: 0.6.8
- **Integration**: CLI
- **Configuration**:
```
newlines.alwaysBeforeTopLevelStatements=true
```

## Steps

I ran scalafmt like this:

```bash
scalafmt
```

## Problem

Scalafmt formats code like this:
```scala
object Main {
  val x = 1
  val y = 2
  val z = 3

  val foo = {
    println("foo")
  }
  val bar = {
    println("bar")
  }
}
```

## Expectation

I would like the formatted output to look like this:
```scala
object Main {
  val x = 1
  val y = 2
  val z = 3

  val foo = {
    println("foo")
  }

  val bar = {
    println("bar")
  }
}
```

## Notes

Related to #892 "Don't add NL before single-line defs"

Daxten:
Hey, I can make a PR for this this weekend

olafurpg:
Please give 0.7.0-RC1 a try and see if the updates to the new option are OK. The main reason I made this release RC1 was to battle test the latest changes to  alwaysBeforeTopLevelStatements  in #904

olafurpg:
I believe with #920, this is now fixed. I opened https://github.com/scalameta/scalafmt/issues/922 which is related, but I prefer not to let that one block v1.0

Issue #892, opened 2017-04-20
Don't add NL before single-line defs


dwijnand:
This template is a guideline, not a strict requirement.
- **Version**: 0.6.8
- **Integration**: CLI
- **Configuration**:
```
newlines.alwaysBeforeTopLevelStatements=true
```

## Steps

I ran scalafmt like this:

```bash
scalafmt
```

## Problem

Scalafmt formats code like this:
```scala
object Main {
  def x = 1

  def y = 2

  def z = 3

  def foo = {
    println("foo")
  }

  def bar = {
    println("bar")
  }
}
```

## Expectation

I would like the formatted output to look like this:
```scala
object Main {
  def x = 1
  def y = 2
  def z = 3

  def foo = {
    println("foo")
  }

  def bar = {
    println("bar")
  }
}
```

Daxten:
Hey, I can make a PR for this this weekend

Daxten:
I'm thinking about generally only adding newlines if the statement is more then 1 line. Atm it adds newlines before toplevel statements.. I will think about this later and comment back

Daxten:
alwaysBeforeTopLevelStatements
alwaysBeforeMultilineStatements

a new option would make sense I think!

olafurpg:
I believe with #920, this is now fixed. I opened https://github.com/scalameta/scalafmt/issues/922 which is related, but I prefer not to let that one block v1.0

PR #899, opened 2017-04-22
Topic/always before multiline statements

Daxten:
new features for #892 #893 and fixes an unreported error with annotations like js.native

I wanted to add this before but didn't find a solution right away but I think this catches all edge-cases now!


Daxten:
urgh, how did that custom cache dir commit from you get in here? I just reverted it for now.. has that to do with the "new"repo? Maybe I have to pull a clean master and cherry pick my commit and make a new PR?

Issue #897, opened 2017-04-22
Get appveyor green again


olafurpg:
Appveyor has been red for a while https://ci.appveyor.com/project/olafurpg/scalafmt/history This doesn't affect windows users, but it does affect contributors on Windows. If someone with a windows machine could take a look at this then I would very grateful. I don't have a Windows machine to reproduce on a local environment, which makes it very slow for me to troubleshoot the problem.

cc/ @mmatloka @mpociecha

olafurpg:
The latest error: https://ci.appveyor.com/project/olafurpg/scalafmt/build/68#L1289

```
warn] one warning found
coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
  com.geirsson:scalafmt-cli_2.12:0.6.8+12-dcfaf6fe:
    not found:
      https://repo.typesafe.com/typesafe/ivy-releases/com.geirsson/scalafmt-cli_2.12/0.6.8+12-dcfaf6fe/ivys/ivy.xml
      https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/com.geirsson/scalafmt-cli_2.12/0.6.8+12-dcfaf6fe/ivys/ivy.xml
      C:\Users\appveyor\.ivy2\local\com.geirsson\scalafmt-cli_2.12\0.6.8+12-dcfaf6fe\ivys\ivy.xml
      https://repo1.maven.org/maven2/com/geirsson/scalafmt-cli_2.12/0.6.8+12-dcfaf6fe/scalafmt-cli_2.12-0.6.8+12-dcfaf6fe.pom
[error] (scalafmtSbt/*:coursierResolution) coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
[error]   com.geirsson:scalafmt-cli_2.12:0.6.8+12-dcfaf6fe:
[error]     not found:
[error]       https://repo.typesafe.com/typesafe/ivy-releases/com.geirsson/scalafmt-cli_2.12/0.6.8+12-dcfaf6fe/ivys/ivy.xml
[error]       https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/com.geirsson/scalafmt-cli_2.12/0.6.8+12-dcfaf6fe/ivys/ivy.xml
[error]       C:\Users\appveyor\.ivy2\local\com.geirsson\scalafmt-cli_2.12\0.6.8+12-dcfaf6fe\ivys\ivy.xml
[...]

PR #900, opened 2017-04-22
alwaysBeforeTopLevelStatements

Daxten:
* keep blocks of singleline statements which start with the same keyword together (case [object|class] / vals / defs / etc)
* fix bug with annotations which are used with a prefix, like js.native
* add newlines to non-toplevel statements if they are multiline

#893 and #892

Even though this adds some details to alwaysBeforeTopLevelStatements, I think this shouldn't go into it's own setting

### DIFF
--- a/core/src/main/scala/org/scalafmt/util/TreeOps.scala
+++ b/core/src/main/scala/org/scalafmt/util/TreeOps.scala
@@ -486,4 +486,12 @@ object TreeOps {
   // procedure syntax has decltpe: Some("")
   def isProcedureSyntax(defn: Defn.Def): Boolean =
     defn.decltpe.exists(_.tokens.isEmpty)
+
+  def isMultiline(tree: Tree): Boolean = tree.tokens.exists(_.is[LF])
+
+  def isAnnotation(tree: Tree): Boolean = tree.parent match {
+    case None => false
+    case Some(e) if e.is[Mod.Annot] => true
+    case Some(e) => isAnnotation(e)
+  }
 }
--- a/core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
+++ b/core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
@@ -9,6 +9,9 @@ package a {
   object a {
       val v1 = 1
       val v2 = 2
+      val v3 = {
+        3
+      }
       case class C1
       sealed trait T1
       final class FC1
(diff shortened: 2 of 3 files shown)


core/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -486,4 +486,12 @@ object TreeOps {
   // procedure syntax has decltpe: Some("")
   def isProcedureSyntax(defn: Defn.Def): Boolean =
     defn.decltpe.exists(_.tokens.isEmpty)
+
+  def isMultiline(tree: Tree): Boolean = tree.tokens.exists(_.is[LF])
```
olafurpg: I am afraid that this approach may cause non-idempotent formatting in some cases. Example

```scala
// maxColumn                  |
def foo = 2
def looongLine = { baar(1111, 2222, 33333))) }

// format once
def foo = 2
def looongLine = {
  baar(1111, 2222, 33333)))
}

// format second time
def foo = 2

def looongLine = {
  baar(1111, 2222, 33333)))
}
```

I don't think it's possible to determine inside Router.scala if a tree is multiline or not. What we can do instead is to handle this inside FormatWriter, which runs after the line wrapping algorithm has chosen where to put newlines and spaces.

olafurpg:
PS. iirc, are you on Windows? If so, does the build fail on your machine like on appveyor? I would love to get it green again cc/ #897

core/src/main/scala/org/scalafmt/util/TreeOps.scala
Daxten: Thought something like this could be a problem!

Daxten:
Yeah I'm on windows, I will take a look later!

olafurpg:
@Daxten FYI I just merged https://github.com/scalameta/scalafmt/pull/902 which makes the build green again on Windows

core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
```diff
@@ -29,25 +35,29 @@ package a {
...
 
+      val v3 = {
+        3
+      }
+
       case class C1
 
       sealed trait T1
 
       final class FC1
```
dwijnand: I don't mean to feature creep your PR, so feel free to de-scope this remark, but ideally (for me) these single line definitions wouldn't be force-split apart by the tool.

dwijnand: these = C1, T1, FC1

olafurpg:
I opened https://github.com/scalameta/scalafmt/pull/904 as a followup of this PR

olafurpg:
Superseded by #904

PR #901, opened 2017-04-23
Fix AvoidInfix for methods with empty parameters list

### LINKED ISSUE #884: AvoidInfix is broken for parameterless methods
- **Version**: 0.6.8
- **Integration**: IntelliJ
- **Configuration**:
```
rewrite.rules = [
  AvoidInfix
]
```

## Steps

I ran scalafmt like this:

```bash
scalafmt
```

## Problem
When formatting 
```scala
object Infix {
  case class Foo(i: Int) {
    def infix(): Foo = Foo(i + 1)
  }
  
  val a = Foo(10) infix()
}
```
```scala
java.util.NoSuchElementException: head of empty list
```
is thrown

## Expectation

```scala
object Infix {
  case class Foo(i: Int) {
    def infix(): Foo = Foo(i + 1)
  }
  
  val a = Foo(10).infix()
}
```

## Notes
Quick look at the `AvoidInfix.scala` made me think this [line](https://github.com/scalameta/scalafmt/blob/3444c16c5493005adcfb62dec3ba6c94d300e276/core/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala#L32) might be the source of this issue.

sugakandrey:
Close #884.

### DIFF
--- a/core/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala
+++ b/core/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala
@@ -29,20 +29,25 @@ case object AvoidInfix extends Rewrite {
           TokenPatch.AddLeft(fstOpToken, ".", keepTok = true)
         )
 
-        val fstArgsToken = args.head.tokens.head
-        val lastArgsToken = args.last.tokens.last
+        val fstArgsToken = args.headOption.flatMap(_.tokens.headOption)
+        val lastArgsToken = args.lastOption.flatMap(_.tokens.lastOption)
         val fstIsNotLeftParenAndBrace = fstArgsToken
-          .isNot[LeftParen] && fstArgsToken
-          .isNot[LeftBrace]
+          .exists(_.isNot[LeftParen]) && fstArgsToken
+          .exists(_.isNot[LeftBrace])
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala
```diff
@@ -29,20 +29,25 @@ case object AvoidInfix extends Rewrite {
           TokenPatch.AddLeft(fstOpToken, ".", keepTok = true)
         )
 
-        val fstArgsToken = args.head.tokens.head
-        val lastArgsToken = args.last.tokens.last
+        val fstArgsToken = args.headOption.map(_.tokens.head)
```
olafurpg: Since we're already going through the effort of wrapping everything in Option, what do you think about `.flatMap(_.tokens.headOption)`?

sugakandrey: yeah, makes sense.

olafurpg:
Great work @sugakandrey, I'm really happy to see both this issue fixed as well as #903, I suspect a lot of people will enjoy them, esp. #903 😄

PR #903, opened 2017-04-23
Fix #735. newlines before and after implicit keyword

### LINKED ISSUE #735: put implicit key word at position of implicit parameters in separate line
Obtained, scalafmt formats code like this:
```scala
implicit def pairEncoder[A, B](
      implicit aEncoder: CsvEncoder[A],
      bEncoder: CsvEncoder[B]
  ): CsvEncoder[(A, B)] =
```

Expected, I would like the formatted output to look like this:
```scala
implicit def pairEncoder[A, B](
      implicit
      aEncoder: CsvEncoder[A],
      bEncoder: CsvEncoder[B]
  ): CsvEncoder[(A, B)] =
```

sugakandrey:
This introduces 2 new settings: `newlines.beforeImplicitKWInVerticalMultiline` and `newlines.afterImplicitKWInVerticalMultiline`. Both are set to `false` by default.

### DIFF
--- a/core/src/main/scala/org/scalafmt/config/Newlines.scala
+++ b/core/src/main/scala/org/scalafmt/config/Newlines.scala
@@ -54,6 +54,48 @@ import metaconfig._
   *   If `always`, it will always add one empty line (opposite of `never`).
   *   If `preserve`, and there isn't an empty line, it will keep it as it is.
   *   If there is one or more empty lines, it will place a single empty line.
+  * @param afterImplicitKWInVerticalMultiline  If true, add a newline after an implicit keyword in function and
+  *                                   class definitions (only for verticalMultiline style)
+  * {{{
+  *   // newlines.afterImplicitKWInVerticalMultiline = true
+  *   def format(
+  *     code: String,
+  *     age: Int
+  *   )(implicit
+  *     ev: Parser,
+  *     c: Context
+  *   ): String
+  *   // newlines.afterImplicitKWInVerticalMultiline = false
+  *   def format(
+  *     code: String,
+  *     age: Int
+  *   )(implicit ev: Parser,
+  *     c: Context
+  *   ): String
+  * }}}
+  * @param beforeImplicitKWInVerticalMultiline If true, add a newline before an implicit keyword in function and
+  *                                   class definitions (only for verticalMultiline style)
+  * {{{
+  *   // newlines.afterImplicitKWInVerticalMultiline = true
+  *   // newlines.beforeImplicitKWInVerticalMultiline = true
+  *   def format(
+  *     code: String,
+  *     age: Int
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/scalafmt/config/Newlines.scala
```diff
@@ -63,7 +105,9 @@ case class Newlines(
     penalizeSingleSelectMultiArgList: Boolean = true,
     alwaysBeforeCurlyBraceLambdaParams: Boolean = false,
     alwaysBeforeTopLevelStatements: Boolean = false,
-    afterCurlyLambda: NewlineCurlyLambda = NewlineCurlyLambda.never
+    afterCurlyLambda: NewlineCurlyLambda = NewlineCurlyLambda.never,
+    afterImplicitKWInVerticalMultiline: Boolean = false,
+    beforeImplicitKWInVerticalMultiline: Boolean = false
```
olafurpg: What is the motivation for this option? I feel like `afterImplicitKWInVerticalMultiline` is a lot more common.

sugakandrey: in #735 @pjan pointed out that this is something he would like to be supported, I also saw this style used in some code bases (e.g. ensime). I don't have a strong opinion on this myself.

olafurpg: OK, fair enough, we can keep that option.

olafurpg:
I'm gonna go ahead and merge this to unblock #898 , I've been planning that PR forever and I was determined to get it in this weekend 😄

pjrt:
Sorry for the late response (got sick while at ScalaDays and haven't recovered). 

Nice! This looks great!

diego-alvarez-hs:
I'd love to have this available for no `verticalMultiline` style

Current:
```scala
def foo(
a: Int
)(implicit executionContext: ExecutionContext): Future[A] = {
```
Wanted:
```scala
def foo(
a: Int
)(
  implicit executionContext: ExecutionContext
): Future[A] = {
```

pjrt:
Oh, I didn't realize that this only applied to verticalMultiline. Now I'm not so sure about it: it couples two things that probably shouldn't be coupled.

olafurpg:
@diego-alvarez-hs That output you can either get with style=IntelliJ or http://scalameta.org/scalafmt/#Configstyle

@pjrt I somewhat agree, we can maybe fix this before the 0.7 final release. The reason I merged it quickly was to clear the PR queue before merging the Scala.js PR, which moved all files around. @sugakandrey do you have an idea how to make this new feature not be tied to vertical multiline?

sugakandrey:
@olafurpg I'll see what I can do.

PR #904, opened 2017-04-23
Multiline top level statement, addresses #893

olafurpg:
Supersedes #900

### DIFF
--- a/core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
+++ b/core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
@@ -9,6 +9,9 @@ package a {
   object a {
       val v1 = 1
       val v2 = 2
+      val v3 = {
+        3
+      }
       case class C1
       sealed trait T1
       final class FC1
@@ -16,6 +19,9 @@ package a {
       def b = v2
       def c: Int
       def d: Int = 2
+      def e: Int = {
+        3
+      }
       def e = a match {
         case 1 => true
         case 2 => false
@@ -29,10 +35,13 @@ package a {
   import x2.b
 
   package b {
-
     object a {
       val v1 = 1
       val v2 = 2
+
+      val v3 = {
+        3
+      }
 
       case class C1
 
@@ -41,12 +50,13 @@ package a {
       final class FC1
 
       def a = v1
-
       def b = v2
-
       def c: Int
-
       def d: Int = 2
+
+      def e: Int = {
+        3
+      }
 
       def e = a match {
         case 1 => true
@@ -135,13 +145,77 @@ object A {
   class B
   @Singleton
   class B @Inject()(val x: Int)
+  @js.native
+  class C() {
+    def c = 5
+  }
 }
 >>>
 object A {
-
   @Annotation
   class B
 
   @Singleton
   class B @Inject()(val x: Int)
+
+  @js.native
+  class C() {
+    def c = 5
+  }
+}
+
+<<< Don't add NL before single-line defs
+object Main {
+  def x = 1
+  def y = 2
+  def z = 3
+
+  def foo = {
+    println("foo")
+  }
+
+  def bar = {
+    println("bar")
+  }
+}
+>>>
+object Main {
+  def x = 1
+  def y = 2
+  def z = 3
+
+  def foo = {
+    println("foo")
+  }
+
+  def bar = {
+    println("bar")
+  }
+}
+
+<<< Add NL before multi-line statements
+object Main {
+  def x = 1
+  def y = 2
+  def z = 3
+  def foo = {
+    println("foo")
+  }
+  def bar = {
+    println("bar")
+  }
+}
+>>>
+object Main {
+  def x = 1
+  def y = 2
+  def z = 3
+
+  def foo = {
+    println("foo")
+  }
+
+  def bar = {
+    println("bar")
+  }
 }
--- a/core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
+++ b/core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
@@ -12,9 +12,6 @@ package a {
       val v3 = {
         3
       }
-      case class C1
-      sealed trait T1
-      final class FC1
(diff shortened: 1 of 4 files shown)


core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
```diff
@@ -144,4 +158,117 @@ object A {
...
+      4
+    }
+  }
+}
+>>>
+object a {
+
+  def foo = {
+
+    def b = {
```
olafurpg: @Daxten @dwijnand do you think a newline should be forced here? We could make it stop for nested defs, but I'm not sure if that makes more sense

core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
```diff
@@ -144,4 +158,117 @@ object A {
...
+  }
+  def bar = 2
+}
+>>>
+object a {
+
+  def foo = {
+    2
+  }
+  def bar = 2
```
olafurpg: IMO, there should be a newline before `bar` because `foo` is multiline, what do you think @Daxten @dwijnand ?

core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
Daxten: I think the question is if the first definition inside (anything) should have a newline like:
```
object A {
  object B {

  }
}
```

I think for anything with lots of pameters it would look better to have a newline before the childs start:
```
class A (val a: Int. 
              val a: Int. 
              val a: Int. 
              val a: Int. 
              val a: Int
            ) {

     class B (val a: Int. 
              val a: Int. 
              val a: Int. 
              val a: Int. 
              val a: Int
            ) 
}
```

so maybe just add it generally?

core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
Daxten: Yes!

olafurpg: OK, we can leave that one for a separate PR. I suspect it will not be too hard to accommodate customizations like this one.
 
I'm gonna quickly merge this PR to unblock #898, which moves almost all source files to a new directory layout.

core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
olafurpg: This new scheme of detecting top level statements is very easy to extend. The trick is to replace this `collect` here https://github.com/scalameta/scalafmt/pull/904/files#diff-4f68f079630ceb866ab2d90e4eacf3e3R182 with a custom `Traverser`. That will make it possible to for example exclude the first statement of `Template`.

We can leave this for a separate PR.

olafurpg:
I'm gonna go ahead and merge this to unblock #898 , I've been planning that PR forever and I was determined to get it in this weekend 😄

core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
```diff
@@ -34,29 +37,37 @@ package a {
...
-
+<<< single line
+object a {
+case class C1
+sealed trait T1
+final class FC1
+}
+>>>
+object a {
+  case class C1
```
olafurpg: BTW @Daxten , following @dwijnand request, I changed the behavior to not force blank lines on single-line class/object/trait. I think this makes sense for ADTs,

```scala
sealed trait MyEnum
object MyEnum {
  case class A() extends MyEnum
  case object Foo extends MyEnum
}
```

core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
```diff
@@ -144,4 +158,117 @@ object A {
...
+}
+<<< Double annotation
+object a {
+  @annot
+  @bannot
+  class A
+}
+>>>
+object a {
+
```
olafurpg: @Daxten Before this PR, iirc, this newline would not have been inserted, right? If so, I'm happy to follow up with another PR changing it back to the old behavior.

core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
dwijnand: Either don't force or force when the previous declaration is multi-line sounds good to me.

core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
dwijnand: Yes!

core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
```diff
@@ -34,29 +37,37 @@ package a {
...
-      case class C1
-
-      sealed trait T1
-
-      final class FC1
-
+      val v3 = {
+        3
+      }
       def a = v1
```
dwijnand: I expect a NL here

core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
```diff
@@ -124,7 +135,6 @@ object A {
 
   /** test */
   def foo(): Int = 0
-
```
dwijnand: I expect this newline. A comment/scaladoc above a single-line makes it "multi-line", for me.

core/src/test/resources/newlines/alwaysBeforeTopLevelStatements.stat
```diff
@@ -144,4 +158,117 @@ object A {
...
+      4
+    }
+  }
+}
+<<< single line following multiline (up for debate)
+object a {
+  def foo = {
+    2
+  }
+  def bar = 2
```
dwijnand: for the same reason below, there should be a NL here because foo is multi-line

PR #905, opened 2017-04-23
Remove homebrew.

olafurpg:
For v1.0, I want to simplify the publish steps. Currently, publishing
takes a lot of time because it involves several flaky steps including

- assembling fatjar (15 mb)
- upload fatjar to github
- download fatjar from github to get sha (I can't run sha against
  original local copy because it changes after uploading to github)
- automatically update homebrew script with dowload url + sha

Almost every time I publish, something in this process breaks and it
takes me 5-10 minutes to fix. After almost 70 scalafmt releases I want
to find a more sustainable release process. From now on, I want to
publish only to Maven/Bintray and use coursier as the main way to
install the cli.

So, instead of `homebrew upgrade scalafmt`, people can run the following
command on any platform (Windows, Linux or Mac)

```
sudo coursier bootstrap --standalone com.geirsson:scalafmt-cli_2.12:0.7.0-RC1 -o /usr/local/bin/scalafmt -f --main org.scalafmt.cli.Cli
```

This command builds a fatjar from maven published artifacts. A nice
side-effect is that coursier will use the existing cache to minimize
downloads. This means that if you upgrade from scalafmt 1.0 to 1.1 and
the underlying scala.meta dependency hasn't changed, you only need to
download the new scalafmt artifacts, saving several megabytes in
downloads.

Another nice benefit to coursier is that users can more easily configure
the JVM options.

I admit that that coursier command is harder to remember than `homebrew
upgrade scalafmt`.  I would like to open a discussion in
coursier/coursier about adding a simpler syntax or support for a
`coursier.conf` configuration to alias maven coordinates, etc. I believe
it will be possible to turn that long command into something like
`coursier install scalafmt` and `coursier upgrade scalafmt 0.7.0`.


ShaneDelmore:
I'll miss the sbt and coursier distributions but can understand you not wanting the maintenance burden.   My team made use of the sbt and homebrew distributions for two separate reasons, any tips on replicating them with coursier would be appreciated.

1.  I liked sbt plugin because I can't get additional software installed and updated on the CI servers at work, sbt plugins are part of the build though.

2.  Brew update && brew upgrade is something all devs on my team just do routinely, even the non-scala devs as everyone uses homebrew.  With coursier it is inevitable that people will not run a separate command to update it and different developers will be on different versions.  This is mildly problematic because we have a git hook preventing commit of files that are out of format, that can be triggered by an old version of scalafmt.  Is there perhaps some sort of versioning I could add to the scalafmt.conf to at least remind people they need to upgrade scalafmt and what command to run to update it?

olafurpg:
@ShaneDelmore 

1. You can continue to use scalafmt in sbt with the workaround explained here http://scalameta.org/scalafmt/#sbt0.13
2. What I use in scalameta/scalameta and scalacenter/scalafix is the "slim bootstrap" script option here http://scalameta.org/scalafmt/#Coursier This script only needs to be generated once and then commited to git. By sharing a script that's committed into VCS you guarantee that everyone on the team use the same scalafmt version for that project (regardless of what OS they're on). My biggest gripe with homebrew is that it only works on Mac. It's amazing for Mac users, but it's not a complete solution for project maintainers or other users (which are many!). I love Homebrew, I only wish we had it available on all platforms.

ShaneDelmore:
Looks good to me.

```
==> Upgrading olafurpg/scalafmt/scalafmt
==> Cloning git://github.com/scalameta/scalafmt.git
Cloning into '/Users/sdelmore/Library/Caches/Homebrew/scalafmt--git'...
remote: Counting objects: 470, done.
remote: Compressing objects: 100% (381/381), done.
remote: Total 470 (delta 8), reused 212 (delta 4), pack-reused 0
Receiving objects: 100% (470/470), 489.12 KiB | 0 bytes/s, done.
Resolving deltas: 100% (8/8), done.
==> Checking out branch master
🍺  /usr/local/Cellar/scalafmt/HEAD-908ca61: 5 files, 27.1KB, built in 3 seconds
 ~/Documents  scalafmt -v
scalafmt 0.7.0-RC1
```

olafurpg:
Amazing! I borrowed the idea from https://github.com/paulp/homebrew-extras/blob/master/coursier.rb I need to update the installation docs to add homebrew again ^^

ShaneDelmore:
I'm pretty happy you managed to come up with such a low effort solution.  Now I don't have to feel bad about you maintaining a plugin you don't use 👍

olafurpg:
Only if it was as easy to support the sbt plugin 😄

PR #907, opened 2017-04-24
Fix SBT 0.13 bootstrap dependency declaration.

nightscape:


nightscape:
Ah, that's the place I was looking for :wink: 
Here we go: https://github.com/scalameta/scalafmt/pull/908

PR #910, opened 2017-04-25
Add support for auto-binpacking of argument lists with idents only

olafurpg:
Following @jvican's suggestion, this commit adds support for automatically binpack argument lists if they only contain plain identifiers.


olafurpg:
Here is an example diff with this change

https://github.com/sbt/sbt/compare/1.0.x...olafurpg:binpack-names?expand=1

Issue #927, opened 2017-05-09
bootstrap "scalafmt --diff" fails with FNFE for build.sbt when run from subdir


kitbellew:
This template is a guideline, not a strict requirement.
- **Version**: 0.7.0-RC1
- **Integration**: bootstrap CLI
- **Configuration**:
```
style = default
maxColumn = 100
lineEndings = unix
continuationIndent.callSite = 2
continuationIndent.defnSite = 4
continuationIndent.extendSite = 4
align.tokens = []
align.arrowEnumeratorGenerator = false
align.openParenCallSite = false
includeCurlyBraceInSelectChains = true
optIn.annotationNewlines = true
optIn.configStyleArguments = true
optIn.breakChainOnFirstMethodDot = true
newlines.alwaysBeforeTopLevelStatements = false
newlines.alwaysBeforeCurlyBraceLambdaParams = true
newlines.afterCurlyLambda = never
newlines.sometimesBeforeColonInMethodReturnType = false
newlines.penalizeSingleSelectMultiArgList = true
newlines.afterImplicitKWInVerticalMultiline = true
newlines.beforeImplicitKWInVerticalMultiline = true
binPack.literalArgumentLists = true
binPack.parentConstructors = false
runner.optimizer.forceConfigStyleOnOffset = 50
runner.optimizer.forceConfigStyleMinArgCount = 3
rewrite.rules = [ExpandImportSelectors, PreferCurlyFors]
danglingParentheses = true
runner.fatalWarnings = true
verticalMultilineAtDefinitionSite = false
```

## Steps

I ran scalafmt like this:

```bash
git clone xxxx/yyy.git
cd yyy # in git top-level dir
mkdir utils
pushd utils
# store scalafmt in a subdirectory
coursier bootstrap com.geirsson:scalafmt-cli_2.11:0.7.0-RC1 --main org.scalafmt.cli.Cli -o scalafmt
popd # back to git top-level dir
vim .scalafmt.conf
pushd spark # go to project subdir
ln -s ../.scalafmt.conf . # this is for Intellij
../utils/scalafmt --diff
```

## Problem

Scalafmt exits with the ```FileNotFoundException``` looking for ```yyy/spark/spark/build.sbt``` (note "spark/" is mentioned twice).

## Expectation

I would like the tool not to fail and find build.sbt in ```yyy/spark/build.sbt``` if necessary.

## Workaround

[...]

Issue #875, opened 2017-04-05
Possible scalafmt (cli) interface changes


pjrt:
Currently the cli is a bit confusing when it comes to changing single files.

* Calling `scalafmt` (no args) will change all files recursively from where scalafmt is called. 
* Calling `scalafmt -f <file>` will not change the file, but will instead print to stdout the formatted files
    * Also, `<file>` is a list of comma separated files, which is odd
* Calling `scalafmt -i -f <file>` will change the files "in place" (ie: not print it to stdout but actually change the file)
* Calling `scalafmt` will change files in place, but adding ANY flag (eg: `--quiet` or `--non-interactive`) will cause it to print to stdout

I personally find this confusing. 

Proposal:

Require at least one file to be passed to scalafmt. If any of the files passed is a directory, the search recursively for scala files and format them in-place. We could add a new flag for the "output to stdout" case. 

`scopt` already supports such a feature. from their README:

```
arg[File]("<file>...").unbounded().optional().action( (x, c) =>
    c.copy(files = c.files :+ x) ).text("optional unbounded args")
```

Changing the `.optional()` to `.required()` would produce what the proposal would need.

pjrt:
> Require at least one file to be passed to scalafmt

Ofc we could just say "if no files are passed, then assume `./`". Using `.optional()` would do that.

olafurpg:
I apologize for how slow I've been at responding to this issue. Thank you for bringing this up.

I whole-heartedly agree that it would be nice to make the `-f` flag optional. The only reason it's not like that is because I couldn't figure out how to do with scopt.

> Require at least one file to be passed to scalafmt. If any of the files passed is a directory, the search recursively for scala files and format them in-place. We could add a new flag for the "output to stdout" case.

I like the idea of spec-ing something consistent. The current design is the result of random design choices that were made from release to release. However, we need to make sure that what we come up with plays nicely with `project.git` and `project.exclude/includeFilters`.

I propose that
- `project.git` is implemented as an `includeFilter`
- every candidate file must match includeFilters AND not match excludeFilters
- and as you say, if no directory is passed, then assume `./`. This keeps compatibility with the current `./scalafmt` interface which is how I run scalafmt 99% of the time tbh

dwijnand:
Came to file something like this. I often do this:

```
$ scalafmt main/src/main/scala/sbt/EvaluateTask.scala
Error: Unknown argument 'main/src/main/scala/sbt/EvaluateTask.scala'
[...]

PR #939, opened 2017-05-22
Remake the CLI interface

pjrt:
Implement #875: Remake scalafmt's CLI

* Remove --files
  * Instead, files are take in as arguments (like a normal cli)
  * If no arguments are passed, `./` is assumed
* Remake how we handle write-modes
  * Write modes can either be `Override` (in-place) or `Stdout` (--stdout)
  * `-i/--in-place` is now the default all of the time
  * `--stdout` will print to stdout the formatted files
* Implement a `--git=[true|flase]` flag that will override `.scalafmt.conf`
* Fix an issue where the default `includeFilters` accepted more than just `scala` and `sbt`
* [Internal] Simplify how we fetch files (now it all goes through one function)
* [Internal] Simplify how we use CliOptions

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
@@ -35,8 +35,8 @@ object CliArgParser {
     new scopt.OptionParser[CliOptions]("scalafmt") {
       override def showUsageOnError = false
 
-      private def printAndExit(inludeUsage: Boolean)(ignore: Unit,
-                                             c: CliOptions): CliOptions = {
+      private def printAndExit(
+          inludeUsage: Boolean)(ignore: Unit, c: CliOptions): CliOptions = {
         if (inludeUsage) showUsage
         else showHeader
         sys.exit
@@ -81,9 +81,8 @@ object CliArgParser {
         .action((opt, c) => c.copy(writeMode = Stdout))
         .text("write formatted files to stdout")
 
-
       opt[Boolean]("git")
-        .action((opt, c) => c.copy(inputGit = opt))
+        .action((opt, c) => c.copy(git = Some(opt)))
         .text("if true, ignore files in .gitignore (default false)")
       opt[Seq[String]]("exclude")
         .action((excludes, c) => c.copy(customExcludes = excludes))
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
@@ -35,7 +35,7 @@ object CliArgParser {
     new scopt.OptionParser[CliOptions]("scalafmt") {
       override def showUsageOnError = false
 
-      def printAndExit(inludeUsage: Boolean)(ignore: Unit,
+      private def printAndExit(inludeUsage: Boolean)(ignore: Unit,
                                              c: CliOptions): CliOptions = {
         if (inludeUsage) showUsage
         else showHeader
@@ -43,15 +43,20 @@ object CliArgParser {
         c
       }
 
-      def readConfigFromFile(file: String, c: CliOptions): CliOptions = {
+      private def readConfigFromFile(file: String, c: CliOptions): CliOptions = {
         readConfig(
           FileOps.readFile(
             AbsoluteFile.fromFile(new File(file), c.common.workingDirectory)),
           c
         )
       }
-      def readConfig(contents: String, c: CliOptions): CliOptions = {
+      private def readConfig(contents: String, c: CliOptions): CliOptions = {
         c.copy(config = Config.fromHoconString(contents).get)
+      }
+
+      private def addFile(file: File, c: CliOptions): CliOptions = {
+        val absFile = AbsoluteFile.fromFile(file, c.common.workingDirectory)
+        c.copy(customFiles = c.customFiles :+ absFile)
       }
 
       head("scalafmt", Versions.nightly)
@@ -61,14 +66,24 @@ object CliArgParser {
       opt[Unit]('v', "version")
         .action(printAndExit(inludeUsage = false))
         .text("print version ")
-      opt[Seq[File]]('f', "files")
-        .action { (files, c) =>
-          c.copy(
-            customFiles =
-              AbsoluteFile.fromFiles(files, c.common.workingDirectory))
-        }
-        .text(
-          "file or directory, in which case all *.scala files are formatted.")
+
+      arg[File]("<file>...")
+        .optional()
+        .unbounded()
+        .action((file, c) => addFile(file, c))
+        .text("file or directory, in which case all *.scala files are formatted.")
+
+      opt[WriteMode]('w', "write-mode")
+        .action((opt, c) => c.copy(writeMode = opt))
+        .text("write mode. Can be stdout or override (default).")
+
+      opt[Unit]('i', "in-place")
+        .action((opt, c) => c.copy(writeMode = Override))
+        .text("alias for --write-mode=override")
+      opt[Unit]("stdout")
+        .action((opt, c) => c.copy(writeMode = Stdout))
+        .text("alias for --write-mode=stdout")
+
       opt[Seq[String]]("exclude")
         .action((excludes, c) => c.copy(customExcludes = excludes))
         .text(
@@ -85,9 +100,6 @@ object CliArgParser {
       opt[String]("assume-filename")
         .action((filename, c) => c.copy(assumeFilename = filename))
         .text("required to format .sbt files with --stdin flag.")
-      opt[Unit]('i', "in-place")
-        .action((_, c) => c.copy(inPlace = true))
-        .text("write output to file, does nothing if file is not specified")
       opt[Unit]("test")
         .action((_, c) => c.copy(testing = true))
         .text("test for mis-formatted code, exits with status 1 on failure.")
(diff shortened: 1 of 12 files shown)


scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
```diff
@@ -35,40 +35,55 @@ object CliArgParser {
...
+      }
+
       head("scalafmt", Versions.nightly)
       opt[Unit]('h', "help")
         .action(printAndExit(inludeUsage = true))
         .text("prints this usage text")
       opt[Unit]('v', "version")
         .action(printAndExit(inludeUsage = false))
         .text("print version ")
-      opt[Seq[File]]('f', "files")
```
olafurpg: Would it be possible to keep this? We can make it hidden so that it won't appear in --help. Removing the option would break a lot of wrapper scripts, for limited benefits IMO.

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
```diff
@@ -179,7 +177,6 @@ class CliTest extends FunSuite with DiffAssertions {
           |""".stripMargin
     val options = getConfig(
       Array(
-        "--files",
         input.path,
         "-i"
```
olafurpg: Remove?

scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
```diff
@@ -35,40 +35,55 @@ object CliArgParser {
...
-        }
+
+      arg[File]("<file>...")
+        .optional()
+        .unbounded()
+        .action((file, c) => addFile(file, c))
         .text(
           "file or directory, in which case all *.scala files are formatted.")
+
+      opt[Unit]('i', "in-place")
```
olafurpg: When is this used?

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
```diff
@@ -135,26 +133,26 @@ class CliTest extends FunSuite with DiffAssertions {
...
     val args = Array(
-      "--files",
       tmpFile.toFile.getPath,
       "--test"
     )
     val formatInPlace = getConfig(args)
     intercept[MisformattedFile] {
       Cli.run(formatInPlace)
     }
+    val str = new String(Files.readAllBytes(tmpFile))
```
olafurpg: `FileOps.readFile()`

scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
```diff
@@ -35,15 +35,13 @@ class GitOpsImpl(workingDirectory: AbsoluteFile) extends GitOps {
...
           "git",
-          "ls-tree",
-          "-r",
-          "HEAD",
-          "--name-only"
+          "ls-files",
+          dir.path
         )
       ).split(OsSpecific.lineSeparator).toSeq.collect {
         case f if new File(f).isFile => workingDirectory / f
```
olafurpg: While you're at it, can you maybe change workingDirectory to rootDir? That should fix https://github.com/scalameta/scalafmt/issues/927

scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
```diff
@@ -35,40 +35,55 @@ object CliArgParser {
...
-      opt[Seq[File]]('f', "files")
-        .action { (files, c) =>
-          c.copy(
-            customFiles =
-              AbsoluteFile.fromFiles(files, c.common.workingDirectory))
-        }
+
+      arg[File]("<file>...")
+        .optional()
+        .unbounded()
```
olafurpg: Whoop whoop 👏 I didn't know this was possible, was always bummed that scopt forced files to be comma separated. cc/ #940

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
```diff
@@ -76,14 +76,12 @@ class CliTest extends FunSuite with DiffAssertions {
   def gimmeConfig(string: String): ScalafmtConfig =
     Config.fromHoconString(string).get
 
-  test("scalafmt -i --file tmpFile") {
+  test("scalafmt tmpFile") {
```
olafurpg: Can we add or modify a test  to check that we can pass in multiple files like `scalafmt Foo.scala Bar.scala`?

scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
pjrt: This is the same as `in-place` from before (I just moved it from further down to up here for organizational purposes. And made it set the write-mode).

scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
pjrt: Yeah, I was thinking that since this is 1.0 that we could break stuff. But now that I re-think it, there is no reason not to have this as a deprecated flag.

olafurpg:
I am not sure what is causing the test failure on Windows, any idea @pjrt ? https://ci.appveyor.com/project/olafurpg/scalafmt/build/127#L3135

```scala
[info] - excludefilters are respected *** FAILED *** (63 milliseconds)
[info]   ===========
[info]   => Obtained
[info]   ===========
[info]   /foo.sbt
[info]   lazy val x = project
[info]   
[info]   /target/FormatMe.scala
[info]   object PleaseFormatMeOtherwiseIWillBeReallySad {}
[info]   
[info]   /target/nested/DoNotFormatMe.scala
[info]   object AAAAAAIgnoreME {}
[info]   
[info]   /target/nested/nested2/DoNotFormatMeToo.scala
[info]   object BBBBBBIgnoreME {}
[info]   
[info]            
[info]   =======
[info]   => Diff
[info]   =======
[info]    /target/nested/DoNotFormatMe.scala
[info]   -object AAAAAAIgnoreME {}
[info]   +object    AAAAAAIgnoreME   {  }
[info]   ∙
[info]    /target/nested/nested2/DoNotFormatMeToo.scala
[info]   -object BBBBBBIgnoreME {}
[info]   +object    BBBBBBIgnoreME   {  } (DiffAssertions.scala:37)
```

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
```diff
@@ -7,7 +7,7 @@ import org.scalafmt.util.OsSpecific
 case class ProjectFiles(
     git: Boolean = false,
     files: Seq[String] = Nil,
-    includeFilters: Seq[String] = Seq(".*\\.scala", ".*\\.sbt"),
+    includeFilters: Seq[String] = Seq(".*\\.scala$", ".*\\.sbt$"),
```
pjrt: This was, I think, a bug. Currently it will pick up, for example, `something.scalatex` files.

olafurpg: Could you maybe add a test for this change?

scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
olafurpg: After some thinking, I think it's good to keep this even though it won't necessarily be used. The --help message mentions this is the default behavior, which can be useful for people exploring the cli.

olafurpg: And I like moving this next to --stdout 👍

pjrt: Ohh, I see now what you mean. Yeah, it doesn't make sense to have it anymore since it is the default. 

However, if we are keeping `-f` then we should keep `-i` (since a lot of current scripts probably have `-i -f`).

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
pjrt: Will do.

pjrt:
I was wondering what that test was about. it passes locally, I hadn't realized it was failing on Windows. 

Not sure what the issue is, I'll investigate.

scalafmt-tests/src/test/scala/org/scalafmt/util/DiffAssertions.scala
```diff
@@ -14,10 +16,11 @@ trait DiffAssertions extends FunSuiteLike {
...
                          expected: String,
                          obtained: String,
-                         diff: String)
+                         diff: String)(implicit pos: Position)
       extends TestFailedException(
-        title + "\n" + error2message(obtained, expected),
-        1)
+        _ => Some(title + "\n" + error2message(obtained, expected)),
+        None: Option[Throwable],
+        pos)
```
pjrt: This looks like a big change but it produces the same output (except that now it will point to the place in CliTests where it failed, not to this file). It just looks this bad because `TestFailedException` has an odd interface (it has many overloaded `this` but it doesn't (surprisingly) have the basic, and probably common, "a message and a Position").

pjrt:
Odd, it failed to compile in PR, but it works locally (cleaned and all).

PR #947, opened 2017-05-26
Enable Travis CI

olafurpg:


olafurpg:
I'm disabling Appveyor too, it simply has too many false negatives caused by resolution/cache/config setup.

olafurpg:
cc/ @pjrt CI is green again 🎉  https://travis-ci.org/scalameta/scalafmt/builds/236548732 I switched to Travis because it seems our Drone server is getting rate limited or something similar. I never got the sftp cache plugin working properly so the build always downloaded way more jars than it needed.

olafurpg:
I also disabled the ci-slow tests, the ones that run scalafmt on millions of lines of code. The Travis servers always kill those jobs because they consume too much resources. ci-slow only tested the `default` style anyways and most PRs touch custom config options or the docs

PR #964, opened 2017-06-04
Enable alignment on case arrows by default

olafurpg:


olafurpg:
@dwijnand this PR adds support for new alignment options
```
align = false # disabled all alignment
align = default # case arrows and open paren ( in defn/call site
align = true # same as defaultWithAlign
align = everything # new experimental option to align as much as reasonably possible
```

Any suggestions for better names? For example, `s/false/nothing/` and `s/true/alot/`, open for suggestions.

dwijnand:
How about

```
align = none # disabled all alignment
align = some # case arrows and open paren ( in defn/call site
align = more # same as defaultWithAlign
align = most # new experimental option to align as much as reasonably possible
```

with `default` being a synonym for `some` (for now 😸)?

olafurpg:
I see what you did there @dwijnand 😉 

screen shot 2017-06-05 at 09 59 23

Nice alignment 👍

`align=more` is kind of weird as a replacement for `style=defaultWithAlign`, however. Any other 4 char proposal?

dwijnand:
```
align = nil # disabled all alignment
align = min # case arrows and open paren ( in defn/call site
align = std # same as defaultWithAlign
align = all # new experimental option to align as much as reasonably possible (also "pro" :D)
```

`align = base`?

I actually don't feel strongly about what you pick. Thank you for accommodating.

olafurpg:
None, Some, More and Most it will be 🎉

PR #972, opened 2017-06-12
update the doc about initial configurations

gigiyy:
the doc has wrong configurations for Align. updated to correct one


gigiyy:
sorry. only found out after reading the history changes

olafurpg:
No worries! Thank you for sending a PR to fix the docs! :)

PR #979, opened 2017-06-17
Use lenient dialect by default

olafurpg:


scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -164,10 +164,7 @@ object Cli {
     }
 
     val sbtOptions = options.copy(
-      config = options.config.copy(
-        runner = options.config.runner.copy(
-          dialect = Sbt0137
-        )))
+      config = options.config.copy(runner = options.config.runner.forSbt))
```
olafurpg: cc/ @pauldraper You might want to use `ScalafmtRunner.forSbt` instead of `dialects.Sbt0137` in ScalafmtFactory. The former will re-use the existing configured dialect while the latter is Scala211 for sbt files.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -238,8 +238,8 @@ object TreeOps {
     case _: Decl.Def | _: Defn.Def | _: Defn.Macro | _: Defn.Class |
         _: Defn.Trait | _: Ctor.Secondary =>
       true
-    case x: Ctor.Primary => x.parent.exists(isDefnSiteWithParams)
-      true
```
xeno-by: Wow

Issue #976, opened 2017-06-16
Off-by-one error in maxColumn


alexnixon:
This template is a guideline, not a strict requirement.
- **Version**: 1.0.0-RC3
- **Integration**: IntelliJ
- **Configuration**:
```
align = true
maxColumn = 80
```

Firstly - I've just started using scalafmt and it seems like a fantastic tool. Thanks!

Given `maxColumn = 80` I would expect scalafmt to *allow* a character in the 80th column. However scalafmt reformats
```
object Wat {
  val a = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
}
```
as
```
object Wat {
  val a =
    "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
}
```
Despite the line not exceeding 80 characters in length. It's particularly noticeable when you have Intellij's margin marker enabled which shows your code as being within bounds:

![image](https://user-images.githubusercontent.com/121612/27227993-bf2d34f6-529d-11e7-897f-df6977fb70fe.png)

olafurpg:
Thank you for reporting! You are correct, I've opened #982 changing it so that scalafmt allows a character on the n-th column.

PR #982, opened 2017-06-17
Fix off-by-one error, #976.

olafurpg:
Interestingly, this had not been pointed out before.


olafurpg:
Review @alexnixon

olafurpg:
This change broke a bunch of tests, which relied on the old behavior, so I took a shortcut and moved the maxColumn setting to 79 for the test suite 😅

olafurpg:
This is quite a big change, but I think it's the right thing to do.

alexnixon:
With 0 knowledge of the codebase...LGTM! Thanks for jumping on this so quickly.

olafurpg:
I added a test case @pjrt

PR #985, opened 2017-06-18
Adding of --no-stderr option

### LINKED ISSUE #967: Scalafmt Cli doesn't work with Appveyor and powerhell script
- **Version**: 0.6.8
- **Integration**: CLI
- **Configuration**: Appveyor and powershell

## Steps

I used coursier to bootstrap jar file.
I ran scalafmt on AppVeyor from powershell script `scalafmt.ps1`:
`&java -jar $SCALAFMT --test`

(`$SCALAFMT` is path to downloaded jar)

It also could be reproduced locally in powershell by command:
`&java -jar $SCALAFMT --test 2>error.log`

(then `error.log` will have an exception)

## Problem

Output looks like that:
`java.exe : [2KLooking for unformatted files... (2.68 %, 18 / 671)`

Appveyor sees raised exception from the script:
```
At C:\bin\scalafmt.ps1:42 char:13
    + CategoryInfo          : NotSpecified: ([2KLooking for...68 %, 18 / 671):String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
```

## Expectation

No exception while outputs strings.
it would be good to have a command line option to not output `[2K` before strings,
that prefix raised native exception and breaks the rest output.

## Workaround

I've found that I can suppress all output to stderr but it also suppresses all output from scalafmt,
that it also interesting question why `scalafmt` outputs everything to stderr? even normal text

muxanick:
Resolves #967 issue,
don't output to StdErr if option set.


scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -132,7 +132,8 @@ object Cli {
                      inputMethods: Seq[InputMethod],
                      msg: String): TermDisplay = {
     val termDisplay = new TermDisplay(
-      new OutputStreamWriter(options.common.err),
+      new OutputStreamWriter(
+        if (options.noStdErr) options.common.out else options.common.err),
```
olafurpg: How about we add a method `CliOptions.info: OutputStream` that performs this logic?

scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -155,9 +155,11 @@ object Cli {
       else "Reformatting..."
     if (options.debug) {
       val pwd = options.common.workingDirectory.jfile.getPath
-      options.common.err.println("Working directory: " + pwd)
-      options.common.err.println("Formatting files: " + inputMethods.toList)
-      options.common.err.println(
+      val out =
+        if (options.noStdErr) options.common.out else options.common.err
+      out.println("Working directory: " + pwd)
```
olafurpg: Let's use `.info` here too.

olafurpg:
Once the compilation error is fixed then we are good to go! https://travis-ci.org/scalameta/scalafmt/jobs/244220714#L312

olafurpg:
btw @muxanick what do you think about enabling this setting by default in windows?

muxanick:
@olafurpg it's your call I think

olafurpg:
I think it makes sense to enable it by default in Windows if it doesn't work without it. Could you open a PR?

muxanick:
@olafurpg what is the best way to check if it's Windows?

olafurpg:
```scala
scala> sys.props("os.name")
res0: String = Mac OS X
```
?

olafurpg:
Honestly, I'm fine with enabling it by default in all OSes as long as the --stdout option is not used. The progress bar should not be used when --stdout is enabled either.

muxanick:
@olafurpg I think it will not work with `--stdin`, right?

olafurpg:
Correct, it won't work with either --stdout or --stdin. That combination is used by integrations with editors like vim.

muxanick:
I'm not sure what is your final decision?

olafurpg:
If --no-stderr solves the problem you faced, then I propose we make it the default behavior in all cases when stdout and stdin are both false.

PR #986, opened 2017-06-18
Re-enable appveyor

olafurpg:
Following advice from @muxanick


olafurpg:
I suspect the cause is the 0.13.15 sbt launcher, is it possible to tell scoop to install 0.13.13? @muxanick

muxanick:
@olafurpg It shouldn't be a problem, it will start 13.13 even if 13.15 is installed

olafurpg:
The error seems to be reported by the sbt launcher before it loads the build. The log doesn't show typical lines like

```
[info] Loading global plugins from /Users/ollie/.sbt/0.13/plugins
[info] Updating {file:/Users/ollie/.sbt/0.13/plugins/}global-plugins...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
```

olafurpg:
Seems like the .jvmopts comments were the culprit 🤔

olafurpg:
I took a shortcut and don't run readme/test. This is a good start however and the downloaded artifacts will hopefully get cached after the merge. It would be good to run the integration tests as well.

PR #987, opened 2017-06-18
fix for jvmopts and enabling no-stderr by default if possible

muxanick:


olafurpg:
You can rebase on top of master to use the fixes from #986

muxanick:
@olafurpg please check latest commit `scalafmt` with changes scala files

olafurpg:
It seems the .scalafmt.conf `project.excludeFilters` setting is not being respected. Those files are not supposed to be formatted. The `/` in the filepaths in .scalafmt.conf should be converted to `\` here https://github.com/scalameta/scalafmt/blob/d204ae1437247aae0f0b8a9901e433368057e678/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala#L14-L15 Any idea why the exclude filters are not respected?

The diff from the fatjar is likely caused by https://github.com/scalameta/scalafmt/pull/982 the codebase is formatted against the latest maven release, not master.

muxanick:
I removed all formatted files, but I don't know why it's not excluded by config. I can try to investigate but I'm not familiar with codebase.

olafurpg:
I apologize the super slow response, this PR got totally lost in my inbox. LGTM 👍 Thank you @muxanick !

PR #998, opened 2017-07-10
Implement option for discarding whitespace between a keyword and left-paren

nrinaudo:
## spaces.afterKeywordBeforeParen
Default: `true`

```scala
// spaces.afterKeywordBeforeParen = false
if(a) foo()
while(a) foo()
for(a <- as) foo()

// spaces.afterKeywordBeforeParen = true
if (a) foo()
while (a) foo()
for (a <- as) foo()
```

### DIFF
--- a/readme/Configuration.scalatex
+++ b/readme/Configuration.scalatex
@@ -379,6 +379,21 @@
       implicit def validatedInstances[E](implicit E: Semigroup[E]): Traverse[
           Validated[E, ?]] with ApplicativeError[Validated[E, ?], E] = 2
 
+  @sect{spaces.afterKeywordBeforeParen}
+    Default: @b(default.spaces.afterKeywordBeforeParen)
+
+    @hl.scala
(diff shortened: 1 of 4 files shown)


hseeberger:
Ain't that misnamed? When spaces... true I'd expect a space to be present and vice versa. But it seems to work the other way round.

nrinaudo:
Ah. I think my example is incorrect, not the actual code, but you're right: something is wrong. I'll amend soon as I get behind a proper computer.

nrinaudo:
@hseeberger turns out this was wrong *everywhere*. Thanks for spotting it!

This is now fixed. I've updated the PR's description to reflect the actual behaviour.

hseeberger:
That was fast ;-)

nrinaudo:
Well, this isn't likely to be a popular PR - I'm apparently alone in the entire universe to prefer that coding style - so I need to show motivation for it to have any chance of being merged :)

Issue #975, opened 2017-06-14
Upgrade Scalameta  dependency


olafurpg:
- [x] remove scalameta paradise macro annotations
- [x] upgrade to v1.8.0 (handle xml literals + new lambdas)
- [x] upgrade to latest metaconfig
- [x] upgrade to scalameta v2

olafurpg:
Given that scalameta trees went through a major refactoring in https://github.com/scalameta/scalameta/pull/907 to be released as v2.0 I suspect scalafmt will stick on scalameta v1.7 for a while longer.

olafurpg:
Full list of breaking changes in v2.0 https://github.com/scalameta/scalameta/releases/tag/v2.0.0

olafurpg:
I consider this issue a ticking timebomb since Scalafmt is unable to enjoy parser improvements in Scalameta. I haven't been motivated to spend time upgrading because I am personally most excited about getting a proper tree printer (https://github.com/scalameta/scalafmt/issues/917) instead of token printer (current scalafmt impl).

I made really nice progress in https://github.com/olafurpg/scala-syntax last october but haven't gotten around to complete the project. I estimate it requires 4-5 days of focused work to be usable and maybe 2-3 more weeks to replace the current scalafmt Router. The new implementation is around 5x faster in my experiments, it can format ~100k loc in the Scala Native repo in ~5 seconds, essentially bottlenecked by the parser.

In comparison, I estimate it would take ~2-3 days to get the current scalafmt impl upgraded to Scalameta v2.0.

olafurpg:
Been taking the first steps towards this direction in #1145. The upgrade is more painful than just the v1 -> v2 changes, we have to replace the usage of the old scalameta paradise macro annotations and handle changes in v1.8.0 related to lambda parsing and tokenization of XML literals. We now "only" have the v1 -> v2 upgrade remaining.

olafurpg:
Pretty stoked to share that I have a WIP and ALL TESTS PASS

> [info] Tests: succeeded 1094, failed 0, canceled 0, ignored 7, pending 0

Will be opening PR soon, this should be fixed in the next release 🎉

PR #1006, opened 2017-07-27
Add JS facade

gabro:
This PR adds a JS facade for scalafmt.

The final purpose is to have a `scalafmt` library published on npm that can power JS-based tools, such as editors and browser extensions.

Here's a first proof of concept of a VSCode extension that uses this API:

Two things to notice:

- it works 😄 
- it's FAST! 🚀  (way faster than running the cli, as we don't pay for JVM startup)

This is just a start, the next step would be to accept options (currently it uses the default ones).


gabro:
By the way, I took the liberty of publishing a beta version on npm (https://www.npmjs.com/package/scalafmt) under the [scalameta org](https://www.npmjs.com/org/scalameta). This makes the vscode extension usable by anyone interested. Here it is: https://github.com/gabro/vscode-scalafmt

gabro:
Pushed support for config and ranges. The code for parsing the config is correct, but currently hindered by this bug: https://github.com/olafurpg/metaconfig/issues/28

gabro:
So, the trickiest part turned out to be the config. metaconfig scalajs support is not fully there yet, and in general it seems destined to be replaced with something else (circe?). So I rolled up my sleeves and built the settings by hand (using shocon for the parsing).

I'm *almost* done, still a few configuration settings to go and this horrible workaround will be ready 😱

olafurpg:
Thank you for this contribution! This is super exciting. I need more time to review this PR. At a glance, I'm not excited about duplicating the config parsing logic. It's a maintenance burden to keep the two parsers in sync.  I think it should not be too hard to improve metaconfig-hocon to support fields like `a.b.c = blah` so we can use the same config parse logic on all platforms. I can try to take a look later this week.

gabro:
>  I'm not excited about duplicating the config parsing logic

Neither am I! 

It's the quickest workaround to the current state of metaconfig-hocon, but if you have some time to spend on it, I'd be more than glad to drop my horrible manual construction of the config object :)

olafurpg:
It seems the Scala.js implementation is not calling `Conf.normalize` which is the "resolve" equivalent in typesafe/config (I'm terrible with naming 😞 ). I customized the MetaconfigParser in the Scala.js side to normalize and then it seems to handle input like `project.git=true`. 

See diff https://github.com/scalameta/scalafmt/commit/29ded6c477a2865c22e292caf1baf1c059a544bf 

The reason there are no Scala.js tests is that I was gonna get Node.js IO to work first so that we run all unit test on JVM+JS. scalameta/io works on Node.js now, but I still haven't gotten around to merge it into scalafmt. In fact, scalafmt is still on scalameta 1.7 due to limited time and motivation to do the upgrade.

As for long-term solutions, I'd like to
- stick with HOCON, despite being only used in the scala community it still 
  - is the most standard choice in the scala community
  - has several desirable features above json when it comes to configuration (comments, compact syntax for nested keys).
- replace metaconfig.Conf with (fast)parsing directly into circe JSON and replace `@DeriveConfDecoder` macro annotations with circe semi-automatic derivation.
  - I might be tempted to roll some extra macros to customize error messages on invalid fields (see https://github.com/circe/circe/issues/374).
  - Manual decoders are not an option IMO, since they add friction for first-time contributors to add new configuration options.
- replace typesafe/config with this fastparse+circe approach, use same configuration implementation on JS/JVM.

bullets 2 and 3 are totally up for debate, but the configuration syntax was already debated to death in https://github.com/scalameta/scalafmt/issues/316 so i'd prefer to stick with HOCON.

olafurpg:
I just enabled metaconfig-hocon for the JVM implementation and found two failing test cases (out of 800 unit tests), the failing test case has configuration like this

```
align.treeCategory {
  Case: Assign
  "Enumerator.Generator": Assign
}
```
which the parser doesn't handle yet (same one as shocon uses).

gabro:
> stick with HOCON, despite being only used in the scala community

That's ok, I agree it's a sensible choice given the context (although I don't like the context :P )

> replace metaconfig.Conf with (fast)parsing directly into circe JSON

nice, I agree using circe derivation would be really nice.

> replace typesafe/config with this fastparse+circe approach, use same configuration implementation on JS/JVM.

👍 it would be really cool to have a single implementation. Any plan to leverage shocon directly for this? 

> which the parser doesn't handle yet (same one as shocon uses).

*almost* the same one, I recently fixed an issue upstream that we may want to backport: https://github.com/unicredit/shocon/pull/21

olafurpg:
Any plan to leverage shocon directly for this?

olafurpg:
> Any plan to leverage shocon directly for this?

If we can implement something like https://github.com/circe/circe-config/blob/d65ae7dc4f69003e7cd66a31fe41936e5157ce5b/src/main/scala/io.circe.config/parser.scala#L51 with shocon then absolutely!

gabro:
Paging @andreaTP on this one

andreaTP:
I'm probably missing some context... What is the missing feature? 🙂

olafurpg:
@andreaTP here are the errors I get when I use replace typesafe/config with shocon in circe-config https://gist.github.com/olafurpg/9b78dff4496d8100e96e148c640915af

gabro:
@andreaTP If I get it right, the final goal is to parse HOCON into the Circe JSON AST (so that we can use circe for the case class deserialization). I would like to leverage shocon instead of rolling yet another implementation of the parser.

andreaTP:
I truly think that is totally doable and we just need a little more love from @evacchi and @mfirry to enable this... Still I see that your errors are about Typesafe API and we can probably totally avoid to use it in this context, I.e. Shocon crosscompile to jvm and to Js and we can use directly the parser API instead of trying to get right a POJ one

andreaTP:
Wdyt?

olafurpg:
Parsing directly into circe Json would be ideal IMO, I never call the com.typesafe.config api directly except to convert it into a json like structure.

olafurpg:
Is this still relevant after scalameta/language-server @gabro ?

gabro:
Mm, I'll have to check from the computer, buy I think I'm using this in vscode-scalafmt (although it's not merged). I still think having a js facade is valuable for many reasons (I've seen some creative web-based uses of prettier).

olafurpg:
>  I still think having a js facade is valuable for many reasons (I've seen some creative web-based uses of prettier).

I agree! I propose we use the changes from https://github.com/scalameta/scalafmt/commit/29ded6c477a2865c22e292caf1baf1c059a544bf and merge this even if it's not 100% compatible with the hocon parser. I opened https://github.com/scalameta/scalafmt/issues/1080 to track further updates on the js hocon parser.

gabro:
I agree! Is the appveyor fail a false positive?

olafurpg:
If you repush I think it should work, I think appveyor was misconfigured when you opened this PR

laughedelic:
I'm also interested in having a scalafmt plugin for Atom. I think this is useful independently of the language server.

mfirry:
So at last I was able to give a go at Shocon version that uses circe Json. Who's willing to try it out?

mfirry:
It's very very very untested... but it's a start. I can probably push some snapshot release? Will that work out?

mfirry:
So @andreaTP is beating me very hard after I've showed him what I did. :)
I knew it though. @olafurpg please lower your expectations on this.
Will keep you posted.

olafurpg:
In the long term for shocon, I think it would actually be most valuable if it supported a fully compatible API with typesafe/config. That would allow you to reuse existing converters like this https://github.com/circe/circe-config/blob/d65ae7dc4f69003e7cd66a31fe41936e5157ce5b/src/main/scala/io.circe.config/parser.scala#L51 this would be most useful for scalafmt, since scalafmt doesn't use circe but a similar converter

evacchi:
chiming in just to say that SHocon lets you access the internal AST representation directly. Would this help?

https://github.com/unicredit/shocon/blob/master/shared/src/main/scala/eu/unicredit/shocon/SHocon.scala#L115-L118

e.g.:

```scala
import shocon
val result = Config.parse("{ a: 1 }") // wrapped Success/Failure
val parsed = Config("{ a: 1 }") // unwrapped
parsed("a").as[Int] // returns Some(1)
```

olafurpg:
@laughedelic scalafmt uses a fairly ancient version of https://github.com/olafurpg/metaconfig which has a different parser for jvm (typesafe/config) and js (fork of shocon). I suspect upgrading to the latest metaconfig would require a lot of work since the `@DeriveConfDecoder` macro annotations are no longer supported
https://github.com/scalameta/scalafmt/blob/master/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala#L119

I'm personally not keen on maintaining two parallel versions of metaconfig (one for scalafix and another for scalafmt)

what applications do you have in mind for scalafmt on js?

gabro:
@laughedelic thanks for reporting that, as @olafurpg said those issues are essentially due to the discrepancies between the JS and the JVM version of metaconfig.

Since this is partially what I'll talk about in Krakow, I may spend some cycles upgrading metaconfig on scalafmt, unless @olafurpg is against it.

@laughedelic as the per the facade name, I believe you're right. I've always been confused on which term to use and indeed it looks facade is the wrong one. I'll give it more thoughts while polishing this up.

olafurpg:
> I may spend some cycles upgrading metaconfig on scalafmt

Metaconfig uses scala.meta.Input which may make it tricky to upgrade because of #975 I could probably tear out the scalameta dependency in metaconfig since `Input` is easy to reimplement

Even if we upgrade we still don't have a solution for handling disrepancy between jvm/js.

An alternative approach would be to support `.scalafmt.json` and users who want to use scalafmt on js would need to write configuration in json. One nice thing about json is that it has really nice editor support in vscode/intellij/atom(?) if a json schema is provided. It should be fairly simple to auto-generate a json schema using the tricks described in https://olafurpg.github.io/metaconfig/#docs

In the long term, I believe betting on json instead of hocon as the configuration format of choice for cross-platform tools (jvm/js/native) is the way to go.

olafurpg:
I'm a bit skeptical of dropping hocon support altogether, there should be at least some sliding window. Supporting both hocon and json in parallel and updating docs would require significantly less effort with the only annoyance that people who use scalafmt js plugins need to migrate to .scalafmt.json. It should be possible for the cli to automatically migrate .scalafmt.conf to .scalafmt.json. If that works out really well we can consider dropping .scalafmt.conf.

ShaneDelmore:
Json is already valid HOCON, are we just missing a js compatible way to desugar HOCON to Json?

laughedelic:
Thanks for your replies, guys! I really like this plan:

> An alternative approach would be to support .scalafmt.json and users who want to use scalafmt on js would need to write configuration in json.

➕ 
> It should be possible for the cli to automatically migrate .scalafmt.conf to .scalafmt.json. If that works out really well we can consider dropping .scalafmt.conf.

----

> what applications do you have in mind for scalafmt on js?

1. An Atom plugin like the one @gabro made for VSCode. I think every editor deserves a simple good formatting plugin. There are a couple Atom plugins based on scalariform and [one on scalafmt](https://github.com/SvenWi/atom-scala-code-formatter), but it's slow and just ignores `.scalafmt.conf`. A JS-based plugin would be a really nice substitution. 

2. I made this over the weekend: [scalafmt-probot](https://github.com/laughedelic/scalafmt-probot), it's a Github bot for checking code formatting with scalafmt. It's just proof-of-concept implementation in JS, not ready to be used for real yet. 
  
    The idea is to have a separate PR status check for code formatting instead of setting up scalafmt to run on Travis. It should be visually more convenient on Github (not just a generic CI failure) and (hopefully) faster (I haven't really compared anything, just guessing). So it looks something like this:

    * screen shot 2018-02-18 at 19 47 57   
        then
    * screen shot 2018-02-18 at 19 48 40   
        or
    * screen shot 2018-02-18 at 19 49 47

    The Github app is closed at the moment, I'm just testing it, but if anybody wants to try it out, let me know, I will make it public (and make another one for development).

3. Another idea is just an idea, I don't think I will have time to implement it any time soon. I think it would be nice to have a web interface where you can paste a piece of code, get it formatted with scalafmt and compare side by side before and after + tweak some config settings and see the difference it makes. It could be useful for choosing configuration that conforms to the users expectations and it could be useful for the scalafmt documentation.

olafurpg:
@laughedelic that is very cool, I would love to use scalafmt-probot in my projects!

Have you tried writing plain json inside .scalafmt.conf? @ShaneDelmore said all json is valid hocon. If that works then we already have a valid workaround today.

laughedelic:
> that is very cool, I would love to use scalafmt-probot in my projects!

I'm glad to hear that you find it useful 😊 


> Have you tried writing plain json inside .scalafmt.conf? @ShaneDelmore said all json is valid hocon. If that works then we already have a valid workaround today.

Yes, you can  JSON for the config indeed. But the problem stays the same, the parser here just cannot handle some fields. Probably it would work with 29ded6c477a2865c22e292caf1baf1c059a544bf?

olafurpg:
@laughedelic if your keys are nested with dots `"a.b.c"` then you need the fix in https://github.com/scalameta/scalafmt/commit/29ded6c477a2865c22e292caf1baf1c059a544bf If you write `{"a": {"b": "c" }}` then it should not be necessary

laughedelic:
Oh, I didn't know that I can use `"a.b.c"` in this JSON, so I rewrote all keys as nested. But still, with this in the config
```json
  "runner": {
    "optimizer": {
      "forceConfigStyleMinArgCount": 3
    }
  }
```

I get this (running a node REPL):

```
> const scalafmt = require('scalafmt')
> scalafmt.format('', false, fs.readFileSync('.scalafmt.conf').toString())
TypeError: Cannot read property 'ba' of null
    at MP.d.si (./node_modules/scalafmt/index.js:2513:40)
    at MP.d.Qb (./node_modules/scalafmt/index.js:2512:461)
    at Gm (./node_modules/scalafmt/index.js:143:481)
    at Yga (./node_modules/scalafmt/index.js:415:128)
    at Object.ca.format (./node_modules/scalafmt/index.js:4408:69)
```

olafurpg:
@laughedelic Hmm 😟 I must admit I never properly tested the JS config support in scalafmt. I'm afraid I won't have time to look into it in the near future.

Closing due to inactivity. Please reopen if you would like to pick this up again @gabro !

laughedelic:
@gabro let me know if you're going to pick up work on this or you need some help. I'm still really interested in pushing this forward, so I want to help, but I'll need some guidance to work on this.

olafurpg:
I believe supporting json in addition to HOCON would be the simplest and
most robust way to push this forward. This would avoid the need to
reimplement all of HOCON in Scala. Json schema would make it possible to
support auto completion, which is a nice plus.

Users who want to run scalafmt.js would need to use json configuration, we
could provide an automatic translation from the current config.

gabro:
@laughedelic I'm afraid I won't have much time to work on this.

The problem lies in the JS implementation of metaconfig, which is not on-par with the JVM parser (correct me if I'm wrong @olafurpg)

As far as the "facade" (bad name, I agree) is concerned, this PR is already ok.
If you want to push this forward, I suggest to focus on the config side.

Also, it may even be if we have a mostly-working implementation, with some well-documented bugs.

olafurpg:
FYI, I opened #1145 which is the first step towards cleaning up the config situation. At least scalafmt and scalafix are now on the same metaconfig version. We still need to fix the inconsistency between JVM/JS, but this is progress.

PR #1005, opened 2017-07-27
Relative config path option

vovapolu:
This PR adds new setting to configure relative path to scalafmt config.

PR #1007, opened 2017-07-27
Use the VCS root as a potential config path in the IntelliJ plugin

stuhood:
Currently if the IntelliJ "project directory" is not located under the root of the repository for a project, two issues occur:

1. Because the plugin was determining the project for a file by comparing `project.getBasePath` to the file path, no project would be detected, and so there was nowhere to look for a scalafmt.conf other than the HOME directory.
1. Because the Project was not located under the root of the repository, even after successfully finding a project, we would not find a relevant config in the project directory.

This change fixes the first issue by using the ProjectFileIndex (as recommended [here](http://www.jetbrains.org/intellij/sdk/docs/reference_guide/project_model/project.html#checking-if-a-file-belongs-to-a-project)), and the second issue by using the Project to locate a configured VCS and look for a config there as well.

### DIFF
--- a/scalafmt-intellij/readme.md
+++ b/scalafmt-intellij/readme.md
@@ -13,7 +13,8 @@
 1. Select `Import module`;
 1. Be sure the `unmanaged-jars` dependency in the `intellij` module is set to `provided`
    inside `Project structure`/`Project settings`/`Modules`/`Dependencies` (btw, setting `provided` inside sbt file gives error);
-   Without this setting, the generated zip file will be >100mb instead of ~20mb.
+   Without this setting you'll get a classpath collision at runtime (and the generated zip file
+   will be >100mb instead of ~20mb).
 1. Right click on top of `intellij-scalafmt` plugin module and select `Prepare Plugin Module 'intellij-scalafmt' for deployment`;
 
 If everything went smoothly, you should have a `intellij-scalafmt/intellij-scalafmt.zip` file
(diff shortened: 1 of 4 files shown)


olafurpg:
Thank you for this contribution! I'm currently on vacation, will review once I'm back August 6th :)

scalafmt-intellij/src/main/scala/org/scalafmt/intellij/IdeaUtils.scala
```diff
@@ -26,28 +28,34 @@ object IdeaUtils {
         Utility.escape(msg),
         notificationType))
 
-  private def getConfigFileInPath(path: String) =
+  private def getConfigFileInPath(path: String) = {
+    emitMigrateConfigWarning(new File(path, ".scalafmt"))
```
olafurpg: This warning now caught an old `$HOME/.scalafmt` lying on my laptop :)

scalafmt-intellij/readme.md
```diff
@@ -13,7 +13,8 @@
 1. Select `Import module`;
 1. Be sure the `unmanaged-jars` dependency in the `intellij` module is set to `provided`
    inside `Project structure`/`Project settings`/`Modules`/`Dependencies` (btw, setting `provided` inside sbt file gives error);
-   Without this setting, the generated zip file will be >100mb instead of ~20mb.
+   Without this setting you'll get a classpath collision at runtime (and the generated zip file
```
olafurpg: 👍

olafurpg:
I just published a version of the plugin on current master to the "nightly" channel of the plugin https://plugins.jetbrains.com/plugin/8236-scalafmt can you please double-check everything is OK before we publish it to the stable branch? there were quite large conflicts with the changes in #1005 and I hope I didn't mess up the new VCS setup.

stuhood:
@olafurpg : Thanks a lot! Looks good from my end. Downloaded and installed `1.2.0` nightly and had success with a {{.scalafmt.conf}} at the git root.

PR #1014, opened 2017-08-11
Fail fast on invalid .scalafmt.conf when running with no args.

olafurpg:
`./scalafmt` would fall back to the default style if .scalafmt.conf
had a typo. Now it fails fast.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,8 @@ version: '{build}'
 os: Windows Server 2012
 install:
   - ps: iex (new-object net.webclient).downloadstring('https://get.scoop.sh')
-  - ps: scoop install sbt
+  - ps: scoop bucket add versions
+  - ps: scoop install sbt0.13
 environment:
   CI_SCALA_VERSION: 2.12.2
 build_script:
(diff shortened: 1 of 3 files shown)


olafurpg:
cc/ @muxanick scoop is failing to download sbt

```
Downloading https://cocl.us/sbt100zip#/dl.zip (-1 B)...
Checking hash of sbt100zip... Hash check failed for 'https://cocl.us/sbt100zip#/dl.zip'.
Expected:
    fb72e5152c8e95972187bbbb0ac3468ac4b48edcead4ccb09295415b5cea3a15
Actual:
    d38f9ea4f9c79d88ab4412dd70574b43d3ff7de318ced3083b99335cd0a8f62b
```

seems it's trying to download sbt 1.0. Any idea how to pin the version to 0.13.16?

muxanick:
@olafurpg I don't know, it breaks my CI as well, seems like scoop doesn't support different versions or I don't know how to switch them.
I will update here if I found something.

olafurpg:
Thanks @muxanick please keep up updated! :)

muxanick:
@olafurpg I contacted person who made the change and he made another change and added 0.13 version as well, I believe now we can do this:
```
scoop bucket add versions
scoop install sbt0.13
```

olafurpg:
I now got a new error

> The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: Cloning into 'C:\Users\appveyor\scoop\buckets\versions'...

Any ideas?

olafurpg:
Merging despite appveyor failure to unblock 1.2.0, hopefully this will resolve soon without disabling appveyor

PR #1019, opened 2017-08-12
Add new setting `includeFixesFrom.v1_2_0 = false`.

olafurpg:
This setting allows users to upgrade integrations such as the IntelliJ
plugin without having their codebase formatted against the latest
scalafmt fixes. Large codebases may prefer to wait upgrading scalafmt to
bundle multiple version upgrades in a single PR.

Any thoughts @stuhood @xeno-by ?

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/IncludeFixesFrom.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/IncludeFixesFrom.scala
@@ -1,0 +1,12 @@
+package org.scalafmt.config
+
+import metaconfig._
+
+@DeriveConfDecoder
+case class IncludeFixesFrom(
+    v1_2_0: Boolean = true
+)
+
+object IncludeFixesFrom {
+  val default = IncludeFixesFrom()
+}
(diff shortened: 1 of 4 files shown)


stuhood:
We're currently  locked on a manually deployed older version of the plugin, but this might make sense for future versions.

Another option would be to decouple the Intellij plugin version from the scalafmt version (don't know what that would look like though... reflection?

olafurpg:
> Another option would be to decouple the Intellij plugin version from the scalafmt version (don't know what that would look like though... reflection?

I've thought about this, it should be possible to download a custom scalafmt version with couriser and classload. I'm not sure how that plays with Jetbrains's plugin terms, however.

I find it beneficial if we can keep everyone on the latest scalafmt version. It simplifies bug reports and other things. That way, formatting unrelated improvements (like performance) can be quickly adopted by everyone. Most of the time it's possible to provide options to opt-out of changes in the core formatting engine, like in this PR. I'm gonna keep this PR open for a while longer to see if there is interest for this.

> We're currently locked on a manually deployed older version of the plugin, but this might make sense for future versions.

Cool, I didn't know that. I've considered moving the intellij plugin into a separate repo, do you think that would make things easier/harder to cross-build the plugin against different scalafmt versions?

olafurpg:
Gonna close this as "not needed". I'm happy to reconsider if someone would find this useful to stay on the latest versions.

PR #1025, opened 2017-08-17
Return non-zero exit code in ./scalafmt if no files match filter

### LINKED ISSUE #876: Return non-zero exit code in ./scalafmt if no files match filter
Recently, we moved files around in scalameta which cause the defined `project.includeFilters` to match no files and our `./scalafmt --test` check in the CI was green because the exit code was zero even if no files matched the `project.` configuration.

A good place to start would be here https://github.com/scalameta/scalafmt/blob/908ca6178d7e14c7f12bf29ffe54cb7004c958f4/scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala#L196

The cli then needs to propagate the error to the `main` function, which might require a bit of refactoring around https://github.com/scalameta/scalafmt/blob/908ca6178d7e14c7f12bf29ffe54cb7004c958f4/scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala#L58

yln:
Intends to fix #876

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
@@ -9,6 +9,7 @@ import java.io.PrintStream
 import java.util.concurrent.atomic.AtomicInteger
 
 import com.martiansoftware.nailgun.NGContext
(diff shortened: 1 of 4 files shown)


olafurpg:
LGTM 👍 Thank you for this fix @yln !

Issue #962, opened 2017-06-02
Make real git tests


pjrt:
Currently `GitOps` is mocked away in the tests, however this shouldn't be the case. Since we are supporting `git` commands to fetch files, we should properly test our integration with git. 

Unsurprisingly (since git is **_the stupid content tracker_**), we can create git repos inside of other repos. So we could make a testing framework where we create a temporary directory (similar to how we do it now) and call `git init` in it. Then we can have the same tests file-system tests we currently have except for git.

Issue #1010, opened 2017-07-31
New files not found when running `scalafmt` with `git = true`


pjrt:
This template is a guideline, not a strict requirement.
- **Version**: 1.1.0
- **Integration**: CLI
- **Configuration**:
```
project.git = true
```

## Steps

```bash
# inside a git repo
cat > Test.scala
  class Test { def     f =    1 }
git add Test.scala
scalafmt
```

## Problem

Test.scala was not picked up (it stayed unformatted)

## Expectation

It should be formatted

## Workaround

If you run `scalafmt --diff-branch HEAD` it will pick it up (because that command uses a different file lookup strategy).

## Notes

More than likely this is related to https://github.com/scalameta/scalafmt/issues/1000, and both are probably caused by my changes to the file lookup strategies on v1.0.

pjrt:
Fixed with afc446

ctoomey:
Thanks for fixing this, I just ran into it for a renamed file.

Issue #1000, opened 2017-07-20
java.io.FileNotFoundException formatting deleted git file


MasseGuillaume:
This template is a guideline, not a strict requirement.
- **Version**: 1.1.0
- **Integration**: bootstraped
- **Configuration**:
```
project.git = true
```

## Steps

I ran scalafmt like this:

```bash
add file to git
commit
delete file from git
run scalafmt
```

## Problem

```
 bin/scalafmt
Exception in thread "main" java.io.FileNotFoundException: /home/gui/scastie/client/src/test/scala/com.olegych.scastie.client/NameChangerSpec.scala (No such file or directory)
        at java.io.FileInputStream.open0(Native Method)
        at java.io.FileInputStream.open(FileInputStream.java:195)
        at java.io.FileInputStream.<init>(FileInputStream.java:138)
        at org.scalafmt.util.FileOps$.readFile(FileOps.scala:63)
        at org.scalafmt.util.FileOps$.readFile(FileOps.scala:51)
        at org.scalafmt.cli.InputMethod$FileContents.readInput(InputMethod.scala:40)
        at org.scalafmt.cli.Cli$.org$scalafmt$cli$Cli$$handleFile(Cli.scala:112)
        at org.scalafmt.cli.Cli$$anonfun$run$1.apply(Cli.scala:173)
        at org.scalafmt.cli.Cli$$anonfun$run$1.apply(Cli.scala:171)
        at scala.collection.Iterator$class.foreach(Iterator.scala:891)
        at scala.collection.AbstractIterator.foreach(Iterator.scala:1334)
        at scala.collection.parallel.ParIterableLike$Foreach.leaf(ParIterableLike.scala:972)
        at scala.collection.parallel.Task$$anonfun$tryLeaf$1.apply$mcV$sp(Tasks.scala:49)
        at scala.collection.parallel.Task$$anonfun$tryLeaf$1.apply(Tasks.scala:48)
        at scala.collection.parallel.Task$$anonfun$tryLeaf$1.apply(Tasks.scala:48)
        at scala.collection.parallel.Task$class.tryLeaf(Tasks.scala:51)
        at scala.collection.parallel.ParIterableLike$Foreach.tryLeaf(ParIterableLike.scala:969)
        at scala.collection.parallel.AdaptiveWorkStealingTasks$WrappedTask$class.internal(Tasks.scala:169)
[...]

PR #1028, opened 2017-08-17
Git testing framework and fix for some git-file-fetching issues

pjrt:
This PR fixes:

#962 by introducing a small testing framework for git and expand GitOpsTest
#1000 and #1010 by changing from `git ls-tree` to `git ls-files`

I remember reading somewhere that ls-tree is better, but I can't recall where or why. All I know is that the tests pass with `ls-files` but not `ls-trees`. 

Please point out any tests that I may have missed.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
@@ -16,42 +16,41 @@ object GitOps {
   def apply(): GitOps = new GitOpsImpl(AbsoluteFile.userDir)
 }
 
-class GitOpsImpl(workingDirectory: AbsoluteFile) extends GitOps {
+class GitOpsImpl(private[util] val workingDirectory: AbsoluteFile)
+    extends GitOps {
 
-  def exec(cmd: Seq[String]): String = {
-    val lastError = new StringBuilder
-    val swallowStderr = ProcessLogger(_ => Unit, err => lastError.append(err))
-    try {
-      sys.process
-        .Process(cmd, workingDirectory.jfile)
-        .!!(swallowStderr)
-        .trim
-    } catch {
-      case NonFatal(e) =>
-        throw new IllegalStateException(
-          s"Failed to run command ${cmd.mkString(" ")}. " +
-            s"Error: ${lastError.toString()}",
-          e)
+  private[util] def exec(cmd: Seq[String]): Try[Seq[String]] = {
+    val gitRes: Try[String] = Try {
+      val lastError = new StringBuilder
+      val swallowStderr = ProcessLogger(_ => Unit, err => lastError.append(err))
+      try {
+        sys.process
+          .Process(cmd, workingDirectory.jfile)
+          .!!(swallowStderr)
+          .trim
+      } catch {
+        case NonFatal(e) =>
+          throw new IllegalStateException(
+            s"Failed to run command ${cmd.mkString(" ")}. " +
+              s"Error: ${lastError.toString()}",
+            e)
+      }
     }
+
+    gitRes.map(_.lines.toSeq)
   }
 
   override def lsTree(dir: AbsoluteFile): Seq[AbsoluteFile] =
     rootDir.fold(Seq.empty[AbsoluteFile]) { rtDir =>
-      Try {
-        exec(
-          Seq(
-            "git",
-            "ls-tree",
-            "-r",
-            "HEAD",
-            "--name-only",
-            "--full-name",
-            dir.path
-          )
-          // DESNOTE(2017-06-02, pjrt): Since the git command above only returns
-          // files, no need to check if they are files
-        ).split(OsSpecific.lineSeparator).toSeq.map(f => rtDir / f)
-      }.getOrElse(Nil)
+      exec(
+        Seq(
+          "git",
+          "ls-files",
+          dir.path
+        )
+        // DESNOTE(2017-06-02, pjrt): Since the git command above only returns
+        // files, no need to check if they are files
+      ).toOption.toSeq.flatten.map(f => rtDir / f)
     }
 
   override def rootDir: Option[AbsoluteFile] = {
@@ -61,7 +60,7 @@ class GitOpsImpl(workingDirectory: AbsoluteFile) extends GitOps {
       "--show-toplevel"
     )
     for {
-      rootPath <- Try(exec(cmd)).toOption
+      Seq(rootPath) <- exec(cmd).toOption
       file <- AbsoluteFile.fromPath(rootPath)
       if file.jfile.isDirectory
     } yield file
@@ -79,7 +78,7 @@ class GitOpsImpl(workingDirectory: AbsoluteFile) extends GitOps {
     // files, no need to check if they are files
     for {
       root <- rootDir.toSeq
-      path <- Try(exec(cmd)).toOption.toSeq.flatMap(_.lines)
+      path <- exec(cmd).toOption.toSeq.flatten
     } yield AbsoluteFile.fromFile(new File(path), root)
   }
 }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
@@ -48,8 +48,6 @@ class GitOpsImpl(private[util] val workingDirectory: AbsoluteFile)
           "ls-files",
           dir.path
         )
-        // DESNOTE(2017-06-02, pjrt): Since the git command above only returns
-        // files, no need to check if they are files
       ).toOption.toSeq.flatten.map(f => rtDir / f)
     }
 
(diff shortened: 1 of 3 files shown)


scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
```diff
@@ -1,12 +1,209 @@
...
+    val f = Files.createTempDirectory(dirName)
+    val absFile = AbsoluteFile.fromPath(f.toString).get
+    val ops = new GitOpsImpl(absFile)
+    init(ops)
+    // initial commit is needed
+    val initF = touch("initialfile")(ops)
+    add(initF)(ops)
+    commit(ops)
+    val t = try test.toNoArgTest(ops)
+    finally f.toFile.delete
```
olafurpg: I think `java.io.File.deleteOnExit()` can do this for you.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
```diff
@@ -16,42 +16,39 @@ object GitOps {
...
-        throw new IllegalStateException(
-          s"Failed to run command ${cmd.mkString(" ")}. " +
-            s"Error: ${lastError.toString()}",
-          e)
+  private[util] def exec(cmd: Seq[String]): Try[Seq[String]] = {
+    val gitRes: Try[String] = Try {
+      val lastError = new StringBuilder
+      val swallowStderr = ProcessLogger(_ => Unit, err => lastError.append(err))
+      try {
+        sys.process
```
olafurpg: FYI, I found the java process builder API to be more powerful than sys.process. I think sys.process may even get pulled out into separate module in 2.13. No need to change here, just a fun fact.

scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
```diff
@@ -1,12 +1,209 @@
...
+  import GitOpsTest._
+
+  val root = AbsoluteFile.userDir
+  val dirName = "gitTestDir"
+
+  override type FixtureParam = GitOpsImpl
+
+  // DESNOTE(2017-08-16, pjrt): Create a temporary git directory for each
+  // test.
+  override def withFixture(test: OneArgTest) = {
```
olafurpg: Oh, one last thing. For native support, we may soon replace ScalaTest with thin FunSuite wrappers around utest. See https://github.com/scalameta/scalameta/issues/772 

It might be worth avoiding scalatest.fixture and Matchers to minimize the usage of scalatest features. We can deal with it once we port the tests to native, however.

scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
pjrt: Wouldn't that delete it when the jvm exits? Which is fine but not the same as what's happening here (where I delete it after the test finishes). 

It might make the suite run faster I guess (since we aren't deleting while running).

scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
pjrt: That would be a nice refactor. That and also the whole "fetching of files". At the moment it is a bit ad-hock, but while I was doing this I realized that GitOpsImpl should really just be an implementation of a higher trait called "FileFetcher" which has GitOps and FileSearch as implementations (names pending).

pjrt: Wait, can we use Java's process builder? Won't that interfere with the ScalaNative push?

pjrt: Created https://github.com/scalameta/scalafmt/issues/1029 to track this

olafurpg:
It seems the new GitOpsTests fail on Max OSX

```
[info] - #1010: lsTree should return staged files *** FAILED *** (193 milliseconds)
[info]   List(/private/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir3040052682994696437/uE3ARBPosN2580305278991432640.ext) did not contain only (/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir3040052682994696437/uE3ARBPosN2580305278991432640.ext) (GitOpsTest.scala:66)
[info] - lsTree should return committed files *** FAILED *** (63 milliseconds)
[info]   List(/private/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir2184956693128142828/k0tAxdZKpZ3476763498564330157.ext) did not contain only (/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir2184956693128142828/k0tAxdZKpZ3476763498564330157.ext) (GitOpsTest.scala:73)
[info] - lsTree should not return commited files that have been deleted (71 milliseconds)
[info] - lsTree should return commited files that have been modified *** FAILED *** (102 milliseconds)
[info]   List(/private/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir6771190511204892308/Oc5KD6roMf3009463898568173025.ext) did not contain only (/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir6771190511204892308/Oc5KD6roMf3009463898568173025.ext) (GitOpsTest.scala:91)
[info] - diff should return modified commited files *** FAILED *** (74 milliseconds)
[info]   List(/private/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir7720017357554995043/4tAzOKj3O42792993858461415370.ext) did not contain only (/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir7720017357554995043/4tAzOKj3O42792993858461415370.ext) (GitOpsTest.scala:103)
[info] - #1000: diff should not return git deleted files (76 milliseconds)
[info] - #1000: diff should not return fs deleted files (72 milliseconds)
[info] - diff should return added files against HEAD *** FAILED *** (70 milliseconds)
[info]   List(/private/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir2241589133121633740/vM910kFi50916460595547242832.ext, /private/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir2241589133121633740/zNCVg9LfYd4492443097765923132.ext) did not contain only (/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir2241589133121633740/vM910kFi50916460595547242832.ext, /var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir2241589133121633740/zNCVg9LfYd4492443097765923132.ext) (GitOpsTest.scala:127)
[info] - diff should return added files against a different branch *** FAILED *** (113 milliseconds)
[info]   List(/private/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir4476308041325973684/QQMbhYTPLi6792830455390268517.ext, /private/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir4476308041325973684/zrsGKc4Osf216756784255788435.ext) did not contain only (/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir4476308041325973684/zrsGKc4Osf216756784255788435.ext, /var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir4476308041325973684/QQMbhYTPLi6792830455390268517.ext) (GitOpsTest.scala:141)
[info] - diff should return added files that are then modified against a different branch *** FAILED *** (91 milliseconds)
[info]   List(/private/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir2746612636398952792/SWkUDfYSn52056108708096769335.ext, /private/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir2746612636398952792/SuXJcr7zay3336737670910157879.ext) did not contain only (/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir2746612636398952792/SuXJcr7zay3336737670910157879.ext, /var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir2746612636398952792/SWkUDfYSn52056108708096769335.ext) (GitOpsTest.scala:156)
[info] - diff should not return removed files against a different branch *** FAILED *** (107 milliseconds)
[info]   List(/private/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir3927730782255776623/60LYP9Svqe3389305970189046741.ext) did not contain only (/var/folders/30/6jlz_xfs46ndvn212mt_wj6m0000gn/T/gitTestDir3927730782255776623/60LYP9Svqe3389305970189046741.ext) (GitOpsTest.scala:171)
```

I suspect it's related to case sensitivity, I'm investigating.

pjrt:
Looking at the result, it seems that the location where the files are being
created is different. One says `/private/var` while the other says `/var`.

Maybe the issue has to do with how Java creates temporary files and
directories in OSX.

PR #1030, opened 2017-08-20
Fix new GitOpsTest on OSX

olafurpg:
Fixes regression from #1028

### DIFF
--- a/scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
+++ b/scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
@@ -37,7 +37,7 @@ class GitOpsTest extends fixture.FunSuite {
 
   def touch(name: String = Random.alphanumeric.take(10).mkString)(
       implicit ops: GitOpsImpl): AbsoluteFile = {
-    val f = File.createTempFile(name, ".ext", ops.workingDirectory.jfile)
+    val f = File.createTempFile(name, ".ext", ops.rootDir.get.jfile)
     AbsoluteFile.fromPath(f.toString).get
   }
 


olafurpg:
I made several attempts to enable OSX on Travis but kept hitting on the same error

> The command "jdk_switcher use default" failed and exited with 127 during .

Logs: https://travis-ci.org/scalameta/scalafmt/jobs/266499443

I tried copying https://github.com/coursier/coursier/blob/10c65c60627b75053045a6718d02c466189517a6/.travis.yml with no luck.

I'm OK for now to not enable OSX on CI, although it could help catch these issues in the future during review.

scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
```diff
@@ -37,7 +37,7 @@ class GitOpsTest extends fixture.FunSuite {
 
   def touch(name: String = Random.alphanumeric.take(10).mkString)(
       implicit ops: GitOpsImpl): AbsoluteFile = {
-    val f = File.createTempFile(name, ".ext", ops.workingDirectory.jfile)
+    val f = File.createTempFile(name, ".ext", ops.rootDir.get.jfile)
```
pjrt: Yeah this was my original plan, but I wanted to avoid dirtying the dev's working dir if anything went wrong. 

But if something goes wrong it isn't the end of the world.

olafurpg: Good point, may be annoying to debug if your workspace suddenly becomes filled with random new files! :D

pjrt: If it gets too bad we can just make the `.ext` into something like `.ignoreme` and add that to the `.gitignore`. More than likely this won't be an issue though.

Issue #722, opened 2017-02-08
[Feature Request] Auto-sort imports (statements)


pjrt:
We currently have a `SortImports` rewrite, but all it does is sort imports in import groups, it does not sort import statements themselves. 

This:
```scala
import com.comp.A
import com.comp.module.B
import scala.collection.mutable
import scala.util
import scalaz.Functor
import java.time.Instant
```

Would be formatted as:
```scala
import java.time.Instant

import scala.collection.mutable
import scala.util

import scalaz.Functor

import com.comp.A
import com.comp.module.B
```

The order is:
* Java stuff
* scala stuff
* 3rd party stuff
* company stuff
   * This would need some setting specifying what the company domain is: `companyDomain = com.comp`

3rd party stuff can be determined by whether it matches anything else. If it isn't Java, Scala or companyDomain, then it is 3rd party. 

I think we can just expand `SortImports` to handle this. We could check if the `companyDomain` setting is set. If so, expand the sorting. Otherwise do what it currently does.

diego-alvarez-hs:
could we also choose the order of import sections?
As in my company we use a different order:

Something similar to intellij:
![image](https://cloud.githubusercontent.com/assets/12563308/22759938/22547cfa-ee09-11e6-830b-1240501ed516.png)

pjrt:
I've never seen that style. Is that the default style in IntelliJ?

diego-alvarez-hs:
I have no idea, my main editor is Emacs, I don't think that's the default though

olafurpg:
Sorting imports requires semantic information to be 100% accurate, since
Scala allows relative imports. If we were to add this rewrite, I suggest to
include it in the SortImports rewrite with an assumeNoRelativeImports
option enabled.

I recently implemented an organize imports rewrite in scalafix. It removes
unused imports. Handles relative imports. Grouping and ordering is
configurable. I think we can reuse almost everything from that except the
semantic features.


On Wed, 8 Feb 2017 at 23:09, Pedro J Rodriguez Tavarez <
notifications@github.com> wrote:

We currently have a SortImports rewrite, but all it does is sort imports in
import groups, it does not sort import statements themselves.

This:

import com.comp.Aimport com.comp.module.Bimport
scala.collection.mutableimport scala.utilimport scalaz.Functorimport
java.time.Instant

Would be formatted as:

import java.time.Instant
import scala.collection.mutableimport scala.util
import scalaz.Functor
[...]

PR #1038, opened 2017-09-09
#722: Implement sorting of import statements

pjrt:
This PR implements #722 

* Adds a new key: `rewrite.importGroups` which takes a list of strings

```
rewrite.rules = [SortImports]
rewrite.importGroups = ["java,scala", "*", "com.example"]
```

The following will sort import statements into three groups: the first will be java AND scala, the third will be imports that start with `com.example`, and the second will be everything else, like so:

```scala
import java.io
import java.net._
import scala.util

import org.scalafmt._

import com.example.foo
```

Note that this **list of strings is NOT a list of regex**, but plain strings. `*` is just a special placeholder that means "everything else". Commas allow multiple kinds of imports to hold the same group. 

If `*` isn't found in the list, then any imports that don't match will be placed on a new group at the end (as if `*` was placed at the end of the list).

If the list is empty or not defined, then all imports are sorted into one group (ie: they are sorted and newlines are removed between them).

* Works for `AsciiSortImports` as well

### Major changes that people need to be aware of:

* This will sort ALL imports in the file
  * Sorting everything was easier than sorting somethings since what the definition of "top" in scala is confusing (since a piece of Scala code can start in a lot of different ways). If this is really undesirable, we can discuss a general style to define "top". 
* This PR also ads a change where spaces are now added before and after all import groups. This is a **default behavior** and is **NOT dependent on any flag being on**. 
  * I feel this behavior should be on by default, but again, we can discuss this.


scalafmt-tests/src/test/resources/rewrite/SortImports.stat
```diff
@@ -35,8 +44,8 @@ import a.{
   bar
 }
 >>>
-import a.{zzzz => _, bar}
-<<< caveat: comments
+import a.{bar, zzzz => _}
+<<< SKIP caveat: comments
```
pjrt: Had to skip this test. For some reason the comments was being removed, but I'm not sure why. 

Though the test is confusing since the comment stays in the same place but is now pointing to the wrong importee.

I don't see the point of having comments in imports but if this is really needed I can dig deeper into what is happening.

pjrt: Oh, I know why the comment is being removed: because I remove all tokens but `Importer.importees` does NOT return comments, so when we do the rewrite, the comment is gone. 

Unn, not sure how this was working before.

scalafmt-tests/src/test/resources/rewrite/SortImports.stat
```diff
@@ -35,8 +44,8 @@ import a.{
   bar
 }
 >>>
-import a.{zzzz => _, bar}
-<<< caveat: comments
+import a.{bar, zzzz => _}
```
pjrt: @olafurpg I "fixed" this. I say "fixed" because there was a comment about "being sure this works". I don't see why it wouldn't so I just changed it.

pjrt:
CI error:
```
method sorted(scala.collection.Seq)scala.collection.Seq in object org.scalafmt.rewrite.AsciiSortImports does not have a correspondent in current version
```

Not sure how to debug that. I prefer not to just add it to the filter....

sjrd:
How does this deal with relative imports?

pjrt:
@sjrd It doesn't take that into account, it just sorts them as if they were full.

For relative imports we would need semantic info, or some sort of style (like, if it starts with a capital letter, then it is relative). 

Now, relative imports don't usually match patterns (for patterns, I expect things like `java` and `scala` and `com.example`). This means that relative imports will probably always fall under the `*`. If there is a group of them, they will be kept in a group (no newline between them) and sorted.

olafurpg:
After some consideration, I am against adding this feature. Currently, running scalafmt is always safe, if scalafmt turns valid code into invalid then it's a bug. With this change to SortImports that will no longer be true. The person running scalafmt must know that there are no relative imports in the codebase. For private projects with a single developer, this may be a fair assumption. For large projects with many contributors, this option/assumption is dangerous and can lead to "scalafmt broke my code" discussions.

We should resist the temptation for convenient solutions that compromise correctness.

On the bright side, scalafix v0.5 is just around the corner and with the semantic API we can handle relative imports (e.g., expand them to fully qualified imports). The fixture to run scalafix is definitely more complex/slower than for scalafmt since it requires a compilation step, but we can also release with confidence that our rewrites work as expected.

Regarding the changes to blank lines around import groups, I'm happy to add it behind a flag. To make it default I would like to go through a similar democratic process as we did with vertical alignment on case arrows (even if I disagreed with the outcome in that case).

pjrt:
@olafurpg Fair enough, when I was making this, relative imports was a point of contention. I don't often use relative imports outside of the scope where they are used (which means this feature would do nothing to them) but I have seem people use them all over the place. 

So yeah, I agree, however idk if Scalafix is a "fix" (ha!) to this since Scalafix is meant to be a tool to fix scalacode, not really reformatting. Is there a plan to allow Scalafmt to do some of the semantic rewrites that Scalafix can do? (ie: give Scalafmt semantic info).

diego-alvarez-hs:
😿 this feature as it's right now would be really usable for me, we don't use relative imports at my company.  Actually, I'd prefer to break the developer's code when they try to use relative imports 😆 

I'd love to have this feature under some experimental configuration. `unsafeRules` or something.

Regarding blank lines around imports, would be great if that can be passed as part of the configuration, so we won't need to pass multiples configurations per item on the list. i.e.:
```scala
rewrite.importGroups = ["java", "scala", "BLANK", "*", "com.example"]

import java.io
import java.net._
import scala.util

import org.scalafmt._
import com.example.foo
```

PR #1044, opened 2017-09-15
PoC remove starting new line on ScalaDocs multiline commens

d1egoaz:
Given that we officially don't have one scala style for multiline comments but 3! https://contributors.scala-lang.org/t/creation-of-a-standard-scala-style-guide/933

I'd like to add a new feature to scalafmt, so at least under a new non default flag, all multiline comments will look the same, using the old  ScalaDoc comments style.

This style is for poor souls like me @eddsteel and @olafurpg that followed the old style, as @olafurpg mentioned it here https://github.com/scala/docs.scala-lang/pull/801#pullrequestreview-44702920

@eddsteel started a discussion on https://contributors.scala-lang.org/t/quiet-changes-to-the-style-guide/923 but I'm not sure how that would end 🤷‍♂️ 

This PoC will match the old published version of the Scala doc style, 
```scala
The general format for a ScalaDoc comment should be as follows:

/** This is a brief description of what's being documented.
  *
  * This is further documentation of what we're documenting.  It should
  * provide more details as to how this works and what it does. 
  */
def myMethod = {}
```

Currently scalafmt doesn't modify these comments:
```scala
/** test
  */
```
or
```scala
/**
  * test
  */
```
I'd like at least to make them consisten to one style.

My current implementation is just a PoC, it doesn't have the flag, it has some corner case where an empty multicomment 
```scala
  /**
    */
```
gets modified to:
```scala
  /**/ // <- bug
```

I just wanted to receive some feedback and check that maintainers are ok with this idea

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -73,9 +73,18 @@ class FormatWriter(formatOps: FormatOps) {
         val spaces: String =
           if (isDocstring && initStyle.scalaDocs) " " * (indent + 2)
           else " " * (indent + 1)
-        leadingAsteriskSpace
+        val str = leadingAsteriskSpace
           .matcher(comment.syntax)
           .replaceAll(s"\n$spaces\\*")
+
+        if (isDocstring && initStyle.scalaDocs) {
+          str
+            .replaceFirst("/\\*\\*\\s+", "/** ") // move second line up if first line just contains /** and spaces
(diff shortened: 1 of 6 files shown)


olafurpg:
Hei! I apologize the terribly slow response, been at Scala World since last week friday.

Just to quickly answer, I am very excited about having an option to enforce proper formatting of docstrings. Ideally, I prefer to only support that automatic enforcement for one official style of docstrings. This setting should be able to port from any other docstring style.

What do you think?

olafurpg:
Closing due to low activity, please re-open if you are interested in continuing with this PR.

PR #1045, opened 2017-09-19
Disabling alignment helps minimizing git diffs

nightscape:


readme/Faq.scalatex
```diff
@@ -64,6 +64,17 @@
        @li
          code is read waaay more often outside of code reviews, for example
          when you are actually coding.
+       @li
+         it helps to disable most alignment options
+         (so that e.g. when you rename the variable with the longest name to become longer,
+         other lines are not impacted):
+         
+         @cliFlags
+            align.tokens = []
```
olafurpg: align = none?

nightscape: I didn't know that one :wink:

nightscape:
I'm open to where to put this. I thought here is a good place because that section already talks about git diffs.
We could put it under Configuration and then link there from the FAQ?

olafurpg:
Under the `align=none` section we can add a "Pro tip. Enable this setting to minimize git diffs/conflicts from renamings and other refactorings." We can also add a new section/question under FAQ "Which configuration options minimize diffs/conflicts?"

PR #1052, opened 2017-09-22
Add setting newlines.beforeCaseClassParameterDefn

broeng:
When verticalMultilineAtDefinitionSite is specified,
this allows for forcing each parameter definition of
case classes on it's own line, and not attempting to
fit it on a single line.

This mimics the IntelliJ code style for method dec-
laration parameters "New line after '('".

### DIFF
--- a/readme/Configuration.scalatex
+++ b/readme/Configuration.scalatex
@@ -381,6 +381,26 @@
       implicit def validatedInstances[E](implicit E: Semigroup[E]): Traverse[
           Validated[E, ?]] with ApplicativeError[Validated[E, ?], E] = 2
 
+  @sect{newlines.beforeCaseClassParameterDefn}
+    Together with @b{verticalMultilineAtDefinitionSite}, this will
+    align parameter definitions vertically, with a line for each, and
+    not attempt to fit the case class definition on a single line.
+
+    Default @b(default.newlines.beforeCaseClassParameterDefn)
+
+    Requires @b{verticalMultilineAtDefinitionSite} set to @b(true).
+
+    @hl.scala
+      // verticalMultilineAtDefinitionSite = true
+      // newlines.beforeCaseClassParameterDefn = false
+      case class A(b: Int, c: Int)
+
+      // verticalMultilineAtDefinitionSite = true
+      // newlines.beforeCaseClassParameterDefn = true
+      case class A(
+          b: Int,
+          c: Int)
+
   @sect{spaces.afterKeywordBeforeParen}
     Default: @b(default.spaces.afterKeywordBeforeParen)
 
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -24,6 +24,7 @@ import org.scalafmt.internal.ExpiresOn.Left
 import org.scalafmt.internal.ExpiresOn.Right
 import org.scalafmt.internal.Length.Num
 import org.scalafmt.internal.Policy.NoPolicy
+import org.scalafmt.util.CaseClass
 import org.scalafmt.util.CtorModifier
 import org.scalafmt.util.StyleMap
 import org.scalafmt.util.TokenOps
@@ -823,13 +824,30 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
       if (isBracket) close // If we can fit the type params, make it so
       else lastParen // If we can fit all in one block, make it so
 
-    Seq(
+    // Determine if we should just allocate a line for each parameter
+    // definition for case classes, and not attempt to fit everything
+    // on a single line.
+    val newlineBefore =
+      !isBracket && style.newlines.beforeCaseClassParameterDefn && owners(
+        ft.left).is[CaseClass]
+
+    val fitInSingleLine =
       Split(NoSplit, 0)
-        .withPolicy(SingleLineBlock(singleLineExpire)),
-      Split(Newline, 1) // Otherwise split vertically
+        .withPolicy(SingleLineBlock(singleLineExpire))
+
+    val splitVertically =
+      Split(Newline, 1)
         .withIndent(firstIndent, close, Right)
         .withPolicy(policy)
-    )
+
+    if (newlineBefore) {
+      Seq(splitVertically)
+    } else {
+      Seq(
+        fitInSingleLine,
+        splitVertically // Otherwise split vertically
+      )
+    }
 
   }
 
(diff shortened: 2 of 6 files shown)


broeng:
Please let me know if you have any comments or requests :-)

readme/Configuration.scalatex
```diff
@@ -381,6 +381,26 @@
...
+    Together with @b{verticalMultilineAtDefinitionSite}, this will
+    align parameter definitions vertically, with a line for each, and
+    not attempt to fit the case class definition on a single line.
+
+    Default @b(default.newlines.beforeCaseClassParameterDefn)
+
+    Requires @b{verticalMultilineAtDefinitionSite} set to @b(true).
+
+    @hl.scala
+      // verticalMultilineAtDefinitionSite = true
```
olafurpg: Why is this required? Seems like this option should be usable without  verticalMultilineAtDefinitionSite=true.

readme/Configuration.scalatex
```diff
@@ -381,6 +381,26 @@
...
+      // newlines.beforeCaseClassParameterDefn = false
+      case class A(b: Int, c: Int)
```
olafurpg: Why only case classes and not regular classes?

broeng:
Thank you for the review, olafurpg!

And sorry - I should probably have included the motivation in my original PR. 

We have a fairly active code base and in order to reduce conflict friction, we try to reduce the scope of any changes as much as possible, to reduce any "noise" in irrelevant parts of the code. This means, that we opt out of alignment of match cases, extends, assignments - and we prefer our members to be declared on a line each, as we find it generally passes easier through automatic merging with git. 

We use **verticalMultilineAtDefinitionSite** for this, which in general works fine, but for declarations fitting a single line, it chooses to do that instead, which is not what we are asking for (and is inconsistent styling with this option), and goes against our efforts of improving automatic merging - though, granted, this is less of a problem with smaller class declarations.

readme/Configuration.scalatex
broeng: Actually, I think this should be the default behavior of **verticalMultilineAtDefinitionSite** :-)

It seems inconsistent to me, that **verticalMultilineAtDefinitionSite** does not do vertical alignment (as specifically asked), if the line is short enough. This is the default behavior now, and I don't expect to get that changed, which is why I added a new setting.

If I make my new setting distinct from  **verticalMultilineAtDefinitionSite**, wouldn't we just have to settings, basicly doing the same, only with different behaviors in the single-line case?

readme/Configuration.scalatex
broeng: Good question. I probably had our use case a bit too in mind. :-)

Would it be better with a multi-option setting?

broeng: Ping? :-)

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -823,13 +824,30 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
...
+    // Determine if we should just allocate a line for each parameter
+    // definition for case classes, and not attempt to fit everything
+    // on a single line.
+    val newlineBefore =
+      !isBracket && style.newlines.beforeCaseClassParameterDefn && owners(
+        ft.left).is[CaseClass]
+
+    val fitInSingleLine =
       Split(NoSplit, 0)
-        .withPolicy(SingleLineBlock(singleLineExpire)),
```
olafurpg: I think you can use `ignoreIf = ...` to conditionally skip `fitInSingleLine` if this setting is enabled.

readme/Configuration.scalatex
```diff
@@ -381,6 +381,26 @@
       implicit def validatedInstances[E](implicit E: Semigroup[E]): Traverse[
           Validated[E, ?]] with ApplicativeError[Validated[E, ?], E] = 2
 
+  @sect{newlines.beforeCaseClassParameterDefn}
```
olafurpg: what do you think about `newlines.alwaysForClassPrimaryConstructor`?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeClasses.scala
```diff
@@ -17,3 +16,14 @@ object CtorModifier {
   def unapply(tree: Tree): Boolean =
     tree.is[Mod.Private] || tree.is[Mod.Protected]
 }
+@classifier
+trait CaseClass
```
olafurpg: I find it's a bit ad-hoc to limit this to case classes, why not primary constructors for all `class`? We could apply it only to Ctor.Primary

readme/Configuration.scalatex
olafurpg: I apologize the slow response, this got totally lost in my inbox.

I am personally not a fan of this option but I can understand the motivation. I'm OK with adding this option, but I'll leave it to @pjrt to LGTM the changes since he is the author of `verticalMultilineAtDefinitionSite`.

readme/Configuration.scalatex
```diff
@@ -381,6 +381,26 @@
...
+    Together with @b{verticalMultilineAtDefinitionSite}, this will
+    align parameter definitions vertically, with a line for each, and
+    not attempt to fit the case class definition on a single line.
```
olafurpg: It might be worth to mention that this might be desirable for cleaner git merges. Can we also mention this option under [this section](http://scalameta.org/scalafmt/#Whichconfigurationoptionsminimizediffs/conflictsinversioncontrol?) since it might be of interest to other people.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
olafurpg: That should make the diff for this PR very small.

pjrt:
I don't understand the purpose of this. The purpose of verticalMultiline was to introduce a new style for vertical defns. However, the implicit of rule of scalafmt is that we should always strive to fit it one line if we can (and only break it when going over the maxColumn). Why would you want to always break stuff into multiple lines?

I don't agree we should be adding this option.

broeng:
@pjrt, fair enough -- I wasn't aware that was an implicit rule of scalafmt. Personally, I disagree that it's a convention that suits all projects. We find that splitting this in multiple lines helps us/git with auto-merging of conflicting regions, and we would prefer to get this assistance as much as possible.

@olafurpg, thank you very much for the reviews, and suggestions (I wasn't aware of the ifIgnore, that would indeed be preferrable - and yeah, really no reason to restrict to case classes).

It seems this won't be merged, so it's probably pointless to implement the changes suggested by @olafurpg - if that changes, feel free to ping me, I'll be happy to update the PR.

olafurpg:
@broeng FWIW, you can use "config style" to manually force a short class definition to break
```scala
class A(
  a: Int
)
```
It requires everyone on the team to agree to use this style, but gives you the flexibility to leave some definitions on a single line when you want it.

Otherwise, I'm sorry @broeng I agree with @pjrt that this goes a bit against how scalafmt works.

PR #1070, opened 2017-10-20
Bump dependencies versions

loskutov:
Older versions of Courser had coursier/coursier#655 which caused problems with builds. Also, Scalatest has brought a couple bug fixes and enhancements since 3.0.1.


olafurpg:
Thank you for this contribution! Seems we can bump up the Scala.js version too to fix the Travis errors. Otherwise LGTM 👍

loskutov:
I've also bumped up the versions of some other plugins. (However, I couldn't make it resolve JetBrains/sbt-idea-plugin as well as the last version of scalatex)

olafurpg:
LGTM 👍 thank you!

Issue #1081, opened 2017-11-25
revamp the sbt plugin


fommil:
The default `sbt-scalafmt` plugin is not fit for purpose: there is no configuration and it blindly formats everything (no concept of unmanaged sources, etc, etc, although this can all be worked around manually). These problems were addressed in `neo-sbt-scalafmt` but it is *really* slow.

Now that sbt 1.0 supports scala 2.12 there is no need for the classloader hacks in https://github.com/lucidsoftware/neo-sbt-scalafmt/blob/master/sbt-scalafmt/src/main/scala/com/lucidchart/sbt/scalafmt/ScalafmtCorePlugin.scala

The 1.0 plugin should depend on `scalafmt-core` directly, no coursier or CLI nonsense, and retain the following `Setting`s / `Task`s (which are scoped per project / config) from the neo plugin:

- `scalafmt`
- `scalafmtOnCompile`
- `scalafmtConfig`
- `scalafmt:test` but renamed to `scalafmtCheck` to avoid creating a config
- `sbt:scalafmt` but renamed to `scalafmtSbt` with a corresponding `scalafmtSbtCheck`. This should format `*.sbt` files (with the appropriate dialect) and the `project/*.scala` files.
- `scalafmtOnly` a new task, which could be easily automated from emacs/vim/etc which takes a single file and formats it. This is ideal for "format on save" hooks.

The first cut can skip all the caching and simply do a brute-force format.

sjrd:
If you plan to base the source detection on `unmanagedSources` (which I approve), how do you suggest dealing with various configurations? I see two options:
* Either `scalafmt` and the other tasks are scoped per config (so it's actually `compile:scalafmt`), which makes it easy to depend on `sources in (Compile, scalafmt)` itself depending on `unmanagedSourceDirectories in (Compile, scalafmt)`, but means I need to explicitly perform `test:scalafmt` to format my test suite,
* Or `scalafmt` and the other tasks are scoped in the `Zero` config, which means I only need `scalafmt` to format `main/` and `test/`, but makes it unclear where to fetch the sources from. In particular, how to deal with the `it:` configuration, or worse, custom configurations?

fommil:
Yup, first choice... `compile:scalafmt` etc. In my projects at work I create an alias `fmt` which formats all known configurations and I use format on compile so I never in reality need to run it.

olafurpg:
Very nice! I approve 👍 @leonardehrenfried may also be interested.

[...]

Issue #917, opened 2017-04-29
New line wrapping backend with paiges


olafurpg:
Background: http://www.scala-lang.org/gsoc/2017.html#new-line-wrapping-algorithm-for-scalafmt
There now exists a Scala implementation of the "Prettier printer" line wrapping algorithm: https://github.com/typelevel/paiges This same algorithms is for example used in https://github.com/prettier/prettier 

I'm opening this issue to track progress on using paiges as an alternative line wrapping backend to [BestFirstFirst](https://github.com/scalameta/scalafmt/blob/908ca6178d7e14c7f12bf29ffe54cb7004c958f4/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala) and [Router](https://github.com/scalameta/scalafmt/blob/908ca6178d7e14c7f12bf29ffe54cb7004c958f4/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala).

I took a quick stab at this a few weeks ago and got amazing results for deeply nested function calls. Example https://gist.github.com/olafurpg/44ee2d0a8c5b8330702cf00d23d0978a What's impressive with paiges is that 
- it took me 10 minutes to get that example output (compared to 2-3 months with BestFirstSearch and it's still unable to handle code with that amount of nesting).
- it's fast! That example gist formats in ~300ms

Here are some (totally solvable) challenges that come with this new approach

### Q: how do we handle comments?
A paiges-based formatter would do a top-down tree traversal. Some comments are not directly associated with a tree, for example

```scala
foo(
  a, // comment
  b
)

def foo() = { // comment 
  // ...
}
```
My guess: we use a similar trick as in https://github.com/scalameta/scalameta/blob/fac0de590bf0633567a12ce38de2cf40c5b1b547/scalameta/contrib/shared/src/test/scala/scala/meta/contrib/AssociatedCommentsTest.scala (this seems similar to what's explained [here](http://jlongster.com/prettier-1.0#Comments))

### Q: What do we do with vertical alignment?

Paiges has an `align` combinator but iiuc it won't handle things like

```scala
case a  => b
case aa => b
```
[...]

PR #1085, opened 2017-11-30
New scalafmt sbt plugin

vovapolu:
https://github.com/scalameta/scalafmt/issues/1081

I renamed old plugin to `sbt-cli-scalafmt`, because it just calls cli interface (and because I couldn't come up with new name for my plugin :D). I also added new scripted tests in addition to old ones. 

Plugin doesn't perform any caching, it just looks to `unmanagedSources` or sbt files and formats them. `*Check` tasks throw exceptions and print error messages when check is failed.


build.sbt
```diff
@@ -87,6 +87,17 @@ lazy val `scalafmt-sbt` = project
...
   .dependsOn(cli)
 
+lazy val `scalafmt-sbt` = project
+  .settings(
+    allSettings,
+    mimaPreviousArtifacts := Set.empty,
+    moduleName := "sbt-scalafmt",
+    isOnly(scala212),
+    sbtPlugin := true,
+    sbtVersion in Global := "1.0.0"
```
sjrd: Don't put settings `in Global` in the settings of a `project`. It is extremely confusing as that affects the whole build, not just the project. It makes no sense.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
-          PlatformTokenizerCache.megaCache.clear()
-          state
+    lazy val scalafmt = TaskKey[Unit]("scalafmt", "Format Scala sources")
+    lazy val scalafmtCheck = TaskKey[Boolean](
+      "scalafmt-check",
+      "Check that Scala sources is formatted properly")
+    lazy val scalafmtOnCompile =
+      SettingKey[Boolean]("scalafmt-on-compile", "Format source when compiling")
+    lazy val scalafmtConfig =
+      TaskKey[File]("scalafmt-config", "Scalafmtter config file")
```
sjrd: typo: Scalafmtter

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
+    lazy val scalafmtSbt = TaskKey[Unit]("scalafmt-sbt", "Format SBT sources")
```
sjrd: The `-` separated names is a relic of sbt 0.12, not to be used in sbt 1 anymore. This should be named `scalafmtSbt`, just like the `val`. In fact, you should probably use `taskKey[Unit]("Format sbt sources")` instead. Same for the other keys (e.g., `scalafmtCheck`).

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
+        (BuildPaths.projectStandard(proj.base) * GlobFilter("*.sbt")).get
+          .filterNot(_.isHidden)
+      rootSbt ++ projectSbt
+    }
+  )
+  private lazy val projectSources = thisProject.map(proj =>
+    (BuildPaths.projectStandard(proj.base) * GlobFilter("*.scala")).get)
+
+  lazy val scalafmtSettings: Seq[Def.Setting[_]] = Seq(
+    scalafmtOnCompile := false,
```
sjrd: This particular default value should be set in the `override def buildSettings` of the `AutoPlugin`, so that it is easy for people to just set it for all their projects with `scalafmtOnCompile in ThisBuild := true`.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
+      val projectSbt =
+        (BuildPaths.projectStandard(proj.base) * GlobFilter("*.sbt")).get
+          .filterNot(_.isHidden)
+      rootSbt ++ projectSbt
+    }
+  )
+  private lazy val projectSources = thisProject.map(proj =>
+    (BuildPaths.projectStandard(proj.base) * GlobFilter("*.scala")).get)
+
+  lazy val scalafmtSettings: Seq[Def.Setting[_]] = Seq(
```
sjrd: I recommend calling this `scalafmtConfigSettings`, to make clear that they are meant to go through an `inConfig(...)` to be meaningful.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
+    scalafmtOnCompile := false,
+    scalafmt := formatSources(
+      unmanagedSources.value,
```
sjrd: You should read `(unmanagedSources in scalafmt).value` instead, so that it is possible to override the set of files processed by scalafmt independently of the sources given to the compiler. In complex builds, this can be necessary.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
+      scalaConfig.value,
+      streams.value.log),
+    scalafmtSbt := {
+      formatSources(sbtSources.value, sbtConfig.value, streams.value.log)
+      formatSources(projectSources.value, scalaConfig.value, streams.value.log)
+    },
+    scalafmtCheck :=
+      checkSources(
+        unmanagedSources.value,
```
sjrd: Same: `(unmanagedSources in scalafmt).value` (not `in scalafmtCheck` in this case, because `scalafmtCheck` should always apply on the same set of sources as `scalafmt`).

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
+        scalaConfig.value,
+        streams.value.log),
+    scalafmtSbtCheck := {
+      checkSources(sbtSources.value, sbtConfig.value, streams.value.log)
+      checkSources(projectSources.value, scalaConfig.value, streams.value.log)
+    },
+    compile := Def.taskDyn {
```
sjrd: The very presence of this `taskDyn` will destroy `inspect tree myProject/compile`. Please don't do this.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
+      val defaultCompile = compile.taskValue
+      if (scalafmtOnCompile.value) {
+        scalafmt.value
```
sjrd: This unconditionally applies `scalafmt` even when `scalafmtOnCompile` is `false`.

build.sbt
vovapolu: old plugin has this setting too :)

sjrd: Then fix the old plugin too ;)

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
olafurpg: I think it's possible to do something like `unmanagedSources.in(scalafmt).or(unmanagedSources).value` so that it defaults to `unmanagedSources` if it's not defined when scoped under the `scalafmt` task.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
+    lazy val scalafmtSbt = TaskKey[Unit]("scalafmt-sbt", "Format SBT sources")
+    lazy val scalafmtSbtCheck = TaskKey[Boolean](
+      "scalafmtSbtCheck",
+      "Check that SBT sources is formatted properly")
+    lazy val scalafmtOnly =
+      InputKey[Unit]("scalafmt-only", "Format only given file")
+  }
+  import autoImport._
+
+  private lazy val scalaConfig =
```
olafurpg: Call this `scalafmtConfigParsed` instead?

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
+    scalafmtConfig.map(Config.fromHoconFile(_).get)
```
olafurpg: `.get` throws an exception on failure with full stacktrace, I think we can `throw new MessageOnlyException(e.getMessage)` to avoid the stack trace

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
+  private lazy val sbtConfig =
+    scalaConfig.map(conf => conf.copy(runner = conf.runner.forSbt))
```
olafurpg: This would actually be nice to have in the public facing API. Can we move this to `ScalafmtConfig.forSbt` as well as add a forwarder method `Scalafmt.configForSbt` exactly like is done here here https://github.com/scalameta/scalafmt/blob/49815ab491ff214363488072ca07ba6a5a524e96/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala#L81
The forwarder method is there for binary compatibility reasons since ScalafmtConfig can break binary compatibility.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
+
+  private type Input = String
```
olafurpg: This does not help readability IMO, either create a value class or use String.

scalafmt-cli-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -0,0 +1,25 @@
...
+import Keys._
+
+object ScalafmtPlugin extends AutoPlugin {
+  override def trigger: PluginTrigger = allRequirements
+  object autoImport {
+    val scalafmt: Command =
+      Command.args("scalafmt", "run the scalafmt command line interface.") {
+        case (state, args) =>
+          org.scalafmt.cli.Cli.main("--non-interactive" +: args.toArray)
+          PlatformTokenizerCache.megaCache.clear()
```
olafurpg: You want to run this in the new plugin too, megaCache will eventually go away but for now it's best to call it on every file. https://github.com/scalameta/scalameta/issues/1068

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
+  private type Output = String
+
+  private def withFormattedSources[T](
+      sources: Seq[File],
+      config: ScalafmtConfig
+  )(
+      onError: (File, Throwable) => T,
+      onFormat: (File, Input, Output) => T
+  ): Seq[Option[T]] = {
+    sources.map(
```
olafurpg: This formats files sequentially, for large projects it may pay off to do `.par.map`, but I'll leave it to you to decide.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,169 @@
...
   override def trigger: PluginTrigger = allRequirements
+
   object autoImport {
-    val scalafmt: Command =
-      Command.args("scalafmt", "run the scalafmt command line interface.") {
-        case (state, args) =>
-          org.scalafmt.cli.Cli.main("--non-interactive" +: args.toArray)
-          PlatformTokenizerCache.megaCache.clear()
-          state
+    lazy val scalafmt = TaskKey[Unit]("scalafmt", "Format Scala sources")
```
fommil: these don't need to be `lazy`, also probably best to use the sbt macros so you don't need to repeat the name as a string.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
fommil: this means that if the config file changes, sbt needs to be restarted. It would be more idiomatic to assign this to a Task that can be cached based on the file.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
fommil: we'll get parallelism for free across projects / configs because of the way sbt works, it's not so clear that parallelism at this level will improve performance. let's measure first.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
sjrd: `(unmanagedSources in scalafmt).value` automatically defaults to `unmanagedSources.value` if the former is not defined. You don't need to do anything for that, it's built in the scopes mechanism.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
olafurpg: I don't think there is any 100% correct way to do this, but I suspect doing the same as sbt-scalariform should be least controversial https://github.com/sbt/sbt-scalariform/blob/b33f32b43bd267a3b5e2b26a5962a77182709f52/src/main/scala/com/typesafe/sbt/SbtScalariform.scala#L82-L87 override `compileInputs` and move the dynTask to only cover the scalafmt-related parts

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
dwijnand: Better yet, in global settings. http://www.scala-sbt.org/1.0/docs/Plugins.html#projectSettings+and+buildSettings

build.sbt
dwijnand: This is effectively defining which sbt version the plugin targets, rather than defaulting whatever sbt version you're using for the whole build as defined in `build.properties`. I've used this in the past to ensure my plugin works on old versions of sbt while using the latest version to build the plugin.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
dwijnand: `scalafmtConfig` is a `TaskKey[File]`, and this is calling map on it, which will return an `Def.Initialize[Task[Configured[ScalafmtConfig]]]` (if I understand correctly) so it's ok.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
sjrd: No, I think this one should be `in ThisBuild`. If you have `ProjectRef`s to other directories outside your build, you don't want those to get reformatted when they're compiled.

dwijnand: Right, that's why `false` is the right default. But rather than define it for every build in the state, just define it once globally, then independent builds (or specific projects) can opt into formatting on compile.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
vovapolu: @dwijnand Why did you put a dislike here?

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
olafurpg: right, since  the plugin works at the granularity of project/config then you already get parallelism there. In the cli we use `.par` to run on all files in the repo

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
dwijnand: Just because I disagree (even if the type `Input` is just an alias, I think it's can be useful to have such aliases).

olafurpg: My view is that you might as well make it a value class then, a type alias like this gives a false sense of type safety

vovapolu: well, it's 2 vs 2, so I leave it :)

scalafmt-cli-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
vovapolu: Call on every file? Your plugin calls it once at start.

vovapolu: And when I should call it? Even when I check source or only when format?

olafurpg: It's a bit trickier to decide compared to the cli, but I think we can do it once per config/project. This should ideally run whenever scalafmt is not running, otherwise it shoudl be a well-behaving cache and the only penalty for clearing it too often is a very minor slowdown hit.

vovapolu:
@olafurpg Can you take a look now? I fixed all comments that were unanimously approved :D

scalafmt-sbt-tests/src/sbt-test/scalafmt-sbt/sbt/test
```diff
@@ -0,0 +1,24 @@
+> p123/compile:scalafmt
```
olafurpg: Should we maybe add a negative check here?

```
-> p123/compile:scalafmtCheck
```

to validate check fails on misformatted files.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,175 @@
...
   object autoImport {
-    val scalafmt: Command =
-      Command.args("scalafmt", "run the scalafmt command line interface.") {
-        case (state, args) =>
-          org.scalafmt.cli.Cli.main("--non-interactive" +: args.toArray)
-          PlatformTokenizerCache.megaCache.clear()
-          state
+    val scalafmt = taskKey[Unit]("Format Scala sources")
+    val scalafmtCheck =
+      taskKey[Boolean]("Check that Scala sources is formatted properly")
```
olafurpg: > Fails if a Scala source is mis-formatted. Does not write to files.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,175 @@
...
+    val scalafmtOnCompile =
+      settingKey[Boolean]("Format source when compiling")
```
olafurpg: Format Scala source files on compile, off by default.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,175 @@
...
+    val scalafmtConfig = taskKey[File]("Scalafmt config file")
```
olafurpg: Location of .scalafmt.conf file

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,175 @@
...
+    val scalafmtSbt = taskKey[Unit]("Format SBT sources")
```
olafurpg: Format `*.sbt` and `project/*.scala` files for this sbt build

PS. it's written "sbt" in lowercase ;)

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,175 @@
...
+    val scalafmtSbtCheck =
+      taskKey[Boolean]("Check that SBT sources is formatted properly")
```
olafurpg: > Fails if a `*.sbt` or `project/*.scala` source is mis-formatted. Does not write to files.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,175 @@
...
+    val scalafmtOnly = inputKey[Unit]("Format only given file")
+  }
+  import autoImport._
+
+  private val scalaConfig = scalafmtConfig.map(Config.fromHoconFile(_) match {
```
olafurpg: Since `scalafmtConfig` is a task key, does that mean this will run on every `scalafmt`? Parsing config is pretty cheap but it's quite wasteful to do it if you run scalafmt on compile. Can we use sbt's caching utilities to avoid redundant work? http://www.scala-sbt.org/1.x/docs/Faq.html#How+can+a+task+avoid+redoing+work+if+the+input+files+are+unchanged%3F

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,175 @@
...
+    case Configured.Ok(conf) => conf
+    case Configured.NotOk(e) => throw new MessageOnlyException(e.msg)
+  })
+  private val sbtConfig = scalaConfig.map(_.forSbt)
+
+  private type Input = String
```
olafurpg: Fair enough, I don't feel strongly about this. I'm guilty of doing this myself sometimes 😏

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,175 @@
...
+  private def withFormattedSources[T](
+      sources: Seq[File],
+      config: ScalafmtConfig
+  )(
+      onError: (File, Throwable) => T,
+      onFormat: (File, Input, Output) => T
+  ): Seq[Option[T]] = {
+    sources.map(
+      file => {
+        val input = IO.read(file)
```
olafurpg: Would it make sense to make this a cached function? http://www.scala-sbt.org/1.x/docs/Faq.html#How+can+a+task+avoid+redoing+work+if+the+input+files+are+unchanged%3F

Scalafmt is around 6x slower than scalariform because it does a lot of work to figure out the best line wrapping (and the best-first algorithm I used is quite a bad fit for this domain, see https://github.com/scalameta/scalafmt/issues/917). Scalafmt performance is usually not a problem for smaller files or if you run scalafmt infrequently but if you run it on compile on large source files it can impact compile times. I'll leave it to you to decide, I would at least keep it in mind.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,175 @@
...
+    scalafmtCheck :=
+      checkSources(
+        (unmanagedSources in scalafmt).value,
+        scalaConfig.value,
+        streams.value.log),
+    scalafmtSbtCheck := {
+      checkSources(sbtSources.value, sbtConfig.value, streams.value.log)
+      checkSources(projectSources.value, scalaConfig.value, streams.value.log)
+    },
+    compileInputs in compile := Def.taskDyn {
```
olafurpg: iiuc, this implementation still breaks `inspect tree compile`, we need to push `Def.taskDyn` into a `scalafmtDoAutoformat` task, like is done here:

https://github.com/sbt/sbt-scalariform/blob/master/src/main/scala/com/typesafe/sbt/SbtScalariform.scala#L82-L87

I may be wrong @dwijnand may know better

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,175 @@
...
+      val sbtFiles = absFiles.filter(_.toString.endsWith(".sbt"))
+      formatSources(sbtFiles, sbtConfig.value, streams.value.log)
+    }
+  )
+
+  override def projectSettings: Seq[Def.Setting[_]] =
+    Seq(Compile, Test).flatMap(inConfig(_)(scalafmtConfigSettings))
+
+  override def buildSettings: Seq[Def.Setting[_]] = Seq(
+    scalafmtConfig := (baseDirectory in ThisBuild).value / ".scalafmt.conf"
```
olafurpg: Will this fail if the file does not exist? I think sbt-scalafmt should work fine out of the box without any .scalafmt.conf, do we have a test for this?

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
vovapolu: well, I just copied it from `neo-scalafmt-sbt` plugin :D

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
vovapolu: Yeah, it's called quite a lot. I'll try to use some caching.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
vovapolu: I can use `default` config when this file doesn't exist. I first used `Option[File]` with default `None`, but changed it thinking `.scalafmt` file **should** exists.

ShaneDelmore: I often use scalafmt on smaller personal projects with the default config without making a .scalafmt file.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
dwijnand: Yes, `Def.taskDyn` "breaks" `inspect` - as in the dependencies don't show up.

sjrd: In addition, note that this could/should be a `settingDyn`, as the test only depends on the value of a setting--the fact that the returned thing is a task is irrelevant. (Too many tasks out there are `taskDyn`s where they could be `settingDyn`s.) If https://github.com/sbt/sbt/issues/3258 was implemented, this problem with breaking `inspect` wouldn't happen.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
vovapolu: The solution in your link only works for generating output files. It doesn't cache any values of arbitrary type `T`.

vovapolu:
@olafurpg I think caching isn't going to be easy to implement. Your link is suitable for `formatSources` task, for example, but not for caching some values computed from files. So I suggest to leave it for next PRs :)

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -15,23 +15,32 @@ object ScalafmtPlugin extends AutoPlugin {
...
-      taskKey[Boolean]("Check that Scala sources is formatted properly")
+      taskKey[Boolean](
+        "Fails if a Scala source is mis-formatted. Does not write to files.")
     val scalafmtOnCompile =
-      settingKey[Boolean]("Format source when compiling")
-    val scalafmtConfig = taskKey[File]("Scalafmt config file")
-    val scalafmtSbt = taskKey[Unit]("Format SBT sources")
+      settingKey[Boolean](
+        "Format Scala source files on compile, off by default.")
+    val scalafmtConfig = taskKey[Option[File]](
```
olafurpg: Does this mean that you must `scalafmtConfig := Some(file(".scalafmt.conf"))` to avoid the default settings? I would prefer to make this non-Option and if the file does not exist then use default settings.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -137,13 +146,16 @@ object ScalafmtPlugin extends AutoPlugin {
       checkSources(sbtSources.value, sbtConfig.value, streams.value.log)
       checkSources(projectSources.value, scalaConfig.value, streams.value.log)
     },
-    compileInputs in compile := Def.taskDyn {
-      val task =
-        if (scalafmtOnCompile.value) scalafmt in resolvedScoped.value.scope
-        else Def.task(())
-      val previousInputs = (compileInputs in compile).value
-      task.map(_ => previousInputs)
+    scalafmtDoFormatOnCompile := Def.taskDyn {
```
olafurpg: As recommended by @sjrd we should make this a settingDyn

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
vovapolu: It searches for `.scalafmt.conf` file by default. And you can redefine it for your custom file.

olafurpg: I see, cool. OK this is fine then 👍

vovapolu:
I'm merging it?

olafurpg:
Please hold a bit, I want to give the plugin a swing locally first. I was traveling to Munich today, I'll try to take a look tomorrow!

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,194 @@
...
+  private def formatSources(
+      sources: Seq[File],
+      config: ScalafmtConfig,
+      log: Logger
+  ): Unit = {
+    withFormattedSources(sources, config)(
+      (file, e) => log.error(s"Error in ${file.toString}: $e"),
+      (file, input, output) => {
+        if (input != output) {
+          log.info(s"Successfully formatted ${file.toString}.")
```
olafurpg: This log entry will be quite noisy for large projects with 100s of sources. I propose we call this once per `formatSources` with a message like "Formatting 1 Scala source"/"Formatting 2 Scala sources" similar to how scalariform does https://github.com/sbt/sbt-scalariform/blob/a378b724a76ee94819939efc0e9adebd896e57d9/src/main/scala/com/typesafe/sbt/Scalariform.scala#L80

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,194 @@
...
+        (unmanagedSources in scalafmt).value,
+        scalaConfig.value,
+        streams.value.log),
+    scalafmtSbtCheck := {
+      checkSources(sbtSources.value, sbtConfig.value, streams.value.log)
+      checkSources(projectSources.value, scalaConfig.value, streams.value.log)
+    },
+    scalafmtDoFormatOnCompile := Def.settingDyn {
+      if (scalafmtOnCompile.value) {
+        scalafmt in resolvedScoped.value.scope
```
olafurpg: This will run scalafmt on **all** sources on **every** `compile`, including nop compiles. It does not take advantage of incrementality. Is this expected/intended behavior? cc/ @fommil I think scalariform uses FileFunction.cached to make this incremental https://github.com/sbt/sbt-scalariform/blob/a378b724a76ee94819939efc0e9adebd896e57d9/src/main/scala/com/typesafe/sbt/Scalariform.scala#L85

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
olafurpg: Output from testing plugin on akka build https://gist.github.com/olafurpg/7a33d069604324a8bf4249cb067b7dd0

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,194 @@
...
+          .filterNot(_.isHidden)
+      rootSbt ++ projectSbt
+    }
+  )
+  private lazy val projectSources = thisProject.map(proj =>
+    (BuildPaths.projectStandard(proj.base) * GlobFilter("*.scala")).get)
+
+  lazy val scalafmtConfigSettings: Seq[Def.Setting[_]] = Seq(
+    scalafmt := formatSources(
+      (unmanagedSources in scalafmt).value,
```
olafurpg: This includes non-Scala source like *.java files. Running on the akka build results in a lot of parse errors like this

```scala
[error] Error in /Users/ollie/dev/akka/akka-testkit/src/test/java/akka/testkit/AkkaJUnitActorSystemResource.java: <input>:48: error: expected class or object definition
[error] public class AkkaJUnitActorSystemResource extends ExternalResource {
[error] ^
[error] Error in /Users/ollie/dev/akka/akka-testkit/src/test/java/akka/testkit/TestActorRefJavaCompile.java: <input>:10: error: expected class or object definition
[error] public class TestActorRefJavaCompile {
[error] ^
```

olafurpg: We should add some *.java files to the scripted tests

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
fommil: yeah it's only supposed to be compiling files that are being compiled, not *all* of them

olafurpg:
I just tried out the plugin on one of my projects and noticed that it formatted files that are excluded with `project.excludeFilters` in .scalafmt.conf. The plugin should use the `(config: ScalafmtConfig).project.matcher.matches(String): Boolean` to skip files. 
https://github.com/scalameta/scalafmt/blob/6cddfe9e217de11738c47240911eaf024cecbb0e/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala#L13

The setting is used here for example: https://github.com/scalameta/language-server/blob/573863cb1db6bf99ac73f35ac17f733b92044beb/.scalafmt.conf#L12-L15

The idiomatic way to do this in sbt-scalafmt would be to customize `unmanagedSources.in(scalafmt)` but that can't be picked up by editor integrations like IntelliJ, so that's why it's supported via .scalafmt.conf.

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
vovapolu: Let's leave caching for next PRs. I would like to implement it entirely, not only in some specific places. I will write a warning in the task description for now :)

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,211 @@
...
+        "Fails if a *.sbt or project/*.scala source is mis-formatted. Does not write to files.")
+    val scalafmtOnly = inputKey[Unit]("Format only given file.")
+  }
+  import autoImport._
+
+  private val scalafmtDoFormatOnCompile =
+    taskKey[Unit]("Format Scala source files if scalafmtOnCompile is on.")
+
+  private val scalaConfig =
+    scalafmtConfig.map(_.map(Config.fromHoconFile(_) match {
```
olafurpg: Could we at least cache this by the timestamp of the file? I feel it's unnecessary to reparse the config for every project/config on every compile.

olafurpg: Here's what the IntelliJ plugin uses https://github.com/scalameta/scalafmt/blob/0f391aa351c59432d1cf6693b5658ab81d4de517/scalafmt-cli/src/main/scala/org/scalafmt/cli/StyleCache.scala

olafurpg: I see, sure we can leave it for a separate PR :)

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
```diff
@@ -1,25 +1,211 @@
...
-          org.scalafmt.cli.Cli.main("--non-interactive" +: args.toArray)
-          PlatformTokenizerCache.megaCache.clear()
-          state
+    val scalafmt = taskKey[Unit]("Format Scala sources.")
+    val scalafmtCheck =
+      taskKey[Boolean](
+        "Fails if a Scala source is mis-formatted. Does not write to files.")
+    val scalafmtOnCompile =
+      settingKey[Boolean](
+        "Format Scala source files on compile, off by default. Warning: It formats all project sources.")
```
olafurpg: Let's remove this warning, it's not clear exactly what it's warning about and we will fix this in a future PR anyways.

fommil:
w00t! I'm really looking forward to format on compile in the follow up :smile: sounds like it's currently not doing the right thing?

olafurpg:
Format on compile seems to work in the current version but it re-formats all of `unmanagedSources.in(scalafmt)`, regardless of incremental compilation

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
vovapolu: >Here's what the IntelliJ plugin uses 

@olafurpg maybe simply reuse that object? It does just what we need

olafurpg: Go ahead, we might want to move it to the `core` module

scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala
vovapolu: @olafurpg that's a wrong solution. I don't know why scalariform uses it. `FileFunction.cached` accepts set of files, if you modify them (and we modify them by formatting), then it thinks that files are changed and runs the function again. So it needs two iteration to "converge" to formatted source and forget about it.

PR #1087, opened 2017-11-30
Fix file not found errors when running from a subdirectory w/ project.git = true

### LINKED ISSUE #1086: Broken paths in CLI when running from a subdirectory of a Git root
- **Version**: 1.3.0
- **Integration**: CLI
- **Configuration**
```
project.git = true
```

## Steps

1. Clone https://github.com/lloydmeta/scalafmt-nested-git-bug
2. Enter the `nested` directory of the project.
3. Run:

```bash
scalafmt
```

## Problem

The tool fails with file not found for any files inside the subdirectory because a directory is missing

```
Exception in thread "main" java.io.FileNotFoundException: /Users/lloyd/Documents/skala/scalafmt-nested-git-bug/build.sbt (No such file or directory)
```

The current subdirectory (`nested`) is missing from that path.

## Expectation

It should not fail.

## Workaround

I've found that it works as long as you are in the main Git root when running `scalafmt`. Another work around is to not have `project.git = true` in your config

lloydmeta:
Closes #1086 

* Add back --full-name when invoking Git 
* Add a test

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
@@ -46,6 +46,7 @@ class GitOpsImpl(private[util] val workingDirectory: AbsoluteFile)
         Seq(
           "git",
           "ls-files",
+          "--full-name",
           dir.path
         )
       ).toOption.toSeq.flatten.map(f => rtDir / f)
--- a/scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
+++ b/scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
@@ -93,14 +93,16 @@ class GitOpsTest extends fixture.FunSuite {
   test(
     "lsTree should return files properly when the working directory is under the git root directory") {
     implicit ops =>
+      val f1 = touch()
+      add(f1)
+
       val newDir = mkDir()
-      println(s"new dir $newDir")
-      val f = touch(dir = Some(newDir))
-      add(f)
+      val f2 = touch(dir = Some(newDir))
+      add(f2)
 
       {
         val innerGitOps = new GitOpsImpl(newDir)
-        ls(innerGitOps) should contain(f)
+        ls(innerGitOps) should contain only f2
       }
   }
 
--- a/scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
+++ b/scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
@@ -100,9 +100,9 @@ class GitOpsTest extends fixture.FunSuite {
 
       {
         val innerGitOps = new GitOpsImpl(newDir)
-        ls(innerGitOps) should contain (f)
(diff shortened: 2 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
```diff
@@ -46,6 +46,7 @@ class GitOpsImpl(private[util] val workingDirectory: AbsoluteFile)
         Seq(
           "git",
           "ls-files",
+          "--full-name",
```
pjrt: What should the behavior be for files not in the directory? As I understand this, this will get ALL files, even ones not in the directory you are in.

lloydmeta: Can you double check that ?

That's not the behaviour that I'm seeing:

1. Using git from the commandline
2. In the refined test I just added

https://github.com/lloydmeta/scalafmt/blob/a5d46c9c6899391f0eab1a310660e73fc8888335/scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala#L96-L105

3. In the man pages

```
--full-name
           When run from a subdirectory, the command usually outputs paths
           relative to the current directory. This option forces paths to be
           output relative to the project top directory.
```

pjrt: Well u added a test to prove that that was the case, that's enough to shut me up :)

scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
```diff
@@ -82,6 +90,22 @@ class GitOpsTest extends fixture.FunSuite {
...
+    implicit ops =>
+      val f1 = touch()
+      add(f1)
+
+      val newDir = mkDir()
+      val f2 = touch(dir = Some(newDir))
+      add(f2)
+
+      {
+        val innerGitOps = new GitOpsImpl(newDir)
```
pjrt: yeah that's super annoying. Really need to refactor GitOpsImpl.

scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
```diff
@@ -82,6 +90,22 @@ class GitOpsTest extends fixture.FunSuite {
...
+    "lsTree should return files properly when the working directory is under the git root directory") {
+    implicit ops =>
+      val f1 = touch()
+      add(f1)
+
+      val newDir = mkDir()
+      val f2 = touch(dir = Some(newDir))
+      add(f2)
+
+      {
```
pjrt: What's with the random `{}`?

lloydmeta: Removed.

In case you were curious though 😆 

* I was originally using an implicit `GitOpsImp`, and thought the brackets would help scope the `ls` call from the outer `GitOpsImpl` (too much Rustlang) but it didn't 
* Then I left it because I thought it made it conveyed nicely that this set of operations was taking place inside a subdirectory

PR #1088, opened 2017-12-01
Add  -r bintray:scalameta/maven to Coursier install commands

lloydmeta:
Add  -r bintray:scalameta/maven to Coursier install commands because
the version resolves to the latest commit hash on master, which is
auto-published to bintray, but *not* to Maven Central.


olafurpg:
Good idea! The version number on the website should actually be the latest stable instead of nightly, I will follow up with another PR to add a badge for the bintray releases.

PR #1090, opened 2017-12-06
Introduce "default", "akka" and "akka-http" as IndentOperator config

daddykotex:
This will allow users to override the current config via a comment and then set
it back to what it was. Something like:

```
// scalafmt: { indentOperator = spray }
get {
  path("hey) {
    complete("yeah!")
  } ~
  path("oh") {
    complete("oh oh")
  }
}
// scalafmt: { indentOperator = default }
```

I've reused an existing test but instead of `indentOperator = spray`, I used the new one: `indentOperator = default` to make sure it works. I'm not entirely sure the test is valuable, let me know otherwise.

Also, if I follow the contributing guidelines, I should be adding stuff to the documentation. IndentOperator are only found, AFAIK, in the `Other` section of the configuration. This section seems to be generated from `default.fields` and I can't figure out where this variable comes from.

There is also a mention of the IndentOperator in the fourth changelog. Should I add a changelog file?

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/IndentOperator.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/IndentOperator.scala
@@ -46,7 +46,11 @@ object IndentOperator {
     ScalafmtConfig.indentOperatorsExcludeAkka)
   implicit val IndentOperatorDecoder: ConfDecoder[IndentOperator] =
     ConfDecoder.instance[IndentOperator] {
-      case Conf.Str("spray") => Ok(IndentOperator.akka)
-      case els => default.reader.read(els)
+      case Conf.Str("spray" | "akka" | "akka-http") =>
+        Ok(IndentOperator.akka)
+      case Conf.Str("default") =>
+        Ok(IndentOperator.default)
+      case els =>
+        default.reader.read(els)
     }
 }


scalafmt-core/shared/src/main/scala/org/scalafmt/config/IndentOperator.scala
```diff
@@ -46,7 +46,11 @@ object IndentOperator {
     ScalafmtConfig.indentOperatorsExcludeAkka)
   implicit val IndentOperatorDecoder: ConfDecoder[IndentOperator] =
     ConfDecoder.instance[IndentOperator] {
-      case Conf.Str("spray") => Ok(IndentOperator.akka)
-      case els => default.reader.read(els)
+      case Conf.Str("spray") | Conf.Str("akka") | Conf.Str("akka-http") =>
```
olafurpg: `Conf.Str("spray" | "akka" | "akka-http")`

scalafmt-tests/src/test/resources/test/DefaultIndentOperator.stat
```diff
@@ -0,0 +1,75 @@
+indentOperator = default
```
olafurpg: Is this test necessary? The default indent operator is tested throughout the test suite for every infix operator.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/IndentOperator.scala
daddykotex: Oh wow, I did not know I could do that

scalafmt-tests/src/test/resources/test/DefaultIndentOperator.stat
daddykotex: From my understanding, it has no effect (because the indentOperator is already default) but it is still parsed and if the `"default"` wasn't part of the `ConfDecoder` it would fail:

```
→ java -jar scalafmt.jar -f /tmp/test.scala
StyleMap:44 [e]:
Type mismatch;
  found    : String (value: "bob")
  expected : IndentOperator
```

daddykotex: This is the only reason I've added the test, to show the config `indentOperator = default` is working.

Should I remove it?

daddykotex: Or maybe it is just not the right place to test for config parsing... I'll have another look to see where this test would make more sense

olafurpg: The pattern match is simple enough so I think it's fine to skip testing the config parsing.

daddykotex: Aight, I removed it

Issue #913, opened 2017-04-25
Too eager at breaking down curly brace lambdas


dwijnand:
This template is a guideline, not a strict requirement.
- **Version**: 0.6.8
- **Integration**: CLI
- **Configuration**:
```
maxColumn = 100
```

## Steps

I ran scalafmt like this:

```bash
scalafmt
```

## Problem

Scalafmt formats code like this:
```scala
class Discovery(baseClasses: Set[String], annotations: Set[String]) {
  def apply(s: Seq[Definition]): Seq[(Definition, Discovered)] =
    s.map { d =>
      (d, apply(d))
    }
}

class LoggerReporter(
    sourcePositionMapper: Position => Position = { p =>
      p
    }
)
```

## Expectation

I would like the formatted output to look like this:
```scala
class Discovery(baseClasses: Set[String], annotations: Set[String]) {
   def apply(s: Seq[Definition]): Seq[(Definition, Discovered)] =
     s.map { d => (d, apply(d)) }
}

class LoggerReporter(
  sourcePositionMapper: Position => Position = { p => p }
)
```

dwijnand:
Any insights as to why this is happening and/or how to fix it?

olafurpg:
The handling of line breaks after `{` is shared for block and lambdas https://github.com/scalameta/scalafmt/blob/3df7c037dd7936ed87de8279dac4b7fc5bc0336f/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala#L178-L236 I've took a stab at this once and struggles to make it work without breaking a lot of tests. It's probably best to handle lambdas separately in another case.

When I started scalafmt I felt like this output here
```scala
foo { 
  loooooooooooooooong =>
    ...
}
```
made sense but in hindsight I think I would probably prefer this today
```scala
foo { loooooooooooooooong =>
    ...
}
```
even if `loooooong` breaks the column limit. This change would also simplify the implementation, since then there are only two alternative layouts instead of three.

olafurpg:
This issue bugs me too almost daily, so I would love if someone took a swing at fixing it. I haven't been able to dedicate much time to scalafmt past months except to review the occasional PR and cut new releases (which takes a while because of IntelliJ plugin).

sjrd:
What I often have in my code is not
```scala
foo {
  loooooooooooooooong =>
    ...
}
```
but rather
```scala
foo.babar.bar(someArgs).map {
  nonSingleCharName =>
    ...
}
```
[...]

PR #1092, opened 2017-12-14
Add an "original code" section the issue template

dwijnand:
This way it's clear what the state of the code was before running scalafmt, so it's clear what scalafmt did, compared to what you expected it to do.

See https://github.com/scalameta/scalafmt/issues/913#issuecomment-351806982 for an example of how the previous template can be misleading.


olafurpg:
Thanks!

PR #1100, opened 2018-01-16
Fix newlines around function blocks Fixes #1097

### LINKED ISSUE #1097: Unwated newline in PartialFunction
- **Version**: 1.4
- **Integration**: IntelliJ/CLI
- **Configuration**: Default

## Steps

Given code like this:
```scala
client.onEventCAlternate { implicit c => {
    case APIMessage.Ready(_)            => println("Foo")
    case APIMessage.ChannelCreate(_, _) => println("Foo")
  }
}
```

```bash
scalafmt -i /path/to/file
```

## Problem

```scala
client.onEventCAlternate { implicit c =>
  {
    case APIMessage.Ready(_)            => println("Foo")
    case APIMessage.ChannelCreate(_, _) => println("Foo")
  }
}
```

## Expectation

```scala
client.onEventCAlternate { implicit c => {
    case APIMessage.Ready(_)            => println("Foo")
    case APIMessage.ChannelCreate(_, _) => println("Foo")
  }
}
```

Katrix-:
Reworked `newlines.afterCurlyLambda` to not insert a newline at all if a function block follows. Was a bit unsure if I should have introduced a new parameter, or just work with the existing one. Decided to work with the existing one as if I didn't, the name wouldn't make as much sense anymore.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -242,16 +242,22 @@ class Router(formatOps: FormatOps) {
           leftOwner.asInstanceOf[Term.Function].body)
         val canBeSpace =
           statementStarts(hash(right)).isInstanceOf[Term.Function]
-        val afterCurlyNewlines =
+
+        val (afterCurlyModification, afterCurlyCost) =
           style.newlines.afterCurlyLambda match {
-            case NewlineCurlyLambda.never => Newline
-            case NewlineCurlyLambda.always => Newline2x
+            case NewlineCurlyLambda.never if right.is[LeftBrace] => (Space, 0)
+            case NewlineCurlyLambda.never => (Newline, 1)
+            case NewlineCurlyLambda.always => (Newline2x, 1)
             case NewlineCurlyLambda.preserve =>
-              if (newlines >= 2) Newline2x else Newline
+              if(newlines == 0) (Space, 0)
+              else if(newlines == 1) (Newline, 1)
+              else (Newline2x, 2)
           }
+
         Seq(
           Split(Space, 0, ignoreIf = !canBeSpace),
-          Split(afterCurlyNewlines, 1).withIndent(2, endOfFunction, Left)
+          Split(afterCurlyModification, afterCurlyCost)
+            .withIndent(2, endOfFunction, Left)
         )
       case FormatToken(arrow @ RightArrow(), right, _)
           if leftOwner.is[Term.Function] =>
--- a/scalafmt-tests/src/test/resources/newlines/afterCurlyLambdaNever.stat
+++ b/scalafmt-tests/src/test/resources/newlines/afterCurlyLambdaNever.stat
@@ -43,3 +43,26 @@ def f = {
     g(x)
   }
 }
+
+<<< Don't put newlines around functions blocks
+foo { c =>
+  {
+    println("Foo")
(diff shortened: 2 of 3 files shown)


sjrd:
TBH, this doesn't make any sense to me. You pair two `{` on the same line with two `}` on different lines, with different levels of indentation.

In addition, in
```scala
foo { c => {
    println("Foo")
  }
}
```
the inner braces are not even necessary, so the correct way to write it would be
```scala
foo { c =>
  println("Foo")
}
```

jozic:
in original issue the example is like
```scala
client.onEventCAlternate { implicit c => {
    case APIMessage.Ready(_)            => println("Foo")
    case APIMessage.ChannelCreate(_, _) => println("Foo")
  }
}
```

which may require inner braces if `onEventCAlternate` takes `C => PartialFunction[_,_]`
outer braces can be replaced with parenthesis though

sjrd:
IMO there are two valid ways to format that piece of code.

The one I would personally use:
```scala
client.onEventCAlternate { implicit c =>
  {
    case APIMessage.Ready(_)            => println("Foo")
    case APIMessage.ChannelCreate(_, _) => println("Foo")
  }
}
```
and another one, closer to your formatting:
```scala
client.onEventCAlternate { implicit c => {
  case APIMessage.Ready(_)            => println("Foo")
  case APIMessage.ChannelCreate(_, _) => println("Foo")
} }
```
In both cases, the indentation of closing brace lines matches the indentation of the corresponding opening brace lines. In your formatting, that is not the case, and there is no other formatting rule in scalafmt that breaks this matching (AFAIK).

Katrix-:
The first one is whats happening currently, which was my goal to get around. Would it be better to add another config entry to control this behavior?

jozic:
for `PartialFunction` i learned to appreciate what scalafmt does now, i.e.
```scala
client.onEventCAlternate { implicit c =>
  {
    case APIMessage.Ready(_)            => println("Foo")
    case APIMessage.ChannelCreate(_, _) => println("Foo")
  }
}
```
Though I didn't like it at first. Now i see it as very explicit sign that a PartialFunction is used here

Tastes are different, so it can be flexible. But as a new parameter probably.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -242,16 +242,22 @@ class Router(formatOps: FormatOps) {
           leftOwner.asInstanceOf[Term.Function].body)
         val canBeSpace =
           statementStarts(hash(right)).isInstanceOf[Term.Function]
-        val afterCurlyNewlines =
+
+        val (afterCurlyModification, afterCurlyCost) =
           style.newlines.afterCurlyLambda match {
-            case NewlineCurlyLambda.never => Newline
-            case NewlineCurlyLambda.always => Newline2x
+            case NewlineCurlyLambda.never if right.is[LeftBrace] => (Space, 0)
```
olafurpg: Might want to guard against rightOwner being a `Term.ParticalFunction`

jozic:
@olafurpg I would prefer keep partial function as it's now, but I understand that others might want it another way. 
Can't we just get a new value for `NewlineCurlyLambda`? like  `NewlineCurlyLambda.preserveIfPF` which will either do the new way or the old one (it doesn't really matter)

olafurpg:
Closing due to inactivity. Please reopen if you would like to pick this up again @Katrix- !

PR #1101, opened 2018-01-19
Add optIn.blankLineBeforeDocstring setting

olafurpg:
Previously, scalafmt always forced a blank line before docstrings. There
was no good reason for this, it was just a random decision I made before
v0.1 that no one has questioned until now. This behavior has in fact bugged
myself.

I think this setting is a good candidate to become default at some
point, but I'd like to receive a bit more feedback before making the
switch.

Review @alexandru

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/OptIn.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/OptIn.scala
@@ -46,6 +46,28 @@ import metaconfig._
   * @param selfAnnotationNewline See https://github.com/scalameta/scalafmt/issues/938
   *                              If true, will force a line break before a self annotation
   *                              if there was a line break there before.
+  * @param blankLineBeforeDocstring
+  *  If true, always insert a blank line before docstrings,
+  *  If false, preserves blank line only if one exists before.
+  *  Example:
+  *  {{{
+  *    // before
+  *    object Foo {
+  *      /** Docstring */
+  *      def foo = 2
+  *    }
+  *    // after, if blankLineBeforeDocstring=true
+  *    object Foo {
+  *      /** Docstring */
+  *      def foo = 2
+  *    }
+  *    // after, if blankLineBeforeDocstring=false
+  *    object Foo {
+  *
+  *      /** Docstring */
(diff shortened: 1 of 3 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/OptIn.scala
```diff
@@ -46,12 +46,36 @@ import metaconfig._
...
+  *  If true, always insert a blank line before docstrings,
+  *  If false, preserves blank line only if one exists before.
+  *  Example:
+  *  {{{
+  *    // before
+  *    object Foo {
+  *      /** Docstring */
+  *      def foo = 2
+  *    }
+  *    // after, if blankLineBeforeDocstring=true
```
alexandru: Shouldn't this be `after, if blankLineBeforeDocstring=false`?

alexandru:
So when can I use it? 😀

scalafmt-core/shared/src/main/scala/org/scalafmt/config/OptIn.scala
```diff
@@ -46,12 +46,36 @@ import metaconfig._
...
 @DeriveConfDecoder
 case class OptIn(
     configStyleArguments: Boolean = true,
     breaksInsideChains: Boolean = false,
     breakChainOnFirstMethodDot: Boolean = true,
     selfAnnotationNewline: Boolean = true,
-    annotationNewlines: Boolean = true
+    annotationNewlines: Boolean = true,
+    // Candidate to become default true at some point.
+    blankLineBeforeDocstring: Boolean = false
```
gabro: Maybe I'm reading it wrong, but doesn't this invert the current default?

olafurpg:
@alexandru It will likely be a while until the next stable Maven release, because publishing the IntelliJ plugin is super tricky. The CI publishes a pre-release to Bintray on merge http://scalameta.org/scalafmt/#Pre-release that you can use to give this a try, but the IJ  plugin cannot use pre-releases

olafurpg:
I set the deadline for v1.5.0 at February 16th, I try to keep releases ~6 weeks apart

PR #1103, opened 2018-01-24
Support trailing commas in sbt files

### LINKED ISSUE #1098: Support trailing commas in sbt files
Currently, the sbt dialect is based of 211, which does not permit trailing commas. We should base it off the default lenient dialect.

dwijnand:
Fixes #1098

Please review mercilessly, I'm happy to iterate.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
@@ -89,6 +89,6 @@ object ScalafmtRunner {
     */
   val statement = default.copy(parser = scala.meta.parsers.Parse.parseStat)
 
-  val sbt = default.copy(dialect = scala.meta.dialects.Sbt0137)
+  val sbt = default.forSbt
 
 }
--- a/scalafmt-core/shared/src/test/scala/org/scalafmt/ScalafmtRunnerTest.scala
+++ b/scalafmt-core/shared/src/test/scala/org/scalafmt/ScalafmtRunnerTest.scala
@@ -1,0 +1,24 @@
+package org.scalafmt
+
+import scala.meta._
+import org.scalafmt.config._
+import org.scalatest.FunSuite
+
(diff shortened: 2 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
```diff
@@ -89,6 +89,6 @@ object ScalafmtRunner {
     */
   val statement = default.copy(parser = scala.meta.parsers.Parse.parseStat)
 
-  val sbt = default.copy(dialect = scala.meta.dialects.Sbt0137)
+  val sbt = default
```
olafurpg: Can we do `forSbt` here?

scalafmt-core/shared/src/test/scala/org/scalafmt/ScalafmtRunnerTest.scala
```diff
@@ -0,0 +1,25 @@
+package org.scalafmt
+
+import scala.meta._
+import org.scalafmt.config._
+import org.scalatest.FunSuite
+
+class ScalafmtRunnerTest extends FunSuite {
+  test("sbt dialect supports trailing commas") {
+    val code = ScalafmtRunner.sbt.forSbt
```
olafurpg: `.sbt.forSbt` looks weird

scalafmt-core/shared/src/test/scala/org/scalafmt/ScalafmtRunnerTest.scala
```diff
@@ -0,0 +1,25 @@
...
+          |  a,
+          |
+          |
+          |  b,
+          |)
+        """.stripMargin
+      )
+      .parse[Source]
+      .get
+    assert(code != null)
```
olafurpg: Remove, parse will never return null

dwijnand: I thought I needed this to make the types align, but guess with scalatest.FunSuite you don't. Done.

PR #1112, opened 2018-02-21
Ensure RedundantParens only removes matching parens - fixes #1111

### LINKED ISSUE #1111: RedundantParens rewrite produces unparseable code
- **Version**: 1.4.0
- **Integration**: CLI
- **Configuration**:
```
maxColumn = 100
rewrite.rules = [AvoidInfix, RedundantBraces, RedundantParens, SortImports]
```

## Steps

Given code like this:
```scala
object Foo {
  val foo = for {
    x <- Seq(1, 2)
    if (true) || (true)
  } yield x

  def main(args: Array[String]) = println(foo.mkString(", "))
}
```

```bash
$ scalafmt Foo.scala
```

## Problem

Scalafmt errors like this:
```
[warn] Error in /Users/.../Foo.scala: <input>:4: error: } expected but ) found
if true) || (true
           ^
```

## Expectation

I would like scalafmt to not fail to parse this, as it is valid syntax.

```bash
$ scalac Foo.scala && scala Foo
1, 2
```

## Workaround

I've found that by removing the second part of the if statement, it formats just fine. e.g.

```scala
// BEFORE scalafmt

object Foo {
  val foo = for {
    x <- Seq(1, 2)
    if (true)
  } yield x

  def main(args: Array[String]) = println(foo.mkString(", "))
}

// AFTER scalafmt

object Foo {
  val foo = for {
    x <- Seq(1, 2)
    if true
  } yield x

  def main(args: Array[String]) = println(foo.mkString(", "))
}
```

## Notes

I have simplified this example from our actual use case as a proof of concept. In reality our use looks more like this, but the point still stands:

```scala
val result = for {
  x <- xs
  y <- ys
  if (foo(x) || bar) && (baz(y) || ban)
} yield {
  // ...
}

### PR DESCRIPTION
**Pull request by bcarl**

### CODE REVIEW

File: scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat [hunk 1]
```diff
@@ -31,3 +31,35 @@ object a {
       if (b ++ b).length >= 2
     } yield a
 }
+<<< compound statement
+object a {
+  def c(b: List[Int]): List[Int] =
+    for {
+      a <- b
+      if (a) || (a)
```
olafurpg: Any idea why this does not have balanced parentheses? I'm a bit uneasy about this fix since I suspect the root cause of the problem is something different.

bcarl: This source code does have balanced parentheses.

However, the bug here is that RedundantParens only trims leading and trailing parentheses and in this case it would output:

```scala
  if a) || (a
```

I decided to go for a safe fix with this PR and check if the resulting code coming from RedundantParens (`innerSegment` above) has balanced parentheses. This may not handle all cases, but it seemed like an easy fix for the issue at hand.

One could probably be clever about actually removing the parentheses correctly, but I wanted to keep this change small to fix the bug and perhaps somebody else could improve RedundantParens in the future.

bcarl:
@olafurpg I added some case statement test cases - looks like they don't get rewritten regardless.

scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat
olafurpg: Aaah, I see! Did you try using RewriteCtx.matchinParens? https://github.com/scalameta/scalafmt/blob/a9cc066127feca7384b03b516821e1840056d0db/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Rewrite.scala#L14

You can use that map to guard against rewriting open/close pairs that don't match.

bcarl: Interesting - I did not know about that. I updated the code to use that instead.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantParens.scala
```diff
@@ -23,8 +23,12 @@ case object RedundantParens extends Rewrite {
           val leftSegment: Tokens = tokens.slice(0, parensToRemoveCount)
           val rightSegment: Tokens =
             tokens.slice(tokens.length - parensToRemoveCount, tokens.length)
-          leftSegment.foreach(tok => builder += TokenPatch.Remove(tok))
-          rightSegment.foreach(tok => builder += TokenPatch.Remove(tok))
+          leftSegment.zip(rightSegment.reverse).foreach {
+            case (left, right) if ctx.isMatching(left, right) =>
```
olafurpg: Great that this worked!

PR #1117, opened 2018-03-03
Fix #1116, insert space in type projection when selecting symbolic members

### LINKED ISSUE #1116: Type projections need a space when the type member name starts with a symbol
The left-right dual of #756:

``` scala
abstract class A {
  // or !_whatever
  type !
}

// the projection A# ! needs the space
// A#! won't compile
def doSomething[A0 <: A]: A# !
```

olafurpg:
Review @eparejatobes


eparejatobes:
:+1:

olafurpg:
If you don't use the IntelliJ plugin you can try out the fix with http://scalameta.org/scalafmt/#Pre-release once CI completes on master.

eparejatobes:
@olafurpg thanks, I tested it on a quite symbol-heavy codebase and it works nicely.

olafurpg:
Sweet 👍

PR #1121, opened 2018-03-13
Correct Homebrew installation terminology

srstevenson:
The scalafmt package is a formula, not a cask.

### DIFF
--- a/readme/Installation.scalatex
+++ b/readme/Installation.scalatex
@@ -78,7 +78,7 @@
         needs to be restarted once in awhile.
 
     @sect{Homebrew}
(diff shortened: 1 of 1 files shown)


readme/Installation.scalatex
```diff
@@ -78,7 +78,7 @@
         needs to be restarted once in awhile.
 
     @sect{Homebrew}
-      You can install scalafmt via homebrew using a custom cask
+      You can install scalafmt via Homebrew using a custom formula
```
olafurpg: I probably meant "tap" https://docs.brew.sh/Taps but formula is good 👍

PR #1122, opened 2018-03-13
Remove redundant braces from if-else and case expressions

japgolly:


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -66,22 +70,71 @@ case object RedundantBraces extends Rewrite {
...
-        val next = nextToken(close)
-        if (next.is[LF] &&
-          close.pos.start.line != bodyStatement.pos.end.line)
-          builder += TokenPatch.Remove(next)
-
+        removeTrailingLF(bodyStatement.pos, close)
         builder += TokenPatch.Remove(open)
         builder += TokenPatch.Remove(close)
+
+      case x: Term.If if settings.ifElseClauses =>
```
olafurpg: I'm wondering if we can generalize the implementation to handle all single statement Term.Blocks in one go. What happens if we do something like the following psuedo code? 
```scala
case b: Term.Block if b.stats.lengthCompare(1) == 0 =>
  for {
    open <- b.tokens.headOption
    if open.is[Token.LeftBrace]
    close <- b.tokens.headOption
    if close.is[Token.RightBrace]
  } yield List(remove(open), remove(left), removeTrailingLF(...))
```

That could handle if + else + case clauses. If there's a wish for more detailed configuration  we could do it against `b.parent.get.productPrefix` which returns for example "Term.If" or "Case"

olafurpg:
to fix the formatting check you can run `./scalafmt` from the project directory. The other test failure in 2.12 is not spurious, I think it's related to `if (true) 1`.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -66,22 +70,71 @@ case object RedundantBraces extends Rewrite {
...
+                              (implicit builder: PatchBuilder, ctx: RewriteCtx): Unit = {
+    import ctx.tokenTraverser._
+    val next = nextToken(close)
+    if (next.is[LF] && close.pos.start.line != bodyEnd.end.line)
+      builder += TokenPatch.Remove(next)
+  }
+
+  private def patchExpr(term: Term, inBlock: Boolean)
+                       (implicit builder: PatchBuilder, ctx: RewriteCtx): Unit = {
+    val open = term.tokens.head
```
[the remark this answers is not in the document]
olafurpg: Actually looks like some Term.Block can also have empty tokens
```scala
ollie-scalafix-scopes@ "case x =>".parse[Case].get.asInstanceOf[Case].body
res10: Term = Term.Block(List())

ollie-scalafix-scopes@ "case x =>".parse[Case].get.asInstanceOf[Case].body.tokens
res11: Tokens = Tokens()
```

japgolly:
Hi @olafurpg. Firstly, thank you for the awesome feedback. This is my first time working with scala{fmt,meta} and appreciate all the help I can get :)

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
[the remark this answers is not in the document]
japgolly: Oh my, what a silly mistake :laughing: I'll check for emptyness. And I like the way you're quickly testing things out in the console, fantastic idea, I'll adopt that. Thanks :)

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
japgolly: That appeals to me much more conceptually but why wasn't it done that way before? The fact that it was only done for a few small cases suggests to me that a holistic approach didn't work... Is that not so? I'd be interested to know the history.

japgolly: Bah, I'm going down this path regardless. It's looking very promising! :D

japgolly:
@olafurpg Please take another look. I think this is much better now

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -1,87 +1,143 @@
...
+        def insideIfThen = parentIf.thenp eq b
+        def parentIfHasAnElse = parentIf.elsep.tokens.nonEmpty
+        def blockIsIfWithoutElse = b.stats.head match {
+          case childIf: Term.If => childIf.elsep.tokens.isEmpty
+          case _ => false
+        }
+        insideIfThen && parentIfHasAnElse && blockIsIfWithoutElse
+
+      case _ =>
+        // No other (known) special cases
```
olafurpg: When I think of it, I'm afraid special cases include whenever blocks are used to influence precedence

```diff
- {1 + 2} * 3
+ 1 + 2 * 3
- case x if { x match { case 1 => true } }
+ case x if x match { case 1 => true }
```

It's possible to detect these cases, but it requires a bit of machinery. 

* label all tree nodes into syntactic groups https://github.com/scalacenter/scala-syntax/blob/f7a6206aa95756855132d3de4e980ace6944a3c5/format/src/main/scala/org/scalafmt/internal/TreeSyntacticGroup.scala
* compare two syntactic groups https://github.com/scalacenter/scala-syntax/blob/f7a6206aa95756855132d3de4e980ace6944a3c5/format/src/main/scala/org/scalafmt/internal/TreeOps.scala#L58

olafurpg: I took the liberty to integrate those syntactic groups to ensure redundant braces does not remove blocks that are required to influence precedence and associativity.

olafurpg:
Sidenote, in Scalafmt v2 (https://github.com/scalameta/scalafmt/issues/917) I want to move to a tree printer where transformations like this should be possible to express as simply as
```scala
tree.transform {
  case Term.Block(stat :: Nil) => stat
}
```
with automatic handling of precedence and other tricky cases. This would help make the scalafix and scalafmt rewrite API so much nicer to deal with, we're working towards this in https://github.com/scalacenter/scala-syntax

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -1,87 +1,143 @@
...
+        exactlyOneStatement && blockSizeIsOk(b) && !retainSingleStatBlock(b)
     }
-    builder.result()
   }
+
+  /** Some blocks look redundant but aren't */
+  private def retainSingleStatBlock(b: Term.Block): Boolean =
+    b.parent.exists {
+      case parentIf: Term.If =>
+        // if (a) { if (b) c } else d
```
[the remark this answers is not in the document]
olafurpg: Turns out this was a bug in https://github.com/scalacenter/scala-syntax/issues/30 😄

japgolly:
Fantastic changes! I didn't think about op precedence at all and I'm very glad you had a solution up your sleeve.

I've added settings to allow the retention of the old behaviour. Back to you @olafurpg

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -1,87 +1,167 @@
...
+        def blockIsIfWithoutElse = b.stats.head match {
+          case childIf: Term.If => childIf.elsep.tokens.isEmpty
+          case _ => false
+        }
+        insideIfThen && parentIfHasAnElse && blockIsIfWithoutElse
+
+      case parent =>
+        val side = parent match {
+          case t: Term.ApplyInfix
+              if t.args.lengthCompare(1) == 0 && (t.args.head eq b) =>
```
japgolly: Note @olafurpg , I've changed this from `case Term.ApplyInfix(_, _, _, arg :: Nil) if arg eq b =>` to

1) be a little more runtime-efficient by bypassing `unappy`s
2) this now works as expected if args isn't a `List`

olafurpg:
This fix should already be available as a pre-release http://scalameta.org/scalafmt/#Pre-release I updated the milestone deadline for v1.5.0 to April 16th https://github.com/scalameta/scalafmt/milestones The IntelliJ plugin is a pain to publish so it's only updated for stable releases.

japgolly:
Awesome and thanks a lot for the help @olafurpg!

olafurpg:
v1.5 is out now! http://scalameta.org/scalafmt/#1.5.0

I ran the rewrite on the scalafmt repo and hit a bug where a non-redundant curly brace was removed. I opened a ticket https://github.com/scalameta/scalafmt/issues/1147 would you be interested in taking a look @japgolly ?

I suspect it might be caused by the SyntacticGroups implementation, so I'm happy to look into it as well.

PR #1132, opened 2018-04-11
Generalize with-chain wrapping for defs and type definitions

### LINKED ISSUE #1125: Long `type` assignments on multiple lines are collapsed to a single long line
We use long `type`s as a way of dependency injection similar to the cake pattern. We generally format the alias with one `with HasX` on each line. Some types might have dozens of dependencies so Scalafmt collapses this `type` assignment into a single very long line, potentially hundreds or thousands of characters long.

- **Version**: 1.4.0
- **Integration**: Commandline
- **Configuration**:
```
maxColumn = 100
```

## Steps

Given code like this:
```scala
object SomeJobService {
  type RequiredServices = Bag
    with HasClockService
    with HasDynamicConfig
    with HasFSLogger
    with HasFutureService
    with HasThrottles
    with HasThrowableNotifier
}
```

```bash
scalafmt SomeJobService.scala
```

## Problem

Scalafmt formats code like this (creates lines potentially much longer than 100 characters):
```scala
object SomeJobService {
  type RequiredServices =
    Bag with HasClockService with HasDynamicConfig with HasFSLogger with HasFutureService with HasThrottles with HasThrowableNotifier
}
```

## Expectation

```scala
object SomeJobService {
  type RequiredServices = Bag
    with HasClockService
    with HasDynamicConfig
    with HasFSLogger
    with HasFutureService
    with HasThrottles
    with HasThrowableNotifier
}
```
Well really I would expect it to be anything that doesn't violate the 100 maxColumn rule that is in place.

## Workaround

No workaround that I know of besides disabling the formatter for the line.

## Notes

Maybe this needs it's own configuration options to wrap it only at maxColumn, or to do one type per line like we have it.

iantabolt:
This fixes #1125 to handle wrapping of "with-chains".

For example, a long chain of

```scala
Foo with Bar with Baz
```

can be wrapped as

```scala
Foo
  with Bar
  with Baz
```

This was already implemented for self-types (`trait Xyz { self: Foo with Bar with Baz => `). This PR expands the behavior to handle type definitions (`type MyType = Foo with Bar with Baz`) and function return types (`def xyz: Foo with Bar with Baz`).

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -24,7 +24,6 @@ import org.scalafmt.internal.ExpiresOn.Right
 import org.scalafmt.internal.Length.Num
 import org.scalafmt.internal.Length.StateColumn
 import org.scalafmt.internal.Policy.NoPolicy
-import org.scalafmt.util.`:parent:`
 import org.scalafmt.util.Delim
 import org.scalafmt.util.CtorModifier
 import org.scalafmt.util.InfixApplication
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -33,7 +33,7 @@ import org.scalafmt.util.Literal
 import org.scalafmt.util.LoggerOps
 import org.scalafmt.util.Modifier
 import org.scalafmt.util.RightParenOrBracket
-import org.scalafmt.util.SelfAnnotation
+import org.scalafmt.util.WithChain
 import org.scalafmt.util.SomeInterpolate
 import org.scalafmt.util.TokenOps
 import org.scalafmt.util.TreeOps
@@ -1021,7 +1021,7 @@ class Router(formatOps: FormatOps) {
               Split(Space, 0),
               Split(Newline, 1).withPolicy(policy)
             )
-          case t @ SelfAnnotation(top) =>
+          case t @ WithChain(top) =>
             val isFirstWith = !t.lhs.is[Type.With]
             if (isFirstWith) {
               val chain = withChain(top)
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeExtractors.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeExtractors.scala
@@ -1,6 +1,7 @@
 package org.scalafmt.util
 
 import scala.collection.immutable.Seq
+import scala.meta.Defn
 import scala.meta.Name
 import scala.meta.Pat
 import scala.meta.Template
@@ -41,6 +42,8 @@ object SelfAnnotation {
   def unapply(t: Type.With): Option[Type.With] =
     TreeOps.topTypeWith(t) match {
       case (top: Type.With) `:parent:` (_: Term.Param) `:parent:` (_: Template) =>
+        Some(top)
+      case (top: Type.With) `:parent:` (_: Defn.Type) =>
         Some(top)
       case _ => None
     }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeExtractors.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeExtractors.scala
@@ -1,6 +1,7 @@
 package org.scalafmt.util
 
 import scala.collection.immutable.Seq
+import scala.meta.Decl
 import scala.meta.Defn
 import scala.meta.Name
 import scala.meta.Pat
@@ -38,12 +39,17 @@ object `:parent:` {
   }
 }
 
-object SelfAnnotation {
+object WithChain {
   def unapply(t: Type.With): Option[Type.With] =
     TreeOps.topTypeWith(t) match {
+      // self types
       case (top: Type.With) `:parent:` (_: Term.Param) `:parent:` (_: Template) =>
         Some(top)
+      // type definitions
       case (top: Type.With) `:parent:` (_: Defn.Type) =>
+        Some(top)
+      // function return types
+      case (top: Type.With) `:parent:` (_: Decl.Def) =>
         Some(top)
       case _ => None
     }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeExtractors.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeExtractors.scala
@@ -45,11 +45,8 @@ object WithChain {
       // self types
       case (top: Type.With) `:parent:` (_: Term.Param) `:parent:` (_: Template) =>
         Some(top)
-      // type definitions
-      case (top: Type.With) `:parent:` (_: Defn.Type) =>
-        Some(top)
-      // function return types
-      case (top: Type.With) `:parent:` (_: Decl.Def) =>
+      // val/def/var/type definitions or declarations
+      case (top: Type.With) `:parent:` (_: Defn | _: Decl) =>
         Some(top)
       case _ => None
     }
--- a/scalafmt-tests/src/test/resources/default/TypeWith.stat
+++ b/scalafmt-tests/src/test/resources/default/TypeWith.stat
@@ -9,3 +9,13 @@
       annotationOwnerLookUp: ScLiteral => Option[
           PsiAnnotationOwner with PsiElement]): Option[PsiAnnotationOwner] = 1
 }
+<<< #1133
+implicit val catsStdInstancesForOption
+   : Traverse[Option] with MonadError[Option, Unit] with Alternative[Option]
+       with CommutativeMonad[Option] with CoflatMap[Option] = ???
+>>>
+implicit val catsStdInstancesForOption: Traverse[Option]
+  with MonadError[Option, Unit]
+  with Alternative[Option]
+  with CommutativeMonad[Option]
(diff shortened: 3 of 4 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1021,7 +1021,7 @@ class Router(formatOps: FormatOps) {
               Split(Space, 0),
               Split(Newline, 1).withPolicy(policy)
             )
-          case t @ SelfAnnotation(top) =>
+          case t @ WithChain(top) =>
```
iantabolt: I don't love this name but I wanted something more general than `SelfAnnotation`

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeExtractors.scala
```diff
@@ -37,11 +39,18 @@ object `:parent:` {
...
 
-object SelfAnnotation {
+object WithChain {
   def unapply(t: Type.With): Option[Type.With] =
     TreeOps.topTypeWith(t) match {
+      // self types
       case (top: Type.With) `:parent:` (_: Term.Param) `:parent:` (_: Template) =>
         Some(top)
+      // type definitions
+      case (top: Type.With) `:parent:` (_: Defn.Type) =>
```
olafurpg: I suspect this pattern can be extended to any Defn or Decl, that will cover val/var/def regardless if they're abstract or not.

olafurpg: I'm not sure when we don't want this to match for Type.With, there are other tree nodes like Term.ApplyType where this is equally relevant.

iantabolt: Yeah I agree for the most part. 

If I open it up to match all `Type.With`s then one test fails (with good reason IMO):

```scala
// Before:
{ def getAnnotationOwner(
       annotationOwnerLookUp: ScLiteral => Option[
         PsiAnnotationOwner with PsiElement]): Option[PsiAnnotationOwner] = 1 }

// After -- breaks in generic type:
{
  def getAnnotationOwner(
      annotationOwnerLookUp: ScLiteral => Option[PsiAnnotationOwner
        with PsiElement]): Option[PsiAnnotationOwner] = 1
}

// Expected:
{
  def getAnnotationOwner(
      annotationOwnerLookUp: ScLiteral => Option[
          PsiAnnotationOwner with PsiElement]): Option[PsiAnnotationOwner] = 1
}
```

From https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/resources/default/TypeWith.stat

iantabolt: But even in that scenario, if the type were long enough we probably want a newline instead of violating the column limit. Maybe we can make this split a low priority or high cost etc?

ceedubs:
This looks like a great change! I'm adding some examples here based on @olafurpg's request in #1133.

In Cats when we have code like:

```scala
  implicit val catsDataInstancesForNonEmptyList: SemigroupK[NonEmptyList]
    with Reducible[NonEmptyList] with Bimonad[NonEmptyList] with NonEmptyTraverse[NonEmptyList] = ...

  implicit val catsDataInstancesForNonEmptyVector: SemigroupK[NonEmptyVector]
    with Reducible[NonEmptyVector] with Bimonad[NonEmptyVector]
    with NonEmptyTraverse[NonEmptyVector] = ...

   implicit val catsStdInstancesForOption
    : Traverse[Option] with MonadError[Option, Unit] with Alternative[Option]
      with CommutativeMonad[Option] with CoflatMap[Option] = ...
```

After formatting it ends up looking like:

```scala
  implicit val catsDataInstancesForNonEmptyList: SemigroupK[NonEmptyList] with Reducible[
    NonEmptyList] with Bimonad[NonEmptyList] with NonEmptyTraverse[NonEmptyList] = ...

  implicit val catsDataInstancesForNonEmptyVector: SemigroupK[NonEmptyVector] with Reducible[
    NonEmptyVector] with Bimonad[NonEmptyVector] with NonEmptyTraverse[NonEmptyVector] = ...

   implicit val catsStdInstancesForOption
    : Traverse[Option] with MonadError[Option, Unit] with Alternative[Option] with CommutativeMonad[
      Option] with CoflatMap[Option] = ...
```

It would be nice to see the style changes exhibited in this PR to extend to the return types of `val`s like these.

iantabolt:
@ceedubs Thanks for the input! I'll add these as test cases. 

I ran the examples through with this change and it ended up with

```scala
implicit val catsDataInstancesForNonEmptyList: SemigroupK[NonEmptyList]
  with Reducible[NonEmptyList]
  with Bimonad[NonEmptyList]
  with NonEmptyTraverse[NonEmptyList] = ???

implicit val catsDataInstancesForNonEmptyVector: SemigroupK[NonEmptyVector]
  with Reducible[NonEmptyVector]
  with Bimonad[NonEmptyVector]
  with NonEmptyTraverse[NonEmptyVector] = ???

implicit val catsStdInstancesForOption: Traverse[Option] with MonadError[Option, Unit]
  with Alternative[Option]
  with CommutativeMonad[Option]
  with CoflatMap[Option] = ???
```

Does that seem about right? I'm not sure why the last one didn't drop `MonadError` to the next line 🤔

ceedubs:
@iantabolt nice! On the last one, maybe it has to do with inlining until you hit the column limit? I'm not super familiar with the particulars of behavior. And does it depend on `binPack.parentConstructors`? Either way, I think that it's a really big improvement and I'd be happy with it.

scalafmt-tests/src/test/resources/unit/Type.stat
```diff
@@ -85,3 +85,29 @@ type Row =
...
+    with HasDynamicConfig with HasThrottles
+    with HasThrowableNotifier
+>>>
+type RequiredServices = Bag
+  with SomeOtherService.RequiredServices
+  with HasDynamicConfig
+  with HasThrottles
+  with HasThrowableNotifier
+<<< abstract def #1125
+trait NeedsServices {
```
iantabolt: This probably shouldn't be in `Type.stat` since there is no `type`

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
olafurpg: This name is good :+1:

scalafmt-tests/src/test/resources/default/TypeWith.stat
```diff
@@ -9,3 +9,13 @@
       annotationOwnerLookUp: ScLiteral => Option[
           PsiAnnotationOwner with PsiElement]): Option[PsiAnnotationOwner] = 1
 }
+<<< #1133
+implicit val catsStdInstancesForOption
+   : Traverse[Option] with MonadError[Option, Unit] with Alternative[Option]
+       with CommutativeMonad[Option] with CoflatMap[Option] = ???
+>>>
+implicit val catsStdInstancesForOption: Traverse[Option]
```
olafurpg: This may be more ambitious that you were planning for (so "no" is an acceptable answer), but while were fixing with chains, could we add a test case for?
```scala
new A with B with C with D ...
```
I've seen surprising behavior in these scenarios and I suspect fixing it likewise is a matter of adding another case to `WithChain`

scalafmt-tests/src/test/resources/unit/Type.stat
olafurpg: `TypeWith.stat` would be good, this is important enough to merit a separate suite.

scalafmt-tests/src/test/resources/default/TypeWith.stat
iantabolt: Yes no problem (I think)

iantabolt: I might need some help with this. I don't think this falls into the `Type.With` case and instead hits the `Template` case.

My thought is to treat this like the `KwExtends` case above and do

```patch
       case tok @ FormatToken(_, right @ KwWith(), _) =>
         rightOwner match {
+          // Anonymous template (new A with B { ... })
+          case template: Template if template.parent.exists(_.is[Term.New]) =>
+            val lastToken = template.tokens.last
+            binPackParentConstructorSplits(
+              Set(template),
+              lastToken,
+              2
+            )
           case template: Template =>
             val hasSelfAnnotation = template.self.tokens.nonEmpty
```

This almost works but it indents after every with

```
new A
  with B
    with C
      with D
```

Any ideas?

iantabolt: I pushed the non-working change mentioned above. I'm assuming the fix is something analogous to the `isFirstWith` logic but I'm not sure how to achieve it with the template

olafurpg: Hmm, the changes seem unrelated enough to merit a separate PR. Your change already improves on the state-of-the-art so I propose you drop the last commit, we merge and move this case to another ticket.

I'll need to locally play around with this case to see how would be best to solve it.

iantabolt: SGTM 👍. I rolled it back

olafurpg:
I'll be attending a conference for the next week, but I hope to have v1.5.0 out this April. Until then, you can already use the pre-releases to Bintray as documented on the website but note those don't include the IntelliJ plugin.

tian000:
What configuration setting do you use to wrap with-chains?

PR #1136, opened 2018-04-15
sbt plugin formatting caching

vovapolu:
This PR adds caching for formatting and for config files.


scalafmt-core/shared/src/main/scala/org/scalafmt/util/FormattingCache.scala
```diff
@@ -0,0 +1,15 @@
+package org.scalafmt.util
+
+import java.io.File
+
+import scala.collection.mutable
+
+object FormattingCache {
+  private val lastFormatting = mutable.Map.empty[File, Long]
+
+  def outdatedFormatting(file: File): Boolean =
```
olafurpg:
```
object Caching {
  def format[T](file: File)(f: () => Unit): Unit = {
    if (isOutdated(file)) {
      f()
      cache(file) = file.lastModified
    }
  }
  import java.io.File
  private val cache = mutable.Map.empty[File, Long]
  private def isOutdated(file: File): Boolean=
    cache.get(file).forall(_ < file.lastModified)
}
```

PR #1140, opened 2018-04-19
Implement SortModifiers rewrite rule

### LINKED ISSUE #1138: Add rewrite rule for sorting modifiers
EDIT:
See final behavior in the corresponding PR: https://github.com/scalameta/scalafmt/pull/1140

This is a proposal for introducing a new rewrite rule `SortPropertyModifiers` (working title)

Config-wise it would look:
```
rewrite {
  rules = [SortPropertyModifiers]
  sortPropertyModifiers {
    # "access" stands for private/protected with their scoped variants
    order = ["access", "abstract", "final", "implicit", "override", "lazy"]
  }
}
```
And the output is fairly self-explanatory. When dealing with `val`s (also when they show up in class constructors), `def`s, all modifiers get sorted.

For initial code of:
```
object test {
  private implicit final def foo: Int = ???
}
```
you get the following given the above rules:
```
object test {
  private final implicit def foo: Int = ???
}
```

It was decided that the rule apply to class, trait, and object definitions as well
~One question remains though:
Should sorting of modifiers on `class`, `object`, `trait` be included in this rule? Or should it be a separate one? Personally I do not feel strongly one way or the other.~

~A WIP implementation can be found on this branch:
https://github.com/lorandszakacs/scalafmt/tree/feature/organize_field_modifiers~

lorandszakacs:
Closes #1138 
Later edit:
I decided to add docs even before feedback.

TODO:
- [x] If the implementation/behavior/defaults are deemed OK, then I will add the appropriate documentation as well

This rewrite rule will sort the modifiers applied to the following language constructs:
- Defn.Def
- Defn.Val
- Defn.Var
- Defn.Type
- Defn.Class
- Defn.Object
- Defn.Trait
- Term.Param

Example config:
```
rewrite {
  rules = [SortModifiers]
  sortModifiers {
    # also the default value, the user has to specify all 8. The message if they fail
    # to do so is very instructive, and the docs will be clear, so hopefully it's user friendly
    order = ["implicit", "final", "sealed", "abstract", "override", "private", "protected", "lazy"]
  }
}
```

Later edit — example with config:
```
maxColumn = 200
rewrite {
  rules = [SortModifiers]
}
```

Before:
```
package test

sealed private[test] trait Test {
  protected def name: String
}

final private[test] object Tests {

  private final type T1 = Int
  final private type T2 = String

  final private implicit case object Test1 extends Test { def name: String = "Test1" }
  implicit private final case object Test2 extends Test { def name: String = "Test2" }
  final implicit private case object Test3 extends Test { def name: String = "Test3" }

  sealed abstract class TestSup(protected override val name: String) extends Test

  final class Test4(implicit final private val impl: String) extends Test {
    final protected implicit override lazy val name: String = impl
  }
}
```

After:
```
package test

sealed private[test] trait Test {
  protected def name: String
}

final private[test] object Tests {

  final private type T1 = Int
  final private type T2 = String

  implicit final private case object Test1 extends Test { def name: String = "Test1" }
  implicit final private case object Test2 extends Test { def name: String = "Test2" }
  implicit final private case object Test3 extends Test { def name: String = "Test3" }

  sealed abstract class TestSup(override protected val name: String) extends Test

  final class Test4(implicit final private val impl: String) extends Test {
    implicit final override protected lazy val name: String = impl
  }
}
```
-----
N.B:
There is some weird behavior that shows up when formatting `SortModifiers_Mod_With_No_Token.source`. The tokens of a Mod are empty (which makes very little sense to me). This case is explicitly handled and no patches are applied to the code. And lo' and behold everything formats as expected anyway. Very, very weird.

Supporting changes:
 - added a method `org.scalafmt.config.ReaderUtil.oneOfIgnoreBackticks`
   to easily define a reader for the config that ignores backticks in
   case object names. So I factored out the implementation of `oneOf`

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ReaderUtil.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ReaderUtil.scala
@@ -7,16 +7,29 @@ import metaconfig.Configured._
 
 object ReaderUtil {
   // Poor mans coproduct reader
-  def oneOf[T: ClassTag](options: sourcecode.Text[T]*): ConfDecoder[T] = {
-    val m = options.map(x => x.source.toLowerCase() -> x.value).toMap
+  def oneOf[T: ClassTag](options: sourcecode.Text[T]*): ConfDecoder[T] =
+    oneOfImpl(lowerCase, options)
+
+  def oneOfIgnoreBackticks[T: ClassTag](
+      options: sourcecode.Text[T]*): ConfDecoder[T] =
+    oneOfImpl(lowerCaseNoBackticks, options)
+
+  private val lowerCase: String => String = s => s.toLowerCase
+  private val lowerCaseNoBackticks: String => String = s =>
+    s.toLowerCase().replace("`", "")
+
+  private def oneOfImpl[T: ClassTag](
+      sanitize: String => String,
+      options: Seq[sourcecode.Text[T]]): ConfDecoder[T] = {
+    val m = options.map(x => sanitize(x.source) -> x.value).toMap
     ConfDecoder.instance[T] {
       case Conf.Str(x) =>
-        m.get(x.toLowerCase()) match {
+        m.get(sanitize(x)) match {
           case Some(y) =>
             Ok(y)
           case None =>
-            val available = options.map(_.source).mkString(", ")
-            val msg = s"Unknown input '$x'. Expected one of $available"
+            val available = m.keys.mkString(", ")
+            val msg = s"Unknown input '$x'. Expected one of: $available"
             ConfError.msg(msg).notOk
         }
     }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ReaderUtil.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ReaderUtil.scala
@@ -14,8 +14,8 @@ object ReaderUtil {
       options: sourcecode.Text[T]*): ConfDecoder[T] =
     oneOfImpl(lowerCaseNoBackticks, options)
 
-  private val lowerCase: String => String = s => s.toLowerCase
-  private val lowerCaseNoBackticks: String => String = s =>
+  private def lowerCase(s: String): String = s.toLowerCase
+  private def lowerCaseNoBackticks(s: String): String =
     s.toLowerCase().replace("`", "")
 
   private def oneOfImpl[T: ClassTag](
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/RewriteSettings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/RewriteSettings.scala
@@ -9,7 +9,8 @@ case class RewriteSettings(
     rules: Seq[Rewrite] = Nil,
     @Recurse redundantBraces: RedundantBracesSettings =
       RedundantBracesSettings(),
-    @Recurse neverInfix: Pattern = Pattern.neverInfix
+    @Recurse neverInfix: Pattern = Pattern.neverInfix,
+    @Recurse sortModifiers: SortSettings = SortSettings.default
 ) {
   Rewrite.validateRewrites(rules) match {
     case Nil => // OK
@@ -20,5 +21,13 @@ case class RewriteSettings(
         )
       )
   }
+
+  if (sortModifiers.order.distinct.length != 8)
+    throw InvalidScalafmtConfiguration(
+      new IllegalArgumentException(
+        "'sortModifiers.order', if specified, it has to contain all of the following values in the order you wish them sorted:" +
+          """["private", "protected" , "abstract", "final", "sealed", "implicit", "override", "lazy"]"""
+      )
+    )
 
 }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
@@ -18,7 +18,7 @@ object SortSettings {
       `override`,
       `private`,
       `protected`,
-      `lazy`,
+      `lazy`
     )
 
   val defaultOrder: Vector[ModKey] =
@@ -34,7 +34,7 @@ object SortSettings {
       `private`,
       `protected`,
       //
-      `lazy`,
+      `lazy`
     )
 
   def default: SortSettings =
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
@@ -1,0 +1,53 @@
+package org.scalafmt.config
+
+import metaconfig._
+
+@DeriveConfDecoder
+case class SortSettings(
+    order: Vector[SortSettings.ModKey]
+)
+
+object SortSettings {
+
+  implicit val SortSettingsModKeyReader: ConfDecoder[ModKey] =
+    ReaderUtil.oneOfIgnoreBackticks[ModKey](
+      `implicit`,
+      `final`,
+      `sealed`,
+      `abstract`,
+      `override`,
+      `private`,
+      `protected`,
+      `lazy`,
+    )
+
+  val defaultOrder: Vector[ModKey] =
+    Vector(
+      `implicit`,
+      //
+      `final`,
+      `sealed`,
+      `abstract`,
+      //
+      `override`,
+      //
+      `private`,
+      `protected`,
+      //
+      `lazy`,
+    )
+
+  def default: SortSettings =
+    SortSettings(defaultOrder)
+
+  sealed trait ModKey
+
+  case object `private` extends ModKey
+  case object `protected` extends ModKey
+  case object `final` extends ModKey
+  case object `sealed` extends ModKey
+  case object `abstract` extends ModKey
+  case object `implicit` extends ModKey
+  case object `override` extends ModKey
+  case object `lazy` extends ModKey
+}
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
@@ -40,7 +40,7 @@ object SortSettings {
   def default: SortSettings =
     SortSettings(defaultOrder)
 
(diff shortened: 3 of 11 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * The case of parameters is a bit more curious, because there the
+     * "val" or "var" in, say:
+     * {{{
+     *   class Test(private final val x: Int)
+     * }}}
+     * are considered Mods, instead of being similar to `Defn.Val`, or `Defn.Var`.
+     */
+    val patchesOfPatches = code.collect {
+      case d: Defn.Def => patchMods(sortMods, d.mods)
```
olafurpg: Missing `Decl.{Def,Val,Var}`, those are the abstract variants (declarations)

```scala
implicit val x: Int
def x: Int
var x: Int
```

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
```diff
@@ -0,0 +1,92 @@
...
+  private def patchMods(
+      sortMods: Seq[Mod] => Seq[Mod],
+      oldMods: Seq[Mod]): Seq[Patch] = {
+    if (oldMods.isEmpty) Nil
+    else {
+      val sortedMods: Seq[Mod] = sortMods(oldMods)
+      sortedMods.zip(oldMods).flatMap {
+        case (next, old) =>
+          if (old.tokens.isEmpty) {
+            //Not sure why this happens, how can you have a mod with no tokens?
```
olafurpg: Yes, there is a bug for example for `def foo(implicit x: Int)`, the `implicit` has no tokens. The user can't do anything about it so it's fine to ignore it.

olafurpg: Please remove the comment, this guard is totally fine.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
```diff
@@ -0,0 +1,53 @@
...
+      `private`,
+      `protected`,
+      //
+      `lazy`
+    )
+
+  def default: SortSettings =
+    SortSettings(defaultOrder)
+
+  sealed trait ModKey
```
olafurpg: `extends Product` allows you to use `.productPrefix` from the docs instead of the `getClass` hack

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
```diff
@@ -0,0 +1,92 @@
...
+      case t: Defn.Trait => patchMods(sortMods, t.mods)
+      case p: Term.Param => patchMods(sortMods, p.mods.filterNot(isValOrVar))
+    }
+    patchesOfPatches.flatten
+  }
+
+  private val isValOrVar: Mod => Boolean = m =>
+    m.is[Mod.ValParam] || m.is[Mod.VarParam]
+
+  private val isCase: Mod => Boolean = m => m.is[Mod.Case]
```
olafurpg: Inline, it's clearer to read `filterNot(_.is[Mod.Case])`

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
```diff
@@ -0,0 +1,92 @@
...
+      case c: Defn.Class => patchMods(sortMods, c.mods.filterNot(isCase))
+      case o: Defn.Object => patchMods(sortMods, o.mods.filterNot(isCase))
+      case t: Defn.Trait => patchMods(sortMods, t.mods)
+      case p: Term.Param => patchMods(sortMods, p.mods.filterNot(isValOrVar))
+    }
+    patchesOfPatches.flatten
+  }
+
+  private val isValOrVar: Mod => Boolean = m =>
+    m.is[Mod.ValParam] || m.is[Mod.VarParam]
```
olafurpg: Ditto, better to inline.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
```diff
@@ -0,0 +1,92 @@
...
+
+import scala.meta.Tree
+import scala.meta._
+
+object SortModifiers extends Rewrite {
+
+  override def rewrite(code: Tree, ctx: RewriteCtx): Seq[Patch] = {
+    val order = ctx.style.rewrite.sortModifiers.order
+
+    val sortMods: Seq[Mod] => Seq[Mod] = { mods =>
```
olafurpg: It's clearer to use methods

```scala
def sortMods(mods: Seq[Mod]): Seq[Mod] = mods.sortWith(orderModsBy(order))
```

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
```diff
@@ -0,0 +1,92 @@
...
+    patchesOfPatches.flatten
+  }
+
+  private val isValOrVar: Mod => Boolean = m =>
+    m.is[Mod.ValParam] || m.is[Mod.VarParam]
+
+  private val isCase: Mod => Boolean = m => m.is[Mod.Case]
+
+  private def patchMods(
+      sortMods: Seq[Mod] => Seq[Mod],
```
olafurpg: Why pass in the function as a parameter  instead of `sortedMods: Seq[Mod]`? Alternatively, it will always be `mods.sortWith(orderModsBy(order))` so that can be inlined at line 51 below

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
lorandszakacs: @olafurpg I did not inline because I didn't want to leave it to JIT to not do extra allocations for this function (if it can even do that).

lorandszakacs:
@olafurpg. Thanks for the comments. I'll get on these asap.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
olafurpg: I see, I  would be surprised if this was a performance issue. It is better to optimize for readability.

lorandszakacs:
@olafurpg applied all suggestions, and added the missing cases 😃

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ReaderUtil.scala
```diff
@@ -7,16 +7,29 @@ import metaconfig.Configured._
...
-  def oneOf[T: ClassTag](options: sourcecode.Text[T]*): ConfDecoder[T] = {
-    val m = options.map(x => x.source.toLowerCase() -> x.value).toMap
+  def oneOf[T: ClassTag](options: sourcecode.Text[T]*): ConfDecoder[T] =
+    oneOfImpl(lowerCase, options)
+
+  def oneOfIgnoreBackticks[T: ClassTag](
+      options: sourcecode.Text[T]*): ConfDecoder[T] =
+    oneOfImpl(lowerCaseNoBackticks, options)
+
+  private val lowerCase: String => String = s => s.toLowerCase
```
olafurpg: Why not `def lowerCase(s: String): String = s.toLowercase`?

There are many reasons to use methods instead of functions. Function declarations are harder to read, produce more cryptic stack traces, `val` in objects are evaluated during global initialization potentially causing NullPointerException, `val` in object creates a JVM field + method while methods creates only one JVM method.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/RewriteSettings.scala
```diff
@@ -21,4 +22,12 @@ case class RewriteSettings(
       )
   }
 
+  if (sortModifiers.order.distinct.length != 8)
```
olafurpg: It's better to write a decoder by hand in the `SortSettings` companion, the values are encoded from `Conf.Lst(values)` where values should be only `Conf.Str`.

The `Rewrite.validateRewrites(rules)` validation above is a hack that should not be repeated, I'm cleaning up configuration in #1145.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
```diff
@@ -0,0 +1,53 @@
+package org.scalafmt.config
+
+import metaconfig._
+
+@DeriveConfDecoder
```
olafurpg: Best to replace this with a hand-written decoder that validates that all modifiers are defined, and produces a helpful error message saying which modifier is missing in case something is wrong.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/RewriteSettings.scala
olafurpg: Actually, this is fine, I can take care of it in #1145 once this PR is merged :)

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
olafurpg: Nevermind this comment.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
```diff
@@ -0,0 +1,53 @@
...
+      `final`,
+      `sealed`,
+      `abstract`,
+      `override`,
+      `private`,
+      `protected`,
+      `lazy`,
+    )
+
+  val defaultOrder: Vector[ModKey] =
```
kitbellew: is there any reason why the default order doesn't follow standard scala style:

https://docs.scala-lang.org/style/declarations.html#modifiers

Specifically, `final` should precede `lazy`.

olafurpg: I have no strong preference, defaulting to the scala style guide sounds good 👍

Issue #1142, opened 2018-04-20
Force verticalMultilineAtDefinitionSite for specific arity threshold


mads-hartmann:
I'd like to be able to configure scalafmt to always enforce multiline formatting for definitions that have a specific arity threshold. Preferable I'd like to distinguish the arity based on the entity that's being defined. I imagine having different values for case class and functions for example.

So a config like the following

```
verticalMultilineAtDefinitionSite = true
verticalMultilineAtDefinitionSite.arityThreshold = 2
```

Should provided with the following code

```scala
final case class Foo(x: String)
final case class Bar(x: String, y: String)
```

Produce this

```scala
final case class Foo(x: String)
final case class Bar(
  x: String,
  y: String
)
```

If this is already doable with the current configuration I apologise, I just couldn't find it. If not I'd be more than happy to help implement.

Cheers

olafurpg:
Thanks for reporting! Would you be open for this addition @pjrt ?

I suspect the relevant place to fix this is here https://github.com/scalameta/scalafmt/blob/433a73b3a5d3dbd85598aa520f71c68c8c7b1d5d/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala#L827 

the single line Split is missing something like `Split(NoSplit, 0, ignoreIf = arity >= config.maxArity)`

mads-hartmann:
I'd be more than happy to give it a go but it won't be before tonight, so if that messes with your release plans then wait for me :)

pjrt:
> configure scalafmt to always enforce multiline formatting for definitions

So that means it will trigger even if the case class (or function) is under maxcolumn? I'm guessing if it goes over maxcolumn the normal behavior is triggered, correct?

mads-hartmann:
@pjrt Definitely. Given the following configuration

```
verticalMultilineAtDefinitionSite = true
verticalMultilineAtDefinitionSite.arityThreshold = 2
```

This

```scala
final case class Foo(x: String)
final case class Bar(x: String, y: String)
case class VeryLongNames(xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx: String)
```

Should result in 
```scala
final case class Foo(x: String)
final case class Bar(
  x: String,
  y: String
)
case class VeryLongNames(
  xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx: String
)
```

[...]

PR #1143, opened 2018-04-21
Arity threshold

mads-hartmann:
This is a proof of concept for #1142. I'm opening a PR so I can ask questions about the code 😄

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
@@ -115,6 +115,9 @@ import org.scalafmt.util.ValidationOps
   *                                          [[maxColumn]]. If a multi-line funcion can fit in a single line, it will
   *                                          make it so. Note that this setting ignores continuation.defnSite,
   *                                          [[binPack.unsafeDefnSite]], and [[align.openParenDefnSite]].
+  * @param verticalMultilineAtDefinitionSiteArityThreshold If set, this will trigger a vertical multi-line formatting as
+  *                                                        described above even thoug the definition falls below the
+  *                                                        [[maxColumn]] width.
   */
 @DeriveConfDecoder
 case class ScalafmtConfig(
@@ -142,6 +145,7 @@ case class ScalafmtConfig(
     danglingParentheses: Boolean = false,
     poorMansTrailingCommasInConfigStyle: Boolean = false,
     verticalMultilineAtDefinitionSite: Boolean = false,
+    verticalMultilineAtDefinitionSiteArityThreshold: Int = 100,
     onTestFailure: String = "",
     encoding: Codec = "UTF-8",
     @Recurse project: ProjectFiles = ProjectFiles()
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
@@ -109,14 +109,14 @@ import org.scalafmt.util.ValidationOps
   *                                           )
   *                                          }}}
   *
-  *                                          All parameters are on their on line indented by four (4), seperation between
-  *                                          parament groups are indented by two (2). ReturnType is on its own line at
-  *                                          then end. This will only trigger if the function would go over
-  *                                          [[maxColumn]]. If a multi-line funcion can fit in a single line, it will
+  *                                          All parameters are on their own line indented by four (4), separation between
+  *                                          parameter groups are indented by two (2). ReturnType is on its own line at
+  *                                          the end. This will only be triggered if the function would go over
+  *                                          [[maxColumn]]. If a multi-line function can fit in a single line, it will
   *                                          make it so. Note that this setting ignores continuation.defnSite,
   *                                          [[binPack.unsafeDefnSite]], and [[align.openParenDefnSite]].
   * @param verticalMultilineAtDefinitionSiteArityThreshold If set, this will trigger a vertical multi-line formatting as
-  *                                                        described above even thoug the definition falls below the
+  *                                                        described above even though the definition falls below the
   *                                                        [[maxColumn]] width.
   */
 @DeriveConfDecoder
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -4,6 +4,7 @@ import scala.annotation.tailrec
 import scala.collection.mutable
 import scala.meta.Case
 import scala.meta.Ctor
+import scala.meta.Decl
 import scala.meta.Defn
 import scala.meta.Import
 import scala.meta.Name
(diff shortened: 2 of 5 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -142,6 +145,7 @@ case class ScalafmtConfig(
     danglingParentheses: Boolean = false,
     poorMansTrailingCommasInConfigStyle: Boolean = false,
     verticalMultilineAtDefinitionSite: Boolean = false,
+    verticalMultilineAtDefinitionSiteArityThreshold: Int = 100,
```
mads-hartmann: I really wanted to make this an Option but there isn't a `ConfDecoder[Option[A]]` defined. Does it make sense to make it an option?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -823,8 +821,16 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
       if (isBracket) close // If we can fit the type params, make it so
       else lastParen // If we can fit all in one block, make it so
 
+    val arity = valueParamsOwner match {
```
mads-hartmann: I'm cheating here. I'd prefer to make this decision for each parameter group (in the case of functions) but I couldn't figure out how.

scalafmt-tests/src/test/resources/test/arityThreshold.source
```diff
@@ -0,0 +1,16 @@
...
+case class Foo(x: String)
+case class Bar(x: String, y: String)
+case class VeryLongNames(xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx: String)
+>>>
+case class Foo(x: String)
+case class Bar(
+  x: String,
+  y: String)
+case class VeryLongNames(
+  xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx: String)
```
mads-hartmann: Dangling parentheses is explicitly ignored for `Ctor.Primary`, `Defn.Class`, and `Trait`. I'd prefer to have them dangling in these cases but that seems like a separate issue :)

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
olafurpg: I can't remember why I never implemented a decoder for Option[T], there was some reason. I agree dummy Int values are not great, but it's consistent with some other settings so it's fine here

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -823,8 +821,16 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
...
+      case d: Decl.Def => d.paramss.foldLeft(0)(_ + _.size)
```
olafurpg: Missing `Defn.Def`, `Defn.Macro`, `Ctor.Secondary` and `Init`

Scalameta trees are quite precise, which is sometimes annoying but other times a blessing

olafurpg: Also, should this rule affect call-site? In that case you need to extract from `Term.Apply`.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
olafurpg: Hmm, not sure I follow. Can you elaborate?

mads-hartmann: Hah, my bad. I'm more confused than I though it seems. Given this code I would have assumed that the following would've triggered a reformat:

```scala
object A {
  def hello(how: String)(are: String)(you: String) = how + are + you
}
```

as it would hit `case d: Decl.Def` and `d.paramss.foldLeft(0)(_ + _.size)` would return `3`.

I extended the test and it does the right thing, I'm just not sure why.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
mads-hartmann: No IMO it shouldn't affect the call-site

mads-hartmann: I can't find `Init` anywhere? :)

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
mads-hartmann: If I add `Defn.Def` to the match case it does indeed end up formatting it as 

```scala
object A {
  def hello(
    how: String
  )(are: String
  )(you: String
  ) = how + are + you
}
```

Which isn't quite what I'd want 😅

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
mads-hartmann: Alright, I'll leave this as it for now then 👍

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
mads-hartmann: I came up with a solution in 4a034ce

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -823,8 +821,19 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
       if (isBracket) close // If we can fit the type params, make it so
       else lastParen // If we can fit all in one block, make it so
 
+    val maxArity = valueParamsOwner match {
+      case d: Decl.Def if d.paramss.nonEmpty => d.paramss.map(_.size).max
```
olafurpg: `Init` came in scalameta v2 but scalafmt is still on v1,

Seems in v1 `Init` is nested `Term.Apply`

```scala
@ q"@a(1)(2) def foo = 1"
res1: Defn.Def = Defn.Def(List(Mod.Annot(Term.Apply(Term.Apply(Ctor.Ref.Name("a"), List(Lit.Int(1))), List(Lit.Int(2))))), Term.Name("foo"), List(), List(), None, Lit.Int(1))
```

It's fine to skip those as they only appear in curried annotations, which is quite rare.

scalafmt-tests/src/test/resources/test/arityThreshold.source
olafurpg: Hmm, is this a peculiarity of `verticalMultilineAtDefinitionSite`? If so I have no opinions, but @pjrt might!

mads-hartmann:
@olafurpg Thanks a lot for reviewing the code 👍

pjrt:
Look great! Good job. Glad to see I'm not the only one using this feature :)

olafurpg:
Great :shipit: 

We'll probably wait with the release until #1140 is in as well.

scalafmt-tests/src/test/resources/test/arityThreshold.source
pjrt: @mads-hartmann Yeah, the feature ignores dangling (and most other options). It is one of those features that don't make sense if it listens to every other scalafmt option. It is a holistic style, instead of a piece-meal one (like dangling or align).

As for why, well it looks ugly if you allow dangling. So it only dangles when the class `extends` something.

PR #1145, opened 2018-04-22
Upgrade to the latest metaconfig

olafurpg:
This PR removes the dependency on scalameta/paradise macro annotations and upgrades to the latest metaconfig version (https://olafurpg.github.io/metaconfig/). This is the first step towards having configuration work the same across JVM, JS and Native.

Most of the diff is replacing the `@GenerateConfDecoder` macro annotation with the semi-automatic `metaconfig.generic.{deriveDecoder,deriveSurface}` def macros. Amazingly, 1031 out of 1032 tests passed on first compile 💪 

The surface macro is nice since it allows automatic generation of all configuration settings with explanation messages, default values and types. Example how we use it in scalafix https://scalacenter.github.io/scalafix/docs/rules/DisableSyntax

The only regression I can't avoid is the "Other" section on the website that did a best-effort to pretty-print all available configuration options as HOCON. The website links to the `ScalafmtConfig` sources instead now.


olafurpg:
All tests are passing now after upgrading to v1.8 :tada: scalafmt can now format code splices inside of xml literals :)

We are now blocked by the lack of a `ConfEncoder[T]` in metaconfig for the website to show configuration examples next to code listings. I started working on this in metaconfig, it doesn't seem too hard but stuck on 
```
[error] (metaconfig-coreJVM/test:compileIncremental) scala.reflect.internal.FatalError: Unknown type: <error>, <error> [class scala.reflect.internal.Types$ErrorType$, class scala.reflect.internal.Types$ErrorType$] TypeRef? false
```
writing the encoder derivation macro

olafurpg:
Opened https://github.com/olafurpg/metaconfig/pull/43 adding `metaconfig.ConfEncoder` typeclass to replace `toHocon`

olafurpg:
The upgrade to scalameta v1.8 is complete, it took a while since the changes around lambdas were quite big. Overall I think it's a nice improvement, for example it can now force the body of a HOF to be on the same line as the parameter list

```scala
// before
(x: Int) =>
  x + 1
// after
(x: Int) => x + 1
```

Now we are only blocked by lack of toHocon.

fommil:
if you're using semi automatic derivation in scalameta, maybe there is an opportunity to use the compiler plugin (without the scalaz bits) from https://gitlab.com/fommil/scalaz-deriving/

olafurpg:
I am indeed using semiauto but with a a twist where I need a base default value passed into the derivation macro. This is required to apply configuration on top of a base setting, for example `// scalafmt: { maxColumn = 200}` comments are applied on top of the active configuration.

This I took a look at scalaz deriving but struggled to figure out what required scalaz and what didn't. It would be helpful to have a section on how to use it without bringing in scalaz.

fommil:
@olafurpg the entire https://gitlab.com/fommil/scalaz-deriving/#compiler-plugin does not require scalaz.

mads-hartmann:
Once this is merged I should be able to use metals on this code-base right?

olafurpg:
@mads-hartmann I haven't tried, there might be other issues but removing scalameta/paradise should definitely improve the situation

olafurpg:
Almost done 😸 we now have a `ConfEncoder` typeclass that we can use to pretty print ScalafmtConfig as hocon

screen shot 2018-05-10 at 00 50 28

This is far cleaner than what we had before which used ugly runtime type matching. The show/hide buttons are still empty for some reason, that's the only remaining blocker to merge this.

olafurpg:
Released metaconfig v0.7.2 that adds a hocon printer and `Conf.patch` method to print out minimal configuration examples given two `ScalafmtConfig`

https://github.com/olafurpg/metaconfig/releases/tag/v0.7.2


- olafurpg.github.io/metaconfig/#confprinthocon
- olafurpg.github.io/metaconfig/#confpatch

olafurpg:
Gonna go ahead and merge this to unblock a followup PR that upgrades scalameta to v3.7

mads-hartmann:
@olafurpg Awesome work. As far as I can tell the show/hide config options works like expected 👍

PR #1150, opened 2018-04-24
Flip to default danglingParentheses = true

dwijnand:
Given the community's vote:
https://twitter.com/dwijnand/status/983979109918347264


olafurpg:
Can we please open a ticket first? I'd like to collect more opinions there, but the results from the twitter poll definitely help strengthen the argument in favor of flipping the default. It's a question if `align.openParen{Call,Defn}Site` needs to be flipped as well to avoid

```scala
foobar(a,
       b
)
```
I'm not sure exactly what the implications are.

dwijnand:
Sure!

dwijnand:
https://github.com/scalameta/scalafmt/issues/1151

PR #1152, opened 2018-04-24
Add (failing) test-case for vertical align formatting

mads-hartmann:
I'm opening this so we can discuss

- How to achieve the expected output
- If it should be triggered by a config parameter or be the default style for vertical align

### DIFF
--- a/scalafmt-tests/src/test/resources/test/arityThreshold.source
+++ b/scalafmt-tests/src/test/resources/test/arityThreshold.source
@@ -10,6 +10,7 @@ case class VeryLongNames(xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx: Str
 object A {
   def foo(x: String, y: String)
   def hello(how: String)(are: String)(you: String) = how + are + you
+  def other(a: String, b: String)(c: String) = a + b + c
 }
 >>>
 case class Foo(x: String)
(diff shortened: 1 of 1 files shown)


mads-hartmann:
Currently it will produce

```scala
  def other(
    a: String,
    b: String
  )(c: String
  ) = a + b + c
```

Which I find to be a bit of an eyesore 😉

scalafmt-tests/src/test/resources/test/arityThreshold.source
```diff
@@ -10,6 +10,7 @@ case class VeryLongNames(xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx: Str
 object A {
   def foo(x: String, y: String)
   def hello(how: String)(are: String)(you: String) = how + are + you
+  def other(a: String, b: String)(c: String) = a + b + c
```
olafurpg: Pro tip 😉 it's good to make this a standalone test case. You can rename arityThreshold.source into arityThreshould.stat and then write unit tests like

```scala
<<< Issue 101
def other(a: String, b: String)(c: String) = a + b + c 
>>>
def other(
  a: String,
  b: String
)(c: String) = a + b + c 
```

mads-hartmann:
I'm working towards a solution for this is #1154 instead, so closing this

Issue #1149, opened 2018-04-24
SortModifiers website examples looks weird


olafurpg:
<img width="1419" alt="screen shot 2018-04-24 at 09 08 48" src="https://user-images.githubusercontent.com/1408093/39171444-239795e8-479f-11e8-8ba3-5710814cfd5e.png">

I think it would be clearer if the code examples were smaller and vertically next to each other
```scala
implicit final val x // before
final implicit val x // after
```

cc/ @lorandszakacs

olafurpg:
This went by review since the SortModifiers PR was open at the same time as the website css change, it was unfortunate they conflicted in this way.

lorandszakacs:
@olafurpg can be closed. Did not use correct `Closes` syntax in the PR 😊

Issue #1147, opened 2018-04-23
New RedundantBraces is too aggressive


olafurpg:
This template is a guideline, not a strict requirement.
- **Version**: 1.5.0
- **Integration**: cli
- **Configuration**:
```
rewrite.rules = [
  RedundantBraces
]
```

## Steps

Given code like this:
```scala

      case FormatToken(open @ LeftParen(), _, _) if {
            val isSuperfluous = isSuperfluousParenthesis(open, leftOwner)
            leftOwner match {
              case _: Term.If | _: Term.While | _: Term.For | _: Term.ForYield
                  if !isSuperfluous =>
                true
              case _ => false
            }
          } =>
```

When I run scalafmt like this:

```bash
scalafmt
```

## Problem

Scalafmt formats code like this:
```scala
>
...
[warn] Error in /Users/ollie/dev/scalafmt/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala: <input>:1036: error: illegal start of simple expression
val isSuperfluous = isSuperfluousParenthesis(open, leftOwner)
            ^
```

## Expectation
I expect the code to format since it has no parse errors.

## Workaround

```
rewrite.redundantBraces.generalExpressions = false
```

## Notes

Regression since https://github.com/scalameta/scalafmt/pull/1122

olafurpg:
I don't know if this is intended behavior or not, but I was surprised to see

```diff
-    while (curr.left.start < end.start && curr != prev) {
-      if (matches(curr.left)) {
-        result += curr.left
-        goToMatching()
-      } else {
-        prev = curr
-        curr = next(curr)
-      }
+    while (curr.left.start < end.start && curr != prev) if (matches(curr.left)) {
+      result += curr.left
+      goToMatching()
+    } else {
+      prev = curr
+      curr = next(curr)
     }
```

Question if curly braces should be kept around while loops and if statements without an else 🤔

olafurpg:
Infix operators are a tough one as well, 


```diff
-            (opensConfigStyle(formatToken) || {
-              forceConfigStyle(leftOwner) && !styleMap.forcedBinPack(leftOwner)
-            }) =>
+            (opensConfigStyle(formatToken) ||
+              forceConfigStyle(leftOwner) && !styleMap.forcedBinPack(leftOwner)) =>
```

It's technically redundant but operator precedence and associativity is difficult sometimes so guiding braces may help readability

olafurpg:
[...]

PR #1153, opened 2018-04-24
SortModifiers rewrite fixes + doc prettyfying

### LINKED ISSUE #1148: New SortModifiers overwrites modifiers of implicit class parameter val/vars
- **Version**: 1.5.0
- **Integration**: IntelliJ, SBT
- **Configuration**:
```
rewrite {
  rules = [SortModifiers]
  sortModifiers {
    #the crucial part of this bug is for modifiers to have higher precedence than "implicit"
    order = ["private", "protected", "implicit", "final", "sealed", "abstract", "override", "lazy"]
  }
}
```

## Steps

Given code like this:
```scala
class Test(private val x: Int)(
  implicit
  private val i1: Int,
  private val i2: String
)
```

When I run scalafmt anywhere (IntelliJ, or SBT).

## Problem

```scala
class Test2(private val x: Int)(
  implicit
  implicit val i1: Int,
  implicit val i2: String
)
```

## Expectation

```scala
class Test2(private val x: Int)(
  implicit
  private val i1: Int,
  private val i2: String
)
```

## Workaround

I've found that this happens only if the class parameters have modifiers that have higher precedence than `implicit`.

lorandszakacs:
Closes #1148 (bug), #1149 (docs)

There is one super weird corner case that is not handled properly by the fix. There's also a test for it:
If you have a sort order where `implicit` has lower order than the rest, then the following code gets formatted weirdly.

Config:
```
rewrite {
  rules = [SortModifiers]
  sortModifiers {
    order = ["private", "protected" , "abstract", "final", "sealed", "implicit", "override", "lazy"]
  }
}
```
Original code:
```
class X(private final val x: Int)(
    implicit final private val i1: Int,
    implicit final private var i2: Int
)
```
Formatted code:
```
class X(private final val x: Int)(
    implicit private final val i1: Int,
    private final implicit var i2: Int
)
```

But this is a super weird corner case where the `implicit` keyword is duplicated, and honestly until I wrote this format I had no idea that was even possible 😅. 

You will also notice that this implementation is actually cleaner than the buggy one, which is a pleasant surprise 😃 
/cc @olafurpg

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
@@ -35,15 +35,14 @@ object SortModifiers extends Rewrite {
       case t: Defn.Trait => sortMods(t.mods)
       case p: Term.Param =>
         sortMods(
-          p.mods.filterNot(m => m.is[Mod.ValParam] || m.is[Mod.VarParam]),
-          code = code)
+          p.mods.filterNot(m => m.is[Mod.ValParam] || m.is[Mod.VarParam])
+        )
     }
     patchesOfPatches.flatten
   }
 
   private def sortMods(
-      oldMods: Seq[Mod],
-      code: Tree = null
+      oldMods: Seq[Mod]
   )(implicit order: Vector[ModKey]): Seq[Patch] = {
     if (oldMods.isEmpty) Nil
     else {
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
@@ -69,7 +69,7 @@ object SortModifiers extends Rewrite {
     * )
     * }}}
     *
-    * `val i1`, and `var i2` have an ``Mod.Implicit`` with empty tokens.
+    * `val i1`, and `var i2` have a ``Mod.Implicit`` with empty tokens.
     * Therefore we want to completely ignore this "mod" whenever we sort,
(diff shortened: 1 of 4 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
```diff
@@ -35,31 +35,48 @@ object SortModifiers extends Rewrite {
...
+    *     private[this] var i2: String
+    * )
+    * }}}
+    *
+    * `val i1`, and `var i2` have a ``Mod.Implicit`` with empty tokens.
+    * Therefore we want to completely ignore this "mod" whenever we sort,
+    * and apply patches
+    */
+  private def isHiddenImplicit(m: Mod): Boolean = {
+    m.tokens.isEmpty && m.is[Mod.Implicit]
```
olafurpg: I think this will return true for `def foo(implicit x: Int)` due to a bug, can you double check?

lorandszakacs: @olafurpg will do.

lorandszakacs: @olafurpg indeed it does return `true`, but luckily no rewrite is necessary in this case, so the final result is OK. Added an explicit test case for this:
```
<<< 1148 — only implicit param
def foo(implicit x: Int)
>>>
def foo(implicit x: Int)
```

olafurpg:
This came out nice 👍 

screen shot 2018-04-24 at 13 37 02

olafurpg:
I'll try to get a hotfix out later this week, I want to address #1147 as well

PR #1154, opened 2018-04-24
Make it configurable when to exclude dangling parens for vertical

mads-hartmann:
The configuration defaults to the previous behavior.

### DIFF
--- a/readme/Configuration.scalatex
+++ b/readme/Configuration.scalatex
@@ -311,7 +311,7 @@
     @p The @code{newlines.*} options are used to configure when and where @code{scalafmt} should
     insert newlines.
 
-    @warning
+    @sidenote
       You might be interested in the @sect.ref{Vertical Multiline} section.
 
     @sect{newlines.alwaysBeforeTopLevelStatements}
@@ -560,7 +560,7 @@
     @p
       If enabled this formats methods such that parameters are on
       their own line indented by
-      @sect.ref{continuationIndent.defnSite} separation between
+      @sect.ref{continuationIndent.defnSite}. Separation between
       parameter groups are indented by two spaces less than
       @sect.ref{continuationIndent.defnSite}. The return type is on
       its own line at then end.
@@ -577,8 +577,7 @@
         }
 
     @warning
-      This setting ignores continuation.defnSite,
-      binPack.unsafeDefnSite, and @sect.ref{align.openParenDefnSite}.
+      This setting ignores binPack.unsafeDefnSite, and @sect.ref{align.openParenDefnSite}.
 
     @p
       The default configuration options are
@@ -611,8 +610,8 @@
       Default: @b(default.verticalMultiline.newlineAfterOpenParen)
 
       @p
-        If true, add a newline aften the opening paren of a parameter
-        list. This can be used to control for formatting of methods that
+        If true, add a newline after the opening paren of a parameter
+        list. This can be used to control the formatting of methods that
         have multiple parameter lists.
 
       @fullWidthDemo(multilineNewlineAfterParen)
--- a/readme/Configuration.scalatex
+++ b/readme/Configuration.scalatex
@@ -584,7 +584,7 @@
 
     @hl.scala(VerticalMultilineDefultConfigStr)
 
-    @sect{verticalMultiline.atDefinitionSite}
+    @sect{verticalMultiline.atDefnSite}
       Default: @b{false}
 
       @p
--- a/readme/Configuration.scalatex
+++ b/readme/Configuration.scalatex
@@ -902,35 +902,6 @@
       There is a Scala.js style that is super experimental and does not
       work for complicated code.
 
-    @p
-      To enable a rewrite rule, add it to the config like this @config{verticalMultiline.arityThreshold = 2}.
-
-    @fullWidthDemo(arityThreshold)
-      case class Foo(x: String)
-      case class Bar(x: String, y: String)
-      case class VeryLongNames(xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx: String)
-      object A {
-        def foo(x: String, y: String)
-        def hello(how: String)(are: String)(you: String) = how + are + you
-      }
-
-  @sect{project}
-    @p
-      Configure which source files should be formatted in this project.
-    @cliFlags
-      # Only format files tracked by git.
-      project.git = true
-      # manually exclude files to format.
-      project.excludeFilters = [
-         regex1
-         regex2
-      ]
-      # manually include files to format.
-      project.includeFilters = [
-        regex1
-        regex2
-      ]
-
   @sect{Other}
 
     Scalafmt has a ton of internal configuration options which may be removed
--- a/readme/Configuration.scalatex
+++ b/readme/Configuration.scalatex
@@ -312,10 +312,7 @@
     insert newlines.
 
     @warning
-      There are three options that are covered in the @sect.ref{Vertical Multiline} section
-      as they are only relevant if that feature is enabled. The options are are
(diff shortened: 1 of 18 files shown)


mads-hartmann:
By the way, this is very much a proof of concept. I'd just like to discuss the behavior for now :)

mads-hartmann:
The previous formatting of 

```scala
case class Foo(x: String) extends Foobar
case class Bar(x: String, y: String) extends Foobar
case class VeryLongNames(xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx: String)
```

Looks like this

```scala
sealed trait Foobar
case class Foo(x: String) extends Foobar
case class Bar(
  x: String,
  y: String)
    extends Foobar
case class VeryLongNames(
  xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx: String)
```

Where the new one looks like this

```scala 
sealed trait Foobar
case class Foo(x: String) extends Foobar
case class Bar(
  x: String,
  y: String
) extends Foobar
case class VeryLongNames(
  xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx: String
)
```

mads-hartmann:
Another interesting approach would be to say that it respects the normal `danglingParentheses` options but then not consider a paren to be dangling if it has an explicit return type or and `extends`  coming after it. 

Which would result in

```scala
sealed trait Foobar
case class Foo(x: String) extends Foobar
case class Bar(
  x: String,
  y: String
) extends Foobar
case class VeryLongNames(
  xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx: String)
```

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -146,6 +146,7 @@ case class ScalafmtConfig(
     poorMansTrailingCommasInConfigStyle: Boolean = false,
     verticalMultilineAtDefinitionSite: Boolean = false,
     verticalMultilineAtDefinitionSiteArityThreshold: Int = 100,
+    verticalMultilineAtDefinitionSiteDanglingExclude: List[String] = List("class", "trait"),
```
olafurpg: I propose we introduce a `VerticalMultiline` case class with the nested settings, similar to `Align` and `Newlines`. We can still keep the old `verticalMultilineAtDefinitionSite` `verticalMultilineAtDefinitionSiteArityThreshold` for backwards compat.

```
verticalMultiline {
  defnSite = true
  arityThreshold = 100
  ...
}
```

mads-hartmann: Definitely, it was getting a bit messy 👍

pjrt:
But why? It is so ugly. You haven't seen the ugliness because the test you used were just plain case classes with no body. 

Try doing `case class X(x: Int) { def hello }` or something like that.

scalafmt-tests/src/test/resources/test/verticalMultilineDangling.source
```diff
@@ -0,0 +1,21 @@
...
+>>>
+sealed trait Foobar
+case class Foo(x: String) extends Foobar
+case class Bar(
+  x: String,
+  y: String
+) extends Foobar
+case class VeryLongNames(
+  xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx: String
+)
```
pjrt: How does it behave for `case class X()`?

mads-hartmann:
Ugliness is relative ☺️ Do you prefer this

```scala
case class Bar(
  x: String,
  y: String)
    extends Foobar
```

Over this?

```scala
case class Bar(
  x: String,
  y: String
) extends Foobar
```

pjrt:
I prefer the second one, which is why `verticalMultiline` does the second one. Note that the dangling does happen if u extend something. It doesn't dangle if you do not extend anything. 

That was a bad example (that's ugly regardless of how u do it). 

More lile

```
case class X(x: Int) {
   def hello
}
```

If we force dangle, then it looks like this:

```
case class X(
    x: Int
  ) {
   def hello
}
```

Which is...ugly. But yes, it is relative. I feel it looks better if the `)` is inline in that case. It is an option though...

olafurpg:
FYI, 
```scala
case class X2(
  x: Int,
  y: Int
) { def hello: String = "Hello" }
```
becomes

```scala
case class X2(
  x: Int,
  y: Int
) {
  def hello: String = "Hello"
}
```
if you insert a line break inside the template.

olafurpg:
One reason I'm excited about https://github.com/olafurpg/scala-syntax is that tree printing makes it easier to enforce "long-distance" consistency

mads-hartmann:
Once #1155 is merged I'll rebase this and write some docs ☺️ Until then, here's two things to consider

- If merged this is a breaking change configuration wise
- I've deleted some of the docs from the source. IMO examples belong in the public docs. WDYT?

mads-hartmann:
screen shot 2018-04-28 at 23 24 37

mads-hartmann:
@olafurpg @pjrt This is now in a state where it makes it possible to

- Configure when to use dangling parens using the excludeDanglingParens option

- Opt-in to adding a newline after open parens. This is useful when formatting methods with multiple parameter lists (This fixes #1152)

Additionally this moves all vertical multiline related configuration options into a single key `verticalMultiline` - this is a breaking change configuration wise.

olafurpg:
> Additionally this moves all vertical multiline related configuration options into a single key verticalMultiline - this is a breaking change configuration wise.


That's great! Is it possible to respect the old configuration options to avoid breaking existing .scalafmt.conf?

mads-hartmann:
@olafurpg Should be, I was just being slightly lazy ;) I'll have a look at making it backwards compatible

mads-hartmann:
@olafurpg Done ☺️

mads-hartmann:
I ran it on our own codebase and there are still a couple of edge-cases I need to handle :)

mads-hartmann:
I don't have any outstanding issues now :)

mads-hartmann:
@olafurpg No more breaking changes and it works-as-expected-on-our-code-base™️ 😉

readme/Configuration.scalatex
```diff
@@ -560,66 +557,104 @@
...
+      Default: @b(default.verticalMultiline.newlineAfterImplicitKW)
+
       @p If true, add a newline after an implicit keyword in function and class definitions.
 
       @fullWidthDemo(verticalAlignImplicitAfter)
         object a {
           def format(code: String, age: Int)(implicit ev: Parser, c: Context): String
         }
 
+    @sect{verticalMultiline.excludeDanglingParens}
```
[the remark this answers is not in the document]
mads-hartmann: I was wondering what a Defn.Trait was 😅I'd definitely love to kill this config altogether. I only included it to preserve backwards compatibility - the previous implementation would accept dangling parens some places (methods) but not everywhere (constructors).

olafurpg:
That sounds good to me @mads-hartmann

olafurpg:
Does this PR impose changed behavior from how people used vertical multiline before? I just unblocked #1145 and I'd like to rebase on top of the changes here to clean up the remaining usage of `toHocon`

mads-hartmann:
@olafurpg In its current state it does not :) However if we remove the `excludeDanglingParens` option in favour of relying on `danglingParens` it will as the previous behaviour would add a dangling paren for methods but not case-class constructors - using `danglingParens` it wouldn't be able to differentiate between the two :)

mads-hartmann:
@olafurpg Awesome, thanks 👍

PR #1155, opened 2018-04-25
Reorganise the documentation for configuration options

mads-hartmann:
I find the existing flat list to be a bit overwhelming so I tried to
group the relevant settings together a bit.

I realise the docs site might be rewritten in the near future but I
still find it valuable to explore how best to group the configuration
options.

### DIFF
--- a/readme/Changelog04.scalatex
+++ b/readme/Changelog04.scalatex
@@ -181,7 +181,7 @@
       For super easy CLI installation in OSX and Linux, download the new
       auto-upgrade script in (EDIT: removed) Linux/OSX section.
     @li
-      NEW! See @sect.ref{rewrite.rules}.
+      NEW! See @sect.ref{Rewrite Rules}.
     @li
       NEW! Dynamic style configuration. You can now update the configuration
       by adding comments to the source code. Example:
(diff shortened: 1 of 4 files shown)


mads-hartmann:
The TOC looks like this after the reorganisation: 
screen shot 2018-04-25 at 21 24 55

mads-hartmann:
This is what nodes styled using `@warning` looks like:

screen shot 2018-04-25 at 23 01 37

mads-hartmann:
@olafurpg I pushed a commit that introduces a `Most Poplar` section right after `style` - how do you feel about that solution?

As for moving project.git

mads-hartmann:
@olafurpg This is ready for a new review ☺️

olafurpg:
btw, vertical multiline still looks weird

screen shot 2018-04-28 at 13 47 54

I'm wondering if it better to show the examples vertically instead of horizontally?

mads-hartmann:
@olafurpg Is that from this branch or master after it was merged? I think this branch was behind master :)

olafurpg:
Right I was confused because I thought it was already fixed, didn't think it was only in your branch :)

PR #1156, opened 2018-04-25
Fix the full-width example styling

mads-hartmann:
Before:

screen shot 2018-04-25 at 22 08 05

After:

screen shot 2018-04-25 at 22 07 42

### DIFF
--- a/readme/resources/custom.css
+++ b/readme/resources/custom.css
@@ -100,9 +100,43 @@ ul.menu-item-list {
   justify-content: space-around;
 }
 
+.scalafmt-rows {
+  background: #f9f9f9;
+  border-left: 10px solid #ccc;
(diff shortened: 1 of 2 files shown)


mads-hartmann:
@olafurpg haha duh. As you know my editor situation is broken right now so didn't have auto-formatting enabled 😅

olafurpg:
format on save in the editor is the best! Although, I have built a habit to run `./scalafmt --diff` on push. Been meaning to setup something like https://prettier.io/docs/en/precommit.html for a while but never got around to try it

mads-hartmann:
@olafurpg Definitely, that would be nice to have. We could probably steal some of their content structure as well to be honest (when we get docusaurus) 😉

PR #1159, opened 2018-04-26
Add size to AvoidInfix

joan38:
The issue I'm trying to resolve is:
```
runningJobs should have size 0
```
reformated to:
```
(runningJobs should have).size(0)
```

This might be an issue with other similarly named functions like:
```
Seq() size
```
will not be reformated to:
```
Seq() size
```

Let me know what you think.


olafurpg:
`Seq() size` is postfix, which is similar but a different thing that is not targeted by AvoidInfix

joan38:
True.
Cheers @olafurpg

PR #1164, opened 2018-05-06
Add section about scalafmt support in mill

rockjam:
This PR adds section in documentation site about scalafmt support in [mill](http://www.lihaoyi.com/mill/) build tool.


rockjam:
Well, to keep it fresh - yes. 
There is a single setting for scalafmt version, users are free to update version in their own builds, before mill's code get updated.

PR #1174, opened 2018-05-11
Add trailingCommas option

### LINKED ISSUE #1060: optIn.configStyleArguments = false breaks with trailing commas
- **Version**: 1.3.0
- **Integration**: IntelliJ
- **Configuration**:
```
optIn.configStyleArguments = false
```

## Steps

I ran scalafmt like this:

```bash
Ctrl + Shift + L
```

## Problem

Formatted this
```scala
object Dummy {
  Seq(
    1,
    2,
    3,
  )
}
```

```scala
object Dummy {
  Seq(1, 2, 3, )
}
```

## Expectation

```scala
object Dummy {
  Seq(1, 2, 3)
}
```

## Notes
Don't actually remember why I disabled configStyleArguments, but I think it had something to do with bad compatibility with IntelliJ style back in the day. Here'a bug report anyway.

### LINKED ISSUE #1173: Enforce or remove trailing commas
I think it would be useful for scalafmt to enforce/remove trailing commas, so that a codebase can adopt a consistent style with those.

Ideally

with this config

```json
{
  "align": "none",
  "danglingParentheses": true,
  "trailingCommas": true
}
```

and given this code

```scala
// column limit   |
case class MyClass(a: String, b: Int)
```

this would happen

```scala
case class MyClass(
  a: String,
  b: String,
)
```

---

Some other cases:

```scala
// column limit                      |
case class MyClass(a: String, b: Int)
```

nothing happens (trailing commas not inserted if not followed by a newline).

---

```json
{
  "align": "none",
  "danglingParentheses": false,
  "trailingCommas": true
}
```

```scala
// column limit   |
case class MyClass(a: String, b: Int)
```

no trailing comma, because danglingParentheses is `false` (and a trailing comma would be invalid syntax)

```scala
case class MyClass(
  a: String,
  b: String)
```

---

I'll flesh out more details as we discuss.

gabro:
Closes #1173 
Closes #1060

Examples, given this config:

```
trailingCommas = always
align = none
danglingParentheses = true
```

<table>
<thead><tr><td><b>Before</b></td><td><b>After</b></td></tr></thead>
<tr>
<td>

```scala
def method(
    a: String,
    b: Int
)
```

</td>
<td>

```scala
def method(
    a: String,
    b: Int,
)
```

</td>
</tr>
<tr>
<td>

```scala
// maxColumn          |
def method(a: String, b: Int)
```

</td>
<td>

```scala
def method(
    a: String,
    b: Int,
)
```

</td>
</tr>
<tr>
<td>

```scala
// maxColumn          |
import x.y.{memberA, memberB}
```

</td>
<td>

```scala
import x.y.{
  memberA,
  memberB,
}
```

</td>
</tr>
<tr>
<td>

```scala
// maxColumn       |
def method[Key, Value](paramA: Int, paramB: Int)
```

</td>
<td>

```scala
// maxColumn       |
def method[
    Key,
    Value,
](
    paramA: Int,
    paramB: Int,
)
```

</td>
</tr>
</table>

Before moving forward, I would appreciate if @olafurpg could take a quick look, just to see whether I'm doing this right 😅


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -43,6 +43,24 @@ class FormatWriter(formatOps: FormatOps) {
                 .getOrElse(token.syntax, token.syntax)
             sb.append(rewrittenToken)
         }
+
+        if (runner.dialect.allowTrailingCommas &&
+          state.splits.last.modification.isNewline &&
```
olafurpg: It would be good to have a parallel that removes the trailing comma if `!modification.isNewline`

scalafmt-tests/src/test/resources/unit/Dialect.stat
```diff
@@ -7,7 +7,7 @@ foo(
 >>>
 foo(
     1,
-    2,
+    2
```
gabro: This PR is currently breaking. I have an idea on how to make this non-breaking by default.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -43,6 +43,24 @@ class FormatWriter(formatOps: FormatOps) {
                 .getOrElse(token.syntax, token.syntax)
             sb.append(rewrittenToken)
         }
+
+        if (runner.dialect.allowTrailingCommas &&
```
olafurpg: Let's move this to a helper method `handleTrailingComma` or something like that

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -43,6 +43,24 @@ class FormatWriter(formatOps: FormatOps) {
...
+          state.splits.last.modification.isNewline &&
+          (formatToken.right.is[Token.RightParen] ||
```
olafurpg: Conventionally we pattern match against `case FormatToken(_: Token.Comma, _: Token.RightParen | _: Token.RightBracket, _)`

olafurpg: Actually, pattern matching might not be a good idea since we'll have to handle comments as well. This is fine 👍

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
gabro: I agree, I was about to do it!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
gabro: Yep, it may be necessary for handling `configStyleArguments = false`

olafurpg: also necessary for
```scala
foo(a,
  b,
)
```

gabro: yep, I have a skipped test for it. Thanks for the examples, it's hard to come up with those :D

olafurpg: The comma should btw be removed in that case regardless of the setting since it's a bug to produce `foo(a, b,)`

gabro:
Ok, not it handles the removal of trailing commas when going from multi-line to single line, e.g.

```scala
// before
foo(
  a,
)

// after
foo(a)
```

or

```scala
// before
foo(a,
  b,
)

// after
foo(a, b)
```

Comments are next!

gabro:
I've added support for handling comments after trailing commas when adding/removing:

```scala
foo(
  a,
  b, // comment
)
```

Unfortunately I hit https://github.com/scalameta/scalameta/issues/1531 which makes the test fail. I was able to test the addition nonetheless, since the bug affects the output, but I'm not confident the removal works, since the test input doesn't parse.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -363,6 +365,89 @@ class FormatWriter(formatOps: FormatOps) {
...
+      // foo(
+      //   a,
+      //   b // comment
+      // )
+      //
+      // Insert a comma after b (before comment)
+      case TrailingCommas.always
+          if left.is[Comment] && !prev(left).is[Comma] && right.is[ClosingBracket] && isNewline =>
+        sb.insert(sb.length - left.syntax.length - 1, ",")
+        sb.append(whitespace)
```
gabro: I had to deal with comments in this way, detecting a comment, looking for a comma next to it, going back in the buffer and removing the comma.

Not sure how safe is this.

I had initially tried to incorporate this into the previous case, but in presence of `[ID][COMMENT][NEWLINE][CLOSING_BRACKET]` I couldn't detect the newline after che comment.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -363,6 +365,89 @@ class FormatWriter(formatOps: FormatOps) {
...
+      // foo(
+      //   a,
+      //   b, // comment
+      // )
+      //
+      // Remove the comma after b (before comment)
+      case TrailingCommas.never
+          if left.is[Comment] && prev(left).is[Comma] && right.is[ClosingBracket] && isNewline =>
+        sb.deleteCharAt(sb.length - left.syntax.length - 1)
+        sb.append(whitespace)
```
gabro: This is the case that is currently untested, since the test input fails to parse.

gabro:
Ok, I've tried using https://github.com/scalameta/scalameta/pull/1532 in a published version of Scalameta, and I've found a couple of bugs with the comments. It seems my approach is not idempotent. I'll investigate!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -43,7 +43,9 @@ class FormatWriter(formatOps: FormatOps) {
                 .getOrElse(token.syntax, token.syntax)
             sb.append(rewrittenToken)
         }
-        sb.append(whitespace)
+
+        handleTrailingCommasAndWhitespace(formatToken, state, sb, whitespace)
```
gabro: I had to incorporate the insertion of the whitespace, since in one case it needs to be skipped, namely when rewriting

```scala
foo(a,
  b,
)
```

to

```scala
foo(a, b)
```

Inserting the white space would result in

```scala
foo(a, b )
```

instead

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -363,6 +365,93 @@ class FormatWriter(formatOps: FormatOps) {
...
+      // foo(
+      //   a,
+      //   b
+      // )
+      //
+      // Insert a comma after b
+      case TrailingCommas.always
+          if !left.is[Comma] && !left.is[Comment] &&
+            right.is[CloseDelim] && isNewline =>
+        sb.append(",")
```
olafurpg: Does this force a comma in this case?

```scala
// before
lst.map { x =>
  x + 1
}
// after
lst.map { x =>
  x + 1,
}
```

I think we may have to limit this behavior to a couple of nodes

- Term.Apply
- Importee/Importer
- Type.Apply

olafurpg: Right, tuples as well 👍 

`FormatOps.owners(Token): Tree` goes to the closest enclosing node of a token

gabro: done! The final list includes also `Defn.Def` and `Decl.Def`. I've added test cases for all of those.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -363,6 +367,99 @@ class FormatWriter(formatOps: FormatOps) {
...
+      formatToken: FormatToken,
+      state: State,
+      sb: StringBuilder,
+      whitespace: String): Unit = {
+
+    import org.scalafmt.config.TrailingCommas
+
+    val owner = owners(formatToken.right)
+    val hasValidOwner =
+      owner.is[Term.Apply] || owner.is[Term.Tuple] ||
```
olafurpg: Can we introduce a classifier named `CommaSeparated` for this group? I think `Ctor.{Primary,Secondary}` are missing for 
```scala
class Foo(
  a: Int,
  b: String,
)
```

scalafmt-tests/src/test/resources/trailing-commas/trailingCommasAlwaysComments.stat
```diff
@@ -0,0 +1,13 @@
+maxColumn = 30
+trailingCommas = always
+
+<<< SKIP should consider comments when adding trailing commas
```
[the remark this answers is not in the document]
olafurpg: It may be a couple weeks until the next meta release so I think it would be good to track this in a ticket.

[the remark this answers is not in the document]
olafurpg: Done https://github.com/scalameta/scalafmt/issues/1182

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
gabro: Nice catch, yes I'll introduce a classifier if you think it's easier to read

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -318,6 +319,11 @@ object TreeOps {
   def isDefnOrCallSite(tree: Tree): Boolean =
```
gabro: this method was already here, I realized!

gabro:
@olafurpg take a look, I've re-used an existing method in `FormatOps`, so now the owner check should be fairly comprehensive. I've also added a bunch of new test cases for stuff like

```scala
val (
  a,
  b
) = someTuple
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -363,6 +367,94 @@ class FormatWriter(formatOps: FormatOps) {
...
+      //
+      // Remove the comma after b
+      case TrailingCommas.never
+          if left.is[Comma] && right.is[CloseDelim] && isNewline =>
+        sb.deleteCharAt(sb.length - 1)
+        sb.append(whitespace)
+
+      // foo(
+      //   a,
+      //   b, // comment
```
olafurpg: 👏 Super nice comments, nice break down

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -124,6 +134,7 @@ case class ScalafmtConfig(
     assumeStandardLibraryStripMargin: Boolean = false,
     danglingParentheses: Boolean = false,
     poorMansTrailingCommasInConfigStyle: Boolean = false,
+    trailingCommas: TrailingCommas = TrailingCommas.preserve,
```
olafurpg: I wonder what is the best default 🤔 Personally, I think `always` might be a good choice since I'd prefer scalafmt to be opinionated where possible. However, that will break code on 2.11.

I'll open a discussion to gauge opinions, until then I agree `preserve` is the safest choice 👍

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
gabro: Those were mostly for myself 😅 They seemed useful so I've left them

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
gabro: I wonder if the default could depend the dialect of the runner

gabro:
CI is 🍏 merging! Thank you so much @olafurpg for the many and quick reviews!

gabro:
> Seems we forgot to document the new setting on the website, I'll write about it in the changelog but I think it would be a good candidate for the "Popular" section

Sort of. I didn't forget, but I didn't want to put energy in writing docs before completing the new website (maybe I'm ambitious :P )

Issue #1002, opened 2017-07-23
Try / finally formatted on a single line


nrinaudo:
- **Version**: 1.1.0
- **Integration**: command line (coursier)
- **Configuration**: default (no configuration specified)

## Steps
Input
```scala
object Test {
  try { foo() }
  finally { bar() }
}
```

## Problem

Scalafmt formats code like this:
```scala
object Test {
  try { foo() } finally { bar() }
}
```

## Expectation

I was expecting `finally` to have a line of its own.

## Notes

This might be, if not a duplicate of, heavily related to #350.

olafurpg:
I suspect this is even related to https://github.com/scalameta/scalafmt/issues/271

I once tried to take a stab at it but scalafmt's design/implementation makes it surprisingly hard to insert a newline there.

olafurpg:
Fixed in #1187 🎉 

Scalafmt will now force try/finally blocks to break on multiple lines

```scala
try { foo() } finally { bar() }
// after
try {
  foo()
} finally {
  bar()
}

```

This is not an optimal solution but avoids the ugly output at least.

PR #1187, opened 2018-05-15
Try / finally formatted on a single line

slavaschmidt:
A possible fix for https://github.com/scalameta/scalafmt/issues/1002

It extends the rule prohibiting single line blocks for `if` blocks to try-catch blocks as well

### DIFF
--- a/scalafmt-tests/src/test/resources/default/Try.stat
+++ b/scalafmt-tests/src/test/resources/default/Try.stat
@@ -58,3 +58,9 @@ object Foo {
     bar()
   }
 }
+<<< try/catch with handler
+try { foo()  } catch handler
+>>>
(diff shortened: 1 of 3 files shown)


scalafmt-tests/src/test/resources/default/Try.stat
```diff
@@ -45,3 +45,16 @@ object Foo {
   try proc.waitFor()
   catch { case e: Throwable => run() }
 }
+<<< try/finally curly no single line blocks
```
olafurpg: Can we add a test case for the case below?

```scala
try { foo() } finally handler
```

slavaschmidt:
Thanks for guiding me through!

Issue #915, opened 2017-04-28
Space on both sides of operator method definition


durban:
Feature request: an option to surround symbolic method names with exactly one space on both sides.

- **Version**: 0.6.8
- **Integration**: sbt
- **Configuration**: default

## Problem

Scalafmt formats code like this:
```scala
trait Test[A] {
  def <=>[B](that: Test[B]): Int
}
```

## Expectation

I would like the formatted output to look like this:
```scala
trait Test[A] {
  def <=> [B](that: Test[B]): Int
}
```

Note the space immediately after the `<=>` method name. It mirrors how it would typically appear at callsites: `x <=> y`. (Obviously the `<=>` operator is just an example.)

olafurpg:
Thank you for reporting! This should be possible to accommodate. What do you think would be a good name for this option? I propose `spaces.aroundSymbolicOperatorDefinitions`.

The tutorial https://github.com/scalameta/scalafmt/blob/master/Tutorial.md should contain all the necessary information to find the place to implement this and test it.

ShaneDelmore:
Doesn't this happen on application, not just definition?  I seem to recall seeing it in scalatest with foo should ===(bar) but can't recall if it has been fixed or not.  If so, then I like a name closer to the original reported issue,  spaces.aroundSymbolicMethodNames to signify the space should be at both the definition site and the usage site.

olafurpg:
@ShaneDelmore You may be referring  to the `spaces.afterTripleEquals = true` setting to get this output`a should be === (x)`. I'm personally very much against that setting because it makes the illusion that `===(x)` is an infix operator when in fact it's a regular Term.Apply. Infix operators always have space at call-site, so it's not necessary to have a setting for that. 

However, now that I think about it maybe `spaces.aroundSymbolOperatorDefnSite` is more consistent with other callSite/defnSite settings.

durban:
I think `spaces.aroundSymbolOperatorDefnSite` is okay. Maybe even `spaces.aroundOperatorDefnSite`. But english is not my first language, so you shouldn't believe me about naming things :-)

olafurpg:
Fixed in #1188 by @slavaschmidt  🎉

PR #1188, opened 2018-05-15
Added a rule for space after an operator method

slavaschmidt:
A possible fix for https://github.com/scalameta/scalafmt/issues/915

It reuses existing `isSymbolicName` to decide if method name is an operator or not.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -374,7 +374,7 @@ class Router(formatOps: FormatOps) {
                 t.tokens.map(_.syntax) == Seq("===") =>
             Space
           case name: Term.Name
-              if isSymbolicName(name.value) && name.parent.exists(isDefDef) =>
+              if style.spaces.afterSymbolicDefs && isSymbolicName(name.value) && name.parent.exists(isDefDef) =>
             Space
           case _ => NoSplit
         }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -36,6 +36,8 @@ import org.scalafmt.util.SomeInterpolate
 import org.scalafmt.util.TokenOps
 import org.scalafmt.util.TreeOps
 import org.scalafmt.util.Trivia
+
+import scala.meta.Decl.Def
 import scala.meta.Lit
 
 // Too many to import individually.
@@ -73,6 +75,8 @@ class Router(formatOps: FormatOps) {
     val leftOwner = owners(formatToken.left)
     val rightOwner = owners(formatToken.right)
     val newlines = newlinesBetween(formatToken.between)
+    val isDef = rightOwner.tokens.headOption.exists(_.is[KwDef])
+
     formatToken match {
       case FormatToken(_: BOF, _, _) =>
         Seq(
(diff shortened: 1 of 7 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -36,6 +36,8 @@ import org.scalafmt.util.SomeInterpolate
 import org.scalafmt.util.TokenOps
 import org.scalafmt.util.TreeOps
 import org.scalafmt.util.Trivia
+
+import scala.meta.Decl.Def
```
olafurpg: By convention we use `Decl.Def` instead of `import Decl.Def; Def`. This convention prevents conflicts since there is another `Defn.Def`

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -73,6 +75,8 @@ class Router(formatOps: FormatOps) {
     val leftOwner = owners(formatToken.left)
     val rightOwner = owners(formatToken.right)
     val newlines = newlinesBetween(formatToken.between)
+    val isDef = rightOwner.tokens.headOption.exists(_.is[KwDef])
```
olafurpg: Let's remove this line. Suggested solution below.

BTW. This returns true for any class/trait/object that contains methods.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -371,6 +375,8 @@ class Router(formatOps: FormatOps) {
               if style.spaces.afterTripleEquals &&
                 t.tokens.map(_.syntax) == Seq("===") =>
             Space
+          case name: Term.Name if isDef && isSymbolicName(name.value) =>
```
olafurpg:
```scala
 case name: Term.Name if isSymbolicName(name.value) && isDefDef(owners(name))
```

olafurpg: We should guard this behavior behind a configuration setting `spaces.afterSymbolicDefs`, see "Spaces.scala" and add an explanation for the configuration option in the docstring.

slavaschmidt: Thanks for the feedback! Just a small question about the requested changes: what would be the correct way to get from `name: Term.Name` to the `Token` to call the `owners` function, something like `isDefDef(owners(formatToken.right))` ?

olafurpg: `name.parent.exists(isDefDef)`

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -371,6 +373,9 @@ class Router(formatOps: FormatOps) {
               if style.spaces.afterTripleEquals &&
                 t.tokens.map(_.syntax) == Seq("===") =>
             Space
+          case name: Term.Name
+              if isSymbolicName(name.value) && name.parent.exists(isDefDef) =>
```
olafurpg: Missing

```scala
if style.spaces.afterSymbolicDefs && ...
```

scalafmt-tests/src/test/resources/default/DefDef.stat
```diff
@@ -303,3 +303,19 @@ def foo =
     val bar = fetchBar()
     bar.map(_ + 10)
   })
+<<< #915 Space on both sides of operator method definition (positive)
+trait Test[A] {
```
olafurpg: We should create a separate `test/afterSymbolicDefs.stat` file with these test cases

```
spaces.afterSymbolicDefs = true
<<< #915 Space on both sides of operator method definition (positive) 
...
>>>
...
````

It would be good to have a test case to validate that by default there is no space.

slavaschmidt:
Sorry it took me so long go get back on that.

PR #1199, opened 2018-05-27
Move website to markdown + docusaurus

gabro:
New website based on Docusuaurus and tut + sbt-site preprocessing.

The setup is identical to the Metal's one, so the setup instructions are exactly the ones published at https://scalameta.org/metals/docs/website.html with the only difference that the website can be browsed locally at http://localhost:3000/scalafmt . We should copy this guide over.

Still WIP, since not all the examples have been ported and there's a few TODOs scattered around.

## TODOs

- [ ] publish via Travis (copy this from metals, it works fine)
- [x] check that all examples in the configuration make sense (there are some inconsistencies)
- [x] check that everything is fine on mobile
- [ ] fix links in the footer
- [x] port installation instructions


olafurpg:
This is awesome @gabro! I am excited to migrate the website to docusaurus + tut 

I will try to update the content, both configuration options and prose.

cc/ @mads-hartmann

mads-hartmann:
This is fantastic 👏

gabro:
Rebased and added a contributing section, similarly to what we have in Metals.

docs/configuration.md
```diff
@@ -0,0 +1,494 @@
...
+> renamings and other refactorings.
+
+#### `align=more`
+
+```tut:passthrough
+website.exampleBlock(
+  s"""|val x = 2 // true for assignment
+      |val xx = 22
+
+      |case object B extends A // false for `extends`
```
gabro: it seems true, instead

docs/configuration.md
```diff
@@ -0,0 +1,494 @@
...
+
+```tut:passthrough
+website.exampleBlock(
+  s"""|x match { // true for case arrows
+      |  case 2  => 22
+      |  case 22 => 222
+      |}
+
+      |def foo(a: Int, // true for defn site open paren
+      |        b: String): Int
```
gabro: this is false now

docs/configuration.md
```diff
@@ -0,0 +1,494 @@
...
+      |foo(a: Int, // true for call site open paren
+      |    b: String): Int
```
gabro: this too

sake92:
Sorry for popping out of nowhere. 😅 If you want to try Scala alternative, you could try https://sake92.github.io/hepek/ (markdown is supported, typesafe layouts, PDF rendering etc).  
It would be almost trivial to integrate typechecking and similar stuff if you need that.

gabro:
@sake92 thanks for the link! It looks like an interesting project, but currently I don't think switching from docusaurus would buy us anything.

Using `tut:passthrough` is good enough for our needs and for the rest docusaurus provides a nice experience out of the box.

mads-hartmann:
@gabro I'd like to help you get this merged if possible - is there something specific I can help with? 😊

olafurpg:
@mads-hartmann The blocking issue is primarily content. I took a stab at migrating some of the contents on the current site but made the mistake of starting to write new content from scratch because I wasn't too happy with the current prose.

Another (too) big improvement I wanted to take on is to automatically generating docs for the configuration options using `@metaconfig.annotation.Description` and other Metaconfig annotations. This involves the `ScalafmtConfig` docstrings into annotations and writing methods to print out markdown for the settings.

I wonder what the minimum effort we can do to publish the new website without a regression from the current website. From there we can gradually improve things.

gabro:
I'm currently on vacation, but I've took a glance at the todo list on the top post and it seems accurate. I would start from there, especially installation instructions and config example correctness

mads-hartmann:
@gabro Thanks, I'll have a look - enjoy your vacation ☀️

mads-hartmann:
Just got the site up and running locally. The docs on how to contribute to the site all worked 👍 This is going to be great 👏 I'll see if I can port over the installation instructions.

I checked it out on mobile and are a few tiny glitches but it's still a huge improvement over the current site so I wouldn't block the merge on that.

mads-hartmann:
Just a note to anyone stumbling upon this thread: I have a PR to port over the installation instructions here https://github.com/gabro/scalafmt/pull/2. I think the largest missing piece is "check that all examples in the configuration make sense (there are some inconsistencies)". I'd be happy to just have parity with the current site - we can always improve them later

olafurpg:
There seems to be a bug in tut that `.stripMargin` is applied to code listings before `:passthrough`. I'm upgrading to the latest tut to see if the problem persists and this means upgrading the build to sbt 1 first. I've merged #1247 and will followup with another build cleanup. I will rebase this PR on top of master once that's done.

olafurpg:
I have a WIP branch to fix the TODO for `assumeStandardLibraryStripMargin` here https://github.com/scalameta/scalafmt/compare/master...olafurpg:new-website the branch rebases on top of #1250 (via merge commit). The pipe characters are still not present
screen shot 2018-08-05 at 18 38 40

I suspect tut is pre-processing with `stripMargin`, will need more time to investigate. I tried updating to the latest tut version (which required upgrading to sbt 1) but that did not do the trick.

gabro:
I suspect that's due to interpolation of stripMargined strings inside others, if that makes any sense

--
Gabriele

olafurpg:
I've opened a ticket in tut with a minimal repro https://github.com/tpolecat/tut/issues/235 the fix should be easy, in the meantime I think we can work around it by using another characters than `|` for strip margin.

olafurpg:
The `stripMargin` bug is caused by the Scala REPL! See https://github.com/tpolecat/tut/issues/235#issuecomment-410619210 It's been fixed in 2.13 so it seems we have no option but to work around it on our side. We can probably fake `|` stripMargin strings by using another syntax like 
```scala
val foo = s"""
  # a
  # b
""".stripMargin
```
and process the `#` into `|`. It's a shame

olafurpg:
I pushed a few commits

- workaround `stripMargin` processing by Scala REPL by using `#` instead of `|`
- merge conflicts with master (rebase was too large)
- fill out "rewrite rules" section


The css for side-by-side code examples look a bit weird at the moment

screen shot 2018-08-13 at 00 29 54

- the left padding for "Before" is HUGE
- the config settings are cut off from the code examples

Any ideas how to fix those?

mads-hartmann:
@olafurpg I should be able to fix both of those issues. I’ll try to make time today 👍🏻

olafurpg:
Thanks @mads-hartmann 

Some before/examples are also too large to fit next to each other horizontally. It would be good to have a combinator for writing before/after examples that lie next to each other vertically

screen shot 2018-08-13 at 11 02 24

It looks nicer on the current website

screen shot 2018-08-13 at 11 03 34

olafurpg:
I pushed several commits that migrate the site to use a new project I've been working on past week: https://github.com/olafurpg/vork. I plan to rename the project into "mdoc" and add a few more missing features before making noise about it so please keep quiet about it for now 😉 

The nicest improvements with this change are

- incremental site generation takes ~200ms instead of 30 seconds
- docusaurus live reload doesn't crash on site generation
- no more fiddly multi-line string wrapping of code examples, we can write the code fences like normal instead using the `vork:scalafmt` and `vork:defaults` modifiers (see configuration.md).

I ported the remaining parts of the configuration section, I left a few uninteresting settings out.

TODOs:

- improve rendering of config code, currently it's wrapped in `/* CONFIG */` comments which is a bit confusing (especially when showing `docstrings` setting).
- split up `configuration.md` into multiple pages? It's HUGE at the moment, although I kinda like large pages.
- general CSS/html polish would be awesome if you see anything weird. We can generate arbitrary html in the `vork:scalafmt` and `vork:defaults` blocks.
- render before/after for rewrite rules (currently, it only shows after)
- improve prose (I don't like the language in some of the sections in the current website)


A lot of progress but still quite a bit more to go before we can merge this 💪

olafurpg:
Pro tip if you want to write a custom modifier:
- update `Main.scala` to use `withIncludePath` targeting a single `foo.md` (where `foo.md` is a tiny file)
- keep `~docs/run -w` running in sbt, whenever you change the modifier code, press enter

gabro:
I've pushed a commit that renders configuration blocks in a collapsible section under the example. Take a look :)


ezgif com-optimize 2

gabro:
Code examples now have a button to toggle displaying the original code vs the formatted code. The implementation not wonderful but it's working well enough.

kapture 2018-08-21 at 18 39 52

We need to rewrite some of the examples to make the original code different from the formatted code now :P

olafurpg:
@gabro 😍 That looks amazing

er1c:
@gabro what can I do to help?

gabro:
@er1c if you have some time to spare, we could use a hand in making the code example significant. Meaning: the original code should be realistic and different from the formatted result. This is already true for some examples, but many aren't.

It would also be interesting to beta test the new docs setup :)

MasseGuillaume:
@gabro this is really cool! One small detail I noted: I would remove the Copy button on the code example. I don't think It's useful.

gabro:
@MasseGuillaume yeah, I haven't spent too much time on it. It's on by default on every snippet on docusaurus (recent versions of it).

I can tweak the script to remove it, wdyt @olafurpg?

olafurpg:
I think it makes sense to have copy button for config examples but it's not necessary for the example Scala code

gabro:
👍 I'll work on it (not super hi-pri, I'll take my time to polish the code a little)

gabro:
@olafurpg I've polished all the configuration section a bit, mostly by tweaking all the examples to show meaningful "original" code (vs "formatted"). 

gabro.github.io/scalafmt contains the latest version :)

Also, I suspect mdoc messes up with formatting a bit, since some indentation in the config blocks is lost (see the align.tokens example). This happens after the modifier has passed, since the code produced by the modifier looks fine.

docs/configuration.md
```diff
@@ -0,0 +1,772 @@
...
+```scala mdoc:scalafmt
+rewrite.rules = [RedundantBraces]
+rewrite.redundantBraces.stringInterpolation = true
+---
+q"Hello ${name}"
+```
+
+Configuration options and default values:
+
+// TODO(olafur): multiline defaults
```
gabro: If you're bored this looks like an easy fix

olafurpg: I'll take a look 👍

olafurpg:
Overall, I think this PR is almost ready to got! The only blocking issue for putting this live IMO is that the "copy" button should only be for config, not Scala code examples
screen shot 2018-09-02 at 21 04 57

I agree there is work to do to improve the "show original/formatted" examples. However, the current PR is already such an improvement over the current site I think it's fine to move it to fix that later.

olafurpg:
Regarding publishing the site on merge, I can take care of that in a followup PR using https://olafurpg.github.io/sbt-docusaurus/ For now I propose we disable publish-site-on-merge and we manually publish the site to begin with.

gabro:
> Overall, I think this PR is almost ready to got! The only blocking issue for putting this live IMO is that the "copy" button should only be for config, not Scala code examples

Done!

gabro:
:shipit: It's online!! 🍾 

https://scalameta.org/scalafmt/

PR #1211, opened 2018-06-05
Don't crash on isWindows check, fixes #1207

### LINKED ISSUE #1207: OsSpecific utils use JVM-specific code
- **Version**: 1.6.0-RC1
- **Integration**: using as a library

## Steps

Calling files filter from Scala.js:
```scala
config.project.matcher.matches(...)
```

Crashes with `Cannot read property 'toLowerCase' of null`.

## Problem

It calls to `OsSpecific.fixSeparatorsInPathPattern`:

https://github.com/scalameta/scalafmt/blob/295cee602b9bd6fb1fd6a2aa4aef5a808f272848/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala#L15-L16

Which is not platform independent, if I understand correctly:

https://github.com/scalameta/scalafmt/blob/645e72b71974ce0c42e9a3b957fbbb75ada0045e/scalafmt-core/shared/src/main/scala/org/scalafmt/util/OsSpecific.scala#L4-L7

But is still a shared code: `scalafmt/scalafmt-core/shared/src/main/scala/org/scalafmt/util/OsSpecific.scala`.

## Expectation

I guess this code should be in `scalafmt/scalafmt-core/jvm/src/` and the JS-specific code cannot use `System.getProperty`.

## Workaround

I just defined a matcher without the `OsSpecific.fixSeparatorsInPathPattern` method:

```scala
val matcher = FilterMatcher(
  config.project.includeFilters,
  config.project.excludeFilters,
)
```

olafurpg:
Review @laughedelic

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/OsSpecific.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/OsSpecific.scala
@@ -2,8 +2,10 @@ package org.scalafmt.util
 
 /** Utils related to differences between various operating systems. */
 object OsSpecific {
-  val isWindows: Boolean =
-    System.getProperty("os.name", "unknown").toLowerCase().startsWith("windows")
+  def isWindows: Boolean =
+    // NOTE: org.scalameta:io implements java.io.File for Node.js so this will work correctly
+    // on Node.js + Window and also in the browser.
+    java.io.File.separatorChar == '\\'
 
   def fixSeparatorsInPathPattern(unixSpecificPattern: String): String =
     if (isWindows) unixSpecificPattern.replace("/", """\\""")
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/OsSpecific.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/OsSpecific.scala
@@ -3,8 +3,8 @@ package org.scalafmt.util
 /** Utils related to differences between various operating systems. */
 object OsSpecific {
   val isWindows: Boolean =
-    System.getProperty("os.name").toLowerCase().startsWith("windows")
   val lineSeparator: String = System.getProperty("line.separator")
+    System.getProperty("os.name", "unknown").toLowerCase().startsWith("windows")
 
   def fixSeparatorsInPathPattern(unixSpecificPattern: String): String =
     if (isWindows) unixSpecificPattern.replace("/", """\\""")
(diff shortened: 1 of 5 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/util/OsSpecific.scala
```diff
@@ -3,8 +3,7 @@ package org.scalafmt.util
 /** Utils related to differences between various operating systems. */
 object OsSpecific {
   val isWindows: Boolean =
-    System.getProperty("os.name").toLowerCase().startsWith("windows")
-  val lineSeparator: String = System.getProperty("line.separator")
+    System.getProperty("os.name", "unknown").toLowerCase().startsWith("windows")
```
laughedelic: What about this check when running on Node.js? It's still relevant, but then `System.getProperty("os.name")` won't work. 

This works on Node (probably there's a simpler way):

```scala
scalajs.js.Dynamic.global.require("os").platform().startsWith("win")
``` 

But won't work when running in a browser. 

Anyway it requires having this piece of code split for JVM and JS.

olafurpg: How about `def isWindows = System.lineSeparator == "\r\n"`?

olafurpg: org.scalameta:io actually implements `java.io.File.separator` https://github.com/scalameta/scalameta/blob/5536eb02db562af944f80c11cb171e19508771c4/scalameta/io/js/src/main/scala/java/io/File.scala#L57-L62 we can use that instead, it does the right thing on node.js and assumes `/` in the browser

laughedelic: Good idea! Seems to work both on Node and in the browser: returns `\n` on Mac, hopefully works on Windows as expected.

laughedelic: Sure, if it's already done in scalameta, why not. Then it probably should be used below in `fixSeparatorsInPathPattern` too

olafurpg: Done!

olafurpg:
We implement `java.io.File` and `java.nio.file.{Path,Paths}` for Node.js in Scalameta since it made it easier to run our test suite in JS

PR #1213, opened 2018-06-05
Avoid lookup up match for non-Token.Left{Brace,Paren}

### LINKED ISSUE #1203: key not found
I've tried 1.5.1 and 1.6.0-RC1 and get the same problem. Screenshot attached of the exception that causes this to appear in sbt

```
[error] Error in /home/fommil/Work/.../Recorder.scala: java.util.NoSuchElementException: key not found: -5260204305981373584
```

for the source code

```scala
package foo

object Recorder {
  sealed abstract class Ast[A]
  final case class Record()
}
```

and configuration

```
```

2018-05-30-115201_1918x1078_scrot

olafurpg:
Fixes #1203 

cc/ @fommil

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -751,10 +751,19 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
       leftTok2tok(matching(token)) match {
         case FormatToken(RightParenOrBracket(), l @ LeftParen(), _) =>
           loop(l)
-        // This case only applies to classes
-        case f @ FormatToken(RightBracket(), mod, _) if mod.is[Modifier] =>
-          loop(next(f).right)
-        case FormatToken(r @ RightParenOrBracket(), _, _) => Some(r)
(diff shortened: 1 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -751,10 +751,19 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
...
-        // This case only applies to classes
-        case f @ FormatToken(RightBracket(), mod, _) if mod.is[Modifier] =>
-          loop(next(f).right)
-        case FormatToken(r @ RightParenOrBracket(), _, _) => Some(r)
+        case f @ FormatToken(left @ RightParenOrBracket(), right, _) =>
+          right match {
+            case Modifier() =>
+              // This case only applies to classes
+              next(f).right match {
+                case x @ (_: Token.LeftParen | _: Token.LeftBracket) =>
```
olafurpg: The fix is ugly but it seems to do the job, feel free to refactor @pjrt I'm not sure exactly what this case is about

pjrt: Unn, I haven't looked at this code in a while. But it doesn't look any less ugly than the other stuff we need to do in this file :).

Issue #1200, opened 2018-05-29
java.lang.NoSuchMethodError: fastparse.all$.CharsWhileIn$default$2()I


fommil:
in 1.6.0-RC1 I'm getting this...

I figure this is a configuration / dependency version problem that can probably be eyeballed and creating a minimisation would be a lot of work for little benefit.

```
> fmt
[error] java.lang.NoSuchMethodError: fastparse.all$.CharsWhileIn$default$2()I
[error] 	at scala.meta.internal.tokenizers.XmlParser.<init>(XmlParser.scala:16)
[error] 	at scala.meta.internal.tokenizers.LegacyScanner.getXml(LegacyScanner.scala:990)
[error] 	at scala.meta.internal.tokenizers.LegacyScanner.fetchLT$1(LegacyScanner.scala:347)
[error] 	at scala.meta.internal.tokenizers.LegacyScanner.fetchToken(LegacyScanner.scala:355)
[error] 	at scala.meta.internal.tokenizers.LegacyScanner.nextToken(LegacyScanner.scala:241)
[error] 	at scala.meta.internal.tokenizers.LegacyScanner.foreach(LegacyScanner.scala:1068)
[error] 	at scala.meta.internal.tokenizers.ScalametaTokenizer.uncachedTokenize(ScalametaTokenizer.scala:153)
[error] 	at scala.meta.internal.tokenizers.ScalametaTokenizer.tokenize(ScalametaTokenizer.scala:31)
[error] 	at scala.meta.internal.tokenizers.ScalametaTokenizer$$anon$2.apply(ScalametaTokenizer.scala:287)
[error] 	at scala.meta.tokenizers.Api$XtensionTokenizeDialectInput.tokenize(Api.scala:21)
[error] 	at scala.meta.tokenizers.Api$XtensionTokenizeInputLike.tokenize(Api.scala:10)
[error] 	at scala.meta.internal.parsers.ScalametaParser.scannerTokens$lzycompute(ScalametaParser.scala:183)
[error] 	at scala.meta.internal.parsers.ScalametaParser.scannerTokens(ScalametaParser.scala:183)
[error] 	at scala.meta.internal.parsers.ScalametaParser.<init>(ScalametaParser.scala:140)
[error] 	at scala.meta.internal.parsers.ScalametaParser$$anon$189.apply(ScalametaParser.scala:3462)
[error] 	at scala.meta.parsers.Api$XtensionParseDialectInput.parse(Api.scala:21)
[...]

PR #1218, opened 2018-06-10
First steps to shading sbt-scalafmt

olafurpg:
Towards #1200 

- adds new scalafmt-cli-shaded module that is a fatjar with all scalafmt dependencies shaded in the package `org.scalafmt.shaded`
- remove sbt plugin from the build since it will be moved into a separate repo


ches:
Thank you for shaving this yak. It has resolved a case of #1200 for me.

olafurpg:
Thanks for reporting back @ches I will close #1200 as fixed then 😊

Issue #1222, opened 2018-06-14
Add braces to multi-line lambda body


iantabolt:
This issue occurs when you have a single lambda argument with a single statement multi-line body. Scalafmt will drop the lambda to a newline and indent the block. Please correct me if this is working as intended or if a configuration already exists.

With default configuration,

```scala
// Before scalafmt (and expected output)
foos.map(foo =>
  foo
    .withBar("bar")
    .toString)

// After scalafmt
foos.map(
  foo =>
    foo
      .withBar("bar")
      .toString)
```

Meanwhile these similar scenarios are unchanged as expected:

```scala
foos.map(foo => {
  // ...
})

foos.map { foo =>
  // ...
}
```

P.S. for our style guide, I would actually prefer a rewrite rule to add curly braces around multiline lambda bodies:

```scala
// Before 
foos.map(foo =>
  foo
    .withBar("bar")
    .toString)

// After hypothetical rewrite rule
foos.map(foo => {
  foo
    .withBar("bar")
    .toString
})
```

Do you think this is doable/worthwhile? I'm always happy to contribute if you think it is a good addition.

olafurpg:
Thanks for reporting! Moving the argument to the first line is what https://github.com/scalameta/scalafmt/issues/667 is about, and it's been open for a while and my attempts a while ago to fix it failed. I your style guide would prefer curly braces I think you can implement a rewrite rule for this case quite easily so that 
```scala
// Before 
foos.map(foo =>
  foo
    .withBar("bar")
    .toString)

// After hypothetical rewrite rule
foos.map(foo => {
  foo
    .withBar("bar")
    .toString
})
```

Implementing a rewrite rule would also be a good primer for you to rewrite `Foo { ... }` into `Foo({...})` 😉 


Here's a small session from the repl that might get you started

```scala
@ import $ivy.`com.geirsson::scalafmt-core:1.6.0-RC3`, scala.meta._, org.scalafmt._, rewrite._
import $ivy.$                                      , scala.meta._, org.scalafmt._, rewrite._

@ val c = config.ScalafmtConfig()
c: config.ScalafmtConfig = ScalafmtConfig(
  "1.6.0-RC3",
  80,
  ScalaDoc,
  OptIn(true, false, true, true, true, false),
  BinPack(false, false, false, true, 5, List(".*"), List("String", "Term.Name")),
  ContinuationIndent(2, 4, 4),
  Align(
    false,
    false,
[...]

Issue #667, opened 2017-01-06
Move lambda arg to first line


olafurpg:
This template is a guideline, not a strict requirement.
- **Version**: 0.5
- **Configuration**:
```
style = default
```

## Unexpected formatting output

Obtained, scalafmt formats code like this:
```scala
    val search: BinarySearch = new BinarySearch(
      guess =>
        this.calculateNumbers(
          state.copy(amount = guess, inputType = InputType.Income)))
```

Expected, I would like the formatted output to look like this:
```scala
    val search: BinarySearch = new BinarySearch(guess =>
      this.calculateNumbers(
        state.copy(amount = guess, inputType = InputType.Income)))
```

ochrons:
Another example
```scala
 button(cls := "btn btn-default", onclick := { () => dispatch(Increase(2)) }, "Increase")
```
gets formatted into

```scala
 button(cls := "btn btn-default", onclick := { () =>
          dispatch(Increase(2))
        }, "Increase")
```
while
```scala
 button(cls := "btn btn-default", onclick := (() => dispatch(Increase(2))), "Increase")
```
stays as it is.

Also returning an anon function from a function results in an unwanted line break:
```scala
  def subscribe[T](cursor: ModelR[M, T])(listener: ModelRO[T] => Unit): () => Unit = {
    this.synchronized {
      listenerId += 1
      val id = listenerId
      listeners += id -> Subscription(listener, cursor, cursor.eval(model))
      () =>
        this.synchronized(listeners -= id)
    }
  }
```

olafurpg:
Related https://github.com/olafurpg/scalafmt/issues/357

PR #1225, opened 2018-06-24
Add braces around multi-line lambda body

iantabolt:
Implements rewrite for #1222

If there is a lambda with multiple line body, it will require braces to wrap the body (even though it is a single statement).

Before
```scala
xs.map(foo =>
  foo
    .withBar(bar)
    .toList)
```
After
```scala
xs.map(foo => {
  foo
    .withBar(bar)
    .toList
})
```

Part of the motivation for this is that the current behavior inserts another line before the lambda argument:
```scala
xs.map(
  foo => 
    foo
      .withBar(bar)
      .toList)
```
This is still an open issue, tracked in #667, but this PR sort of adds an alternate solution.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/RedundantBracesSettings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/RedundantBracesSettings.scala
@@ -2,6 +2,19 @@ package org.scalafmt.config
 
 import metaconfig._
 
+/**
+  * @param methodBodies if true, removes redundant braces around method bodies
+  * @param includeUnitMethods if true, also removes redundant braces around method bodies
+  *                           with unit return types
+  * @param maxLines the maximum block size to process (anything larger will be ignored)
+  * @param stringInterpolation if true, removes redundant braces around values interpolated
+  *                            in strings, as in s"Hello, ${name}" => s"Hello $name"
+  * @param generalExpressions if true, removes redundant braces around basically any
+  *                           expression (danger, possibly superfluous)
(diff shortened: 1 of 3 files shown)


iantabolt:
Thanks! I just tested it by running it on our codebase and found some scenarios where it is broken. So I guess consider this a WIP 😞. I'll add more tests and fix the issues.

iantabolt:
I found a pretty important edge case here that is going to be hard to catch. If there are lines that could be interpreted as separate lines, then adding curly braces will make the compiler think they are separate lines (as opposed to without the curly braces where they are forced to be a single line).

```scala
List(10, 11, 12).map(x =>
  x
  - 5
)
// = List(5, 6, 7)

List(10, 11, 12).map(x => {
  x
  - 5
})
// = List(-5, -5, -5)
```

I'm going to close this issue until I take care of this and some other edge cases.

olafurpg:
Good catch @iantabolt ! Infix operators are indeed tricky, the problem generalizes to other non-infix cases however

```scala

foo(a,
b 

(- 5)
)

// vs.

foo(a,
{b

(-5)
})
```
I can't think of a smart solution at the top of my head. A simple but slow solution would be something like

```scala
val candidateBody = termFunction.body
val wrappedInBraces = "{" + candidateBody.syntax + "}"
val isSafeToWrap =
  wrappedInBraces.parse[Term].structure ==
    Term.Block(candidateBody :: Nil).structure
if (isSafeToWrap) // trigger rewrite
else // do nothing
```

I suspect there's a better way to detect when the block parser would newlines differently from the expression parser. I'll ask around

PR #1230, opened 2018-06-29
Fix 1228

### LINKED ISSUE #1228: Removing trailing comma also removes closing parens when next line starts directly with a comment
OK, I know the title is long and a bit weird, but so is this bug.

- **Version**: 1.6.0-RC3
- **Integration**: SBT or CLI
- **Configuration**: none (all default, .scalafmt.conf is empty)

## Steps

Given code like this:
```scala
class Stuff {
  def id(str: String): String = str
  val fields = Seq(
    id("name"),
// this is a comment
  )
}
```
Notice that the comment starts on the very first character of the line (no space before it).

I run scalafmt through CLI or SBT:
```bash
scalafmt
sbt scalafmt
```

## Problem

```scala
class Stuff {
  def id(str: String): String = str
  val fields = Seq(
    id("name"
// this is a comment
  )
}
```
A parens has been lost !

## Expectation

```scala
class Stuff {
  def id(str: String): String = str
  val fields = Seq(
    id("name")
// this is a comment
  )
}
```

The trailing comma has been removed, but the parens is still there. I don't really care what happens the comment.

## Workaround

In the source file, just having a space before the comment helps.

Or in the configuration:
```
trailingCommas = preserve
```

gabro:
Fix #1228

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -194,6 +194,15 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
   def nextNonComment(curr: FormatToken): FormatToken =
     nextNonCommentWithCount(curr)._2
 
+  def nextNonBlank(curr: FormatToken): FormatToken = {
+    if (!(curr.right.is[Space] || curr.right.is[Tab])) curr
+    else {
+      val tok = next(curr)
+      if (tok == curr) curr
+      else nextNonBlank(tok)
+    }
+  }
+
   @tailrec
   final def rhsOptimalToken(start: FormatToken): Token = start.right match {
     case Comma() | LeftParen() | RightParen() | RightBracket() | Semicolon() |
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -194,15 +194,6 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
   def nextNonComment(curr: FormatToken): FormatToken =
     nextNonCommentWithCount(curr)._2
 
-  def nextNonBlank(curr: FormatToken): FormatToken = {
-    if (!(curr.right.is[Space] || curr.right.is[Tab])) curr
-    else {
-      val tok = next(curr)
-      if (tok == curr) curr
-      else nextNonBlank(tok)
-    }
-  }
-
   @tailrec
   final def rhsOptimalToken(start: FormatToken): Token = start.right match {
     case Comma() | LeftParen() | RightParen() | RightBracket() | Semicolon() |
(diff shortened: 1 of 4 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -194,6 +194,15 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
   def nextNonComment(curr: FormatToken): FormatToken =
     nextNonCommentWithCount(curr)._2
 
+  def nextNonBlank(curr: FormatToken): FormatToken = {
+    if (!(curr.right.is[Space] || curr.right.is[Tab])) curr
```
olafurpg: Can this ever return true? `FormatToken.{right,left}` are always non-whitespace.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -411,6 +412,7 @@ class FormatWriter(formatOps: FormatOps) {
       // Insert a comma after b
       case TrailingCommas.always
           if !left.is[Comma] && !left.is[Comment] &&
+            !nextNonBlank(formatToken).right.is[Comment] &&
```
olafurpg: Does this fail with `!formatToken.right.is[Comment]`?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -445,9 +450,10 @@ class FormatWriter(formatOps: FormatOps) {
       //
       // Remove the comma after b (before comment)
       case TrailingCommas.never
-          if left.is[Comment] && prev(formatToken).left.is[Comma] &&
+          if left.is[Comment] && prevFormatToken.left.is[Comma] &&
             right.is[CloseDelim] && isNewline =>
-        sb.deleteCharAt(sb.length - left.syntax.length - 2)
+        sb.deleteCharAt(
```
olafurpg: I am wondering if it's possible to prevent printing the trailling comma in the first place instead of deleting the characters afterwards 🤔

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
gabro: Right, I've made that fix before knowing it, so I'm sure there's a better way. I'll try.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
gabro: ah! That works, thanks. I was trying with `next(formatToken)` but that failed

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
gabro: I've tried a bit when sketching the first implementation, but then I got too confused 😓. I agree it's not airtight, but at least now we're computing the index dynamically instead of hardcoding numbers...

Issue #1224, opened 2018-06-19
Use trailingCommas=always with verticalMultilineAtDefinitionSite creates a comma where it shouldn't


lespea:
This template is a guideline, not a strict requirement.
- **Version**: 1.6.0-RC3
- **Integration**: SBT
- **Configuration**:
```
verticalMultilineAtDefinitionSite = true
trailingCommas = always
```

## Steps

Given code like this:
```scala
case class Test()(a1: Int, a2: Int, a3: Int, a4: Int, a5: Int, a6: Int, a7: Int, a8: Int, a9: Int)
```

## Problem

Scalafmt formats code like this:
```scala
case class Test(,
  )(a1: Int,
    a2: Int,
    a3: Int,
    a4: Int,
    a5: Int,
    a6: Int,
    a7: Int,
    a8: Int,
    a9: Int)
```

## Expectation

I would like the formatted output to look like this:
```scala
case class Test()(a1: Int,
    a2: Int,
    a3: Int,
    a4: Int,
    a5: Int,
    a6: Int,
    a7: Int,
    a8: Int,
    a9: Int)

```

The first parens are empty so shouldn't have a comma in them

olafurpg:
Thanks for reporting! This looks like a valid bug. A workaround is to use `trailingCommas = preserve`. If someone wants to take a stab a this, a good place to start is likely this condition here

https://github.com/scalameta/scalafmt/blob/4b198a1865f62c73e8930522274ecc323d203135/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala#L395-L396

It's possible to query the number of arguments with this method here

https://github.com/scalameta/scalafmt/blob/2a7dc2ad849ea982a7774a93062403db6e2f043a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala#L364

olafurpg:
I am able to reproduce with `verticalMultilineAtDefinitionSite = true` but not without it.

PR #1241, opened 2018-07-29
Fix combination of trailingCommas=always and verticalMultilineAtDefnSite=true

tanishiking:
Fixes https://github.com/scalameta/scalafmt/issues/1224

In the original issue, the expected formatted code is something like the below one. (doesn't put newline in the first empty parentheses). However, whether putting a newline in the empty parentheses is an issue of the verticalMultiline.
This PR fixes trailingCommas=always not to break compilation by checking the target formatToken is surrounded by empty parentheses

```scala
case class Test()(a1: Int,
    a2: Int,
    a3: Int,
    a4: Int,
    a5: Int,
    a6: Int,
    a7: Int,
    a8: Int,
    a9: Int)
```


tanishiking:
Thank you for reviewing! Rebased 😄

Issue #1242, opened 2018-07-29
Align comments in the middle of method chain


tanishiking:
This template is a guideline, not a strict requirement.
- **Version**: 1.6.0-RC4
- **Integration**: sbt
- **Configuration**:
```
align=more
```

I want scalafmt (with align=some) to be able to align comments in a method chain like scalafmt do for `case` or `var/val`

## Steps

Given code like this:
```scala
foo
  .x() // comment1
  .xx() // comment2
  .xxx() // comment3
```

When I run scalafmt like this:

```bash
$ sbt scalafmt
```

## Problem

Scalafmt formats code like this:
```scala
foo
  .x() // comment1
  .xx() // comment2
  .xxx() // comment3
```

## Expectation

I would like the formatted output to look like this:
```scala
>>>
foo
  .x()     // comment1
  .xx()   // comment2
  .xxx() // comment3

```

## Workaround

There are no ways to do that as far as I know.

PR #1243, opened 2018-07-29
Align comments in the middle of method chain

tanishiking:
This PR addresses https://github.com/scalameta/scalafmt/issues/1242

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -704,7 +704,7 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
   private final def vAlignDepthUnCached(tree: Tree): Int = {
     val count: Int = tree match {
       // infix applications don't count towards the length, context #531
-      case _: Term.ApplyInfix => 0
+      case _ @(_: Term.Apply | _: Term.ApplyInfix | _: Term.Select) => 0
       case _ => 1
     }
     tree.parent match {
--- a/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
+++ b/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
@@ -420,3 +420,13 @@ Seq(
   "org.scalatest"      % "scalatest_2.11"               % "3.0.0" % Test,
   "org.scalamock"      %% "scalamock-scalatest-support" % "3.2.2" % Test
 )
+<<< align coments in method chain
+foo
+  .x() // comment1
+  .xx() // comment2
+  .xxx() // comment3
+>>>
+foo
+  .x()   // comment1
+  .xx()  // comment2
+  .xxx() // comment3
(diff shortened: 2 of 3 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -704,7 +704,7 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
   private final def vAlignDepthUnCached(tree: Tree): Int = {
     val count: Int = tree match {
       // infix applications don't count towards the length, context #531
-      case _: Term.ApplyInfix => 0
+      case _ @(_: Term.Apply | _: Term.ApplyInfix | _: Term.Select) => 0
```
tanishiking: Without this change, line2 and line3 of the below code [won't be in same block](https://github.com/scalameta/scalafmt/blob/38de09b9e7b32c8bc5b2e658b721e2d64d96570b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala#L333), because `vAlignDepth` of line2 and line3 are not equal, and `columnsMatch` would return 0
https://github.com/scalameta/scalafmt/blob/38de09b9e7b32c8bc5b2e658b721e2d64d96570b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala#L286

```scala
foo
  .x() // comment1
  .xx() // comment2
```

scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
```diff
@@ -420,3 +420,13 @@ Seq(
   "org.scalatest"      % "scalatest_2.11"               % "3.0.0" % Test,
   "org.scalamock"      %% "scalamock-scalatest-support" % "3.2.2" % Test
 )
+<<< align coments in method chain
```
olafurpg: Can we add more tests to see if the change has unintended consequences? For example, I'm curious to know what happens in the following cases

```scala
// Test 1: separate statements
val ys = xs
  .filter(_ > 2) // comment1
  .map(x => x + 1) // comment2
ys.map(y => y + 1) // comment3

// Test 2: Missing handling for Term.ApplyType
foo
   .x[T]() // comment1
   .xx[T]() // comment2

// Test 3: iirc, this didn't align before but maybe it's been fixed now
a.b(
  1, // comment1
  1 // comment2
)

// Test 4: alignment of random dots

a.b.c.d
aa.bb.cc.dd

// Test 4: alignment of random parentheses

a().b().c().d
aa().bb().cc().dd
```

Have you tried running this change on your codebase? I would be interested to see a large diff from a representative codebase. I personally don't use vertical alignment so I don't know what behaviour makes most sense.

tanishiking: OK, I'll try running this chage on some codebases 👍  (I've done only on my private project yet...) and add some tests like you mentioned! Thank you for your advice.

tanishiking:
Hmm, I tried this change on some projects and found that this contains some significant regressions... (these tests all fails https://github.com/scalameta/scalafmt/pull/1243/commits/8e6ecb663ae3b8bec642ed897bcf9bbfe23b8571)
I'll close this PR this time and try to fix it. Thank you for your comments!

olafurpg:
Thanks for investigating @tanishiking I believe my proposal in https://github.com/scalameta/scalafmt/pull/1243#pullrequestreview-142073794 to special case `Token.Comment` might address the regressions

er1c:
@tanishiking did the latest comments from @olafurpg make sense? (they don't to me)  just curious if you tried it to resolve the issues in this PR

Issue #1254, opened 2018-08-14
trailingCommas: always behaves incorrectly for comments with alignment whitespace


er1c:
Using latest master branch 2018-08-15

[Update], I identified a more simple test/issue.  If there is a comment with additional alignment spacing, the position of the comma is calculated incorrectly.

I added the following to `trailingCommasAlwaysComments.stat`:

```
<<< should consider comments with extra whitespace when adding trailing commas
def method(
    abc: String, // a comment
    d: String   // a comment
) = ???
>>>
def method(
    abc: String, // a comment
    d: String, // a comment
) = ???
```

diff:

```
diff --git a/scalafmt-tests/src/test/resources/trailing-commas/trailingCommasAlwaysComments.stat b/scalafmt-tests/src/test/resources/trailing-commas/trailingCommasAlwaysComments.stat
index 43c0023b..986eb09b 100644
--- a/scalafmt-tests/src/test/resources/trailing-commas/trailingCommasAlwaysComments.stat
+++ b/scalafmt-tests/src/test/resources/trailing-commas/trailingCommasAlwaysComments.stat
@@ -12,6 +12,18 @@ def method(
     b: String, // a comment
 ) = ???

+
+<<< should consider comments with extra whitespace when adding trailing commas
+def method(
+    abc: String, // a comment
+    d: String   // a comment
+) = ???
+>>>
+def method(
+    abc: String, // a comment
+    d: String, // a comment
+) = ???
+
 <<< should consider indented comments on the next line when adding trailing commas
 def method(
     a: String,
```

This generates:

```
[info] - trailing-commas/trailingCommasAlwaysComments.stat: should consider comments with extra whitespace when adding trailing commas| *** FAILED ***
[info]   org.scalafmt.Error$FormatterOutputDoesNotParse: Formatter output does not parse:
[info] def method(
[info]     abc: String, // a comment
[info]     d: Stri,ng // a comment
[info] ) = ???
[info] ^
```

See how the comma "," gets positioned incorrectly by the two "extra" whitespace characters in the original `d: String[   ]// a comment`.  If you add more spaces, the comma gets shifted by exactly the N "extra" space characters.

er1c:
I think this is symptomatic of some other off-by N issue for adding the trailing comma.  Here are some other examples where the comma is getting added in the incorrect place:

From:
```
  private def lookupPartAndFitmentsByLegacyUrlKey(
    legacyUrlKey: String,
    includeFitments: Boolean
[...]

PR #1255, opened 2018-08-15
Add two more trailingCommas tests for when comments have additional whitespace alignment

er1c:
Related issue: https://github.com/scalameta/scalafmt/issues/1254

Related code: https://github.com/scalameta/scalafmt/blob/fd0df2a8d6ce94815fef87ff8257eb5f5e8d63e6/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala#L412-L462

It probably wouldn't hurt to think about some more complicated tests that include alignment formatting too?  I'm just not super familiar enough with scalafmt to create them yet.  This is just an example of scratching my own itch as we are exploring adding scalafmt to our codebase, and this causes the resulting code to not compile if `trailingCommas: always` is set.

Passes the existing formatTests

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -31,6 +31,7 @@ class FormatWriter(formatOps: FormatOps) {
   def mkString(splits: Vector[Split]): String = {
     val sb = new StringBuilder()
     var lastState = State.start // used to calculate start of formatToken.right.
+    var previousLeftWhitespace = ""
     reconstructPath(tokens, splits, debug = false) {
       case (state, formatToken, whitespace) =>
         formatToken.left match {
@@ -52,7 +53,16 @@ class FormatWriter(formatOps: FormatOps) {
             sb.append(rewrittenToken)
         }
 
-        handleTrailingCommasAndWhitespace(formatToken, state, sb, whitespace)
+        handleTrailingCommasAndWhitespace(
+          formatToken,
+          state,
+          sb,
+          whitespace,
+          previousLeftWhitespace)
+
+        // Need to keep track of the actual previous left token whitespace,
+        // Since prevFormatToken.between.length is the original length
+        previousLeftWhitespace = whitespace
 
         formatToken.right match {
           // state.column matches the end of formatToken.right
@@ -387,7 +397,8 @@ class FormatWriter(formatOps: FormatOps) {
       formatToken: FormatToken,
       state: State,
       sb: StringBuilder,
-      whitespace: String): Unit = {
+      whitespace: String,
+      previousLeftWhitespace: String): Unit = {
 
     import org.scalafmt.config.TrailingCommas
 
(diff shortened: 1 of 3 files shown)


gabro:
@er1c thanks for taking a stab at this. I've been on vacation and I'm currently catching up with some work, but I'll try to take a look at this soon :)

er1c:
@gabro I was attempting to figure out if the more "correct" approach to fix the `prevFormatToken.between` was possible, but this: https://github.com/scalameta/scalafmt/blob/fd0df2a8d6ce94815fef87ff8257eb5f5e8d63e6/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala#L157 would need to have a better way to keep track of the "corrected" prevToken.  Perhaps some type of `foldLeft` instead of `foreach` in order to keep track of the "previous token"/update it with it's resulting whitespace, and then use that instead of `val previous = locations(Math.max(0, i - 1))` ? to update the current token to a new previous token? or some such?

gabro:
That's an interesting problem and I'm not super familiar with the internals myself.

For this specific case, maybe we could assume that comments are always normalized to have exactly one space, so we could hardcode 1. This breaks when there are comments on newlines involved, however, so it may be necessary so special-case them.

Paging @olafurpg on this one

gabro:
Here's a horrible fix that implements the solution I've mentioned above.

https://github.com/scalameta/scalafmt/compare/master...gabro:fix-1254

er1c:
I'm trying to figure out if there is an option for comment alignment.  My original code aligns the comments, making it a little nicer to read.  If not, then this would work.  I might try creating tests that include `align = most` and check the behavior next week

gabro:
That's a good idea.
And it would indeed be better to not rely on the implicit knowledge of the comments not being aligned, which it may change in the future.

er1c:
@gabro 

```
maxColumn = 30
trailingCommas = always
align = most

<<< should consider comments with extra whitespace when adding trailing commas
def method(
    abc: String, // abc comment
    d: String    // d comment
) = ???
>>>
def method(
    abc: String, // abc comment
    d: String,   // d comment
) = ???
```

returns back:

```
[info] - trailing-commas/trailingCommasAlwaysVerticalAlignComments.stat: should consider comments with extra whitespace when adding trailing commas| *** FAILED ***
[info]   Idempotency violated
[info]
[info]   ===========
[info]   => Obtained
[info]   ===========
[info]   def method(
[info]       abc: String, // abc comment
[info]       d: String,   // d comment
[info]   ) = ???
[info]
[info]
[info]   =======
[info]   => Diff
[info]   =======
[info]        abc: String, // abc comment
[info]   -    d: String,   // d comment
[info]   +    d: String, // d comment
[info]    ) = ??? (FormatTests.scala:79)
```


I'm a little confused because the "Obtained" looks like it has the alignment, but I'm not sure why the diff shows the single space and the "Idemptency violated" error

gabro:
The error message is not super clear, but it means that running scalafmt again on the obtained output didn't yield the same result, hence idempotency is violated.

For a test to pass running scalafmt twice on the original code should be identical to running it once.

er1c:
Interesting, as another reference, there is whitespace alignment in this test: https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat#L171-L182

er1c:
@gabro this: https://github.com/scalameta/scalafmt/pull/1260 will break your https://github.com/scalameta/scalafmt/compare/master...gabro:fix-1254

gabro:
That's for the best, it was a horrible fix anyway 😅

er1c:
I'm going to close this in favour of https://github.com/scalameta/scalafmt/compare/master...tanishiking:fix-trailing-comma-with-comment?expand=1

PR #1257, opened 2018-08-18
Add ignore and allElementsOf to AvoidInfix

joan38:


joan38:
Thanks @olafurpg 
(Ahah, I guess it feels weird that it tries to imitate the english language but sometimes it can't 😄)

olafurpg:
My goto solution for comparing data is multiline string comparison like here https://github.com/scalameta/metals/blob/b2700d293af96f07150389897346b86d315ad8f9/metals/src/test/scala/tests/provider/DiagnosticsProviderTest.scala#L65-L138 This produces the most helpful error messages in my experience, and makes for readable tests 😄

PR #1260, opened 2018-08-20
Vertical align adjacent single lines of comment

tanishiking:
Addresses https://github.com/scalameta/scalafmt/issues/1242

The reason why scalafmt doesn't align the adjacent single line of comments which appears in method chains or parameters with `align=more` is that [columnsMatch](https://github.com/scalameta/scalafmt/blob/bcfe2b296144e6bf6fac1c253b8c12fc7031ca12/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala#L277-L294) in `FormatWriter.scala` doesn't matches the two adjacent `//`s.

There are mainly two factors that makes `columnsMatch` doesn't match two `//`s.

---

The first one is that [vAlignDepth(row1Owner) == vAlignDepth(row2Owner)](https://github.com/scalameta/scalafmt/blob/bcfe2b296144e6bf6fac1c253b8c12fc7031ca12/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala#L285-L286
) become usually `false` when we are trying to match the two `//`s which appears in method chains or parameters like below.

```scala
x
  .foo // comment1
  .bar  // comment2

def method(
  a: String, // comment3
  b: Int // comment4
)
```

I tried to adjust the depth of them in https://github.com/scalameta/scalafmt/pull/1243 by stop incrementing the depth on `Term.Apply`, `Term.Select` and so on.
However, as @olafurpg advised in https://github.com/scalameta/scalafmt/pull/1243#pullrequestreview-142073794, it is more natural and safe to just ignore the vAlignDepth of the rowOwners when we try to align single-line comments.

---

The another one is that the `key` of `row1.formatToken.right` and `row2.formatToken.right` sometimes differs when we try to align single-line comments which appear in method chains, because some of them are `Term.Apply` and some are `Term.Select`.

https://github.com/scalameta/scalafmt/blob/bcfe2b296144e6bf6fac1c253b8c12fc7031ca12/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala#L287

I tried to adjust it in https://github.com/scalameta/scalafmt/pull/1243 by grouping the keys of `Term.Apply` and `Term.Select`.
However, as advised in https://github.com/scalameta/scalafmt/pull/1243#pullrequestreview-142073794, its better to just ignore the tree owners.

---

Therefore, in this PR, we just ignore the `vAlignDepth` and tree owners in `columnsMatch` when we try to align single-line comments.

### DIFF
--- a/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
+++ b/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
@@ -420,3 +420,67 @@ Seq(
   "org.scalatest"      % "scalatest_2.11"               % "3.0.0" % Test,
   "org.scalamock"      %% "scalamock-scalatest-support" % "3.2.2" % Test
 )
+<<< align comments in method chain
+foo
+  .x() // comment1
+  .xx() // comment2
+  .xxx() // comment3
+>>>
+foo
+  .x()   // comment1
+  .xx()  // comment2
+  .xxx() // comment3
+<<< should align comments on separate statements
+object test {
+  val ys = xs
+    .filter(_ > 2) // comment1
+    .filter(x => x + 1) // comment2
+  ys.map(y => y + 1) // comment3
+}
+>>>
+object test {
+  val ys = xs
+    .filter(_ > 2)      // comment1
+    .filter(x => x + 1) // comment2
+  ys.map(y => y + 1)    // comment3
+}
+<<< align comments on type infix
+foo
+  .x[T]() // comment1
+  .xx[T]() // comment2
+>>>
+foo
+  .x[T]()  // comment1
+  .xx[T]() // comment2
+<<< align comments on params
+a.b(
+  1, // comment1
+  123  // comment2
+)
+>>>
+a.b(
+  1,  // comment1
+  123 // comment2
+)
+<<< should align comments on params of methods
+def method(
+    abc: String, // abc comment
+    d: String // d comment
+) = ???
+>>>
+def method(
+    abc: String, // abc comment
+    d: String    // d comment
+) = ???
+<<< shouldn't align comments on the line which only has a comment
+object x {
+  println("foo bar") // comment
+  // comment
+  println("foo") // comment
+}
+>>>
+object x {
+  println("foo bar") // comment
+  // comment
+  println("foo") // comment
+}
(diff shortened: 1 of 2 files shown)


er1c:
@tanishiking thank you!

tanishiking:
> I apologize the slow responses :pray:

That's alright:+1: Thank you for your quick checking. I know you seem so busy with your projects including scalafmt.
I'll help you by addressing the issues if I can :)

scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
```diff
@@ -420,3 +420,67 @@ Seq(
...
+    d: String // d comment
+) = ???
+>>>
+def method(
+    abc: String, // abc comment
+    d: String    // d comment
+) = ???
+<<< shouldn't align single lines of comment if a comment doesn't have owner
+object x { // comment
+  println("foo bar") // comment
```
olafurpg: My original intuition was this test should fail. The reason the comments don't align is not because they have no owner (all tokens have an owner) but instead

- the `x {` split is an "alignment candidate" causing line 476 and 477 to have mismatching number of alignment candidates
- it seems the alignment algorithm doesn't support aligning tokens that start new lines


Here is a test case that failed before this PR and passes after this PR.

```scala
<<< align comments with different owners and different depth
{
  val a =
    1 // comment
  foo() // comment
}
>>>
{
  val a =
    1   // comment
  foo() // comment
}
```

Are you OK with this behavior? I have generally tried to avoid false positives for alignment but I don't personally use vertical alignment.

tanishiking: >  because they have no owner (all tokens have an owner) but instead
> - the x { split is an "alignment candidate" causing line 476 and 477 to have mismatching number of alignment candidates
> - it seems the alignment algorithm doesn't support aligning tokens that start new lines

Aah, I misunderstood the cause, that's right.
I think what I want to test is the latter one. (I think we don't have to test the former one case) so I'll change the test case to below one.

```scala
<<< shouldn't align comments on the line which only has a comment
object x {
  println("foo bar") // comment
  // comment
  println("foo") // comment
}
>>>
object x {
  println("foo bar") // comment
  // comment
  println("foo") // comment
}
```

---

> Are you OK with this behavior? 

Yes, it's alright with the behavior, because

- Although the comments have different owners, we can infer they are somehow related each other if they are vertically adjacent, and I think it's acceptable to align them.
- it is quite difficult to avoid the behavior :(
- gofmt (which can align the comments) also has such a behavior (https://play.golang.org/p/UQbqCFY_iFN)

olafurpg: > gofmt (which can align the comments) also has such a behavior 

Interesting. 

I'm happy with going this route. I think we might be able to avoid

```scala
val x =
  2     // a
foo()   // a
```

but keep

```scala
val x =  2 // a
foo()      // a
```

But maybe there's no need for that, maybe people just like it that way 🤷 We can discuss it if people complain about over-aggressive alignment.

PR #1270, opened 2018-09-02
Add website landing page

### LINKED ISSUE #1269: Website landing page
In #1199 we've introduced a new website which brings directly to the Docs section. 

It would be nice to have a landing page alà docusaurus, with a summary of the main features.

gabro:
Closes #1269

Still WIP, and just a scaffold for now.
I've brought back the original docusaurus landing page along with the users page.


website/pages/en/index.js
```diff
@@ -0,0 +1,213 @@
...
+        content: 'Talk about trying this out',
+        image: imgUrl('docusaurus.svg'),
+        imageAlign: 'left',
+        title: 'Try it Out',
+      },
+    ]}
+  </Block>
+);
+
+const Description = () => (
```
gabro: I think most of the content that currently lives in "Introduction" should be move to the landing page. This block is a good place where to put it.

website/siteConfig.js
```diff
@@ -4,6 +4,15 @@
...
 const baseUrl = '/scalafmt/';
 
+const users = [
+  {
+    caption: 'buildo',
+    image: 'https://www.buildo.io/static/media/logo.60e4e7fc.svg',
+    infoLink: 'https://www.buildo.io',
+    pinned: true,
+  },
+];
```
gabro: I've picked a (not-so-random) company just to showcase this feature.

I think it makes sense for Scalafmt to showcase its users! We can start with a shortlist, then tweet about it and have people to add themselves (the edit button is already in place in the full "users" page.

website/siteConfig.js
```diff
@@ -4,6 +4,15 @@
 const repoUrl = 'https://github.com/scalameta/scalafmt';
 const baseUrl = '/scalafmt/';
 
+const users = [
+  {
+    caption: 'buildo',
+    image: 'https://www.buildo.io/static/media/logo.60e4e7fc.svg',
+    infoLink: 'https://www.buildo.io',
+    pinned: true,
```
gabro: `pinned` controls whether a user ends up on the homepage. It probably makes sense to have some famous projects/companies here.

website/siteConfig.js
olafurpg: We can start with this list here https://github.com/scalameta/scalafmt/blob/72def4a91ae4b68d6d1bcfc17669b7f1ae663b47/readme/src/main/scala/org/scalafmt/readme/Adopters.scala

We can then ping the people who contributed the original PR to that file to update the logo URL. Let's not add the logos ourselves, since some companies have strict policies about where their logo is used.

website/pages/en/index.js
olafurpg: I'm personally not too happy with the contents in the current introduction. I would prefer to write new stuff. The main points I'd want to highlight are

- saves you time and energy. Focus on code instead of formatting.
- works with many editors and build tools
- simple to enforce in CI

olafurpg:
Can you take a look @gabro ?

docs/introduction.md
```diff
@@ -3,51 +3,4 @@ id: introduction
...
-time. Bug reports, feature requests, questions and PRs are welcome. Complaints
-and unfriendly attitude is not welcome.
-
-Curious to learn more about scalafmt? Check out this talk:
-
-<iframe src="https://player.vimeo.com/video/165929294" width="100%" style="height: 28em;" frameborder="0" webkitallowfullscreen mozallowfullscreen allowfullscreen></iframe>
-
-[bob]:
-  http://journal.stuffwithstuff.com/2015/09/08/the-hardest-program-ive-ever-written/
+Visit [installation docs](installation.md)
```
gabro: Do we even need this page?

olafurpg: This is the current active landing page so there may be links to it. Better to have something than 404s. The page is not listed in the sidebar.

gabro: Makes sense. I'll do another pass later today and I'll merge

PR #1271, opened 2018-09-03
Publish website with CI

### LINKED ISSUE #1267: Use sbt-docusaurus for the website
After #1199, we should migrate to [sbt-docusaurus](https://github.com/olafurpg/sbt-docusaurus) for reducing boilerplate and automating the publishing via Travis

gabro:
Closes #1267 

Yet untested in the CI, but I've successfully published with `sbt docusaurusPublishGhPages` locally.


bin/bintray.sh
```diff
@@ -1,22 +0,0 @@
-#!/usr/bin/env bash
```
gabro: I took the chance of removing a bunch of these scripts, as the seemed unused.

Let me know if I've touched something that I shouldn't have, @olafurpg

.travis.yml
```diff
@@ -18,6 +18,15 @@ jobs:
       script: sbt ci-test
     - stage: release
       script: sbt ci-release
+    - env:
```
gabro: not sure about introducing a second job.
Maybe it's better if we publish the website only if the release succeeds?

bin/bintray.sh
olafurpg: Very nice to see these scripts gone 🙏

.travis.yml
olafurpg: Let's start with this, maybe we'll join the two later.

olafurpg:
Thank you @gabro !

PR #1273, opened 2018-09-03
Use sbt-docusaurus 0.1.2

gabro:
Let's see whether this fixes the CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ jobs:
     - env:
         - USER_EMAIL=olafurpg@gmail.com
       stage: release
-      script: sbt docusaurusPublishGhpages
+      script: sbt docs/docusaurusPublishGhpages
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     - env:
         - USER_EMAIL=olafurpg@gmail.com
       stage: release
-      script: sbt docusaurusPublishGhpages
+      script: sbt docs/docusaurusPublishGhpages
```
olafurpg: Have you checked if this works? The 0.1.2 release didn't do changes here.

gabro: nope, I'm using this as a placeholder. I suspect 0.1.2 isn't published yet, is it?

olafurpg: It's already released, might take a while to sync to Maven Central

Issue #1223, opened 2018-06-14
Migration from 1.3.0 to 1.6.0-RC3 bad continuationIndent


mkrajc:
Hello I am migrating scalafmt  for our project from `1.3.0` to `1.6.0-RC3`. I want the upgrade to be as compatible as possible but I am failing on some files, that has different indentation like in previous version.

## Configuration 1.3.0 
```
style = intellij
maxColumn = 120
lineEndings = unix
align.tokens = [off]
rewrite.rules = [RedundantBraces, RedundantParens, SortImports, PreferCurlyFors]
newlines.sometimesBeforeColonInMethodReturnType = false
align.openParenCallSite = false
align.openParenDefnSite = false
unindentTopLevelOperators = true
spaces.inImportCurlyBraces = true
danglingParentheses = true
```
## Configuration 1.6.0-RC3
I just removed ones that should be defaulted in intellij style and turn on preserve for `trailingCommas` so 
update will have minimal impact on code base.
```
style = intellij
maxColumn = 120
lineEndings = unix
align.tokens = [off]
rewrite.rules = [RedundantBraces, RedundantParens, SortImports, PreferCurlyFors]
newlines.sometimesBeforeColonInMethodReturnType = false
unindentTopLevelOperators = true
spaces.inImportCurlyBraces = true
trailingCommas = preserve
``` 

## Steps

Given snippets like these formatted with `1.3.0`: (note 2 space indent):
```scala
    val workerFactory: WorkerFactory = connectorFactory andThen SearchWorker.props(
      articleStore,
      config.articleStore.maxParallelism
    )
```
or
```scala
      val addressColumns = addressLines :: List(
        address.city,
        address.postalCode,
        address.stateOrProvince,
        address.country
      ).map(_.getOrElse(""))
```
are reformatted with 4 space indent with 1.6.0-RC3, although intellij style have 2 spave for both call and defn site continuationIndent. 
Mismatch is only on some files (like 40 out of 1200) and usually the bad formatted code is part of some expression like `andThen` or `::` as in example above

## Problem

Scalafmt formats code like this:
```scala
    val workerFactory: WorkerFactory = connectorFactory andThen SearchWorker.props(
        articleStore,
        config.articleStore.maxParallelism
      )
```
```scala
      val addressColumns = addressLines :: List(
          address.city,
          address.postalCode,
          address.stateOrProvince,
          address.country
        ).map(_.getOrElse(""))
```

## Expectation
[...]

PR #1282, opened 2018-09-14
Fix for #1223 regression in unindentTopLevelOperators/continuationIndent

er1c:
All previous tests pass, and added the new tests based upon the PR.  Fixes regression #1223

### DIFF
--- a/scalafmt-tests/src/test/resources/test/IndentOperator.stat
+++ b/scalafmt-tests/src/test/resources/test/IndentOperator.stat
@@ -112,3 +112,25 @@ val x =
 val x = 1 +
   2 +
   3
+<<< #1223 infix with call
+val x = y :: List(
+a,
+b,
+c
+)
+>>>
+val x = y :: List(
+  a,
+  b,
+  c
+)
+<<< #1223 infix operator two
+  val x = y andThen List(
+  a,
+  b
+  )
+>>>
+val x = y andThen List(
+  a,
+  b
+)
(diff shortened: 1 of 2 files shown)


scalafmt-tests/src/test/resources/test/IndentOperator.stat
```diff
@@ -112,3 +112,25 @@ val x =
...
+b,
+c
+)
+>>>
+val x = y :: List(
+  a,
+  b,
+  c
+)
+<<< #1223 infix operator two
```
olafurpg: The following test case passed before this PR and fails after this change
```
<<< ONLY #1223 2 infix operator two
  val x = 1 + 2 +
    3
>>>
val x = 1 + 2 +
  3
```

The failure message
```
[info] - test/IndentOperator.stat: #1223 2 infix operator two                  | *** FAILED ***
[info]   ===========
[info]   => Obtained
[info]   ===========
[info]   val x = 1 + 2 +
[info]   3
[info]
[info]
[info]   =======
[info]   => Diff
[info]   =======
[info]    val x = 1 + 2 +
[info]   -3
[info]   +  3 (FormatTests.scala:81)
```

olafurpg:
I opened https://github.com/scalameta/scalafmt/pull/1286 to revert the original regression. Thank you @er1c for the help! I wish we could reach a better solution but I'm afraid until somebody is willing to take maintainership of `unindentTopLevelOperators=true` (the option is not even documented on the website) I prefer to keep it unchanged from v1.5.1. I don't know what the expected behavior of this option should be.

scalafmt-tests/src/test/resources/test/IndentOperator.stat
er1c: I'm looking at this again, and this behavior might actually be correct/desireable.  I think the infix operators (+, *, /, -) just need to be added to the default exclude list.

PR #1283, opened 2018-09-20
Scalameta 4.0.0

eed3si9n:
I would like a Scalameta 4 version of Scalafmt.


olafurpg:
I'm not happy about having to bring in ScalaPB for 2.11. We have discussed separating the syntactic Scalameta APIs into an isolated module and maybe even separate repo. Until then, we'll have to live with a slightly overweight Scalafmt fatjar.

Issue #1252, opened 2018-08-09
Add explicit dependency scala-reflect


olafurpg:
Reported by @pjrt on Gitter against latest RC4
```
λ coursier bootstrap --standalone com.geirsson:scalafmt-cli_2.12:1.6.0-RC4 \
                    -r bintray:scalameta/maven \
                    -o ~/.local/bin/scalafmt-cli -f
λ scalafmt-cli --version
Exception in thread "main" java.lang.NoClassDefFoundError: scala/reflect/api/TypeCreator
    at fansi.ColorCategory.$anonfun$Full$1(Fansi.scala:851)
    at fansi.ColorCategory.$anonfun$Full$1$adapted(Fansi.scala:850)
    at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
    at scala.collection.immutable.Range.foreach(Range.scala:156)
    at scala.collection.TraversableLike.map(TraversableLike.scala:234)
    at scala.collection.TraversableLike.map$(TraversableLike.scala:227)
    at scala.collection.AbstractTraversable.map(Traversable.scala:104)
    at fansi.ColorCategory.<init>(Fansi.scala:850)
    at fansi.Color$.<init>(Fansi.scala:717)
    at fansi.Color$.<clinit>(Fansi.scala)
    at fansi.Attr$.<init>(Fansi.scala:623)
    at fansi.Attr$.<clinit>(Fansi.scala)
    at fansi.Str$.<init>(Fansi.scala:373)
    at fansi.Str$.<clinit>(Fansi.scala)
    at org.scalafmt.config.OptIn$.$anonfun$surface$1(OptIn.scala:86)
    at pprint.TPrint$$anon$2.render(TPrint.scala:21)
    at org.scalafmt.config.OptIn$.surface$lzycompute(OptIn.scala:86)
    at org.scalafmt.config.OptIn$.surface(OptIn.scala:86)
    at org.scalafmt.config.OptIn.<init>(OptIn.scala:82)
    at org.scalafmt.config.ScalafmtConfig$.apply$default$4(ScalafmtConfig.scala:117)
    at org.scalafmt.config.ScalafmtConfig$.<init>(ScalafmtConfig.scala:192)
    at org.scalafmt.config.ScalafmtConfig$.<clinit>(ScalafmtConfig.scala)
    at org.scalafmt.cli.CliOptions$.apply$default$1(CliOptions.scala:77)
[...]

PR #1284, opened 2018-09-22
Add explicit dependency on org.scala-lang/scala-reflect

tanishiking:
Address https://github.com/scalameta/scalafmt/issues/1252

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -56,6 +56,8 @@ lazy val core = crossProject(JVMPlatform, JSPlatform)
     libraryDependencies ++= Seq(
       metaconfig.value,
       scalameta.value,
+      // scala-reflect is an undeclared dependency of fansi, see #1252.
+      // Scalafmt itself does not require scala-reflect.
       "org.scala-lang" % "scala-reflect" % scalaVersion.value,
     )
   )
--- a/build.sbt
+++ b/build.sbt
@@ -55,7 +55,8 @@ lazy val core = crossProject(JVMPlatform, JSPlatform)
     buildInfoSettings,
     libraryDependencies ++= Seq(
       metaconfig.value,
-      scalameta.value
+      scalameta.value,
+      "org.scala-lang" % "scala-reflect" % scalaVersion.value,
     )
   )
   .jsSettings(


build.sbt
```diff
@@ -55,7 +55,8 @@ lazy val core = crossProject(JVMPlatform, JSPlatform)
     buildInfoSettings,
     libraryDependencies ++= Seq(
       metaconfig.value,
-      scalameta.value
+      scalameta.value,
+      "org.scala-lang" % "scala-reflect" % scalaVersion.value,
```
olafurpg: Can we add a comment explaining why this is needed? 

> scala-reflect is an undeclared dependency of fansi, see https://github.com/scalameta/scalafmt/issues/1252. Scalafmt itself does not require scala-reflect.

It's a big dependency and it would be great to avoid it in the future.

tanishiking: Sure! Added https://github.com/scalameta/scalafmt/pull/1284/commits/2a13a4b01301f1ccc86381b8ec3b83587f8e79a1

> It's a big dependency and it would be great to avoid it in the future.

I agree.

build.sbt
```diff
@@ -55,7 +55,10 @@ lazy val core = crossProject(JVMPlatform, JSPlatform)
     buildInfoSettings,
     libraryDependencies ++= Seq(
       metaconfig.value,
-      scalameta.value
+      scalameta.value,
+      // scala-reflect is an undeclared dependency of fansi, see #1252.
```
olafurpg: I generally try to include the full link since it makes it easier to open from the editor.

olafurpg: Aah, I see the issue link got shortened from my review comment, the original markup had the full link.

olafurpg: It's fine, don't worry about it 😉

MasseGuillaume:
Hi @olafurpg would it be possible to release 1.6.0-RC5, I just ran into this bug. Thanks!

ShaneDelmore:
I hit it yesterday as well for the first time.

sake92:
Is this the same stuff? https://github.com/sake92/hepek-starter/issues/4

poslegm:
@sake92 may be. Please create bug report into [sbt-scalafmt](https://github.com/scalameta/sbt-scalafmt) repo

PR #1285, opened 2018-09-22
Enable now previews

gabro:
Testing this


olafurpg:
Did it work?

gabro:
Nope, I don't think it's worth it. The docker build takes too long anyway because it builds scalafmt-core from scratch. If we want preview, it's better to do custom integration from the existing CI job.

Closing for now

olafurpg:
Sounds good to me 👍 I find it very convenient to update `siteConfig.js` to publish a live demo to my personal site. Structural changes to the website are infrequent enough that I can live with a manual approach.

PR #1295, opened 2018-10-13
Upgrade mdoc and generate docs in PRs

olafurpg:


tenorviol:
Brilliant, thank you!

PR #1304, opened 2018-10-24
add JDK 11 to Travis-CI matrix

SethTisue:
also update sbt to 1.2.6

work around scala/bug#11125 as needed

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/Error.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/Error.scala
@@ -42,7 +42,7 @@ object Error {
                        |=====================
                        |$diff
                        |=====================
-                       |${output.lines.toVector.take(10).mkString("\n")}
+                       |${output.linesIterator.toVector.take(10).mkString("\n")}
                        |=====================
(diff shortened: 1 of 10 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/Error.scala
```diff
@@ -42,7 +42,7 @@ object Error {
                        |=====================
                        |$diff
                        |=====================
-                       |${output.lines.toVector.take(10).mkString("\n")}
+                       |${output.linesIterator.toVector.take(10).mkString("\n")}
```
SethTisue: I used `linesIterator` here instead of `augmentString` because `augmentString` made the line too long and then scalafmt tried to reformat it 💥

(and I figured in this context the deprecation warning that using `linesIterator` normally incurs, until we undeprecate it in 2.12.8, wouldn't matter)

olafurpg: Fatal warnings are not enabled, I'm fine with the deprecation warning until 2.12.8 :)

PR #1315, opened 2018-11-10
Clarify "use intellij formatter"

olafurpg:


docs/installation.md
```diff
@@ -44,14 +47,23 @@ To try out the latest pending releases for the Scalafmt plugin:
 
 ### Continue using IntelliJ formatter
```
olafurpg: @yurique does this make it clearer?

dwijnand:
This is unfortunate.  Is the built-in IntelliJ support going to be removed, or what's the plan there?

PR #1319, opened 2018-11-16
Preferences changed to Settings.

P3trur0:
In IntelliJ the menu you currently name "**Preferences**", actually is "**Settings**". So, this small PR changes the word properly.
Hope it helps :smile_cat:


P3trur0:
Hi @olafurpg,

I didn't know of this label difference between the two versions. 
Indeed, I stumbled upon this typo because at first sight I did not find that menu on my Windows version of IntelliJ, so I had to figure out which one you're referring to.

The easiest alternative that comes to my mind is to add a small reference to the first occurrence of `Preferences` pointing to a margin note where you could highlight the difference. Although this approach'd introduce accidental complexity to the documentation itself.

Alternatively, given the scenario you're describing, you could stay with the current version :smiley:

docs/installation.md
```diff
@@ -10,7 +10,7 @@ You can use Scalafmt from your editor, build tool or terminal.
 To install the
 [Scalafmt IntelliJ plugin](https://plugins.jetbrains.com/plugin/8236?pr=)
 
-- open `Preferences > Plugins`
```
olafurpg: How about?

> open `Preferences > Plugins` (for Windows/Linux it is called `Settings` instead of `Preferences`)

P3trur0:
what about this alternative instead?

docs/installation.md
```diff
@@ -6,6 +6,9 @@ title: Installation
 You can use Scalafmt from your editor, build tool or terminal.
 
 ## IntelliJ
+*For the following paragraphs about IntelliJ setup, please note that the referred (macOS) `Preferences` dialog is named `Settings ` for Windows and Linux.*
```
olafurpg: This will be the first thing the user sees after clicking "Getting started" from the landing page

screenshot 2018-11-16 at 14 18 43

I prefer to mention it right next to where "Preferences" is mentioned first. How about using "Preferences/Settings > ..." instead?

> `Preferences/Settings > Plugins`

olafurpg: This matches the IntelliJ page you linked.

P3trur0: Ok then, so I'd stay with you first suggestion, so I'll decorate the first occurrence as: `Preferences > Plugins` (for Windows/Linux it is called `Settings` instead of `Preferences`)

Does that sound good?

olafurpg: Sounds good, either way is fine by me 👍

P3trur0: here you go :smile:

Issue #966, opened 2017-06-05
verticalMultilineAtDefnSite breaks if the last parameter on a class has a comment


pjrt:
This template is a guideline, not a strict requirement.
- **Version**: 0.5
- **Integration**: IntelliJ
- **Configuration**:
```
verticalMultilineAtDefinitionSite = true
```

## Steps

I ran scalafmt like this:

```bash
scalafmt
```

## Problem

Original code:

```scala
class Y(
    x: Int,
    y: Int,
    z: Int // Some Comment
)
```

Scalafmt formats code like this:
```scala
class Y(
    x: Int,
    y: Int,
    z: Int // Some Comment)
```
Notice the `)` in the comment. This is no longer valid Scala. 

## Expectation

I would like the formatted output to look like this:
```scala
class Y(
    x: Int,
    y: Int,
    z: Int) // Some Comment
```

## Workaround

Putting the `)` manually before the comment fixes this.

pjrt:
@olafurpg Not sure how to go about this since this requires a rewrite (the comment and the `)` need to be swapped and inlined). Do we do something like this currently? Could you point me to the right place if so?

olafurpg:
Correct, this would be a rewrite. You can look at the other rewrites for examples, it seems this one will need to find those comments that ought to be swapped and then remove the comment token and add the comment syntax to the right of the closing `)` token. The patch api is a bit outdated at the moment, I'm planning to switch soon to the scalafix patch api which I've been polishing for the past months 😎

pjrt:
I see. Should I wait until that's complete then?

olafurpg:
Nah, I will port the rewrites to the new system when I find time to do it. Can't promise when that will happen though

pjrt:
@olafurpg Do we have a system for "required" rewrites? Currently, all rewrites are optional (the user ads them) but this rewrite would need to be mandatory (if `verticalMultilineAtDefinitionSite` is defined). 

If we don't, then I guess I will need to make one.

olafurpg:
Fixed in #1323

PR #1323, opened 2018-12-06
avoid generating invalid scala code

woparry:
#966 

The suggested fix in the ticket is a rewrite rule. While I think that would be reasonable, it's more complex. This patch is the absolute minimal version that fixes the invalid code generation. If you want the code formatted as suggested in the linked issue, scalafmt will preserve that style as long as you do the edit (move the parenthesis) yourself.


pjrt:
Understandable. I hit my head against a wall when I was trying to do a rewrite + format, since that isn't something we currently do in scalafmt.

Thanks for the fix!

PR #1331, opened 2018-12-19
Expand on the align=none protip

dwijnand:
Mention vertical alignment mitigation strategies, so that users can
better decide whether they want to align and use the strategies, or not
align and therefore not need them.

### DIFF
--- a/readme/Configuration.scalatex
+++ b/readme/Configuration.scalatex
@@ -138,7 +138,10 @@
           foo(a: Int, // false for call site
               b: String): Int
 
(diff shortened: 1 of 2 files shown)


readme/Configuration.scalatex
```diff
@@ -138,7 +138,10 @@
           foo(a: Int, // false for call site
               b: String): Int
 
-        Pro tip: Enable this setting to minimize git diffs/conflicts from renamings and other refactorings.
```
olafurpg: I this page is not used actually, will remove the scalatex sources separately.

PR #1337, opened 2019-01-07
Add scalafmt-dynamic module for easier build tool integrations.

olafurpg:
Currently, IntelliJ, sbt, gradle, Metals, ..., integrate with the library API of Scalafmt in slightly different and incompatible ways. The Scalafmt library API hasn't been well documented and the API itself frankly isn't any good either.

This commit introduces a new module 'scalafmt-dynamic' that should hopefully improve the situation. This module takes care of many important details such as respecting 'version' and 'project' setting in .scalafmt.conf. The API is written as pure Java interfaces and will remain binary compatible for all future versions of Scalafmt.


ekrich:
As far as I know, unless you use a method with a vararg, a Scala trait would work just fine rather than using a Java Interface.

olafurpg:
With a Scala trait you can't support calling from Java without scala-library https://scalameta.org/scalafmt/docs/installation.html#calling-from-java

ekrich:
Right, right, minor large detail - sorry.

PR #1339, opened 2019-01-07
Change publish organization to org.scalameta from com.geirsson

olafurpg:
Scalafmt has historically been published from com.geirsson to avoid
breaking changes. Now that we're releasing v2 and scalafmt-dynamic
can take care of adjusting the organization it's time to bite the bullet
and switch.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ def scala212 = "2.12.8"
 
 inThisBuild(
   List(
-    organization := "com.geirsson", // not org.scalameta because that's a breaking change
+    organization := "org.scalameta",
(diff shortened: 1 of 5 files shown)


dwijnand:
Doing this without changing the package name risks running into classpath ordering problems whenever someone has `com.geirsson:scalafmtcore_2.12:1.5.1` and `org.scalameta:scalafmtcore_2.12:2.0.0` on the same classpath.

I would advise you to rename the package to `scalafmt2`.  Oh, actually it's `org.scalafmt`.  Hmm, then I don't know but I still advise changing it 🙂

PR #1345, opened 2019-01-19
Micro optimalizations

borissmidt:
I made this pull request, Because i wanted to learn some optimalization in scala and was reading the  scalaftm documentation that you would like to have some microoptimalizations.

My test method was using the ScalafmtProps program for performance testing and using visualVM.

In summery i changed some Map implementation to either a mutable implementation or a better specialized version of Map.

I changed many sequences to Array since most of these where acesed in order. 
Added specialized constructors to Split to avoid copies  with the 'withXXX' 
Added more constants in the split operations to avoid initialization in the router.

If i could i would change the List in the Tree in scala.meta to an array as well.

after these optimalizations the ScalafmtProps ran in 
66746ms
before it was:
491728ms
My computer specs.
core I5 8250u 4 core and 16GB ram.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -69,7 +69,7 @@ class Router(formatOps: FormatOps) {
   import TreeOps._
   import formatOps._
 
-  private def getSplits(formatToken: FormatToken): Seq[Split] = {
+  private def getSplits(formatToken: FormatToken): Array[Split] = {
     val style = styleMap.at(formatToken)
     val leftOwner = owners(formatToken.left)
     val rightOwner = owners(formatToken.right)
@@ -77,11 +77,11 @@ class Router(formatOps: FormatOps) {
 
     formatToken match {
       case FormatToken(_: BOF, _, _) =>
-        Seq(
+        Array(
           Split(NoSplit, 0)
         )
       case FormatToken(_, _: EOF, _) =>
-        Seq(
+        Array(
           Split(Newline, 0) // End files with trailing newline
         )
       case FormatToken(start @ Interpolation.Start(), _, _) =>
@@ -90,7 +90,7 @@ class Router(formatOps: FormatOps) {
         val policy =
           if (isTripleQuote(start)) NoPolicy
           else penalizeAllNewlines(end, BreakSingleLineInterpolatedString)
-        Seq(
+        Array(
           // statecolumn - 1 because of margin characters |
           Split(NoSplit, 0, ignoreIf = !isStripMargin)
             .withPolicy(policy)
@@ -103,7 +103,7 @@ class Router(formatOps: FormatOps) {
           Interpolation.SpliceStart(),
           _,
           _) =>
-        Seq(
+        Array(
           Split(NoSplit, 0)
         )
       case FormatToken(
@@ -111,17 +111,17 @@ class Router(formatOps: FormatOps) {
           Interpolation.Part(_) | Interpolation.End() |
           Interpolation.SpliceEnd(),
           _) =>
-        Seq(
+        Array(
           Split(NoSplit, 0)
         )
       case FormatToken(LeftBrace(), RightBrace(), _) =>
-        Seq(
+        Array(
           Split(NoSplit, 0)
         )
       // Import
       case FormatToken(Dot(), open @ LeftBrace(), _)
           if parents(rightOwner).exists(_.is[Import]) =>
-        Seq(
+        Array(
           Split(NoSplit, 0)
         )
       // Import left brace
@@ -143,7 +143,7 @@ class Router(formatOps: FormatOps) {
             SingleLineBlock(close)
         }
 
-        Seq(
+        Array(
           Split(if (style.spaces.inImportCurlyBraces) Space else NoSplit, 0)
             .withPolicy(policy),
           Split(
@@ -156,14 +156,14 @@ class Router(formatOps: FormatOps) {
       // Interpolated string left brace
       case FormatToken(open @ LeftBrace(), _, _)
           if leftOwner.is[SomeInterpolate] =>
-        Seq(
+        Array(
           Split(NoSplit, 0)
         )
       case FormatToken(_, close @ RightBrace(), _)
           if parents(rightOwner).exists(_.is[Import]) ||
             rightOwner.is[SomeInterpolate] =>
         val isInterpolate = rightOwner.is[Term.Interpolate]
-        Seq(
+        Array(
           Split(
             if (style.spaces.inImportCurlyBraces && !isInterpolate) Space
             else NoSplit,
@@ -171,7 +171,7 @@ class Router(formatOps: FormatOps) {
         )
       case FormatToken(Dot(), underscore @ Underscore(), _)
           if parents(rightOwner).exists(_.is[Import]) =>
-        Seq(
+        Array(
           Split(NoSplit, 0)
         )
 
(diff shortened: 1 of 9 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -371,15 +403,15 @@ class Router(formatOps: FormatOps) {
           case _: Mod => Space
           case t: Term.Name
               if style.spaces.afterTripleEquals &&
-                t.tokens.map(_.syntax) == Seq("===") =>
+                t.tokens.map(_.syntax) == Array("===") =>
```
olafurpg: Are you sure this preserves the original semantics? 

```scala
@ Array("==") == Array("==")
res0: Boolean = false
```

olafurpg: Use `sameElements` instead
```scala
(t.tokens.map(_.syntax) sameElements Array("==="))
```

borissmidt: I guess you are right i will look into this since array is like a primitive type with corner case semantics.

borissmidt:
I have selectively checked out the sources from master to find the code that is causing the problem it certainly is caused by the router or the split.

I have corrected the Split specialized functions because it didn't handle the penalty Boolean correctly.

But there seems to be one of the router rule to be off and i haven't found it yet even after compairing the router to the master branch multiple times.

borissmidt:
one of the statements newLine was without capitalization and i replaced them with a regex that was not capital sensitive.
now it should be fixed.

olafurpg:
@borissmidt nice job tracking it down. For large changes like these I also manually  verify there are no regressions from running scalafmt on a large corpus https://github.com/olafurpg/scala-repos. If you want to try it yourself, you can 

- check out the [v1.6.0-RC1](https://github.com/olafurpg/scala-repos/tree/v1.6.0-RC1) branch
- checkout scalafmt master and run `sbt cli/assembly`
- run `java -jar /path/to/scalafmt/scalafmt-cli/target/scala-2.12/scalafmt.jar /path/to/scala-repos/`
- commit diffs,  if any
- checkout this branch and re-run `sbt cli/assembly`
- run `java -jar scalafmt.jar` again and see if there are any diffs

Regardless, I will try to run this step myself later this week.

borissmidt:
I ran the fmt props, but i saw that it had too many idempotent bugs. So i
guess one of the changes failed.

I also had some time to look at the memory usage and found that most of the
garbage was the extended api of scala.meta. i.e. token.is[TokenType]. But i
have no idea if it has any performance impact.

olafurpg:
@borissmidt I experimented with making the `is` and `isNot` methods static and it did yield a minor improvement parsing 720k lines of code.
```
[info] Benchmark            Mode  Cnt     Score    Error  Units
[info] Parser.parseSources    ss   10  7895.383 ± 70.968  ms/op (AnyVal)
[info] Parser.parseSources    ss   10  8188.405 ± 65.772  ms/op (non-AnyVal)
```
That's a total of 3.7% speedup in the parser, which would likely result in <1% speedup for scalafmt formatting so I don't think it will have a big impact. The speedup is not noticeable in benchmarks when parsing a single 4k line of code source file.

borissmidt:
Cool, did it have any impact on the memory consumption and gc part?
If i find the time this week i will have a look at error generated when i
run the longer test.

olafurpg:
I didn't profile memory usage but I suspect it has an impact there. Especially, since the cpu difference from the change is only measurable in the benchmarks parsing the large 700k line corpus but not a single 4k line source file.

borissmidt:
I ran the commands you asked me and it gave no diffs. So i guess that is good news.

borissmidt:
I had one more question what target are you aiming for on the performance front? Or is this speedup enough for now?

borissmidt:
I came across this yesterday: 
https://developer.lightbend.com/blog/2018-11-01-the-scala-2.12-2.13-inliner-and-optimizer/index.html

this article states that @inline doesn't do anything unless the opt:inline is activated.

olafurpg:
@borissmidt very nice! I left a couple comments on that PR

borissmidt:
Is there anything left that i should do for this pull request?
I could also merge create another PR with the other changes i made if those are worth it?

olafurpg:
@borissmidt I haven't gotten around to run run the manual tests in https://github.com/olafurpg/scala-repos yet. This is a large change and I need more time to validate this doesn't introduce unintended regressions.

borissmidt:
Oke no problem, i had run it and there where no differences between this and version 1.6-RC

borissmidt:
I have a small update about the inlining optimizations which you can enable in the compiler, it doesn't seem to make any real difference in the performance of the application when doing a long running process, also enabled the warning to see if it failed to inline code, there where many places where it just couldn't resolve to inline libary code.

I also changed all defs marked with @inline to final otherwise the compiler cannot inline work.

 Graal Native build
> ~/D/scala-repos ❯❯❯ time /home/boris/Desktop/scalafmt/scalafmt-cli/target/scala-2.11/scalafmt
> Reformatting...
>   100.0% [##########] 27500 source files formatted
> /home/boris/Desktop/scalafmt/scalafmt-cli/target/scala-2.11/scalafmt  2218,91s user 244,08s system 218% cpu 18:48,80 total

No optimazations 2.11
> ~/D/scala-repos ❯❯❯ time java -jar /home/boris/Desktop/scalafmt/scalafmt-cli/target/scala-2.11/scalafmt.jar .
> Reformatting...
>   100.0% [##########] 27500 source files formatted

 -opt flags on core and cli 2.12:
> java -jar  .  1454,64s user 4,95s system 575% cpu 4:13,71 total
> ~/D/scala-repos ❯❯❯ time java -jar /home/boris/Desktop/scalafmt/scalafmt-cli/target/scala-2.12/scalafmt.jar .
> Reformatting...
>   100.0% [##########] 27500 source files formatted
> java -jar  .  1612,11s user 7,12s system 662% cpu 4:04,30 total

 -No optimization 2.12:
>  time java -jar /home/boris/Desktop/scalafmt/scalafmt-cli/target/scala-2.12/scalafmt.jar .
> Reformatting...
>   100.0% [##########] 27500 source files formatted
> java -jar  .  1590,72s user 5,66s system 693% cpu 3:50,12 total

olafurpg:
Thanks for the detailed analysis! Do I understand correctly that the `-opt` flag on 2.12 slows down performance?

```diff
--- -opt
+++ no -opt
- java -jar . 1612,11s user 7,12s system 662% cpu 4:04,30 total
+ java -jar . 1590,72s user 5,66s system 693% cpu 3:50,12 total
```

borissmidt:
It was a little bit faster,

But i was also using my computer during the profiling. So i guess most of
the inlining is done by the jvm hotspot on long running processed.

So i guess it could make a difference on short running tasks where the jvm
doesn't have the time to warm up.

borissmidt:
Is there something i could still do for this open pull request?

olafurpg:
@borissmidt I'm sorry, I haven't been able to carve out enough time to properly validate this PR yet. I usually work on Scalafmt in intense short periods a few times a year and for the past months I've been busy with other projects. I try to prioritize small changes when they come in but large changes like this need more careful testing/validation.

borissmidt:
Oke no problem, if you need any help with a different project then let me
know. It is always intresting to learn from different projects.

tanishiking:
I'm going to take over the review and validation, and try to test it this (or next) weekend
(by https://github.com/scalameta/scalafmt/pull/1345#issuecomment-456079507) :) FYI @olafurpg @borissmidt

tanishiking:
Let me close this PR for inactivity. I really appreciate your contribution, and maybe we'll get back to this and cherry-picking your some commits at some point.

PR #1347, opened 2019-01-30
Added Arch Linux installation instruction on website

poslegm:
I created Arch Linux scalafmt [package](https://aur.archlinux.org/packages/scalafmt-native) built with GraalVM and going to maintain it. Do you think installation instructions should be presented on the site? Or is it excessively for one Linux distro?

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -242,6 +242,15 @@ scalafmt --version // should be @STABLE_VERSION@
 brew upgrade scalafmt
 ```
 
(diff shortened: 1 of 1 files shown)


poslegm:
@olafurpg ok, I'll keep you up to date about maintainance status

olafurpg:
You can watch the repo to get only notifications for new releases
screenshot 2019-01-30 at 14 04 27

PR #1352, opened 2019-01-31
Clarify scalafmtConfig's default vs None

dwijnand:


dwijnand:
I eventually understood what it meant, so I thought maybe it would be clearer reworded like this.

docs/installation.md
```diff
@@ -101,9 +101,9 @@ addSbtPlugin("com.geirsson" % "sbt-scalafmt" % "1.5.1")
 
 ### Customize configuration location
 
-- `scalafmtConfig: Option[File]`: Optional location of `.scalafmt.conf` file. If
-  `None` the default config is used. By default, `.scalafmt.conf` file on the
-  project root will be used.
+- `scalafmtConfig: Option[File]`: The location of the `.scalafmt.conf`
+  configuration file. Defaults to the `.scalafmt.conf` file at the root of
+  the project. `None` will make it use Scalafmt's default config.
```
olafurpg: I think this is incorrect, `None` means Scalafmt first tries to use `.scalafmt.conf` in the base directory if it exists with fallback to the default config. If I were to do it again I would remove the fallback and require a `.scalafmt.conf` file like the new `scalafmt-dynamic` module does (which Metals uses).

dwijnand: Ah, so my understanding was wrong.  Good, updated and pushed.

> If I were to do it again I would remove the fallback and require a `.scalafmt.conf` file like the new `scalafmt-dynamic` module does (which Metals uses).

If you mean default to requring the file, allowing the key to be overridden to `None`, then I agree.  It would avoid the setting defaulting based on external state.

PR #1355, opened 2019-01-31
Switch to non-legacy align = more

dwijnand:
Also, use "presets" instead of "defaults".

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -9,7 +9,7 @@ using [HOCON](https://github.com/typesafehub/config) syntax.
 Here is an example `.scalafmt.conf`:
 
 ```scala config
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -9,7 +9,7 @@ using [HOCON](https://github.com/typesafehub/config) syntax.
 Here is an example `.scalafmt.conf`:
 
 ```scala config
-align = true    // For pretty alignment.
+align = true    // For pretty alignment. (equivalent to align=more)
```
olafurpg: Let's write `align = more` instead, the true/false variants are only kept for legacy reasons

dwijnand: Sure.

PR #1359, opened 2019-01-31
sbt

dwijnand:
Sorry, my PRs have resorted to single word commits by now.  Next up are just grunts.  Or maybe just punctuation.


dwijnand:
Yep, I reckoned as much :)

PR #1365, opened 2019-02-06
Minor tweak for java11 .lines not getting picked up via StringOps in scala-dynamic testfile DynamicSuite.scala

er1c:


er1c:
Eh, this might be more complicated: https://github.com/scala/bug/issues/11125, looks like the better fix is the `scala.Predef.augmentString`

er1c:
I was going to switch to `.linesIterator` but didn't want to deal with the deprecation warning in case an older version was getting compiled

Issue #1318, opened 2018-11-14
Editor vs Build tool version syncing


ShaneDelmore:
It would be helpful if the editor integration and the build tool integration could be synced on the version of scalafmt they use.  In my use case for example pants has a version of scalafmt, and IntelliJ has another version.  I try to keep them in sync, but they are on different release schedules and go out of sync sometimes.

Any solution works for me but one idea I had was to add a config option specifying the invocation syntax to use for scalafmt.  For example 
```
scalafmtExecTemplate = "./pants scalafmt {{fileOrDir}}
```
or 
```
scalafmtExecTemplate = "./bin/scalafmt {{fileOrDir}}
```

olafurpg:
Thanks for reporting! I agree this is a problem.

In Metals we opted to launch the scalafmt version based on the `version=1.5.1` setting in `.scalafmt.conf`. The IntelliJ plugin and cli integration don't respect this setting, but they could for example refuse to format or report a warning if the version number in `.scalafmt.conf` is a mismatch with the running scalafmt instance.

olafurpg:
I opened #1337 that introduces a new `scalafmt-dynamic` module which has a stable binary API and automatically fetches the Scalafmt version matching the 'version' setting in `.scalafmt.conf`. The hard work remains for build tool integrations like sbt/pants/intellij to this module, I will make sure at least sbt-scalafmt, IntelliJ Scalafmt and Metals use it.

tanishiking:
Introducing scalafmt-dynamic (that enable us to switch running scalafmt version using `setting` field of `.scalafmt.conf`) to each integration is working on progress. 
Currently, the following integrations have introduced scalafmt-dynamic.

- [x] sbt-scalafmt (2.0.0-RC5 ~)
  - https://github.com/scalameta/sbt-scalafmt/pull/20
- [x] scalafmt-cli (2.0.0-RC5 ~)
  - https://github.com/scalameta/scalafmt/pull/1366
- [x] intellij-scalafmt (2.0.0-RC5 ~)
  - https://github.com/scalameta/scalafmt/pull/1368
- [x] metals (0.4.0 ~)
  - https://scalameta.org/metals/blog/2019/01/24/tin.html
  - https://github.com/scalameta/metals/pull/429
- [x] gradle-scalafmt
  - https://github.com/alenkacz/gradle-scalafmt/pull/25
- [ ] and more ...?

ches:
[...]

Issue #1346, opened 2019-01-26
Add an API that format multiple programs to scalafmt-dynamic


tanishiking:
Currently, `scalafmt-interface` (and `scalafmt-dynamic`) provides an [API that format single program](https://github.com/scalameta/scalafmt/blob/d1fc11d25dba1df4997787baa6c4ef1fe531641c/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala#L54). Thanks to this API, `scalameta/metals` is capable of formatting a single program via [textDocument/formatting](https://microsoft.github.io/language-server-protocol/specification#textDocument_formatting) using [scalafmt-dynamic](https://github.com/scalameta/metals/blob/70fedca0d355eec93f148f20964fe60ba13fcdd1/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala#L170-L178) which has a stable binary API and automatically fetches the Scalafmt version matching the 'version' setting in `.scalafmt.conf`.

To [sync the version of scalafmt for multiple scalafmt clients](https://github.com/scalameta/scalafmt/issues/1318) (like sbt-scalafmt, scalafmt-cli, IntelliJ-scalafmt), it is necessary to integrate those tools to use `scalafmt-dynamic` instead of depending on a specific version of (pre-resolved) scalafmt.

However, current API of `scalafmt-dynamic` is insufficient for integrating sbt-scalafmt and scalafmt-cli (and some other build tool integration) to it because 

- [`format` method](https://github.com/scalameta/scalafmt/blob/d1fc11d25dba1df4997787baa6c4ef1fe531641c/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala#L54) load scalafmt instance every time format a single file (from [inmemory cache](https://github.com/scalameta/scalafmt/blob/d1fc11d25dba1df4997787baa6c4ef1fe531641c/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala#L25), coursier cache (disk), or remote repository using coursier). Loading scalafmt instance every time formatting a file may cause a performance issue when formatting multiple scala programs.
[...]

PR #1366, opened 2019-02-07
Make scalafmt-cli to use scalafmt-dynamic module instead of depending on specific version of scalafmt-core.

tanishiking:
This PR makes scalafmt-cli to use `scalafmt-dynamic` and enable it to switch scalafmt version using `version` setting in a `.scalafmt.conf`.

## Related issue
https://github.com/scalameta/scalafmt/issues/1318

### Remove `--config-str` option
Since `scalafmt-interface#format` receive the config as `Path`, we cannot parse the config on the client side as long as we use `scalafmt-dynamic` as a backend.

### When missing `.scalafmt.conf
If scalafmt couldn't find `.scalafmt.conf` the following message will appear repeatedly.

```
file does not exist: /home/tanishiking/dev/src/github.com/scalacenter/scalafix/.scalafmt.conf
```

### When .scalafmt.conf missing `version` setting
If `.scalafmt.conf` doesn't contain `version` setting, scalafmt will tell the following message repeatedly. ( this warning will appear every time scalafmt-cli try to format a file)

```
missing setting 'version'. To fix this problem, add the following line to .scalafmt.conf: 'version=1.6.0-RC4'.: /home/tanishiking/dev/src/github.com/scalacenter/scalafix/.scalafmt.conf
```

### When failed to parse config
Some error message like the following one will appear repeatedly

```
Invalid fields: gt: /home/tanishiking/dev/src/github.com/scalacenter/scalafix/.scalafmt.conf
```

## Testing performance and diff
I built a new version of CLI and tested on some repositories.

```
$ sbt:scalafmtRoot> publishLocal
...
$ coursier bootstrap -f org.scalameta:scalafmt-cli_2.12:1.6.0-RC4+...-SNAPSHOT -o scalafmtCli --main org.scalafmt.cli.Cli
```

- Performance: Unfortunately, a new version of scalafmt-cli is slower than the previous scalafmt-cli because the previous one pre-resolved the scalafmt jar.
- diff: `scalafmtCli` (that use the 1.5.1 version of scalafmt) doesn't cause any diff with the former version of scalafmt(1.5.1). :+1:

<details>
<summary>Tested on scalafix</summary>

```
[tanishiking] ~/dev/src/github.com/tanishiking/scalafix
$ ./scalafmt --version
scalafmt 1.5.1

[tanishiking] ~/dev/src/github.com/scalacenter/scalafix
$ /usr/bin/time ./scalafmt
25.62user 0.53system 0:04.20elapsed 621%CPU (0avgtext+0avgdata 821168maxresident)k
0inputs+64outputs (0major+210917minor)pagefaults 0swaps
```

```
[tanishiking] ~/dev/src/github.com/scalacenter/scalafix
$ git diff -u
diff --git a/.scalafmt.conf b/.scalafmt.conf
index ea1573f6..8dc55bab 100644
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,3 +1,4 @@
+version = 1.5.1
 project.git = true
 docstrings = Javadoc
 optIn.blankLineBeforeDocstring = true

[tanishiking] ~/dev/src/github.com/scalacenter/scalafix
$ /usr/bin/time ./scalafmtCli
Reformatting...
  100.0% [##########] 399 source files formatted
28.40user 0.42system 0:08.60elapsed 334%CPU (0avgtext+0avgdata 657196maxresident)k
0inputs+2424outputs (0major+175552minor)pagefaults 0swaps

[tanishiking] ~/dev/src/github.com/scalacenter/scalafix
$ ./scalafmtCli --test
  100.0% [##########] 399 source files formatted
All files are formatted with scalafmt :)

```

</details>

<details>
<summary>Tested on scalafmt</summary>

fat jar

```
[tanishiking] ~/dev/src/github.com/tanishiking/scalafmt
$ ./scalafmt --version
scalafmt 1.5.1

[tanishiking] ~/dev/src/github.com/tanishiking/scalafmt
$ /usr/bin/time ./scalafmt
25.14user 0.53system 0:04.78elapsed 537%CPU (0avgtext+0avgdata 691688maxresident)k
9384inputs+64outputs (1major+173936minor)pagefaults 0swaps
```

dynamic

```
[tanishiking] ~/dev/src/github.com/tanishiking/scalafmt
$ git diff -u
diff --git a/.scalafmt.conf b/.scalafmt.conf
index f93f7ae3..06c8cefd 100644
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version=1.6.0-RC4
+version=1.5.1
 project.git = true
 project.excludeFilters = [
   scalafmt-benchmarks/src/resources,

[tanishiking] ~/dev/src/github.com/tanishiking/scalafmt
$ /usr/bin/time ./scalafmtCli
Reformatting...
  100.0% [##########] 425 source files formatted
25.48user 0.44system 0:07.66elapsed 338%CPU (0avgtext+0avgdata 680364maxresident)k
20808inputs+2464outputs (2major+176225minor)pagefaults 0swaps

[tanishiking] ~/dev/src/github.com/tanishiking/scalafmt
$ ./scalafmtCli --test
  100.0% [##########] 425 source files formatted
All files are formatted with scalafmt :)

```

</details>

### Separate scalafmt.util from scalafmt-core
Though scalafmt-cli successfully stops to depend on a specific version of scalafmt, scalafmt-cli still depends on scalafmt-core because scalafmt-cli uses modules in `scalafmt.util` which is located in scalafmt-core.
I believe that removing scalafmt-core from scalafmt-cli will reduce the binary size of scalafmt (command).

### Remove Scalafmt210 ?
Should we leave Scalafmt210 module that depends on scalafmt-core?
As long as I know, no module depends on this.

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
@@ -1,22 +1,10 @@
 package org.scalafmt.cli
 
 import com.martiansoftware.nailgun.NGContext
-import java.io.{InputStream, OutputStreamWriter, PrintStream}
-import java.nio.file.Paths
-import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
-import java.util.function.UnaryOperator
+import java.io.{InputStream, PrintStream}
+import org.scalafmt.Versions
+import org.scalafmt.util.AbsoluteFile
 
-import metaconfig.Configured
-import org.scalafmt.Error.MisformattedFile
-import org.scalafmt.Error.NoMatchingFiles
-import org.scalafmt.config.{FilterMatcher, ScalafmtConfig}
-import org.scalafmt.{Formatted, Scalafmt, Versions}
-import org.scalafmt.interfaces.{Scalafmt => ScalafmtInterface}
-import org.scalafmt.util.{AbsoluteFile, FileOps, OsSpecific}
-
-import scala.meta.internal.tokenizers.PlatformTokenizerCache
-import scala.meta.parsers.ParseException
-import scala.meta.tokenizers.TokenizeException
 import scala.util.control.NoStackTrace
 
 object Cli {
@@ -84,65 +72,6 @@ object Cli {
     CliArgParser.scoptParser.parse(args, init).map(CliOptions.auto(args, init))
   }
 
-  /** Returns file paths defined via options.{customFiles,customExclude} */
-  private def getFilesFromCliOptions(
-      options: CliOptions,
-      filter: Option[FilterMatcher]): Seq[AbsoluteFile] = {
-    def canFormat(f: AbsoluteFile): Boolean =
-      filter.map(_.matches(f)).getOrElse(true)
-    val files = options.fileFetchMode match {
-      case m @ (GitFiles | RecursiveSearch) =>
-        val fetchFiles: AbsoluteFile => Seq[AbsoluteFile] =
-          if (m == GitFiles) options.gitOps.lsTree(_)
-          else FileOps.listFiles(_)
-
-        options.files.flatMap {
-          case d if d.jfile.isDirectory => fetchFiles(d).filter(canFormat)
-          // DESNOTE(2017-05-19, pjrt): A plain, fully passed file will (try to) be
-          // formatted regardless of what it is or where it is.
-          case f => Seq(f)
-        }
-      case DiffFiles(branch) =>
-        options.gitOps.diff(branch).filter(canFormat)
-    }
-    val excludeRegexp = options.excludeFilterRegexp
-    files.filter { f =>
-      excludeRegexp.findFirstIn(f.path).isEmpty
-    }
-  }
-
-  private def getInputMethods(
-      options: CliOptions,
-      filter: Option[FilterMatcher]): Seq[InputMethod] = {
-    if (options.stdIn) {
-      Seq(InputMethod.StdinCode(options.assumeFilename, options.common.in))
-    } else {
-      val projectFiles: Seq[AbsoluteFile] =
-        getFilesFromCliOptions(options, filter)
-      projectFiles.map(InputMethod.FileContents.apply)
-    }
-  }
-
-  def newTermDisplay(
-      options: CliOptions,
-      inputMethods: Seq[InputMethod],
-      msg: String): TermDisplay = {
-    val termDisplay = new TermDisplay(
-      new OutputStreamWriter(options.info),
-      fallbackMode =
-        options.nonInteractive ||
-          TermDisplay.defaultFallbackMode
-    )
-    if (!options.quiet &&
-      (options.inPlace || options.testing) &&
-      inputMethods.length > 5) {
-      termDisplay.init()
-      termDisplay.startTask(msg, options.common.workingDirectory.jfile)
-      termDisplay.taskLength(msg, inputMethods.length, 0)
-    }
-    termDisplay
-  }
-
   private[cli] def run(options: CliOptions): ExitCode = {
     val termDisplayMessage =
       if (options.testing) "Looking for unformatted files..."
@@ -157,12 +86,13 @@ object Cli {
     // - `scalafmt-dynamic` if the specified `version` setting doesn't match build version.
     // - `scalafmt-core` if the specified `version` setting match with build version
     //   (or if the `version` is not specified).
-    val exit = options.version match {
-      case None => runScalafmt(options, termDisplayMessage)
+    val runner: ScalafmtRunner = options.version match {
+      case None => ScalafmtCoreRunner
       case Some(v) if v == Versions.version =>
-        runScalafmt(options, termDisplayMessage)
-      case _ => runDynamic(options, termDisplayMessage)
+        ScalafmtCoreRunner
+      case _ => ScalafmtDynamicRunner
     }
+    val exit = runner.run(options, termDisplayMessage)
 
     if (options.testing) {
       if (exit.isOk) {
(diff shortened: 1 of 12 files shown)


scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -195,50 +167,62 @@ object Cli {
...
-    inputMethods.par.foreach { inputMethod =>
-      val inputConfig =
-        if (inputMethod.isSbt || inputMethod.isSc) sbtOptions else options
-      val code = handleFile(inputMethod, inputConfig)
-      exitCode.getAndUpdate(new UnaryOperator[ExitCode] {
-        override def apply(t: ExitCode): ExitCode =
-          ExitCode.merge(code, t)
-      })
+
+    inputMethods.foreach { inputMethod =>
```
tanishiking: Unfortunately, `scalafmt-dynamic` has lower performance when it runs parallelly.
https://github.com/scalameta/scalafmt/issues/1346#issuecomment-457840603

scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -195,50 +167,62 @@ object Cli {
...
+        options.customFiles.contains(file)
+      }
+    }
+    val scalafmtInstanceIgnoreFilters =
+      if (fqpns.isEmpty) scalafmtInstance
+      else
+        Scalafmt
+          .create(this.getClass.getClassLoader)
+          .withReporter(reporter)
+          .withRespectProjectFilters(false)
```
tanishiking: When `CliOptions` has a `customFile` that specify the fully qualified path name, `scalafmt-cli` uses the `scalafmt-dynamic` instance that ignores project filters for formatting the file regardless of what it is or where it is

scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -120,44 +111,24 @@ object Cli {
...
-          ExitCode.Ok // do nothing
-        } else {
-          e match {
-            case e @ (_: ParseException | _: TokenizeException) =>
-              options.common.err.println(e.toString)
-              ExitCode.ParseError
-            case _ =>
-              new FailedToFormat(inputMethod.filename, e)
-                .printStackTrace(options.common.out)
-              ExitCode.UnexpectedError
```
tanishiking: `ExitCode` will be caluculated on `ScalafmtCliReporter` instead.

scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -195,50 +167,62 @@ object Cli {
...
     }
 
-    val sbtOptions = options.copy(
-      config = options.config.copy(runner = options.config.runner.forSbt))
+    val reporter = new ScalafmtCliReporter(options)
+    val scalafmtInstance = Scalafmt
+      .create(this.getClass.getClassLoader)
+      .withReporter(reporter)
+
+    // DESNOTE(2017-05-19, pjrt): A plain, fully passed file will (try to) be
```
tanishiking: What is `DESNOTE` ?

olafurpg:
First of all, thanks for working on this! I'll try to quickly answer a few questions now and review the code in more detail later.

> When .scalafmt.conf missing `version` setting

This is OK.

> Remove `--config-str` option

I think it might be worth adding a `format(String config, ...)` overload so that we can continue to support `--config-str`. The flag is been useful in bug reports and I have received feedback it's a useful flag when trying out different options.

> Testing performance and diff

I think we can overcome the performance regression by making the cli depend on a "basic" Scalafmt instance that uses no reflection. We can implement the this `Scalafmt` class using the same APIs as the current cli and then we can do something like this

```scala
val scalafmt: Scalafmt =
  if (version == BuildInfo.currentVersion) new NoReflectionScalafmt()
  else Scalafmt.newInstance(this.getClass.getClassLoader)
```

Sidenote: you might want to try the JMH benchmark fixture to measure performance. This could provide more reliable numbers and also makes it easier to validate performance improvements in my experience.

https://github.com/scalameta/scalafmt/blob/f2154330afa0bc4a0a556598adeb116eafecb8e3/scalafmt-benchmarks/src/main/scala/org/scalafmt/benchmarks/MacroBenchmark.scala


> Remove Scalafmt210 ?

Let's keep it. It's not used anywhere as far as I know but it's been since ~v0.2 and it's probably the most stable binary API for reflectively invoking scalafmt.

> Separate scalafmt.util from scalafmt-core

This won't be necessary, I think it's OK to let the the cli depend on Scalafmt especially since I think it may be helpful for performance.

tanishiking:
Thank you so much for the reviewing!

> I think it might be worth adding a `format(String config, ...)` overload so that we can continue to support `--config-str`

OK, I agree it is worth adding a `format(String config, ...)` :+1:
I will work on it on another branch.

> I think we can overcome the performance regression by making the cli depend on a "basic" Scalafmt instance that uses no reflection.

Looks better!

> you might want to try the JMH benchmark fixture to measure performance

Thanks! I'll try it!

> Let's keep it.

OK.

scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -180,55 +143,37 @@ object Cli {
...
-        override def apply(t: ExitCode): ExitCode =
-          ExitCode.merge(code, t)
-      })
-      PlatformTokenizerCache.megaCache.clear()
-      termDisplay.taskProgress(termDisplayMessage, counter.incrementAndGet())
+    val exit = options.version match {
+      case None => runScalafmt(options, termDisplayMessage)
+      case Some(v) if v == Versions.version =>
+        runScalafmt(options, termDisplayMessage)
+      case _ => runDynamic(options, termDisplayMessage)
```
tanishiking: Use `scalafmt-dynamic` only when `version` setting doesn't match build version.

scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -237,8 +182,163 @@ object Cli {
...
+        options: CliOptions,
+        config: ScalafmtConfig): ExitCode = {
+      try unsafeHandleFile(inputMethod, options, config)
+      catch {
+        case MisformattedFile(_, diff) =>
+          options.common.err.println(diff)
+          ExitCode.TestError
+      }
+    }
+    def unsafeHandleFile(
```
tanishiking: (Mostly) copied from original `unsafeHandleFile`

scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -106,60 +105,24 @@ object Cli {
...
+    val excludeRegexp = options.excludeFilterRegexp
+    files.filter { f =>
+      excludeRegexp.findFirstIn(f.path).isEmpty
+    }
   }
 
-  private def getInputMethods(options: CliOptions): Seq[InputMethod] = {
+  private def getInputMethods(
+      options: CliOptions,
+      filter: Option[FilterMatcher]): Seq[InputMethod] = {
```
tanishiking: Accept `Option[FilterMatcher]` for the following cases.

- `runDynamic` doesn't want to filter files on client side (scalafmt-dynamic will do that instead).
- `runScalafmt` want to filter files here as the original `getInputMethods` did.

scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala
```diff
@@ -92,13 +92,42 @@ case class CliOptions(
     gitOpsConstructor: AbsoluteFile => GitOps = x => new GitOpsImpl(x),
     noStdErr: Boolean = false
 ) {
+  private[this] val DefaultGit = false
+  private[this] val DefaultFatalWarnings = false
+  private[this] val DefaultIgnoreWarnings = false
+  private[this] val DefaultEncoding = Codec.UTF8
```
tanishiking: These default values are borrowed from here https://github.com/scalameta/scalafmt/blob/f2154330afa0bc4a0a556598adeb116eafecb8e3/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala#L160

Though it is not reasonable to manage default values in both `scalafmt-cli` and `scalafmt-core`, we need to refer those configuration values from cli side...

scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala
```diff
@@ -92,13 +92,42 @@ case class CliOptions(
...
+
+  private[this] val tempConfigPath: Option[Path] = configStr.map { s =>
+    val file = Files.createTempFile(".scalafmt", ".conf")
+    Files.write(file, s.getBytes)
+    file
+  }
```
tanishiking: Create a temporary file that contains configuration string specified by `--config-str`. This temporary file will be passed to `scalafmt-dynamic`.
See https://github.com/scalameta/scalafmt/pull/1367#issuecomment-464744077

tanishiking:
Sorry for my late response, I was busy these two weeks...

Though I've not yet tried JMH, I've adopted your feedbacks: 
- > continue to support `--config-str`.
- > We can implement the this `Scalafmt` class using the same APIs as the current cli

Now, it's ready for the review :)
Primary changes are done by https://github.com/scalameta/scalafmt/pull/1366/commits/994e7c0aa50f851b30ea87d59287d385693e7d2c and https://github.com/scalameta/scalafmt/pull/1366/commits/10093d1bc6d88ba857563a8281d60994014e7c2e, and https://github.com/scalameta/scalafmt/pull/1366/commits/6f54714023e4965232e2bc4d97eff7d784d33978 did some refactorings.

scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtCoreRunner.scala
```diff
@@ -0,0 +1,104 @@
...
+object ScalafmtCoreRunner extends ScalafmtRunner {
+  override private[cli] def run(
+      options: CliOptions,
+      termDisplayMessage: String
+  ): ExitCode = {
+    options.scalafmtConfig match {
+      case Configured.NotOk(e) =>
+        if (!options.quiet) options.common.err.println(s"${e.msg}")
+        ExitCode.UnexpectedError
+      case Configured.Ok(scalafmtConf) =>
```
tanishiking: Mostly copied from original `Cli#run` method

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
```diff
@@ -123,509 +123,626 @@ abstract class AbstractCliTest extends FunSuite with DiffAssertions {
...
+              !out.contains(unexpected)
+          )
+        }
+      )
+    }
   }
 }
+
+class CliTest extends AbstractCliTest with CliTestBehavior {
+  testsFor(testCli("1.6.0-RC4")) // test for runDynamic
```
olafurpg: Nice! 👏

tanishiking:
@olafurpg Thank you for reviewing! I added some tests that ensure running `scalafmt-core` if `.scalafmt.conf` is missing, or `version` setting is missing.
https://github.com/scalameta/scalafmt/pull/1366/commits/3dea40fc48063b20a2fa11f6c1ab875865770eea

tanishiking:
Merging :rocket:

PR #1367, opened 2019-02-16
Add a format(String config, ...) method to scalafmt-interface

tanishiking:
This PR adds a method `String format(String config, Path file, String code)` to `scalafmt-interface` (and scalafmt-dynamic) so that we can continue to support `--config-str` option of scalafmt-cli
when it begins to use scalafmt-dynamic module.
See: https://github.com/scalameta/scalafmt/pull/1366#issuecomment-461782061

### DIFF
--- a/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala
+++ b/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala
@@ -5,13 +5,15 @@ import com.geirsson.coursiersmall.Dependency
 import com.geirsson.coursiersmall.Repository
 import com.geirsson.coursiersmall.ResolutionException
 import com.geirsson.coursiersmall.Settings
-import com.typesafe.config.ConfigException
-import com.typesafe.config.ConfigFactory
 import java.lang.reflect.InvocationTargetException
 import java.net.URLClassLoader
-import java.nio.file.Files
-import java.nio.file.Path
+import java.nio.charset.StandardCharsets
+import java.nio.file.{Files, Path, Paths}
+import java.security.MessageDigest
+
+import com.typesafe.config.{ConfigException, ConfigFactory}
 import org.scalafmt.interfaces._
+
 import scala.collection.concurrent.TrieMap
 import scala.collection.mutable
 import scala.concurrent.duration.Duration
@@ -22,7 +24,7 @@ final case class ScalafmtDynamic(
     respectVersion: Boolean,
     respectExcludeFilters: Boolean,
     defaultVersion: String,
-    fmts: mutable.Map[Path, ScalafmtReflect]
+    fmts: mutable.Map[String, ScalafmtReflect]
 ) extends Scalafmt {
   override def clear(): Unit = {
     fmts.values.foreach(_.classLoader.close())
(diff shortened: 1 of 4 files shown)


scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala
```diff
@@ -22,7 +24,7 @@ final case class ScalafmtDynamic(
     respectVersion: Boolean,
     respectExcludeFilters: Boolean,
     defaultVersion: String,
-    fmts: mutable.Map[Path, ScalafmtReflect]
+    fmts: mutable.Map[String, ScalafmtReflect]
```
tanishiking: Now, the key of `fmts` is sha1 hash of the content of config of scalafmt.

scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala
```diff
@@ -51,7 +53,29 @@ final case class ScalafmtDynamic(
   override def withDefaultVersion(defaultVersion: String): Scalafmt = {
     copy(defaultVersion = defaultVersion)
   }
+
+  override def format(config: String, file: Path, code: String): String = {
+    val configFile = Paths.get("stdin.conf")
```
tanishiking: This is just a dummy configuration path that will be passed to `ScalafmtReporter`.

scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtReflect.scala
```diff
@@ -82,18 +75,12 @@ case class ScalafmtReflect(
     }
   }
 
-  private var lastTimestamp = FileTime.fromMillis(0)
-
   def format(filename: Path, code: String): String = {
-    val currentTimestamp = Files.getLastModifiedTime(configFile)
-    if (currentTimestamp.compareTo(lastTimestamp) != 0) {
+    if (config == null) {
```
tanishiking: Since `ScalafmtDynamic` create the instance of `ScalafmtReflect`  for each different configuration (sha1 hash of config string) instead of each path of `.scalafmt.conf`, now, we don't need to rely on last modified time for the cache in a instance.

scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala
olafurpg: Can we avoid this change by creating a temporary file to host the string config contents?

tanishiking:
@olafurpg 
Thank you for replying!

> What do you think about writing the string contents to a temporary file? That would allow the rest of the implementation to work unchanged.

You are right. For preserving the current implementation as far as possible, it would be better to create a temporary file that contains the string configuration contents. 
What do you think about creating a temporary file on the client (of scalafmt-dynamic) side instead?
By doing that, we can continue to support `--config-str` option of scalafmt-cli without changing the current scalafmt-dynamic implementation.

Although we can create a temporary file on scalafmt-dynamic, if scalafmt-dynamic has a responsibility to manage a temporary config file, it will create a temporary file every time it runs `format(config: String, ...)` 
Certainly, we can avoid this by managing the key-value store for contents to the temporary file (`Map[String, Path]`), but it is much easier and straightforward to create a temporary file on the client side.

olafurpg:
I like the idea of letting the client create a temporary file, that keeps the dynamic interface small 👍

PR #1368, opened 2019-02-19
1344 api changes to integrate with intellij

unkarjedy:


unkarjedy:
Main changes:

- ScalafmtDynamic is now responsible for both config caching and scalafmt caching
- ScalafmtDynamic is now responsible for checking for checking whether source file matches config includes/excludes filter
- ScalafmtReflect now simply represents a wrapper around Scalafmt instance of a specific version, providing us an interface that encapsulates all reflection stuff
- All config-specific methods were moved to `ScalafmtReflectConfig` class that represents a wrapper around native `ScalafmtConf` object

For details please see commits detailed messages

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
```diff
@@ -151,7 +151,7 @@ class BestFirstSearch(
       stop: Token,
       depth: Int = 0,
       maxCost: Int = Integer.MAX_VALUE): State = {
-    val Q = new PriorityQueue[State]()
+    val Q = new mutable.PriorityQueue[State]()
```
olafurpg: Is this change intentional? I think we have a custom `PriorityQueue` implementation that uses `java.util.PriorityQueue` instead of the one from the Scala collections, this was necessary for Graal native-image support and is not tested in the CI.

scalafmt-dynamic/src/test/scala/tests/DynamicSuite.scala
```diff
@@ -167,6 +164,29 @@ class DynamicSuite extends FunSuite with DiffAssertions {
...
+    "1.5.0",
+    "1.4.0",
+    "1.3.0",
+    "1.2.0",
+    "1.1.0",
+    "1.0.0-RC4",
+    "1.0.0"
+  )
+
+  def checkExhaustive(name: String)(fn: (Format, String) => Unit): Unit = {
```
olafurpg: Nice 👍

scalafmt-interfaces/src/main/java/org/scalafmt/interfaces/ScalafmtReporter.java
```diff
@@ -25,6 +25,18 @@
...
 
+    /**
+     * An exception occurred while trying to process this file.
+     *
+     * @param file can be either a Scala source file or .scalafmt.conf.
+     * @param message additional error message
+     * @param e the exception that occurred, has type {@link PositionException}
+     *          when the error appeared as a position.
+     */
+    default void error(Path file, String message, Throwable e) {
```
olafurpg: Is this addition necessary? It seems we can introduce a new `MessageOnlyException` class instead to represent message errors. I don't feel strongly about it, just want to limit the surface of the interfaces.

olafurpg:
LGTM 👍 Great work @unkarjedy 👏 I look forward to using the scalafmt-dynamic integration in IntelliJ 🎉

Issue #1370, opened 2019-02-22
Default Indent Operator & unindentTopLevelOperators = true failing test


er1c:
- **Version**: master
- **Integration**: CLI
- **Configuration**:
```
unindentTopLevelOperators = true
indentOperator = default
```
## Problem

Original Code:

```scala
      val foo = 1 +
      2 -
      2 *
      3 \
      4
```

Scalafmt Output:
```scala
       val foo = 1 +
         2 -
         2 *
           3 \
             4
```

## Expectation

```scala
       val foo = 1 +
         2 -
         2 *
         3 \
         4
```

## Notes

Failing Test PR:  https://github.com/scalameta/scalafmt/pull/1369

tanishiking:
This issue seems already resolved by https://github.com/scalameta/scalafmt/issues/1370
Thank you so much for contribution @er1c !
Closing

PR #1372, opened 2019-02-22
Indent Operator Fix

er1c:
This builds ontop of vanilla tests https://github.com/scalameta/scalafmt/pull/1371 to show the desired behavior changes.  The updated code makes the `unindentTopLevelOperators` consistent with the `style.indentOperator .includeRegexp` and `style. indentOperator.excludeRegExp`

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -504,7 +504,15 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
       formatToken.between,
       rightIsComment = formatToken.right.isInstanceOf[Comment])
     val indent = {
-      if ((style.unindentTopLevelOperators ||
+      if (style.unindentTopLevelOperators &&
+        !isTopLevelInfixApplication(owner) &&
+        style.indentOperator.includeRegexp
+          .findFirstIn(formatToken.left.syntax)
+          .isDefined &&
+        style.indentOperator.excludeRegexp
+          .findFirstIn(formatToken.left.syntax)
+          .isEmpty) 2
+      else if ((style.unindentTopLevelOperators ||
         isTopLevelInfixApplication(owner)) &&
         (style.indentOperator.includeRegexp
           .findFirstIn(op.tokens.head.syntax)
--- a/scalafmt-tests/src/test/resources/test/UnindentTopLevelOperatorsOperatorSpray.stat
+++ b/scalafmt-tests/src/test/resources/test/UnindentTopLevelOperatorsOperatorSpray.stat
@@ -15,6 +15,12 @@ object TestRoutes {
       3 \
       4
 
+      val foo = 1 +
+      2 -
+      2 *
+      3 \
+      4
+
       a >
       b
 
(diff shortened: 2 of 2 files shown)


er1c:
I back ported the tests to 1.5.1 and they do break 1.5.1 slightly - 1.5.1 doesn't indent the block infix `()` However, I think the change is desireable and more consistent.

```
[info] - test/UnindentTopLevelOperatorsOperatorSpray.stat: #1223 assignment + regular application| *** FAILED *** (8 milliseconds)
[info]   ===========
[info]   => Obtained
[info]   ===========
[info]   val x = a + B(
[info]     a,
[info]     2
[info]   )
[info]
[info]
[info]   =======
[info]   => Diff
[info]   =======
[info]    val x = a + B(
[info]   -  a,
[info]   -  2
[info]   -)
[info]   +    a,
[info]   +    2
[info]   +  ) (FormatTests.scala:81)
[info] - test/UnindentTopLevelOperatorsOperatorSpray.stat: #1223 assignment + right-associative| *** FAILED *** (2 milliseconds)
[info]   ===========
[info]   => Obtained
[info]   ===========
[info]   val x = a :: B(
[info]     a,
[info]     2
[info]   )
[info]
[info]
[info]   =======
[info]   => Diff
[info]   =======
[info]    val x = a :: B(
[info]   -  a,
[info]   -  2
[info]   -)
[info]   +    a,
[info]   +    2
[info]   +  ) (FormatTests.scala:81)
```

er1c:
In the case of:

```
val foo = 1 +
2 +
3 +
4
```

The formatToken.left is the "=" so it's a specialized code path to evaluate that token because the rest of the code path was simply looking at the `+` as the "op".  The code path was never getting invoked with "=" as an he op, so basically setting the indent to be the "2" in this case which causes the rest of the `+` ops to still return "0" in the unindent code path, making the behavior act like we would expect.  @olafurpg 's comments are also weird, however the tests he added seemed to suggest the lack-of-indenting as a "bug" or a "known issue" with the implementation.

scalafmt-tests/src/test/resources/test/OperatorSpray.stat
```diff
@@ -90,9 +93,9 @@ object TestRoutes {
     pathPrefix("foo") {
       ???
     } ~
-    pathPrefix("bar") {
-      ???
-    }
+      pathPrefix("bar") {
```
olafurpg: iirc, the original motivation for the "spray" indentation was to keep `pathPrefix` at the same indentation level. Do I understand correctly that the output introduced in this PR is the following

```scala
    pathPrefix("foo") {
      ???
    } ~
      pathPrefix("bar") {
        ???
      }
```

regardless whether `unindentTopLevelOperators` is true or false?

er1c: Hmm, I'll review this more, let's hold off on merging then.  I might also be able to tackle the https://github.com/scalameta/scalafmt/issues/1370 issue as well.

er1c: This isn't a code-related change from this PR, it simply indicates what happens in the current master branch.  This PR is "ready to merge".  This test was merged to help document/show the current behavior.  I'll create a separate issue/PR to figure out why that additional `~` is currently being indented on master.

er1c: I think you were looking at the wrong part of the file, the formatted test shows it unindented in the expected: https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/resources/test/OperatorSpray.stat#L46-L51

PR #1374, opened 2019-02-26
Fix an ordering bug/typo for case formatting

er1c:
Ironically, this doesn't change the tests, there is another bug in the indentation that is masking this effect.  I uncovered it in my upcoming branch/PR.


er1c:
Eh, I'll close this for now and have the larger PR

PR #1375, opened 2019-02-27
Add tests for compact nested function calls

aij:
It is fairly common practice to place the last (or only)
argument on the same line as the function it is being
passed to even when that argument spans multiple lines.

(Especially for lambdas or single-argument functions)

Current scalafmt seems to greedily insert newlines in
some cases (when curlies are not used?) resulting in more
deeply indented expressions, which in turn can require
additional wrapping to stay under the column limit.

Are there any existing options that would make these pass?


tanishiking:
Let me close this PR for inactivity

PR #1376, opened 2019-03-03
Added DifferentLambdaBraces rule

### LINKED ISSUE #1027: Different styles for single- and multi-lines literal functions
A coding style I've sometimes seen (and personally use) has:
- `(` for single-line literal functions
- `{` for multi-line literal functions

For example:

```scala
// Single line
xs.map(i => i + 1)

// Multi line
xs.map { i =>
  println(i)
  i + 1
}
```

This was discussed in [gitter](https://gitter.im/scalameta/scalafmt?at=59955261578b44a046ba7629) and seemed to generate some enthusiasm.

poslegm:
Now scalafmt formats long lambda functions this way:

```
<<<
parametersList.map(parameter => applySettings(parameter))
>>>
parametersList.map(
  parameter =>
    applySettings(parameter)
)
```

Because of it you have to use "Convert to block expression" in Intellij for pretty result:
```
parametersList.map { parameter =>
  applySettings(parameter)
}
```

I add `DifferentLambdaBraces` rewrite rule which will convert parens to braces for lambdas longer than `maxColumn` setting.

```
<<< replaces parens to curly braces after line break addition
parametersList.map(parameter => applySettings(parameter))
>>>
parametersList.map { parameter =>
  applySettings(parameter)
}
```

Also it will replace braces to parens for one-line short lambdas
```
<<< replaces curly braces to parens for single-line lambda  
xs.map { x => x + 1 }
>>>
xs.map(x => x + 1)
```

This PR partially fixes #1027 (I didn't realize all ideas from this issue).

Please review it **carefully**! I don't have expertise in scalafmt and could do some things wrong.
Tests/documentation/implementation/naming issues are welcome!


olafurpg:
I took a stab at improving the formatting of lambdas and it seems we might be able to fix it with a small change, see https://github.com/scalameta/scalafmt/commit/577b6f0b1b87c55c4997f4acf32216b3faa3245a (branch https://github.com/scalameta/scalafmt/compare/master...olafurpg:lambda-paren?expand=1). There are several failing tests and all of them look like an improvement excluding "default/Lambda.stat: migrate2hocon". @poslegm Please take a look and see what you think

poslegm:
> rewrite rules run before formatting and we don't know if a lambda fits on a single line until after formatting

Yes, I know this problem. I made assumption that cases when formatting breaks this rule are rare. My assumption may be incorrect and I don't know what will be more annoying to the user: use "convert to block expression" every time or in some cases.

Anyway my solution is just better-than-nothing. I don't like your solution _completely_ because I want blocks for multiline lambdas. But it still makes formatting prettier and will works definitely more stable than mine :smile: 

Can we apply rewrite rule for lambdas after formatting or is it against scalafmt architecture?
And is there a possibility to add this rule as experimental or unstable with corresponding remarks in the documentation?

Regardless of the decision by this PR I hope that your improvement will be in next release!

olafurpg:
The problem with the approach in this PR is that it makes formatting non-idempotent, formatting the same source twice produce different output from the first format. To prevent that you would need a stable heuristic to determine which lambdas to rewrite, currently it uses column numbers in the original source but we need something that doesn’t change after formatting. We have the same idempotency problem if we rewrite after formatting, curly braces impact indentation differently compared to parentheses.

poslegm:
Thank you, I got it. Is it make sense to further develop this solution? Or would it be better to close this PR and make it part of the history?

poslegm:
@olafurpg there are a **lot** of merge conflicts with this branch now. Should we still keep the PR open?

poslegm:
Closed because of irrelevance. If you are interested in rule draft implementation see https://github.com/scalameta/scalafmt/pull/1376/commits/1ba2a697e46a41930bfcda9e4f1bba73a8e84c67

PR #1377, opened 2019-03-05
Update website for sbt-scalafmt 2.0.0-RC5

tanishiking:
sbt-scalafmt v2.0.0-RC5 is out
https://github.com/scalameta/sbt-scalafmt/releases/tag/v2.0.0-RC5

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -83,21 +83,25 @@ It is not possible to reset this setting for all existing projects.
 
 ```scala
 // In project/plugins.sbt. Note, does not support sbt 0.13, only sbt 1.0.
-addSbtPlugin("com.geirsson" % "sbt-scalafmt" % "1.5.1")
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -83,21 +83,25 @@ It is not possible to reset this setting for all existing projects.
 
 ```scala
 // In project/plugins.sbt. Note, does not support sbt 0.13, only sbt 1.0.
-addSbtPlugin("com.geirsson" % "sbt-scalafmt" % "1.5.1")
+addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.0.0-RC5")
```
olafurpg: This should be `@STABLE_VERSION@` so that it updated automatically with new releases, it was hardcoded to 1.5.1 because the sbt plugin hadn't been released yet for v2.

PR #1378, opened 2019-03-06
Avoid newlines after yield keyword.

olafurpg:
```scala
// Before
for (a <- as)
yield
  Future {
    // ...
  }
// After
for (a <- as)
yield Future {
  // ...
}
```

It's possible to get the old behavior with `newlines.avoidAfterYield = false`.

While working on this change, I observed that
`style.indentYieldKeyword=true` (default) has undesirable (and
unexpected) behavior for val definitions.
```scala
// Before
val x = for (a <- as)
  yield a
// After
val x =
  for (a <- as)
    yield a
```
The new behavior is more consistent with how we handle other val
definition bodies. By forcing a line break before `for`, we make
it easier to associate the `for` keyword with its accompanying `yield`.

### DIFF
--- a/scalafmt-tests/src/test/resources/test/IndentYieldKeyword.stat
+++ b/scalafmt-tests/src/test/resources/test/IndentYieldKeyword.stat
@@ -47,11 +47,26 @@ yield
 >>>
 val sizes =
   for (row <- table)
+  yield (for (cell <- row)
+         yield
+           if (cell == null) 0
+           else cell.toString.length)
+<<< break before `for` 2
+val sizes = for (row <- table) yield (
+  for (cell <- row)
   yield
-    (for (cell <- row)
-     yield
-       if (cell == null) 0
-       else cell.toString.length)
+    if (cell == null) 0
+    else cell.toString.length
+)
+>>>
+val sizes =
+  for (row <- table)
+  yield (
+    for (cell <- row)
+    yield
+      if (cell == null) 0
+      else cell.toString.length
+  )
 <<< normal way #592
 val sizes = for (row <- table) yield {
     for (cell <- row)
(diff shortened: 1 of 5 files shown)


olafurpg:
cc/ @lihaoyi

scalafmt-tests/src/test/resources/test/IndentYieldKeyword.stat
```diff
@@ -47,11 +47,26 @@ yield
...
-     yield
-       if (cell == null) 0
-       else cell.toString.length)
+    if (cell == null) 0
+    else cell.toString.length
+)
+>>>
+val sizes =
+  for (row <- table)
+  yield (
```
olafurpg: @lihaoyi notice that you need to manually insert a line break after `(` and before `)` to trigger this behavior, otherwise it defaults to the kinda funky output above
```
val sizes =
  for (row <- table)
  yield (for (cell <- row)
         yield
           if (cell == null) 0
           else cell.toString.length)
```

scalafmt-tests/src/test/resources/unit/For.stat
```diff
@@ -23,12 +23,28 @@ row <- table) yield ( for (
...
+  for (cell <- row)
+  yield
+    if (cell == null) 0
+    else cell.toString.length
+)
+>>>
+val sizes =
+  for (row <- table)
+    yield (
+        for (cell <- row)
```
olafurpg: FTR, the default indent is 2 spaces but for legacy reasons the tests use 4 space continuation indent

lihaoyi:
looks good to me i guess

olafurpg:
Let's give this a spin.

PR #1381, opened 2019-03-06
Upgrade IntelliJ plugin

olafurpg:


olafurpg:
This will hopefully be unnecessary soon once the built-in scalafmt support support dynamic version selection + format on file save (I think it's already merged in nightly).

PR #1384, opened 2019-03-13
Use a simple cache based on futures to avoid redownloading on concurrent usage

jrudolph:


scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/utils/ReentrantCache.scala
```diff
@@ -0,0 +1,61 @@
...
+      cache.get(key) match {
+        case Some(fut) => Right(fut)
+        case None =>
+          val p = Promise[V]()
+          cache += key -> p.future
+          Left(p)
+      }
+    } match {
+      case Right(fut) =>
+        /*fut.value match {
```
jrudolph: I left the printlns in for now but can remove them if required.

tanishiking:
Though CI failed (test for Scala 2.11, and scalafmtCheck), it looks nice at a glance.
I'll review this in a couple of days.

scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/utils/ReentrantCache.scala
olafurpg: I normally don't leave printlns in the code but I don't have a strong preference against them. Let's remove it here for consistency with the rest of the codebase

scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/utils/ReentrantCache.scala
```diff
@@ -0,0 +1,61 @@
+package org.scalafmt.dynamic.utils
+
+import scala.annotation.tailrec
+import scala.concurrent.Await
+import scala.concurrent.Future
+import scala.concurrent.Promise
+import scala.concurrent.duration._
+import scala.util.Try
+
+class ReentrantCache[K, V] {
```
olafurpg: Can we add a simple explanation of this data structure? For example "This is like a map except ..."

scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/utils/ReentrantCache.scala
```diff
@@ -0,0 +1,61 @@
...
+      case Right(fut) =>
+        /*fut.value match {
+          case Some(Success(value)) =>
+          //println(s"Value for $key already cached")
+          case Some(Success(value)) =>
+            //println(s"Previous calculation for $key failed")
+          case None =>
+            //println(s"Previous calculation for $key still ongoing")
+        }*/
+        val result = Await.result(fut, 1.minute) // get or wait for other thread to finish download
```
olafurpg: Is this a 1 minute timeout limit to complete the download? If so, I think we need to increase the timeout since we can't expect everybody to have the same internet connection speed. I would go with a large value like 10 minutes, users see the download progress in the console output and can cancel the process with ctrl-c.

scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/utils/ReentrantCache.scala
olafurpg: Could we use `Collections.synchronizedMap(...)` and `computeIfAbsent` instead of `ReentrantCache`? To limit synchronization we could have another `HashMap` cache in front of that where we first try with `get()`

tanishiking: I agree with @olafurpg; we may we can use Collection.synchronizedMap and `computeIfAbsent` instead.

What we want to do on downloading the ScalafmtReflect (concurrently) is doing the check-then-act operation atomically, and Java's synchronized (or concurrent) collection can seem to play the role.
(ConcurrentHashMap may be better in terms of performance).

Anyway, thank you so much for working on this concurrency issue!

tanishiking: I found that `ReentrantCache` is a helpful abstraction of the series of operations (check the result of download and then cache or not)).
Therefore I think it isn't necessarily required to replacing this data structure with built-in data structures (synchronized or concurrent HashMap) as long as the performance doesn't matter.

It would be better to replace this with synchronizedMap or something while keeping the ScalafmtDynamic.scala simple, but I don't have any ideas now...

tanishiking:
@olafurpg FYI @jrudolph I tweaked this PR a bit so that it passes compilation on 2.11 and `scalafmtCheck`. What do you think about merging it and release the latest version (as a pre-release version)?
And if this fixes https://github.com/scalameta/scalafmt/issues/1399, it would be a time to release scalafmt 2.0.0, isn't it?

tanishiking:
Thanks :)

> I'm still concerned about the 1 minute timeout

That's right... I'll merge this PR after modifying the timeout to 10 minutes for now. (Maybe we want to refactor it to use `synchronizedMap` later). After that, I'm going to release the new version in a few days.

Issue #1385, opened 2019-03-14
Some results from profiling


jrudolph:
Hi @olafurpg,

I spent a bit of time looking into the performance of scalafmt. See the
flame-graph below. I found a few things:

### VirtualFile.value

That one was a mystery to me first, it amounts to 8% of total time. I guess the problem here is that `Input` is used as a key to many of the Maps. For some reason, scalafmt creates only `VirtualFile` which means that every Map access with it as the key has to hash the complete code again.

### TreeOps.getOwners

Removing the recursion seems to help and using a java.util.HashMap instead of a mutable.Map builder.

### FormatOps.iter

Basically the same as `TreeOps.getOwners`.

### getSplits

![Screenshot from 2019-03-14 11-17-25](https://user-images.githubusercontent.com/9868/54349080-c8afcf00-464a-11e9-8057-ee79484aa246.png)

That red "Interpreter" frame is `getSplits`. It seems to mean that the JIT-compiler gave up on that method completely... It may be as you say:

https://github.com/scalameta/scalafmt/blob/8111bb296abf36c8b758a1e43f8ce28b758c03e5/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala#L59

Not sure how the pattern match could be restructured to make it easier for the JIT and for performance.

I'll open a PR for ideas about to improve the first three things. That will probably already improve performance somewhat.

![flame-graph](https://jrudolph.github.io/scalafmt-flames.svg)

olafurpg:
@jrudolph Thank you so much for this thorough investigation. I believe the Scalamt codebase is ripe for microoptimizations 👍 The code was written when I was just starting Scala a few years ago and hasn't changed so much since then 😅 

I have also investigated several low-hanging opportunities to speed up Scalameta

- delay setting `Tree.parent` until after parser completes, this seems to speed up parsing ~35% https://github.com/scalameta/scalameta/issues/1130#issuecomment-458899150
- change `Tree` from trait into abstract class, I haven't measured the difference but I suspect we're paying a lot for virtual interface calls.

I personally don't have much time to pursue these ideas but if somebody is interesting in making a contribution I will be happy to provide feedback and review.

kitbellew:
@jrudolph can you share the steps how you generated this flame graph (and how you read it)?

olafurpg:
You can generate flame graphs with https://github.com/jvm-profiling-tools/async-profiler

PR #1386, opened 2019-03-14
Some performance improvements stubs

jrudolph:
These are more ideas than finished solutions. I'll probably don't have time to work further on these.

See #1385

I only ran this one benchmark but there it's a ~ 20% improvement.

Before:

```
Micro.ExtraLarge.scalafmt  avgt   25  2177.608 ± 99.453  ms/op
```

After:

```
Micro.ExtraLarge.scalafmt  avgt   25  1756.014 ± 64.865  ms/op
```


olafurpg:
Superseded by #1564

PR #1395, opened 2019-04-03
Add flags to support apache/spark style of configuration

lihaoyi-databricks:
Add a `verticalAlignMultilineOperators` boolean flag to vertically align multiline operator chains

Make `danglingParentheses` a case class so it can be specified separately for def and call sites (we only enable it for callsites)


scalafmt-core/shared/src/test/scala/org/scalafmt/ScalafmtTest.scala
```diff
@@ -51,5 +51,87 @@ class ScalafmtTest extends org.scalatest.FunSuite {
        |""".stripMargin,
     config.ScalafmtConfig.default40
   )
+  check(
```
olafurpg: Instead of adding tests, let's create a new `danglingParenthesesCallSite.stat` file similar to this one here https://github.com/scalameta/scalafmt/blob/8111bb296abf36c8b758a1e43f8ce28b758c03e5/scalafmt-tests/src/test/resources/test/ContinuationIndentNoDanglingParens.stat

Docs on the `*.stat` syntax https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/resources/readme.md

PR #1416, opened 2019-04-30
Add the CLI option to fetch only recently changed files for formating.

stremlenye:
The premise is the desire to reduce testing/reformatting time by limiting the workload surface.
While working as part of the team on the same repo the master's head could go quite significantly farther potentially enlarging the `diff` which causes the formatting slowdown. The impact is quite dramatic if the `scalafmt --diff --test` is used as part of pre-commit hook.

Current PR revisits the `--diff` option turning it into `--mode [diff/changed]` to switch between two modes.
While `--mode diff` preserves the behaviour of `scalafmt --diff` the `--mode changed` utilises the `git status` to fetch the files for formatting limiting the surface down to the very minimum of the recent changes.

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
@@ -15,7 +15,8 @@ object CliArgParser {
        |         # 2. .scalafmt.conf inside root directory of current git repo
        |         # 3. no configuration, default style
        |scalafmt --test # throw exception on mis-formatted files, won't write to files.
-       |scalafmt --diff # Format all files that were edited in git diff against master branch.
+       |scalafmt --mode diff # Format all files that were edited in git diff against master branch.
+       |scalafmt --mode changed # Format files listed in `git status` (latest changes against previous commit.
        |scalafmt --diff-branch 2.x # same as --diff, except against branch 2.x
        |scalafmt --stdin # read from stdin and print to stdout
        |scalafmt --stdin --assume-filename foo.sbt < foo.sbt # required when using --stdin to format .sbt files.
@@ -134,14 +135,18 @@ object CliArgParser {
         .text(
           """migrate .scalafmt CLI style configuration to hocon style configuration in .scalafmt.conf"""
         )
-      opt[Unit]("diff")
-        .action((_, c) => c.copy(diff = Some("master")))
-        .text("If set, only format edited files in git diff against master.")
+      opt[FileFetchMode]("mode")
+        .action((m, c) => c.copy(mode = Option(m)))
+        .text(
+          s"""Sets the files to be formatted fetching mode.
+             |Options:
+             |        diff - format files listed in `git diff` against master
+             |        changed - format files listed in `git status` (latest changes against previous commit)""".stripMargin)
       opt[String]("diff-branch")
         .action((branch, c) => c.copy(diff = Some(branch)))
-        .text(
-          "If set, only format edited files in git diff against provided branch."
-        )
+        .text("If set, only format edited files in git diff against provided branch. Has no effect if mode set to `changed`.")
+      opt[Unit]('c', "changed-only")
+          .action((_, c) => c.copy())
       opt[Unit]("build-info")
         .action({
           case (_, c) =>
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
@@ -135,6 +135,12 @@ object CliArgParser {
         .text(
           """migrate .scalafmt CLI style configuration to hocon style configuration in .scalafmt.conf"""
         )
+      opt[Unit]("diff")
+        .action((_, c) => c.copy(mode = Option(DiffFiles("master"))))
+        .text(
+          s"""Format files listed in `git diff` against master.
+             |Deprecated: use --mode diff instead""".stripMargin
+        )
       opt[FileFetchMode]("mode")
         .action((m, c) => c.copy(mode = Option(m)))
         .text(
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
@@ -141,10 +141,13 @@ object CliArgParser {
           s"""Sets the files to be formatted fetching mode.
              |Options:
              |        diff - format files listed in `git diff` against master
-             |        changed - format files listed in `git status` (latest changes against previous commit)""".stripMargin)
+             |        changed - format files listed in `git status` (latest changes against previous commit)""".stripMargin
+        )
       opt[String]("diff-branch")
         .action((branch, c) => c.copy(diff = Some(branch)))
-        .text("If set, only format edited files in git diff against provided branch. Has no effect if mode set to `changed`.")
+        .text(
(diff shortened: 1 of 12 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
```diff
@@ -76,7 +77,31 @@ class GitOpsImpl(private[util] val workingDirectory: AbsoluteFile)
...
+      file <- getFileFromGitStatusLine(statusLine).toSeq
+    } yield AbsoluteFile.fromFile(file, root)
+  }.filter(_.exists)
+
+  private final val statusCodesSubstringLength = 3
+  private final val moveStatusArrowDelimiter = "-> "
+
+  private def getFileFromGitStatusLine(s: String): Option[File] =
+    s.substring(statusCodesSubstringLength)
+      .split(moveStatusArrowDelimiter) match {
```
tanishiking: In my environment (details below), this method sometimes failed to extract the file correctly.

<details>

```
$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.2 LTS"

$ git version
git version 2.21.0
```

</details>

Because, in some cases, the status line shows something like following (status codes occupy only one column), and, as a result, `s.substring(3)` returns `calafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala`, instead of `scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala`.

```
M scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
```

I'm not sure why this is happening ... (sry)

---

What do you think about extracting the file names by this way instead?

```scala
s.trim.split(" ").tail.mkString
```

scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
```diff
@@ -5,10 +5,11 @@ import java.nio.charset.StandardCharsets
 import java.nio.file.{Files, Paths}
 
 import scala.util._
-
 import org.scalatest._
 import org.scalactic.source.Position
 
+import scala.reflect.io.Directory
```
tanishiking: not used?

scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
```diff
@@ -134,13 +135,18 @@ object CliArgParser {
         .text(
           """migrate .scalafmt CLI style configuration to hocon style configuration in .scalafmt.conf"""
         )
-      opt[Unit]("diff")
-        .action((_, c) => c.copy(diff = Some("master")))
-        .text("If set, only format edited files in git diff against master.")
```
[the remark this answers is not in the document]
stremlenye: Sure. Makes sense.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
stremlenye: That inconsistency is quite unfortunate.
Using the `split` to parse the path out was my initial take on the problem but it fails if the path has white spaces in it.

If you don't mind that would help a lot if you could run the script below and share the output of `gitlog` file from the `status_test` directory.

```shell
#!/usr/bin/env bash

function commit() {
  commit_message="$1"
  git add .
  echo "$commit_message" >> gitlog
  echo "-------git status-------" >> gitlog
  echo "$(git status)" >> gitlog
  echo "-------git status --porcelain-------" >> gitlog
  echo "$(git status --porcelain)" >> gitlog
  echo "-------git status --porcelain=1-------" >> gitlog
  echo "$(git status --porcelain=1)" >> gitlog
  echo "-------git status --porcelain=2-------" >> gitlog
  echo "$(git status --porcelain=2)" >> gitlog
  echo "-------git status -z-------" >> gitlog
  echo "$(git status -z)" >> gitlog
  echo "-------git status -z --porcelain=1-------" >> gitlog
  echo "$(git status -z --porcelain=1)" >> gitlog
  echo "-------git status -z --porcelain=2-------" >> gitlog
  echo "$(git status -z --porcelain=2)" >> gitlog
  echo "--------------" >> gitlog
  echo "--------------" >> gitlog
  git commit -am ${commit_message}
}

test_dir="status_test"

mkdir ${test_dir}
cd ${test_dir}
git init .
touch gitlog
commit "Init"
mkdir dir1
mkdir dir2
mkdir "dir 1"
echo "blah blah" > dir1/file1.txt
echo "blah blah" > dir2/file2.txt
echo "blah blah" > dir\ 1/file3.txt
commit "c1"

echo "blah blah" >> dir1/file1.txt
mv dir2/file2.txt  dir\ 1/file2.txt
rm dir\ 1/file3.txt
commit "c2"
```

stremlenye:
@tanishiking 
I've made few changes which suppose to address the issue you had with path parsing as well as the backwards compatibility. Could you please validate them on your system? I was not able to reproduce the case neither on my machine nor on my peer's Ubuntu one.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
```diff
@@ -76,7 +76,42 @@ class GitOpsImpl(private[util] val workingDirectory: AbsoluteFile)
...
+      root <- rootDir.toSeq
+      statusLine <- exec(cmd).getOrElse(Seq.empty)
+      file <- getFileFromGitStatusLine(statusLine).toOption.toSeq
+    } yield AbsoluteFile.fromFile(file, root)
+  }.filter(_.exists)
+
+  private final val renameStatusCode = "R"
+  private final val renameStatusArrowDelimiter = "-> "
+
+  private def extractPathPart(s: String): Try[String] =
```
tanishiking: Could you add some comments that explain the behavior (and include a reference to https://git-scm.com/docs/git-status#_short_format)?
Now it's a bit complicated and worth explaining.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
```diff
@@ -76,7 +76,42 @@ class GitOpsImpl(private[util] val workingDirectory: AbsoluteFile)
...
+
+  private def trimQuotes(s: String): String =
+    s.replaceAll("^\"|\"$", "")
+
+  private def getFileFromGitStatusLine(s: String): Try[File] =
+    for {
+      pathRaw <- extractPathPart(s)
+      path = trimQuotes(pathRaw)
+      file <- Try(new File(path))
+    } yield file
```
tanishiking: Can we remove the `Try` block from `Try(new File(path))` and change it to following? I think we don't have to catch the exceptions for `new File` because [`File` constructor throws exception only when `path` is null](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#File-java.lang.String-)

```scala
  // ...
  path = trimQuotes(pathRaw)
} yield new File(path)
```

scalafmt-core/shared/src/main/scala/org/scalafmt/util/GitOps.scala
stremlenye: Sure 👍

stremlenye:
I honestly have zero ideas why different environments act different in there ways of treating parsed lines but could be something related to the way the `Process` treats the output. The one could be merely trimming the line like ` M path/to/file.scala` ending up with `M path/to/file.scala`.
Anyway whatever – if it works, it works 😄

stremlenye:
No worries and thanks for the review ✋

PR #1422, opened 2019-05-20
 Prints out the paths of files subjected to formating before the beginning of formatting in debug mode

stremlenye:
That's a minor change I've used to debug few cases locally. Thought it could come in handy for the broader audience.

P.S. Sorry for the typos in the commit message.

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
@@ -79,11 +79,9 @@ object Cli {
     val termDisplayMessage =
       if (options.testing) "Looking for unformatted files..."
       else "Reformatting..."
-    if (options.debug) {
-      val pwd = options.common.workingDirectory.jfile.getPath
-      val out = options.info
-      out.println("Working directory: " + pwd)
-    }
+    options.common.info.println(
+      "Working directory: " + options.common.workingDirectory.jfile.getPath
+    )
 
     // Run format using
     // - `scalafmt-dynamic` if the specified `version` setting doesn't match build version.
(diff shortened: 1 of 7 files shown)


tanishiking:
I'll look at this tomorrow :eyes:

scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala
```diff
@@ -66,11 +77,19 @@ object CliOptions {
   }
 }
 
+object NoopOutputStream extends OutputStream { self =>
+  override def write(b: Int): Unit = ()
```
tanishiking: Would you override two other `write` methods with `()` just in case?
https://docs.oracle.com/javase/8/docs/api/java/io/OutputStream.html

(LIke this https://github.com/apache/commons-io/blob/d3137782060046872e1ca15f244cbe6d4142b7cf/src/main/java/org/apache/commons/io/output/NullOutputStream.java#L36-L67)

scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -79,11 +79,9 @@ object Cli {
...
       if (options.testing) "Looking for unformatted files..."
       else "Reformatting..."
-    if (options.debug) {
-      val pwd = options.common.workingDirectory.jfile.getPath
-      val out = options.info
-      out.println("Working directory: " + pwd)
-    }
+    options.common.info.println(
+      "Working directory: " + options.common.workingDirectory.jfile.getPath
+    )
```
tanishiking: Hmm, I think this message should be written to `debug` stream.

scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtCliReporter.scala
```diff
@@ -56,17 +56,14 @@ class ScalafmtCliReporter(options: CliOptions) extends ScalafmtReporter {
     }
   }
 
-  override def excluded(file: Path): Unit = {
-    if (options.debug) options.common.out.println(s"file excluded: $file")
-  }
+  override def excluded(file: Path): Unit =
+    options.common.debug.println(s"file excluded: $file")
```
tanishiking: awesome :+1:

PR #1421, opened 2019-05-17
Add test for issues #1420 and #1227

FabioPinheiro:
Add tests for redundantBraces.stringInterpolation:
- identifier started with underscore
- literal identifier

https://github.com/scalameta/scalafmt/issues/1420
https://github.com/scalameta/scalafmt/issues/1227

Issue #1420, opened 2019-05-17
RedundantBraces fails on String interpolation when the identifier started with underscore 


FabioPinheiro:
This template is a guideline, not a strict requirement.
- **Version**: 2.0.0-RC7
- **Integration**: sbt / any?
- **Configuration**:
```
rewrite {
  rules = [ RedundantBraces ]
  redundantBraces.stringInterpolation = true
}
```

## Steps

Given code like this:
```scala
object a {
  s"Hello ${_id}"
}
```

## Problem

`error: error in interpolated string: identifier, `this' or block expected`

Scalafmt formats code like this:
```scala
object a {
  s"Hello $_id"
}
```

## Expectation
If identifier started with underscore it should be no a block expected
The formatting output should not remove the braces on that case

## Workaround
object a {
  val workaround = _id
  s"Hello $workaround"
}

## Notes
O found a related ticket https://github.com/scalameta/scalafmt/issues/1227
@gaelrenoux

poslegm:
@FabioPinheiro please try it with scalafmt [2.0.0-RC8](https://github.com/scalameta/scalafmt/releases/tag/v2.0.0-RC8)

tanishiking:
It had already fixed by https://github.com/scalameta/scalafmt/pull/1425, thank you so much for @poslegm and @FabioPinheiro !

Issue #1227, opened 2018-06-28
RedundantBraces fails on String interpolation with backquoted variable name


gaelrenoux:
- **Version**: 1.6.0-RC3
- **Configuration**:
```
rewrite {
  rules = [ RedundantBraces ]
  redundantBraces.stringInterpolation = true
}
```

## Steps

This is the code to format:
```scala
class Thing(`type`: String) {
  def label = s"type=${`type`}"
}
```

Note that the bug has nothing to do with `type` being a keyword (any backquoted name fails), but having a variable named as a keyword is usually the reason why you put backquotes.

I run scalafmt using the CLI or SBT (no option):

```bash
scalafmt
sbt scalafmt
```

## Problem

Scalafmt fails:
```
/Users/gael/Repositories/Local/tests/src/main/scala/Thing.scala:2: error: `$$', `$'ident, `$'this, `$'BlockExpr`$'_
  def label = s"type=$`type`"
```

It looks like scalafmt tries to remove the braces.

## Expectation

It shouldn't fail, and it should leave the braces in place (they are necessary).

## Workaround

Using an intermediate value solves the problem:
```scala
class Thing(`type`: String) {
  private val typ = `type`
  def label = s"type=${typ}"
}
```

When I run scalafmt on this code, the curly braces inside the interpolation are removed (as expected).

olafurpg:
Thanks for reporting! I am able to reproduce, this is a bug. The rewrite rule should not be triggered for backticked identifiers.

poslegm:
@gaelrenoux please try it with scalafmt [2.0.0-RC8](https://github.com/scalameta/scalafmt/releases/tag/v2.0.0-RC8)

poslegm:
Fixed in 2.0.0-RC8

PR #1425, opened 2019-05-30
 Fixed bug with RedundantBraces for string interpolation

poslegm:
Fixes (#1420)
Fixes (#1227)

Also includes tests from #1421


poslegm:
Hmm, I don't understand what's happening with commit difference :thinking: 

Has the master been rewritten? Should I create new fork or is it ok?

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -32,9 +32,16 @@ case object RedundantBraces extends Rewrite {
     def isIdentifierAtStart(value: String) =
       value.nonEmpty && (Character.isLetterOrDigit(value.head) || value.head == '_')
 
+    // hack for literal identifiers detection (`type`, `val`, ...)
+    def isLiteralIdentifier(arg: Term.Name): Boolean =
```
poslegm: Is there better solution for it?

tanishiking: How about this?

```scala
arg.syntax.startsWith("`") && arg.syntax.endsWith("`")
```

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -32,9 +32,16 @@ case object RedundantBraces extends Rewrite {
...
+      (arg.pos.end - arg.pos.start) - arg.value.length() == 2
+
+    def shouldTermBeEscaped(arg: Term.Name): Boolean =
```
tanishiking: Could we add a comment that refers to https://github.com/scalameta/scalafmt/issues/1420 ?
This solution is a bit hacky and worth explaining why this condition is required :)

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
poslegm: Oh, I didn't know about `Syntax` typeclass. Thank you very much!

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
poslegm: done

Issue #1391, opened 2019-03-29
spaces.beforeContextBoundColon=IfMultipleBounds doesn't work correctly with ClassTag


anilkumarmyla:
This template is a guideline, not a strict requirement.
- **Version**: 2.0.0-RC5
- **Integration**: `sbt scalafmt`
- **Configuration**:
```
version = 2.0.0-RC5
spaces.beforeContextBoundColon = IfMultipleBounds
```

## Steps

Given code like this:
```
final class Test[A <: B: ClassTag]
```

When I run scalafmt like this:

```
sbt scalafmt
```

## Problem

Scalafmt formats code like this:
```
final class Test[A <: B: ClassTag]
```

## Expectation

I would like the formatted output to look like this:
```
final class Test[A <: B : ClassTag]
```

## Notes
I've found that
`final class Test[A <: B: C: ClassTag]` formats correctly to `final class Test[A <: B : C : ClassTag]`
`final class Test[A: B: ClassTag]` formats correctly to `final class Test[A : B : ClassTag]`

poslegm:
@anilkumarmyla please try it with scalafmt [2.0.0-RC8](https://github.com/scalameta/scalafmt/releases/tag/v2.0.0-RC8)

anilkumarmyla:
it works, thanks! On a sidenote, the PR diff in #1428 is so concise that anyone can contribute only if know the internals. Is there developer documentation to understand scalafmt itnernals and design?

poslegm:
@anilkumarmyla There is [doc](https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/resources/readme.md) about formatting tests. And a little bit information [here](https://scalameta.org/scalafmt/docs/contributing-scalafmt.html).
I use `println` instead of docs when I want to contribute something :)

PR #1428, opened 2019-06-04
Fixed spaces.beforeContextBoundColon=IfMultipleBounds behavior for case with subtyping

poslegm:
Fixes (#1391)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -883,7 +883,8 @@ class Router(formatOps: FormatOps) {
         val mod: Modification = rightOwner match {
           case tp: Type.Param =>
             val contextOption = style.spaces.beforeContextBoundColon
-            if (contextOption.isIfMultipleBounds && tp.cbounds.size > 1 || contextOption.isAlways)
+            val summaryTypeBoundsCount = tp.tbounds.lo.size + tp.tbounds.hi.size + tp.cbounds.size
```
tanishiking: :+1:

Issue #1407, opened 2019-04-24
"key not found" with includeNoParensInSelectChains = true


elijahsmith:
- **Version**: 2.0.0-RC6
- **Integration**: sbt, cli
- **Configuration**:
```
style = default
version=2.0.0-RC6
project.git = true
includeNoParensInSelectChains = true
```

## Steps

Given code like this:
```scala
package com.foo.bar.api.controllers.admin

import play.api.http.{ContentTypeOf, ContentTypes, Writeable}

object CirceWritable {
  implicit def contentTypeOf_CirceJson(): ContentTypeOf[io.circe.Json] =
    ContentTypeOf[io.circe.Json](Some(ContentTypes.JSON))
}
```

When I run scalafmt like this:

```bash
scalafmt
```

## Problem

I get a stacktrace that looks like this (in this file, and many others):
```
org.scalafmt.cli.FailedToFormat: /Users/elijahsmith/apollo/pikachu/src/main/scala/com/apolloagriculture/pikachu/services/sms/SMSClient.scala
Caused by: java.util.NoSuchElementException: key not found: 5350276406439838429
	at scala.collection.MapLike.default(MapLike.scala:235)
	at scala.collection.MapLike.default$(MapLike.scala:234)
	at scala.collection.AbstractMap.default(Map.scala:63)
	at scala.collection.MapLike.apply(MapLike.scala:144)
	at scala.collection.MapLike.apply$(MapLike.scala:143)
	at scala.collection.AbstractMap.apply(Map.scala:63)
	at org.scalafmt.internal.FormatOps.getSelectsLastToken(FormatOps.scala:484)
	at org.scalafmt.internal.FormatOps.chainOptimalToken(FormatOps.scala:503)
	at org.scalafmt.internal.Router.getSplits(Router.scala:976)
	at org.scalafmt.internal.Router.$anonfun$getSplitsMemo$1(Router.scala:1488)
	at scala.collection.mutable.HashMap.getOrElseUpdate(HashMap.scala:86)
	at org.scalafmt.internal.Router.getSplitsMemo(Router.scala:1487)
	at org.scalafmt.internal.BestFirstSearch.$anonfun$routes$1(BestFirstSearch.scala:41)
	at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:36)
	at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:33)
[...]

PR #1429, opened 2019-06-06
Fixed error with includeNoParensInSelectChains and full qualified names

poslegm:
Fixes (#1407)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -481,7 +481,11 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
     var curr = next(leftTok2tok(dot))
     while (isOpenApply(curr.right, includeCurly = true, includeNoParens = true) &&
       !statementStarts.contains(hash(curr.right))) {
-      curr = leftTok2tok(matchingParentheses(hash(curr.right)))
+      if (curr.right.is[Dot]) {
+        curr = next(leftTok2tok(curr.right))
+      } else {
+        curr = leftTok2tok(matchingParentheses(hash(curr.right)))
+      }
```
tanishiking: Ah, `matchingParentheses` expects its arguments `TokenHash` of **parentheses** (`(Left|Right)(Brace|Paren|Bracket)`), but `curr.right` can be `Dot` as `isOpenApply` returns true for dot. Nice catch!! :+1:

Issue #1439, opened 2019-06-27
Wrong indentation


danieletorelli:
- **Version**: 2.0.0-RC8
- **Integration**: IntelliJ + SBT plugin
- **Configuration**:
```
version = 2.0.0-RC8
style = intellij
maxColumn = 120
lineEndings = unix
align = most
align.openParenCallSite = false
align.openParenDefnSite = false
align.tokens = [
  {code = "<-"},
  {code = "=>", owner = "Case"},
  {code = "%", owner = "Term.ApplyInfix"},
  {code = "%%", owner = "Term.ApplyInfix"}
]
continuationIndent.defnSite = 2
rewrite.rules = [RedundantBraces, RedundantParens, SortImports, SortModifiers, PreferCurlyFors]
rewrite.sortModifiers.order = ["override", "private", "protected", "implicit", "final", "sealed", "abstract", "lazy"]
newlines.sometimesBeforeColonInMethodReturnType = false
unindentTopLevelOperators = true
spaces.inImportCurlyBraces = true
danglingParentheses = true
includeCurlyBraceInSelectChains = false
assumeStandardLibraryStripMargin = true

```

## Steps

Given code like this:
```scala
class TestSpec extends FlatSpec with Matchers {

"Foo" should "do things" in {
def bar: String = "test"
bar should endWith("st")
}

}
```

When I run scalafmt like this:

```bash
sbt test:scalafmt
```

or through IntelliJ IDEA.

## Problem

Scalafmt formats code like this:
```scala
class TestSpec extends FlatSpec with Matchers {

  "Foo" should "do things" in {
      def bar: String = "test"
      bar should endWith("st")
    }

}
```

## Expectation

I would like the formatted output to look like this:
```scala
class TestSpec extends FlatSpec with Matchers {

  "Foo" should "do things" in {
    def bar: String = "test"
    bar should endWith("st")
  }

}
```

poslegm:
It works fine without `unindentTopLevelOperators = true` in config :thinking:

tanishiking:
I think this has already fixed in v2.1.0-RC1 with https://github.com/scalameta/scalafmt/pull/1440 🙏

PR #1440, opened 2019-06-27
Fixed unindentTopLevelOperator behavior without indentOperator=spray

poslegm:
Fixes (#1439)

I discovered that fix in https://github.com/scalameta/scalafmt/pull/1372 works fine only with `indentOperator = spray` configuration. Otherwise it breaks correct code.

I **don't like** this solution, it looks hacky. Any ideas how to do it better?

Also I don't use `unindentTopLevelOperator` configuration in my codebase. Can anyone provide more test cases for it? I am afraid that my fix can break anything else :smile:

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
@@ -202,8 +202,6 @@ object ScalafmtConfig {
 
   val indentOperatorsIncludeAkka = "^.*=$"
   val indentOperatorsExcludeAkka = "^$"
-  val indentOperatorsIncludeDefault = ".*"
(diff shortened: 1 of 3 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -202,8 +202,6 @@ object ScalafmtConfig {
 
   val indentOperatorsIncludeAkka = "^.*=$"
   val indentOperatorsExcludeAkka = "^$"
-  val indentOperatorsIncludeDefault = ".*"
```
poslegm: Unused values

poslegm:
@tanishiking yes, you are right. https://github.com/scalameta/scalafmt/pull/1440/commits/8aedfc7cb7d318de383819204ab735417f298af7 looks like working and more general solution!

Issue #1434, opened 2019-06-19
Publish for 2.13


gabro:
I know it's more common to use Scalafmt as a standalone tool, but I have some projects that use it programmatically and it would be nice to cross publish for 2.13 too.

I can take a look at it in the next few days.

PR #1443, opened 2019-06-30
Coursier-small library replaced with Coursier

poslegm:
Now scalafmt depends on [coursier-small](https://github.com/olafurpg/coursier-small) library ― tiny wrapper around coursier. But it is archived and no longer supported.

I see some problems with it:
1. Scala 2.13 update #1434 
2. Old version of coursier
3. coursier features support https://github.com/scalameta/scalafmt/pull/1436

I tried to replace it with coursier itself. Strict code review is necessary! I tested it locally with sbt plugin and cli client, but we need more complex testing. Any ideas how to do it?

@olafurpg what do you think about this replacement?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -57,7 +57,6 @@ lazy val dynamic = project
     buildInfoObject := "BuildInfo",
     libraryDependencies ++= List(
       "io.get-coursier" %% "coursier" % coursier,
-      "io.get-coursier" %% "coursier-cache" % coursier,
       "com.typesafe" % "config" % "1.3.3",
       scalatest.value % Test,
       scalametaTestkit % Test
--- a/build.sbt
+++ b/build.sbt
@@ -56,7 +56,8 @@ lazy val dynamic = project
     buildInfoPackage := "org.scalafmt.dynamic",
     buildInfoObject := "BuildInfo",
     libraryDependencies ++= List(
-      "com.geirsson" %% "coursier-small" % "1.3.1",
+      "io.get-coursier" %% "coursier" % coursier,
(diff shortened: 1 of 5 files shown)


olafurpg:
Thank you for this contribution! I’m ok with replacing coursier-small with the coursier “interface” module which is a java library that shades coursier and Scala-library. It’s a good idea. It’s not ok however to depend on coursier-cache directly since that will cause binary compatibility issues in the sbt plugin.

poslegm:
Thank you very much for review! I am on a vacation without strong internet connection for now. I'll get to work on this in a couple of weeks.

tanishiking:
Don't worry about that! Enjoy your vacation 🌴 :)

build.sbt
```diff
@@ -56,7 +56,7 @@ lazy val dynamic = project
     buildInfoPackage := "org.scalafmt.dynamic",
     buildInfoObject := "BuildInfo",
     libraryDependencies ++= List(
-      "com.geirsson" %% "coursier-small" % "1.3.1",
+      "io.get-coursier" %% "coursier" % coursier,
```
olafurpg: This dependency needs to be replaced with the coursier “interface” module, which is a pure java library with no risk of causing binary incompatibility issues in sbt-Scalafmt. The api of the interface module should be quite similar to the main Scala api.

poslegm:
Currently blocked by https://github.com/coursier/interface/issues/57, I  will work on it

olafurpg:
I opened #1511 porting the usage of coursier-small to the java interface

poslegm:
Closing because of #1511

Issue #1447, opened 2019-07-09
--quiet option doesn't suppress output


laughedelic:
- **Version**: 2.0.0
- **Integration**: CLI

## Steps

When I run scalafmt like this:

```bash
scalafmt --quiet --test
```

## Problem

Scalafmt outputs the diffs (as it would without `--quiet` option). Help says:

```
--quiet				don't print out stuff to console
```

## Expectation

I would like it to exit silently (communicating only the exit-code).

## Workaround

I can alway redirect output to `/dev/null`, but why do we need the `--quiet` option then?

laughedelic:
Closed by #1449

PR #1449, opened 2019-07-10
Suppress cli output in case  option was set

stremlenye:
This is the fix for the bug reported here #1447

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala
@@ -53,14 +53,18 @@ object CliOptions {
       common = parsed.common.copy(
         out = guardPrintStream(parsed.quiet)(parsed.common.out),
         info = guardPrintStream(parsed.quiet)(auxOut),
-        debug = guardPrintStream(parsed.quiet)(if (parsed.debug) auxOut else init.common.debug),
+        debug = guardPrintStream(parsed.quiet)(
+          if (parsed.debug) auxOut else init.common.debug
+        ),
         err = guardPrintStream(parsed.quiet)(parsed.common.err)
       )
     )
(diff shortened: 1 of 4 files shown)


scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala
```diff
@@ -51,30 +51,37 @@ object CliOptions {
...
-      path = tryDirectory(options)(rootDir)
-      configFilePath <- if (path.toFile.isFile) Some(path) else None
+      configFilePath <- Option(tryDirectory(rootDir)).filter(_.toFile.isFile)
     } yield configFilePath
   }
 
-  private def tryCurrentDirectory(options: CliOptions): Option[Path] = {
-    val configFilePath = tryDirectory(options)(options.common.workingDirectory)
-    if (configFilePath.toFile.isFile) Some(configFilePath) else None
-  }
```
tanishiking: 👍

scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala
```diff
@@ -51,30 +51,37 @@ object CliOptions {
...
       config = style,
       common = parsed.common.copy(
-        info = auxOut,
-        debug = if (parsed.debug) auxOut else init.common.debug
+        out = guardPrintStream(parsed.quiet)(parsed.common.out),
+        info = guardPrintStream(parsed.quiet)(auxOut),
+        debug = guardPrintStream(parsed.quiet)(
+          if (parsed.debug) auxOut else init.common.debug
+        ),
+        err = guardPrintStream(parsed.quiet)(parsed.common.err)
```
tanishiking: 🚀

laughedelic:
Thank you @stremlenye

PR #1450, opened 2019-07-15
fix include resolution for cli and sbt plugin

nadavwr:
`ConfigFactory.parseFile` is able to resolve relative includes of the form `include "my/dir/my-defaults.conf"`. However, Scalafmt sticks to reading the main config and then parses it as text—without resolving includes.

This PR alters how config files are parsed. `ConfigFactory.parseFile` is used, followed by `.root.render()` to obtain configuration text with fully resolved includes.

Note that caching consideration within `ScalafmtDynamic` have not yet been addressed—configuration modification date will have to be obtained from all included configuration files. (this can be achieved by using a custom `ConfigIncluder`).

### DIFF
--- a/scalafmt-tests/src/test/scala/org/scalafmt/cli/CliOptionsTest.scala
+++ b/scalafmt-tests/src/test/scala/org/scalafmt/cli/CliOptionsTest.scala
@@ -94,7 +94,7 @@ class CliOptionsTest extends FunSuite {
   }
 
   test(
-    ".scalafmtConfig returns default ScalafmtConfig is configuration file is missing"
+    ".scalafmtConfig returns default ScalafmtConfig if configuration file is missing"
   ) {
     val configDir = Files.createTempDirectory("temp-dir")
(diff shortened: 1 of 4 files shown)


scalafmt-tests/src/test/scala/org/scalafmt/cli/CliOptionsTest.scala
```diff
@@ -94,7 +94,7 @@ class CliOptionsTest extends FunSuite {
   }
 
   test(
-    ".scalafmtConfig returns default ScalafmtConfig is configuration file is missing"
+    ".scalafmtConfig returns default ScalafmtConfig if configuration file is missing"
```
tanishiking: 👍

ches:
> Note that caching consideration within `ScalafmtDynamic` have not yet been addressed—configuration modification date will have to be obtained from all included configuration files. (this can be achieved by using a custom `ConfigIncluder`).

Perhaps an issue should be opened to track this as a known limitation.

PR #1455, opened 2019-07-19
Fix nailgun setup instructions

ArturGajowy:


ArturGajowy:
Apparently I've been running some strange nailgun version. The one obtained freshly from `brew install nailgun` uses the currently documented syntax.

PR #1467, opened 2019-08-06
do not cache for ever invalid configs

### LINKED ISSUE #1442: Regression: duplicate "missing Scalafmt version" error
When upgrading to scalafmt-dynamic v2.0.0 in Metals we encountered a regression from 2.0.0-RC4 where a "missing scalafmt version" warning is reported for a config file that has been updated to include a scalafmt version
https://dev.azure.com/olafurpg/Metals/_build/results?buildId=646&view=logs&jobId=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&taskId=5494183b-801e-5084-3b4b-5451c6a14ee1&lineStart=3604&lineEnd=3606&colStart=1&colEnd=126

Steps to reproduce:

- format with .scalafmt.conf that is missing a `version` settings, a "missing version" warning is reported as expected
- add `version` setting and format again, an invalid "missing version" warning is reported when I expected the format to succeed since the config file has been updated.

It looks like `ReentrantCache.getOrAddToCache` returns stale results (it doesn't pick up changes in .scalafmt.conf files).

bjaglin:
fixes #1442

### DIFF
--- a/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala
+++ b/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala
@@ -117,7 +117,7 @@ final case class ScalafmtDynamic(
       configsCache
         .getOrAddToCache(
           configPath,
-          _.right.exists(_._2.compareTo(currentTimestamp) != 0)
+          _.fold(_ => true, _._2.compareTo(currentTimestamp) != 0)
         ) { () =>
           resolveConfigWithScalafmt(configPath).map { config =>
             reporter.parsedConfig(configPath, config.version)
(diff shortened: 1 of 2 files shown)


scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala
```diff
@@ -117,7 +117,7 @@ final case class ScalafmtDynamic(
       configsCache
         .getOrAddToCache(
           configPath,
-          _.right.exists(_._2.compareTo(currentTimestamp) != 0)
+          _.fold(_ => true, _._2.compareTo(currentTimestamp) != 0)
```
bjaglin: A unfortunate scala 2.11 compatible alias of the more readable:
```suggestion
          _.forall(_._2.compareTo(currentTimestamp) != 0)
```

Issue #1484, opened 2019-08-25
New comment alignment in 2.1.0-RC1 can be ugly


adampauls:
- **Version**: 2.1.0-RC1
- **Integration**: IntelliJ
- **Configuration**:
```
version = "2.1.0-RC1"
maxColumn = 120 
rewrite.rules = [PreferCurlyFors, SortModifiers, AvoidInfix]
trailingCommas = always
assumeStandardLibraryStripMargin = true
optIn.breakChainOnFirstMethodDot = false
spaces.inImportCurlyBraces = true
includeCurlyBraceInSelectChains = true
spaces.beforeContextBoundColon = Always
// These are copied from  https://scalameta.org/scalafmt/docs/configuration.html#other
rewrite.neverInfix.excludeFilters = [
  until
  to
  by
  eq
  ne
  "should.*"
  "contain.*"
  "must.*"
  in
  ignore
  be
  taggedAs
  thrownBy
  synchronized
  have
  when
  size
  only
  noneOf
  oneElementOf
  noElementsOf
  atLeastOneElementOf
  atMostOneElementOf
  allElementsOf
  inOrderElementsOf
  theSameElementsAs
  // these are new
  throws
  returns
]
```

## Steps

Given code like this:
```scala
"x"
    .+("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx")
    // single line comment
    .+("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx")
```

When I run scalafmt with IntelliJ

## Problem

Scalafmt formats code like this:
```scala
"x"
    .+("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx")
      // single line comment
    .+("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx")
```

## Expectation

I would like the formatted output look like the original. Note that this particularly egregious in for multi-line comments:

```scala
"x"
[...]

PR #1488, opened 2019-08-29
Make comment blocks line up.

adampauls:
Partial fix for #1484. I don't know this code all that well so I'm not sure if there are other cases worth testing. Thanks!

UPDATE: I think I fixed the rest of #1484. Please take another look if this isn't your first time.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -176,7 +176,13 @@ class FormatWriter(formatOps: FormatOps) {
     locations.zipWithIndex.foreach {
       case (FormatLocation(tok, split, state), i) =>
         val previous = locations(Math.max(0, i - 1))
-        val next = locations.lift(i + 1)
+        // TODO this could get slow for really long comment blocks. If that
+        //   becomes a problem, we could also precompute these locations.
+        lazy val nextNonComment = locations
+          .drop(i + 1)
+          .dropWhile(_.formatToken.right.isInstanceOf[Comment])
+          .headOption
+
         val whitespace = split.modification match {
           case Space =>
             val previousAlign =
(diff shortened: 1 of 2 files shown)


adampauls:
cc @droptheplot

adampauls:
Friendly ping @tanishiking. It would be nice to have an RC2 with this bug fix soon!

tanishiking:
Thank you for working on this :) I'll look at this tomorrow.

adampauls:
I agree, that seems much more robust, although I don’t know the code well enough to make that change on my own. I can take a shot if that’s the only way this change will get made though!

adampauls:
Also, IMHO it seems a little weird to leave the broken implementation in place. Seems like you could merge this fix and refactor later (with now-passing tests to help catch problems!), or you could revert #1470. Blocking this fix on doing it the right way feels a bit let like letting perfect be the enemy of the good?

tanishiking:
I think what we need to do to fix it is
- delete [Split#adapt](https://github.com/scalameta/scalafmt/blob/741406c482ace0faedd78d359f052f5885c3aeb3/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala#L37-L41)
- revert https://github.com/scalameta/scalafmt/pull/1470

It would be helpful if you work on that, but if you don't have time for working on that, I'm willing to implement it instead (maybe this weekend) :)
Though I really appreciate your effort, I can't merge this PR as it is, sorry.

droptheplot:
@adampauls 
Thanks for pointing out missing cases!

@tanishiking 
While deleting `adapt` and reverting previous commit works for current test case, it doesn't work for many others and will fail 9 other tests (including that [feature](https://github.com/scalameta/scalafmt/blob/741406c482ace0faedd78d359f052f5885c3aeb3/scalafmt-tests/src/test/resources/unit/Comment.stat#L68-L79)). Probably `adapt` should stay?

tanishiking:
Thank you for your comment!
Yeah, that's the point. 

Personally, I feel these changes(add indents for line comments in method chains) have a sense of inconsistency as to the position of line comments.

If we keep the adapt method and keep the line comments stay at the beginning of lines, I think scalafmt should keep **all** line comments left without any exceptions.
If we add an indentation for line comments, I think scalafmt should treat all line comments in the same way. For pursuing a simplicity and consistency.

That's why I personally think we probably should

- Just revert the previous commit
- Remove adapt method and add indents for all line comments
  - it will drop the feature which keep line comments at the first column and break some of existing tests, but I think all of them related to the feature.
  - I think it's inevitable for keeping indentation for line comments, without a bit of invoncistency.
  - Since there could be an unpected regression for deleting adapt method, we need a thorough investigation before carrying out this plan (if we gonna follow this plan, maybe we should merge this PR for now because, as @adampauls said, it could take some time)

What do you think about this?

tanishiking:
Though I have some concerns about this, as @adampauls said, maybe it would be better to merge this for now. About `adapt` method, I'll raise another ticket.
Anyway, thank you very much for these contributions @adampauls @droptheplot ! This PRs kicked us off to fix the problems.

PR #1494, opened 2019-09-02
Updated sbt plugin version on the website

poslegm:


poslegm:
I don't like "x.x.x" in documentation because I can't simply copy-paste it into my `build.sbt` 😄

tanishiking:
It makes sense 👍 Hmm, I hope we can update this automatically somehow. 🤔

PR #1510, opened 2019-09-26
Use coursier/interface instead of coursier-small

olafurpg:
Review @poslegm


olafurpg:
Closing in favor of #1511

PR #1515, opened 2019-10-04
Improve imports rewriters

zeal18:
Removing redundant braces around imports has been added.

Grouping imports rewriter has been added. Groups and their order could be customized using settings. It's useful to have imports grouping and sorting functionality outside Intellij world.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -606,7 +606,7 @@ import java.{bar, foo}
 
 import scala.fmt
 
-import com.my.class
+import com.my.clazz
 ```
 
 ## Vertical Multiline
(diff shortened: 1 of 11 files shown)


olafurpg:
There is BTW a Scalafix rewrite that supports similar functionality https://github.com/NeQuissimus/sort-imports

I can understand that the rewrite is still super useful despite having the limitation of not being able to handle relative improts. Some people may be OK with the rewrite producing invalid code, for example if their code style prohibits relative imports.

However, for Scalafmt it's important to never produce invalid code. Many people run formatting on file save or on compile and formatting should always be safe to run.

zeal18:
Yeah, I see. It seems, I'm 2 years late

Issue #1521, opened 2019-10-08
Scalafmt not respecting SBT resolvers setting


ceedubs:
This template is a guideline, not a strict requirement.
- **Version**: 2.0.4
- **Integration**: SBT

## Problem

This is my understanding of the problem that I'm seeing. I apologize if I've misdiagnosed what's happening.

The Scalafmt SBT plugin seems to use a hard-coded list of resolvers for dynamic resolution, ignoring the resolvers that have been configured the standard SBT ways (ex: `resolvers` or `externalResolvers`). It looks to me like the relevant code is [here](https://github.com/scalameta/scalafmt/blob/c1923bc7e6b621872f06a2b652e34679e61fa06f/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamicDownloader.scala#L78-L84). This shows up as a problem in an environment where network configuration prevents hitting public repositories (Maven Central, Sonatype Releases) directly. It also causes issues if a noisy neighbor causes Maven Central to temporarily ban your IP address...

## Expectation

When resolving artifacts, I would expect Scalafmt to only try to hit the repositories that are returned by the `sbt resolvers` command.

## Workaround

So far I don't know of any.

ceedubs:
Oops I just realized that https://github.com/scalameta/sbt-scalafmt also exists. The code that I linked to in the description is within this repository, so I'm not sure where to open this issue. Let me know if I should move it over there.

poslegm:
It looks like related to https://github.com/scalameta/scalafmt/issues/1415. scalafmt uses coursier for dynamic resolution. So you can try to update sbt plugin to 2.0.6 and set environment variable COURSIER_REPOSITORIES.

If COURSIER_REPOSITORIES doesn't work with sbt-scalafmt v2.0.6 — it's a bug.

ceedubs:
@poslegm thanks I'll give that a try. I still think that it would be nice if sbt-scalafmt had a way of passing along the settings that are configured in the standard way within SBT. We have a somewhat convoluted way of populating the repositories and credentials within SBT. We can probably set up environment variables for them, but it is going to be a little bit of a hassle.

henricook:
I don't think COURSIER_REPOSITORIES is honoured. I don't know why that's an sbt-scalafmt problem, please tell me if i'm missing something with my PR

mihaisoloi:
[...]

PR #1524, opened 2019-10-11
Use Coursier defaults, which also adds support for COURSIER_REPOSITORIES

henricook:
This is another crack at https://github.com/scalameta/scalafmt/pull/1436 which I think is a bit easier now we've got the Coursier interface on board.

I believe that fetching Coursier's default repositories also prompts Couriser to check COURSIER_REPOSITORIES env var and `couriser.repositories` property - so by using defaults here we get COURSIER_REPOSITORIES compatibility

Previously the default repositories were being totally overwritten by the hard coded repository list

I believe this fulfils https://github.com/scalameta/scalafmt/issues/1521

### DIFF
--- a/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamicDownloader.scala
+++ b/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamicDownloader.scala
@@ -1,20 +1,14 @@
 package org.scalafmt.dynamic
 
-import java.io.PrintWriter
+import java.io.OutputStreamWriter
 import java.net.URL
-import java.nio.file.Path
 
 import coursierapi._
-
-import scala.collection.JavaConverters._
 import org.scalafmt.dynamic.ScalafmtDynamicDownloader._
 
+import scala.collection.JavaConverters._
 import scala.concurrent.duration.Duration
 import scala.util.Try
-import java.io.OutputStream
-import java.io.PrintStream
-import java.io.OutputStreamWriter
-import scala.concurrent.JavaConversions._
 
 class ScalafmtDynamicDownloader(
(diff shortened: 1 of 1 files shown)


scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamicDownloader.scala
```diff
@@ -75,14 +71,16 @@ class ScalafmtDynamicDownloader(
...
-    Repository.central(),
-    Repository.ivy2Local(),
-    MavenRepository.of(
-      "https://oss.sonatype.org/content/repositories/snapshots"
-    ),
-    MavenRepository.of("https://oss.sonatype.org/content/repositories/public")
-  )
+  private def repositories: Array[Repository] = {
+    // Default repositories are ivy2local, central and also anything in COURSIER_REPOSITORIES overrides
+    Repository.defaults().asScala.toArray ++ Array(
```
tanishiking: defined here
https://github.com/coursier/interface/blob/1eaf27586ccc17cb2c03d9911a41872b42c9547d/interface/src/main/scala/coursier/internal/api/ApiHelper.scala#L23-L28

https://github.com/coursier/coursier/blob/a5d572031e155a6278b2ce41f7a8beef27d851f8/modules/coursier/jvm/src/main/scala/coursier/PlatformResolve.scala#L17-L56

henricook:
Thanks @tanishiking ! We're super keen to use this, are there nightly builds still? and/or when do you next plan to release?

Thanks both for the speedy review

tanishiking:
OK! Since this PR is merged, It would be good timing for releasing 2.1.1, and I'll release it now :)

henricook:
Ultimate speed :muscle: - thanks!

PR #1529, opened 2019-10-13
Fix inconsistent spaces with verticalMultiline and spaces.inParentheses

sirmax:
- **Configuration**:
```
version = 2.1.0
spaces.inParentheses = true
maxColumn = 40
```

## Steps

Given code like this:
```scala
def x(foo: Int) = 2
```

## Problem

```scala
def x(foo: Int ) = 2
```

## Expectation

```scala
def x( foo: Int ) = 2
```

### DIFF
--- a/scalafmt-tests/src/test/resources/vertical-multiline/withSpacesInParentheses.stat
+++ b/scalafmt-tests/src/test/resources/vertical-multiline/withSpacesInParentheses.stat
@@ -1,0 +1,7 @@
+spaces.inParentheses = true
+maxColumn = 40
+verticalMultiline.atDefnSite = true
+<<< def
+def x(foo: Int) = 2
+>>>
+def x( foo: Int ) = 2
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -956,8 +956,11 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
       else NoSplit
 
     Seq(
-      Split(singleLineModification, 0, ignoreIf = !isBracket && aboveArityThreshold)
-        .withPolicy(SingleLineBlock(singleLineExpire)),
+      Split(
+        singleLineModification,
+        0,
+        ignoreIf = !isBracket && aboveArityThreshold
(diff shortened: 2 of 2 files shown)


scalafmt-tests/src/test/resources/vertical-multiline/withSpacesInParentheses.stat
```diff
@@ -0,0 +1,7 @@
+spaces.inParentheses = true
+maxColumn = 40
+verticalMultiline.atDefnSite = true
+<<< def
+def x(foo: Int) = 2
+>>>
+def x( foo: Int ) = 2
```
poslegm: Can you please provide test case for multiline definitions?

e.g.

```
<<< multiline def
def f(
  name: String,
  anotherName: Int)
>>>
def f(
  name: String,
  anotherName: Int)
```

We need to verify that spaces adds only into singleline definitions.

sirmax: Hmm… not adding a space in a multiline def is actually inconsistent with default scalafmt behaviour (i.e. when `verticalMultiline.atDefnSite = false`).

Unfortunately there's no specific test for defs in [`spaces/InParentheses.stat`](https://github.com/scalameta/scalafmt/blob/v2.2.1/scalafmt-tests/src/test/resources/spaces/InParentheses.stat) but this one seems pretty close
```
<<< maxColumn apply
def foo = bar(aaaaaaaaaaaaa, bbbbbbbbbbbbbb, cccccccccc)
>>>
def foo =
  bar(
    aaaaaaaaaaaaa,
    bbbbbbbbbbbbbb,
    cccccccccc )
```

I'd like to emphasise that the fix is about the _leading_ space, not the trailing one.

I may add a test for a case that was already working, just for the sake of consistency and regression testing. Should I?
```
<<< call site
x(foo)
>>>
x( foo )
```

poslegm: Oh, sorry, I was sure it worked differently. Then everything is fine 👍

poslegm:
@sirmax Thank you for this bug fix!

scalafmt-tests/src/test/resources/vertical-multiline/withSpacesInParentheses.stat
sirmax: No worries, better safe than sorry :)

Issue #1539, opened 2019-10-23
verticalMultiline disables optIn.configStyleArguments


neko-kai:
- **Version**: 2.2.1
- **Integration**: IntelliJ
- **Configuration**:
```
version = 2.2.1
maxColumn = 100

verticalMultiline.atDefnSite = true
optIn.configStyleArguments = true
```

## Steps

Given code like this:
```scala
class Abc(
  x: Int, 
  b: Int
)
```

## Problem

With `verticalMultiline.atDefnSite = true` scalafmt formats code like this:
```scala
class Abc(x: Int, b: Int)
```
i.e. the `optIn.configStyleArguments=true` option is ignored - the config list is not preserved on re-format.

## Expectation

However, if verticalMultiline is disabled with `verticalMultiline.atDefnSite = true`, the output is:
```scala
class Abc(
  x: Int,
  b: Int
)
```
`optIn.configStyleArguments=true` works again – the config list is not squashed into one line despite being smaller than `maxColumn`

## Workaround

None

kitbellew:
@neko-kai vertical multiline, apparently, intends to fully control formatting regardless of source newlines.

given this, what would you expect to happen? what's the *identical* way it should format both
```
  class A(b, c)
  class D(
    e,
    f
  )
```

you can change its parameters to mimic config style, by reducing arity to 2, setting the flag to break after the opening parenthesis and, in this case, removing class from their exclusion list for f dangling closing parenthesis.

neko-kai:
@kitbellew 
> @neko-kai vertical multiline, apparently, intends to fully control formatting regardless of source newlines.
> given this, what would you expect to happen? what's the identical way it should format both

Clearly this is unsolvable because I want these two to be *non-identical*, i.e. I'm turning on the `optIn` option to control what to break into multi-line and what to keep horizontal. For verticalMultiline that would mean that if the braces are already on separate lines (config-style) it can no longer reformat them back to horizontal and must assume minimal arity threshold for that section.

kitbellew:
> Clearly this is unsolvable because I want these two to be _non-identical_, i.e. I'm turning on the `optIn` option to control what to break into multi-line and what to keep horizontal. For verticalMultiline that would mean that if the braces are already on separate lines (config-style) it can no longer reformat them back to horizontal and must assume minimal arity threshold for that section.

if i may ask, why do you want to use vertical multiline if you'd like to control line breaks? is something missing in the non-vertical-multiline mode?

[...]

PR #1541, opened 2019-10-28
Remove `avoidEmptyLinesAroundBlock` option

olafurpg:
Followup from https://github.com/scalameta/scalafmt/pull/1431. This setting is already covered by the `edition` option. 

An important aspect of editions is that users will not be able to pick and choose individual changes to the default formatting behavior. The motivation for this restriction is to

* make the project more maintainable: it's difficult to reason how all possible combinations of off-by-default settings work in practice
* improve the user experience: currently, users can mix and match individual settings that hit on bugs such as #1539.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Newlines.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Newlines.scala
@@ -110,8 +110,7 @@ case class Newlines(
     beforeImplicitKWInVerticalMultiline: Boolean = false,
     alwaysBeforeElseAfterCurlyIf: Boolean = false,
     alwaysBeforeMultilineDef: Boolean = true,
-    avoidAfterYield: Boolean = true,
-    avoidEmptyLinesAroundBlock: Option[Boolean] = None
+    avoidAfterYield: Boolean = true
 ) {
   val reader: ConfDecoder[Newlines] = generic.deriveDecoder(this).noTypos
 }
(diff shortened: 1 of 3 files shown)


neko-kai:
Without an option or an `optIn` to turn off empty line stripping, I'd have to permanently stay on an old edition, since empty lines at start of definitions are frequently used in my codebases to reduce vertical clamping  – and ostensibly lose access to further changes, so I'm quite resistant to this change.

olafurpg:
> empty lines at start of definitions are frequently used in my codebases to reduce vertical clamping

Can you elaborate on this? Does it apply to all blocks or only object/class definitions?

neko-kai:
@olafurpg 
Mainly objects, but also defs that start with a long val assignment or an inner def, e.g.

e.g. [object example](https://github.com/7mind/izumi/blob/cc344a10d310a34c5a92622b54fc85828ed0adc3/distage/distage-config/src/main/scala/izumi/distage/config/ConfigProvider.scala#L219), [def example](https://github.com/7mind/izumi/blob/194ac6c37b7da838ef96e641869da475b2f6b6ec/distage/distage-model/src/main/scala/izumi/distage/model/definition/DIResource.scala#L413)

olafurpg:
Thanks for the links. I see what you mean and I agree it might be desirable to allow blank lines in those cases. I just looked at what Prettier does in this situation and it seems to strip away both the opening and closing blank lines.

I'm open to start a separate discussion whether we should restrict the changes in #1431 to only strip the closing blank line and allowing users to manually keep the opening blank line.

This PR is only to complete the implementation of editions in #1431. Keeping around hidden/undocumented settings for people who dig into the source code is not a sustainable way to maintain this project.

Issue #1542, opened 2019-10-28
Configuration for keep empty lines at the start of code blocks


poslegm:
Default behavior was changed in https://github.com/scalameta/scalafmt/pull/1431 and https://github.com/scalameta/scalafmt/pull/1541. 

Now scalafmt removes empty lines at the end and start of code blocks. But empty lines at start of definitions can be used to reduce vertical clamping.

Should we introduce `newlines.keepEmptyLineAtBlockStart` option for keeping them?

neko-kai:
I don't see how this reduces code complexity vs. just keeping `newlines.avoidEmptyLinesAroundBlock` public, especially since the configuration is in the source code anyway because of ediition switching.

olafurpg:
@neko-kai the problem with supporting `N` boolean configuration options is that they can be combined in `2^N` number of ways, of which most combinations have never been tested. An example issue where settings are combined in unexpected ways is https://github.com/scalameta/scalafmt/issues/1539.

olafurpg:
I'm open to partially revert the changes from #1431 so that it only strips closing blank lines while still allowing users to control the opening blank lines, as discussed in https://github.com/scalameta/scalafmt/pull/1541#issuecomment-546910497. However, I'm not sure we should go ahead with that partial revert right away. For now, users can stay on `edition = 2019-10` if they dislike the new change and report feedback in this issue that they are blocked from upgrading to the latest edition.

olafurpg:
The new behavior introduced by #1431 mirrors the formatting behavior in Prettier. From what I've seen, Prettier doesn't have an option to control this specific behavior so I think it would be great if we can avoid it as well.

neko-kai:
@olafurpg 
Why is Prettier related? It may be popular among open source library authors, but e.g. Intellij/WebStorm JavaScript formatter has a large amount of options, as expected.

>  the problem with supporting N boolean configuration options is that they can be combined in 2^N number of ways, of which most combinations have never been tested. An example issue where settings are combined in unexpected ways is #1539.

The alternative is compromising on functionality – the Scala community is plural, with multiple divergent writing styles, the fact that basically everyone tweaks scalafmt's defaults and uncovers these bugs supports that – a less configurable formatter excludes members of the community with styles that aren't replicable by defaults. tbh personally I'd rather see broken combinations – with an opportunity for contributors to fix them if they want to use, rather than a blanket ban on options – that also excludes contributors from adding the options that they do want to use.

[...]

PR #1546, opened 2019-11-05
AvoidEmptyLinesAroundBlock becomes configurable

poslegm:
Release with unconfigurable default `newlines.avoidEmptyLinesAroundBlock` received a lot of negative feedback in https://github.com/scalameta/scalafmt/issues/1542

I agree with decision to not add too much "off-by-default" settings, but exactly this behaviour may be too opinionated for unconfigurable default.

@tanishiking @olafurpg what do you think?


poslegm:
Yes, users can turn off this setting with `edition = 2019-10`. But it will block them from next editions update. I want this setting to be merged to prevent mass usage of legacy edition.

> dropping the "edition" setting. Having two ways to configure the same formatting behavior is more confusing than helpful IMO

Editions will be helpful in future. The documentation describes only one way to configure empty lines: `newlines.avoidEmptyLinesAroundBlock`. So I don't think that users will be confused by `edition = 2019-10` (they won't even know about it).

> documenting our policy for introducing new configuration options

Fully agreed! I think about 10 (the exact number may vary) reactions in issue as threshold.

olafurpg:
Judging by the feedback in #1542, it seems like it would be better to revert #1431. I believe we it would be less controversial to merge a fix that always removes blank lines before closing braces in method implementations

```scala
//  before
def foo = {
  println(42)

}
//  after
def foo = {
  println(42)
}
```

Another change I think would be worthwhile, is enforcing a blank line before the closing brace if there's a blank line after the opening brace.

```scala
//  before
class Foo {

  def bar = 42
}
//  after
class Foo {

  def bar = 42

}
```

We really should make an effort to improve the default behavior.

poslegm:
I am not sure that more complicated behaviour you described would be good**unconfigurable** default ¯\_(ツ)_/¯
It will force user to think about it while coding:
"aha... if I leave an empty line here than scalafmt will add line at the end of block..."

------

> We really should make an effort to improve the default behavior.

Do we _really_ have problems with default behaviour on empty lines? I think it's a matter of flavour and should be delegated to little config tuning not great changes of defaults.

------

Anyway, I'd rather have `newlines.avoidEmptyLinesAroundBlock` under configuration flag than fully reverted.

olafurpg:
The feedback in #1542 is mostly about having the option to let code "breathe". My proposal in https://github.com/scalameta/scalafmt/pull/1546#issuecomment-551203432 would implement the behavior where the user still has the choice to let the code "breathe" while 

* avoiding the undesirable behavior of having a blank line before the closing brace of a method (I don't think anyone has raised an issue against this)
* enforcing that when you let the code "breathe", there's always a balanced empty line before and after the opening/closing braces. I don't think anyone has expressed the desire for the following output
```scala
object Foo {

  def bar = 42
}
// or
object Foo {
  def bar = 42

}
```

dwijnand:
> * enforcing that when you let the code "breathe", there's always a balanced empty line before and after the opening/closing braces. I don't think anyone has expressed the desire for the following output

Big 💯 here for that symmetry. ❤️

dwijnand:
> Yes, users can turn off this setting with `edition = 2019-10`. But it will block them from next editions update. I want this setting to be merged to prevent mass usage of legacy edition.

I agree, it's too big a hammer.  I didn't realise this when Olaf was proposing the edition system (which I support) but one big difference in how Rust did it and how Scalafmt did it is that in Rust you _opt-in_ to Rust edition 2018 (retroactively edition was defined as have a 2015 default).  In Scalafmt instead the new version by default used the new edition with the changed behaviour, and allows you to opt-out.

The difference is significant because in Rust you upgraded the compiler and then at some point (maybe immediately) you try to upgrade to Rust 2018, and give up if things don't work out for whatever reason.  In Scalafmt ScalaSteward upgraded everyone's Scalafmt which brings the new edition and the new behaviour, out the box.  If they checked formatting in PR validation then it would've failed if they had code like this one.

poslegm:
After reading this PR https://github.com/cb372/cats-retry/pull/112 I realized that current behavior of `newlines.avoidEmptyLinesAroundBlock` is a horrible mistake. 

Most of removed empty lines located not in function blocks but around [traits](https://github.com/cb372/cats-retry/pull/112/files#diff-50ae8ff2639e50848ea47680267ef743) and classes bodies. I don't see any reason to remove it.

---

Benefit of [balancing](https://github.com/scalameta/scalafmt/pull/1546#issuecomment-551203432) empty lines _everywhere_ is not clear. Common case for empty line at the start of function is another function inside:
```scala
//  before
def wow(): Unit = {

  def iter(i: Int): Unit = {
    if (i <= 0) ()
    else {
      println("such recursion")
      iter(i - 1)
  }

  iter(10)
}
//  after
def wow(): Unit = {

  def iter(i: Int): Unit = {
    if (i <= 0) ()
    else {
      println("such recursion")
      iter(i - 1)
  }

  iter(10)
                         // <--- what is benefit of this line?
}
```

Balancing empty lines only in traits and classes seems reasonable. But is it really necessary for default behavior? It's still gonna lead to reformat PRs, merge conflicts and users on old edition. How many users really need it?

---

So now I think that best solution for us is fully revert this behavior and don't do anything with empty lines until a sufficient number of users provide clear and argued proposal.

poslegm:
I'm leaning towards closing this PR in favor to https://github.com/scalameta/scalafmt/pull/1548

PR #1548, opened 2019-11-10
Revert avoidEmptyLinesAroundBlock feature

poslegm:
After discussions in #1542 and #1546 I realized that full revert of avoidEmptyLinesAroundBlock would be the best solution.

Further suggestions for default behavior on empty lines can be discussed in issues.

What do you think?


tdidriksen:
I must admit that I'm bit sorry to see this being removed completely, instead of being left as a configurable option. But I guess the sheer number of already configurable options was one of the main points of the whole discussion.

PR #1549, opened 2019-11-11
RedundantBraces: remove nested braces in lambda

kitbellew:
Example diffs: https://github.com/kitbellew/scala-repos/commits/scalafmt-1549

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
@@ -192,6 +192,13 @@ case class ScalafmtConfig(
     continuationIndent.callSite,
     continuationIndent.defnSite
   )
+
+  // Edition-specific settings below
+  def activeFor(edition: Edition): Boolean =
+    Edition.ordering.gteq(this.edition, edition)
+
+  // Edition 2019-11
+  val activeForEdition_2019_11: Boolean = activeFor(Edition(2019, 11))
 }
 
 object ScalafmtConfig {
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -130,6 +130,10 @@ case object RedundantBraces extends Rewrite {
         !isProcedureSyntax(d) &&
         !disqualifiedByUnit
 
+      case p: Term.Function
+          if ctx.style.activeForEdition_2019_11 && isBlockFunction(p) =>
+        settings.methodBodies
+
       case _ =>
         settings.generalExpressions &&
           exactlyOneStatement &&
--- a/scalafmt-tests/src/test/resources/rewrite/RedundantBraces.stat
+++ b/scalafmt-tests/src/test/resources/rewrite/RedundantBraces.stat
@@ -154,3 +154,57 @@ object a {
   def x: Int = // comment
     2
 }
+<<< fixes nested block in multi-line lambda 1
+object a {
+  method { x => { // 1
+        2
+        3
+        4
+     }
+  }
+}
+>>>
+object a {
+  method { x => // 1
+    2
+    3
+    4
+  }
+}
+<<< fixes nested block in multi-line lambda 2
+object a {
+  method { x =>
+     { // 1
+        2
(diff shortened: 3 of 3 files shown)


kitbellew:
@olafurpg @tanishiking my first submission, please take a look.

I am also working on another but will likely need some guidance, can't figure out splits/policies/decisions. I will send a draft for comments shortly.

poslegm:
LGTM 👍 

I want to test it for regressions on large codebase with #1551. I'll try to do it by the end of the week.

kitbellew:
@poslegm @tanishiking thanks for taking a look. On my side, I tested on `scala-repos`, by adding `rewrite.rules = [RedundantBraces]` to its `.scalafmt.conf` and running a binary built from the master first, then using the new one on it as well. There were no surprises after this second iteration.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -130,6 +130,9 @@ case object RedundantBraces extends Rewrite {
         !isProcedureSyntax(d) &&
         !disqualifiedByUnit
 
+      case p: Term.Function if isBlockFunction(p) =>
```
olafurpg: Can we guard this against edition 2019-11? The idea is that people should be able to upgrade to the latest Scalafmt version without having large diffs. Sometimes you want to upgrade to the latest version to enjoy for example a parser fix but need more time to get the formatting changes merged into your codebase.

kitbellew: will do

kitbellew:
@olafurpg @tanishiking @poslegm here's a fork of `scala-repos` showing the differences:

https://github.com/kitbellew/scala-repos/commits/scalafmt-1549

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -130,6 +132,10 @@ case object RedundantBraces extends Rewrite {
         !isProcedureSyntax(d) &&
         !disqualifiedByUnit
 
+      case p: Term.Function
+          if isBlockFunction(p) && ctx.style.activeFor(Edition(2019, 11)) =>
```
kitbellew: as requested, added an `Edition` barrier

olafurpg: Let’s bind this to a val to avoid repeated evaluation

kitbellew: Done, uploading now.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Rewrite.scala
```diff
@@ -15,6 +17,8 @@ case class RewriteCtx(
   implicit val dialect = style.runner.dialect
   def isMatching(a: Token, b: Token) =
     matchingParens.get(TokenOps.hash(a)).contains(b)
+
+  val activeForEdition_2019_11: Boolean = style.activeFor(Edition(2019, 11))
```
kitbellew: as requested, created a field for this flag

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -192,6 +192,13 @@ case class ScalafmtConfig(
     continuationIndent.callSite,
     continuationIndent.defnSite
   )
+
+  // Edition-specific settings below
```
kitbellew: if we want to keep predefined edition flags (rather than comparing two editions for every token, as @olafurpg pointed out), it probably makes sense to put them here.

after we merge this, i will rebase #1551 as it uses editions as well.

kitbellew:
@olafurpg @poslegm @tanishiking at the risk of sounding impatient, i wonder if there's anything else you'd like to see. i would like to merge this and #1551 first, and then work on the other changes (I have half a dozen in pipeline, with a couple already submitted for review).

poslegm:
@olafurpg I took the responsibility to merge this PR because your requested changes completed. If you have another remarks we can do it separately.

PR #1551, opened 2019-11-11
Router: format lambda in 1-arg call as with braces

### LINKED ISSUE #667: Move lambda arg to first line

kitbellew:
N.B. fork of `scala-repos` showing the difference: https://github.com/kitbellew/scala-repos/commits/scalafmt-1551

Fixes #667

It is desirable to have an option to format paren lambda in calls just
like a curly-brace lambda. For instance, currently:

```
  call { parameter =>
    long_body
  }
```

but

```
  call(
    parameter =>
      long_body
  )
```

instead of the similar (and more compact)

```
  call(parameter =>
    long_body
  )
```

Obviously, when the call refers to multiple arguments (which is never
the case with curly-brace lambda), this formatting wouldn't make sense
and would continue to be

```
  call(
    parameter =>
      long_body,
    another_argument
  )
```

N.B. Using the existing `alwaysBeforeCurlyBraceLambdaParams` configuration parameter is also possible but it magically fixed an error with `align/ArrayIndexOutOfBounds800.source` test (even if I explicitly flipped the `alwaysBeforeCurlyBraceLambdaParams` setting) which, I think, is intended to fail.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
@@ -199,6 +199,8 @@ case class ScalafmtConfig(
 
   // Edition 2019-11
   val activeForEdition_2019_11: Boolean = activeFor(Edition(2019, 11))
+  val newlinesBeforeSingleArgParenLambdaParams: Boolean =
+    newlines.alwaysBeforeCurlyBraceLambdaParams || !activeForEdition_2019_11
 }
 
 object ScalafmtConfig {
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -697,6 +697,21 @@ class Router(formatOps: FormatOps) {
           else if (right.is[T.Comment]) newlines2Modification(between)
           else NoSplit
 
+        val keepConfigStyleSplit =
+          style.optIn.configStyleArguments && newlines != 0
+        val splitForAssignToken =
+          if (defnSite || isBracket || keepConfigStyleSplit) None
+          else
+            getAssignAtSingleArgCallSite(leftOwner).map { assign =>
+              val assignToken = assign.rhs match {
+                case b: Term.Block => b.tokens.head
+                case _ => assign.tokens.find(_.is[T.Equals]).get
+              }
+              Split(NoSplit, 1 + nestedPenalty + lhsPenalty)
+                .withPolicy(newlinePolicy.andThen(SingleLineBlock(assignToken)))
+                .withOptimalToken(assignToken)
+            }
+
         Seq(
           Split(noSplitModification, 0, policy = noSplitPolicy)
             .withOptimalToken(expirationToken)
@@ -705,7 +720,7 @@ class Router(formatOps: FormatOps) {
             newlineModification,
             (1 + nestedPenalty + lhsPenalty) * bracketMultiplier,
             policy = newlinePolicy.andThen(singleLine(4)),
-            ignoreIf = args.length > 1 || isTuple
+            ignoreIf = args.length > 1 || isTuple || splitForAssignToken.isDefined
           ).withOptimalToken(expirationToken)
             .withIndent(indent, close, Right),
           Split(
@@ -724,7 +739,7 @@ class Router(formatOps: FormatOps) {
             ignoreIf = singleArgument || isTuple
           ).withOptimalToken(expirationToken)
             .withIndent(indent, close, Right)
-        )
+        ) ++ splitForAssignToken.toSeq
 
       // Closing def site ): ReturnType
       case FormatToken(left, T.Colon(), _)
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -460,6 +460,39 @@ class Router(formatOps: FormatOps) {
         verticalMultiline(leftOwner, ft)(style)
 
       // Term.Apply and friends
+      case FormatToken(T.LeftParen(), _, _)
+          if style.optIn.configStyleArguments &&
+            !style.newlinesBeforeSingleArgParenLambdaParams &&
+            getLambdaAtSingleArgCallSite(leftOwner).isDefined => {
+        val lambda = getLambdaAtSingleArgCallSite(leftOwner).get
+        val (lambdaIsABlock, lambdaToken) = lambda.body match {
+          case b: Term.Block =>
+            true -> b.tokens.head
+          case _ =>
+            val arrow = lambda.tokens.find(_.is[T.RightArrow]).get
+            next(arrow).is[T.LeftBrace] -> arrow
+        }
+
+        val close = matchingParentheses(hash(formatToken.left))
+        val splitOnClosePolicy =
+          if (style.danglingParentheses.callSite && !lambdaIsABlock)
+            newlineOnTokenPolicy(close)
+          else
+            Policy(Policy.emptyPf, close.end)
+        val noSplitBeforeArrowPolicy =
+          splitOnClosePolicy.andThen(SingleLineBlock(lambdaToken))
+
+        val newlinePenalty = 3 + nestedApplies(leftOwner)
+        Seq(
+          Split(NoSplit, 0, policy = SingleLineBlock(close))
+            .withOptimalToken(close),
+          Split(NoSplit, 0, policy = noSplitBeforeArrowPolicy)
+            .withOptimalToken(lambdaToken),
+          Split(Newline, newlinePenalty, policy = splitOnClosePolicy)
+            .withIndent(style.continuationIndent.callSite, close, Right)
+        )
+      }
+
       case FormatToken(T.LeftParen() | T.LeftBracket(), right, between)
           if style.optIn.configStyleArguments &&
             (isDefnSite(leftOwner) || isCallSite(leftOwner)) &&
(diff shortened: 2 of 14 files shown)


kitbellew:
@olafurpg @tanishiking @poslegm this is the other proposal I mentioned in #1549 , please take a look. I *think* it works as I intended.

poslegm:
Thank you for it!

I will read the code later. Anyway it is a fix for https://github.com/scalameta/scalafmt/issues/667 (old and annoying issue) 👍 

Configuration flag is unnecessary because the issue marked as `should`. 
It should be just enhancement for scalafmt behavior, but with `edition` support (read more about editions and how to add it into your PR: https://github.com/scalameta/scalafmt/pull/1431).

**TODO** document editions on https://scalameta.org/scalafmt/docs/contributing-scalafmt.html

But we need to test changes very thoroughly. May be https://github.com/olafurpg/scala-repos would be helpful for regression testing.

kitbellew:
@poslegm кто ж знал, что не только нас это беспокоило... посмотрю на editions.

п.с. а что насчет #1549?

kitbellew:
@poslegm @tanishiking fyi: implemented editions but now working on that `scala-repos` test. on it, the changes are subtle but I am trying to minimize them. Will update when done.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -461,6 +462,39 @@ class Router(formatOps: FormatOps) {
...
+        val close = matchingParentheses(hash(formatToken.left))
+        val splitOnClosePolicy =
+          if (style.danglingParentheses.callSite && !lambdaIsABlock)
+            newlineOnTokenPolicy(close)
+          else
+            Policy(Policy.emptyPf, close.end)
+        val noSplitBeforeArrowPolicy =
+          splitOnClosePolicy.andThen(SingleLineBlock(lambdaToken))
+
+        val newlinePenalty = 3 + nestedApplies(leftOwner)
```
kitbellew: This part is voodoo. I started with `1` but ended up with this after testing on `scala-repos`.

scalafmt-tests/src/test/resources/newlines/beforeConfigSingleArgParenLambdaParamsF_danglingF.stat
```diff
@@ -216,9 +193,9 @@ def f = {
 >>>
 EventFilter[ActorInitializationException](occurrences = 1) intercept {
   intercept[akka.actor.ActorInitializationException] {
-    wrap(
-      result ⇒
-        actorOf(Props(new OuterActor(
+    wrap(result ⇒
+      actorOf(
```
kitbellew: This split is the only thing I couldn't fix, played with penalties in that match branch where I added the "preserve no-split in assign" logic.

Specifically, the sum `nestedPenalty + lhsPenalty` seemed to be constant for each of these nested calls which doesn't indicate a preference between them.

I tried `2 * nestedPenalty + lhsPenalty` to try and split earlier but it ultimately produced too many changes. I also tried `nestedPenalty + 2 * lhsPenalty` to prefer a later split, but it was also not working as well. So I gave up.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Newlines.scala
```diff
@@ -113,6 +113,9 @@ case class Newlines(
     avoidAfterYield: Boolean = true
 ) {
   val reader: ConfDecoder[Newlines] = generic.deriveDecoder(this).noTypos
+
+  def beforeSingleArgParenLambdaParams(implicit edition: Edition): Boolean =
```
olafurpg: Let's make this a val in `ScalafmtConfig` and remove the implicit edition variable.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1535,3 +1581,19 @@ class Router(formatOps: FormatOps) {
       Seq(Split(Space, 0), Split(Newline, 1))
     }
 }
+
+private object Router {
+
+  def getAssignAtSingleArgCallSite(tree: Tree): Option[Term.Assign] =
```
olafurpg: Let's move these to `TreeOps`. The `Router.scala` file is already too big :)

scalafmt-tests/src/test/resources/default/SearchState.stat
```diff
@@ -455,8 +455,8 @@ EventsResponse.fromJson(json, run).success.value.events should be(
     ))
 <<< react #458
   val TodoApp = ReactComponentB[Unit]("TodoApp")
-    .initialState(State(Nil, ""))
-    .renderS(($, s) =>
+    .initialState(State(Nil, "")).renderS(
```
olafurpg: Is this an desirable change? This case looks like a regression compared to before.

scalafmt-tests/src/test/resources/default/SearchState.stat
```diff
@@ -469,16 +469,16 @@ EventsResponse.fromJson(json, run).success.value.events should be(
...
-      ($, s) =>
-        <.div(<.h3("TODO"),
-              TodoList(s.items),
-              <.form(
-                  ^.onSubmit ==> $._runState(handleSubmit), // runState runs a state monad and applies the result.
-                  <.input( // _runState is similar but takes a function-to-a-state-monad.
-                      ^.onChange ==> $._runState(acceptChange), // In these cases, the function will be fed the JS event.
-                      ^.value := s.text),
-                  <.button("Add #", s.items.length + 1))))
+  .renderS(($, s) =>
```
olafurpg: nice 👍

scalafmt-tests/src/test/resources/newdefault/SearchState.stat
```diff
@@ -1,6 +1,7 @@
 style = default
 <<< forAll
-forAll(table)((reqProto, headReq, reqCH, resProto, resCH, resCD, renCH, close) ⇒
+forAll(table)(
+  (reqProto, headReq, reqCH, resProto, resCH, resCD, renCH, close) ⇒
```
olafurpg: Is this also a desirable change?

scalafmt-tests/src/test/resources/newlines/beforeConfigSingleArgParenLambdaParamsF_danglingF.stat
```diff
@@ -0,0 +1,270 @@
...
+  something.call (
+     (x, y) => {
+       g(x)
+       g(y)
+     }
+  )
+}
+>>>
+def f = {
+  something.call((x, y) => {
```
olafurpg: Nice!

scalafmt-tests/src/test/resources/newlines/beforeConfigSingleArgParenLambdaParamsF_danglingT.stat
```diff
@@ -0,0 +1,289 @@
...
+            actorOf(Props(promiseIntercept(new FailingInnerActor)(result)))
+          )
+        )
+      )
+    }
+  }
+
+  contextStackMustBeEmpty()
+}
+<<< 16: prefer split on equals vs split on arrow in includedTerms
```
olafurpg: What is this case testing? I'm unable to spot the relevant line

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
olafurpg: Mostly everything in the router is vodoo 😅 it's normal to reach some magic penalty based on running the formatter on scala-repos 👍

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Newlines.scala
kitbellew: will do

kitbellew:
@olafurpg @tanishiking @poslegm here's a fork of `scala-repos` showing the difference:

https://github.com/kitbellew/scala-repos/commits/scalafmt-1551

scalafmt-tests/src/test/resources/default/SearchState.stat
olafurpg: Whoops, for some reason I didn't see that. Looks great!

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -192,6 +192,12 @@ case class ScalafmtConfig(
     continuationIndent.callSite,
     continuationIndent.defnSite
   )
+
+  def activeFor(edition: Edition): Boolean =
+    Edition.ordering.gteq(this.edition, edition)
+
+  def newlinesBeforeSingleArgParenLambdaParams: Boolean =
```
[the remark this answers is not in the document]
poslegm: Nitpicking: it should be a `val` because it's a constant value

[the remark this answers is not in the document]
kitbellew: will do in a minute

kitbellew:
@poslegm amended to use `val`, agree it's better. also rebased on top of merged 1549.

PR #1522, opened 2019-10-09
Update Scala to 2.13

poslegm:
Fixes #1434 

Aaaand now pipeline is green :rocket: 

1. Please take a look at [ScalafmtVersion](https://github.com/scalameta/scalafmt/pull/1522/files#diff-03988867feb59caad531803ea8b69f21) introduced in this PR. Now it's difficult to compare scalafmt versions in `ScalafmtDynamicDownloader` only with manual string parsing (e.g. `.startsWith("0.1")`).

2. Hard-coded scalafmt version with Scala 2.13 in `ScalafmtDynamicDownloader` is 2.2.0. Is it ok? **UPD: 2.1.2**

3. There are a lot of garbage commits in my branch. I don't like to force push into already published branches, so squash it on merge, please :smile:

PR #1552, opened 2019-11-13
Make it easier to test scalafmt local snapshot build

tanishiking:
While I tried to test https://github.com/scalameta/scalafmt/pull/1549 locally, I found that scalafmt has mainly two things that make it cumbersome to test locally.

## scala build version mismatch
- Before this commit, `sbt publishLocal` publish the build to local ivy with the version something like `2.1.0+xxxx-SNAPSHOT` (if people didn't fetch the latest tags to local)
- when scalafmt-dynamic tries to download scalafmt-core whose version was `2.1.0+xxxx-SNAPSHOT`, it looks for scalafmt built with 2.12 (see: https://github.com/scalameta/scalafmt/blob/778c75f728a7a039fa63499ae3f97c69b4a018f0/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamicDownloader.scala#L68-L71)
- However, the default scala version is now 2.13 (as of https://github.com/scalameta/scalafmt/pull/1522) and `sbt publishLocal` would build scalafmt with 2.13 and publish it as `2.1.0+xxxx-SNAPSHOT`.
- As a result, scalafmt-dynamic cannot resolve the local build.
  - because `sbt publishLocal` publish 2.13 build with `2.1.0+xxxx-SNAPSHOT` to local, but if we specify `version=2.1.0+xxxx-SNAPSHOT`, scalafmt would look for 2.12 build.

I know we can avoid this situation by building snapshot build with 2.12 instead of 2.13, but it would be handy if we can test the snapshot version only with `sbt publishLocal` instead of `sbt +publishLocal`.

### Solution
https://github.com/scalameta/scalafmt/commit/c416668ee8c313fc18a3f083ce29aaf581578ede
Introduced `localSnapshotVersion` for specifying the version in local publishing.
The idea of localSnapshotVersion comes from [metals](https://github.com/scalameta/metals/blob/05017dc44e384ab8e0a8202ebd128f697594d686/build.sbt#L1-L19)

## scalafmt-dynamic doesn't accept SNAPSHOT version
scalafmt-dynamic introduced `ScalafmtVersion.scala` in https://github.com/scalameta/scalafmt/pull/1522. It parses the scalafmt version specified in `.scalafmt.conf`, and if the version string is invalid, it rejects to resolve with an error.

For example, if `.scalafmt.conf` contains `version=2.2.3-SNAPSHOT` for testing local build, scalafmt-dynamic parse `2.2.3-SNAPSHOT` as invalid version.

### Solution
So with that, I tweaked the regex for scalafmtVersion https://github.com/scalameta/scalafmt/commit/67f6f0a152ef80d577e65d61a1739539b94a2f19 so that scalafmt-dynamic can parse snapshot version as a valid scalafmtVersion, and we can test the local builds.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,7 @@
 import Dependencies._
 import sbtcrossproject.CrossPlugin.autoImport.crossProject
+def localSnapshotVersion = "2.2.3-SNAPSHOT"
+def isCI = System.getenv("CI") != null
 
 def scala211 = "2.11.12"
 def scala212 = "2.12.8"
@@ -7,6 +9,10 @@ def scala213 = "2.13.1"
 
 inThisBuild(
   List(
+    version ~= { dynVer =>
+      if (isCI) dynVer
+      else localSnapshotVersion // only for local publishng
+    },
     organization := "org.scalameta",
     homepage := Some(url("https://github.com/scalameta/scalafmt")),
     licenses := List(
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,12 @@
 import Dependencies._
 import sbtcrossproject.CrossPlugin.autoImport.crossProject
-def localSnapshotVersion = "2.2.3-SNAPSHOT"
+
+def parseTagVersion: String = {
+  import scala.sys.process._
+  // drop `v` prefix
+  "git describe --abbrev=0 --tags".!!.drop(1).trim
+}
+def localSnapshotVersion: String = s"$parseTagVersion-SNAPSHOT"
 def isCI = System.getenv("CI") != null
 
 def scala211 = "2.11.12"
--- a/scalafmt-dynamic/src/test/scala/tests/ScalafmtVersionSuite.scala
+++ b/scalafmt-dynamic/src/test/scala/tests/ScalafmtVersionSuite.scala
@@ -12,6 +12,15 @@ class ScalafmtVersionSuite extends FunSuite {
       ScalafmtVersion.parse("2.0.0-RC4") == Right(ScalafmtVersion(2, 0, 0, 4))
     )
     assert(ScalafmtVersion.parse("2.1.1") == Right(ScalafmtVersion(2, 1, 1, 0)))
+    assert(
+      ScalafmtVersion
+        .parse("2.2.3-SNAPSHOT") == Right(ScalafmtVersion(2, 2, 3, 0))
+    )
+    assert(
+      ScalafmtVersion.parse("2.0.0-RC1-SNAPSHOT") == Right(
+        ScalafmtVersion(2, 0, 0, 1)
+      )
+    )
   }
 
   test("fail on invalid versions") {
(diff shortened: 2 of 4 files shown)


build.sbt
```diff
@@ -1,12 +1,18 @@
 import Dependencies._
 import sbtcrossproject.CrossPlugin.autoImport.crossProject
+def localSnapshotVersion = "2.2.3-SNAPSHOT"
+def isCI = System.getenv("CI") != null
```
tanishiking: It will be true in TravisCI https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

build.sbt
```diff
@@ -1,12 +1,18 @@
 import Dependencies._
 import sbtcrossproject.CrossPlugin.autoImport.crossProject
+def localSnapshotVersion = "2.2.3-SNAPSHOT"
```
olafurpg: I am a bit concerned that this version needs to be bumped up every time we cut a new release. It would be nice to avoid that.

scalafmt-dynamic/src/test/scala/tests/ScalafmtVersionSuite.scala
```diff
@@ -51,6 +55,11 @@ class ScalafmtVersionSuite extends FunSuite {
       ScalafmtVersion
         .parse("2.1.0-RC1-M4") == Left(InvalidVersionException("2.1.0-RC1-M4"))
     )
+    assert(
+      ScalafmtVersion.parse("2.0.0-RC1-SNAPSHOT") == Left(
+        InvalidVersionException("2.0.0-RC1-SNAPSHOT")
```
poslegm: Why is it invalid?

build.sbt
poslegm: We can parse it from git tags

```scala
def parseTagVersion: String = {
    import scala.sys.process._
    "git describe --abbrev=0 --tags".!!.drop(1).strip
  }

def localSnapshotVersion: String = s"$parseTagVersion-SNAPSHOT"
```

tanishiking: Ah, sorry I just realized that it(my local published build with `2.1.0+xxx-SNAPSHOT`) happened just because I haven't fetched tags from a remote repository for a while... oh my ignorance...

> I am a bit concerned that this version needs to be bumped up every time we cut a new release. It would be nice to avoid that.

Now, I agree with that. 👍 

---

> We can parse it from git tags

Awesome! With this `localSnapshotVersion`, we can simplify the regex in `ScalafmtDynamicDownloader` 👍

scalafmt-dynamic/src/test/scala/tests/ScalafmtVersionSuite.scala
tanishiking: Oh, sorry, it's my mistake, it should be valid.

build.sbt
poslegm: May be document [here](https://github.com/scalameta/scalafmt/blob/master/docs/contributing-scalafmt.md) that it is recommended to do `git fetch` before local building?

tanishiking: It might be helpful though it won't be a problem for most people who are new to this project :)
BTW, since that documentation is becoming older and things have changed a bit (like release process, document locations), we have to update it in another PR.

poslegm:
@tanishiking is there another changes you want to do or we can merge it now?

tanishiking:
Thanks!

> @tanishiking is there another changes you want to do or we can merge it now?

That's all, we can merge it now

PR #1554, opened 2019-11-19
Router: treat try/finally the same as if/else #350

### LINKED ISSUE #350: Try-Finally without curlies formats funny
Original:

``` scala
object Foo {
  try loop(blocks, startingImports, Imports(), wrapperIndex = 1 )
  finally scriptImportCallback = outerScriptImportCallback
}
```

Scalafmt:

``` scala
object Foo {
  try loop(blocks, startingImports, Imports(), wrapperIndex = 1) finally scriptImportCallback =
    outerScriptImportCallback
}
```

``` scala
object Foo {
  try loop(blocks, startingImports, Imports(), wrapperIndex = 1 )
  finally scriptImportCallback = outerScriptImportCallback
}
```

Using:
- 0.2.8
- `scalafmt -i -f Foo.scala`

Note that this seems peculiar to `try`/`finally`: I couldn't reproduce it with `if(true)`/`else`

kitbellew:
Fixes #350

See examples of changes in:

https://github.com/kitbellew/scala-repos/commit/6199bb327e98be07db46d43f2f4b8dffeabeba3e?w=1

### DIFF
--- a/scalafmt-tests/src/test/resources/default/Try.stat
+++ b/scalafmt-tests/src/test/resources/default/Try.stat
@@ -91,41 +91,48 @@ try {
 <<< #350 assignment: short single line
 val template: String = try source.mkString finally source.close()
 >>>
-val template: String = try source.mkString
-finally source.close()
+val template: String =
+  try source.mkString
+  finally source.close()
 <<< #350 assignment: long single line with catch handler and finally
 val template: String = try source.mkString catch veryLongHandlerToExceedLineLength finally source.close()
 >>>
-val template: String = try source.mkString
-catch veryLongHandlerToExceedLineLength
-finally source.close()
+val template: String =
+  try source.mkString
+  catch veryLongHandlerToExceedLineLength
+  finally source.close()
 <<< #350 assignment: long single line with catch and finally
 val template: String = try source.mkString catch { case e if NonFatal(e) => log.error("failed to make string", e) } finally source.close()
 >>>
-val template: String = try source.mkString
-catch { case e if NonFatal(e) => log.error("failed to make string", e) } finally source
-  .close()
+val template: String =
+  try source.mkString
+  catch { case e if NonFatal(e) => log.error("failed to make string", e) } finally source
+    .close()
 <<< #350 assignment: short single line with catch handler and finally
 val template: String = try foo catch bar finally baz
 >>>
-val template: String = try foo
-catch bar
-finally baz
+val template: String =
+  try foo
+  catch bar
+  finally baz
 <<< #350 assignment: short single line with catch and finally
 val template: String = try foo catch { case e => bar } finally baz
 >>>
-val template: String = try foo
-catch { case e => bar } finally baz
+val template: String =
+  try foo
+  catch { case e => bar } finally baz
 <<< #350 assignment: short single line with multi-catch and parens
 val template: String = ( try foo catch { case e => bar; case c => baz } )
 >>>
-val template: String = (try foo
-catch { case e => bar; case c => baz })
+val template: String =
+  (try foo
+  catch { case e => bar; case c => baz })
 <<< #350 assignment: short single line with multi-catch and finally
 val template: String = try foo catch { case e => bar; case c => baz } finally qux
 >>>
-val template: String = try foo
-catch { case e => bar; case c => baz } finally qux
+val template: String =
+  try foo
+  catch { case e => bar; case c => baz } finally qux
 <<< #350 statement: short single line with multi-catch and finally
 try foo catch { case e => bar; case c => baz } finally qux
 >>>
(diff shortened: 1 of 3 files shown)


kitbellew:

scalafmt-tests/src/test/resources/default/Try.stat
```diff
@@ -64,3 +64,9 @@ try { foo()  } catch handler
 try {
   foo()
 } catch handler
+<<< #350 long single line
+val template: String = try source.mkString finally source.close()
+>>>
+val template: String =
+  try source.mkString
+  finally source.close()
```
poslegm: Please try to test

```
<<< ONLY #350 long single line with catch and finally
val template: String = try source.mkString catch { case e if NonFatal(e) => log.error("failed to make string", e) } finally source.close()
>>>
val template: String =
  try source.mkString
  catch { case e if NonFatal(e) => log.error("failed to make string", e) }
  finally source.close()
```

On your branch it formats as
```
val template: String =
  try source.mkString
  catch { case e if NonFatal(e) => log.error("failed to make string", e) } finally source
    .close()
```

kitbellew: thank you, will try to fix

kitbellew:

scalafmt-tests/src/test/resources/default/Try.stat
```diff
@@ -43,7 +43,9 @@ object Foo {
 >>>
 object Foo {
   try proc.waitFor()
-  catch { case e: Throwable => run() }
```
olafurpg: I'm not sure this diff here would be popular, it's quite common to define a catch in one line. Is there a way to preserve the old formatting?

kitbellew: i will try again. earlier, couldn't get it to work.

kitbellew: done

kitbellew:
@olafurpg @poslegm all comments addressed, the `scala-repos` link in the description updated with new diffs.

scalafmt-tests/src/test/resources/default/Try.stat
```diff
@@ -58,9 +58,71 @@ object Foo {
...
+object Foo {
+  try { foo()  }
+  catch { case bar => baz }
+  finally {  qux()  }
+}
+>>>
+object Foo {
+  try {
+    foo()
+  } catch { case bar => baz }
```
poslegm: This formatting confusing me a little. Current behavior is not better:
```scala
object Foo {
  try {
    foo()
  } catch { case bar => baz } finally {
    qux()
  }
}
```

Should scalafmt format it like this?

```scala
// option 1
object Foo {
  try { foo() }
  catch { case bar => baz }
  finally { qux() }
}
// option 2
object Foo {
  try {
    foo()
  } catch {
    case bar => baz
  } finally { 
    qux()
  }
}
```

Anyway this is not a regression because current formatting is more stranger

kitbellew:
> @kitbellew Thank you! That is much better. I found a few more cases of undesirable diffs in scala-repos
> ...
> It looks like we're now forcing a newline after the single-line catch expressions. Would it be possible to preserve those `NoSplit`?

@olafurpg fixed in the new submission. the incremental diff in `scala-repos` is: https://github.com/kitbellew/scala-repos/commit/4448d22752ffa7d6365bd444aacd10913a3cb2d4?w=1

as usual, a lot of it is a personal preference, but i find that

```
val a =
  (try { b }
  catch { c })
```

looks better if formatted as:

```
val a = (
  try { b }
  catch { c }
)
```
or, if applicable,
```
val a = (try { b } catch { c })
```

i couldn't find any flags which would be consistent with this behaviour (vertical multiline seems to refer to method definitions only, as does dangling parenthesis). for now, kept as you requested.

scalafmt-tests/src/test/resources/default/Try.stat
kitbellew: @poslegm i have already implemented and will submit option 1, but was planning to do it separately, after this change is merged (as it claims to treat try/finally as if/else).

let me know if you want me to include it here (as it also modifies if/else).

olafurpg: This PR does not introduce a regression for this particular case so I'm fine with merging as is and we can followup separately to use option 1 or 2 (I don't have a preference)

poslegm: It's great! Better to do it separately

kitbellew:
@olafurpg @poslegm interesting thing. running this merged code against the previously checked `scala-repos` introduces a couple of minor changes, so there's some flipping around.

#1562 is intended as a fix for that.

PR #1555, opened 2019-11-20
FormatWriter: insert newline after packages #1069

### LINKED ISSUE #1069: Insert blank line between package declaration and imports section
- **Version**: 1.3.0
- **Integration**: standalone library
- **Configuration**:
```
```

## Steps

I ran scalafmt like this:

```scala
import org.scalafmt.Scalafmt
import org.scalafmt.config.{Newlines, ScalafmtConfig}

object ScalaFormat {

    val source =
      """package com.foo.bar
import collection.JavaConverters._
import com.foo.baz.Grunka
trait Entity {
  val kind: Kind
  val uri: String
}"""

    val scalafmtConfig = ScalafmtConfig(newlines = Newlines(alwaysBeforeTopLevelStatements = true))
    val formattedSource = Scalafmt.format(source, scalafmtConfig).get
    print(formattedSource)
  }
}
```

## Problem

```scala
package com.foo.bar
import collection.JavaConverters._
import com.foo.baz.Grunka

trait Entity {
  val kind: Kind
  val uri: String
}
```

## Expectation

```scala
package com.foo.bar

import collection.JavaConverters._
import com.foo.baz.Grunka

trait Entity {
  val kind: Kind
  val uri: String
}
```

## Notes

It would be nice to be able to insert a blank line between `package` declaration and imports section.

kitbellew:
Fixes #1069

This commit produces no diffs on `scala-repos`.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -9,7 +9,7 @@ import scala.annotation.tailrec
 import scala.meta.tokens.Token
 import scala.meta.tokens.{Token => T}
 import scala.meta.transversers.Traverser
-import scala.meta.{Importer, Mod, Term, Tree}
+import scala.meta.{Importer, Mod, Pkg, Term, Tree}
 
 /**
   * Produces formatted output from sequence of splits.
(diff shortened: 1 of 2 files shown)


kitbellew:

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -258,13 +258,31 @@ class FormatWriter(formatOps: FormatOps) {
       case x => x
     }
     initStyle.newlines.alwaysBeforeTopLevelStatements && {
-      topLevelTokens.contains(hash(toks(i).formatToken.right)) && {
-        val (distance, FormatToken(_, nextNonComment, _)) =
-          nextNonCommentWithCount(toks(i).formatToken)
-        isMultiline(actualOwner(nextNonComment).tokens.last, i + distance + 1)
+      val formatToken = toks(i).formatToken
+      def checkPackage: Option[Boolean] = {
```
olafurpg: We should guard this change against `edition = 2019-11`

kitbellew: will add the edition here, and post the diffs from scala-repos later today.

for now, working on #1557, i think the solution is incomplete, but chains are surprisingly difficult.

olafurpg: Chains are very difficult with the current router design. I spent a LOT of time trying to break on plain `a.b` without exploding the search state but failed.

kitbellew:
@olafurpg rebased, added edition, as requested. updated description regarding `scala-repos`, this change had no effect on that codebase.

PR #1556, opened 2019-11-20
Router: fix def body indent with a comment #1240

### LINKED ISSUE #1240: newlines.alwayasBeforeMultilineDef=false create wrong indentation if a previous line ends with comment
- **Version**: 1.6.0-RC4
- **Integration**: sbt
- **Configuration**:
```
newlines.alwayasBeforeMultilineDef = false
```

## Steps

Given code like this: (comment and newline between `=` and body)
```scala
def foo(a: Int): Int = // a comment
  ???
```

```bash
$ sbt scalafmt
```

## Problem

```scala
def foo(a: Int): Int = // a comment
???
```

## Expectation

```scala
def foo(a: Int): Int = // a comment
  ???
```

## Workaround

`newlines.alwaysBeforeMultilineDef = true`

## Notes

Here is failing test case https://github.com/tanishiking/scalafmt/commit/9af668827e9e5dbb3e9140a1bd992af37d4bc38c

kitbellew:
Fixes #1240 

NB: this change produces no diffs on `scala-repos`.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -438,15 +438,18 @@ class Router(formatOps: FormatOps) {
             // The block will take care of indenting by 2.
             Seq(Split(Space, 0))
           case _ =>
+            val rhsIsComment = isSingleLineComment(right)
             Seq(
               Split(
                 Space,
                 0,
-                ignoreIf = newlines > 0 && !rhsIsJsNative,
+                ignoreIf = rhsIsComment || newlines > 0 && !rhsIsJsNative,
                 policy =
                   if (!style.newlines.alwaysBeforeMultilineDef) NoPolicy
                   else SingleLineBlock(expire, exclude = exclude)
               ),
+              Split(Space, 0, ignoreIf = newlines != 0 || !rhsIsComment)
+                .withIndent(2, expire, Left),
               Split(Newline, 1, ignoreIf = rhsIsJsNative)
                 .withIndent(2, expire, Left)
             )
--- a/scalafmt-tests/src/test/resources/newlines/alwaysBeforeMultilineDef.stat
+++ b/scalafmt-tests/src/test/resources/newlines/alwaysBeforeMultilineDef.stat
@@ -18,3 +18,32 @@ def foo = Future({
   val bar = fetchBar()
   bar.map(_ + 10)
 })
+<<< Indent without comment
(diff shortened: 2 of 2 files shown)


kitbellew:

scalafmt-tests/src/test/resources/newlines/alwaysBeforeMultilineDefF.stat
```diff
@@ -0,0 +1,49 @@
+maxColumn = 40
```
olafurpg: Is the rename of this file intentional?

scalafmt-tests/src/test/resources/newlines/alwaysBeforeMultilineDefF.stat
```diff
@@ -0,0 +1,49 @@
...
+      ???
+>>>
+def foo(a: Int): Int = // a comment
+  ???
+<<< Indent after comment with block
+def foo(a: Int): Int = // a comment
+{
+  ??? }
+>>>
+def foo(a: Int): Int = // a comment
```
olafurpg: 👍

scalafmt-tests/src/test/resources/newlines/alwaysBeforeMultilineDefF.stat
kitbellew: it was, although i can't find the change which contains the test with `T`...

olafurpg: What happens if you rename the file back to the original name?

kitbellew: done

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -438,6 +438,7 @@ class Router(formatOps: FormatOps) {
             // The block will take care of indenting by 2.
             Seq(Split(Space, 0))
           case _ =>
+            val rhsIsComment = isSingleLineComment(right)
             Seq(
               Split(
```
olafurpg: What happens if we `ignoreIf = rhsIsComment` on this split? I'm a bit uncomfortable with having three branches here.

kitbellew: it worked, good eye. resubmitted.

poslegm:
LGTM

@olafurpg do you have any notes?

PR #1557, opened 2019-11-20
Router: apply select indent after a comment #1334

### LINKED ISSUE #1334: Wired alignment when chain contains comment / no parenthesis call / with parenthesis call
- **Version**: 1.5.1 (scalafmt plugin version)
- **Integration**: IntelliJ
- **Configuration**:
```
maxColumn = 150
docstrings = JavaDoc
align = none
```

## Steps

Given code like this:
```scala
class Foo {
  val vv = v.aaa //
    .aaa
    .bbb()
}
```

```bash
<press Reformat with scalafmt key>
```

## Problem

```scala
class Foo {
  val vv = v.aaa //
  .aaa
    .bbb()
}
```

## Expectation

```scala
class Foo {
  val vv = v.aaa //
    .aaa
    .bbb()
}
```
Or like this:
```scala
class Foo {
  val vv = v
    .aaa //
    .aaa
    .bbb()
}
```

## Notes

I found a config key `includeNoParensInSelectChains` in the [document](https://scalameta.org/scalafmt/docs/configuration.html#includenoparensinselectchains) may related with this issue.
But after I added this key to my conf, intellij IDEA reports:
```
12:33	Scalafmt
			Failed to read /................/.scalafmt.conf. 
			Invalid fields: includeNoParensInSelectChains
```
Meanwhile the `Show original` / `Show formatted` button in the [document](https://scalameta.org/scalafmt/docs/configuration.html#includenoparensinselectchains) for this config key was not functional either at this time.

kitbellew:
Fixes #1334

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1340,6 +1340,16 @@ class Router(formatOps: FormatOps) {
           Split(Space, 0, policy = singleLine),
           Split(Newline, 1).withPolicy(penalizeNewlineByNesting(cond, arrow))
         )
+      case FormatToken(_, dot @ T.Dot(), _) =>
+        // allow keeping a newline (and indent) within a select chain
+        val noNewLine = !style.activeForEdition_2019_11 ||
+          newlines == 0 || rightOwner.isNot[Term.Select]
+        val newlinePenalty = if (noNewLine) 0 else 10 + treeDepth(rightOwner)
+        Seq(
+          Split(NoSplit, 0),
+          Split(Newline, newlinePenalty, ignoreIf = noNewLine)
+            .withIndent(2, dot, Left)
+        )
       // Inline comment
       case FormatToken(_, c: T.Comment, between) =>
         Seq(Split(newlines2Modification(between), 0))
@@ -1419,10 +1429,6 @@ class Router(formatOps: FormatOps) {
           Split(NoSplit, 0)
         )
       // Fallback
-      case FormatToken(_, T.Dot(), _) =>
-        Seq(
-          Split(NoSplit, 0)
-        )
       case FormatToken(left, T.Hash(), _) =>
         Seq(
           Split(if (endsWithSymbolIdent(left)) Space else NoSplit, 0)
--- a/scalafmt-tests/src/test/resources/default/Select.stat
+++ b/scalafmt-tests/src/test/resources/default/Select.stat
@@ -277,3 +277,78 @@ object a {
   keys.clear() // we need to remove the selected keys from the set, otherwise they remain selected
 >>>
 keys.clear() // we need to remove the selected keys from the set, otherwise they remain selected
+<<< #1334 continue chain indent after a comment
+class Foo {
+  val vv = v.aaa //
+.bbb
+.ccc()
+}
+>>>
+class Foo {
+  val vv = v.aaa //
+  .bbb.ccc()
+}
+<<< #1334 continue chain indent after a comment, a bit longer
+class Foo {
+  val vv = v.aaa //
+  .bbb //
+  .ccc //
+  .ddd
+  .eee()
+}
+>>>
+class Foo {
+  val vv = v.aaa //
+  .bbb //
+  .ccc //
+  .ddd.eee()
+}
+<<< #1334 continue chain indent after a comment with apply
+class Foo {
+  val vv = v.aaa //
+  .bbb() //
+  .ccc //
+  .ddd()
+}
+>>>
+class Foo {
+  val vv = v.aaa //
+    .bbb() //
+    .ccc //
+    .ddd()
+}
+<<< #1334 continue chain indent after a comment with apply, longer
+class Foo {
+  val vv = v.aaa //
+  .bbb() //
+  .ccc() //
+  .ddd()
+}
+>>>
+class Foo {
+  val vv = v.aaa //
+    .bbb() //
+    .ccc() //
+    .ddd()
+}
+<<< #1334 very long unbroken select chain
+class Foo {
+  val vv = v.aaaaaaaaaa.bbbbbbbbbb.cccccccccc.dddddddddd.eeeeeeeeee.ffffffffff.gggggggggg.hhhhhhhhhh.iiiiiiiiii
+}
+>>>
+class Foo {
+  val vv =
+    v.aaaaaaaaaa.bbbbbbbbbb.cccccccccc.dddddddddd.eeeeeeeeee.ffffffffff.gggggggggg.hhhhhhhhhh.iiiiiiiiii
+}
+<<< #1334 very long broken select chain
+class Foo {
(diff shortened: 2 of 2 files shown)


olafurpg:
Thank you for this contribution! I really appreciate you taking a look at these long standing issues 🙏 

At a quick glance this looks great but I won't be able to review properly until next week earliest. In the meantime, can you please update the PR descriptions to include before/after examples. I personally try to structure my PR descriptions in the following way

> Previously, Scalafmt formatted code like this: EXAMPLE CODE. Now, Scalafmt formats the same code like this instead: NEW FORMAT OUTPUT.

You can also include "Fixes #ISSUE" in the description so that the issue is automatically closed when merged.

scalafmt-tests/src/test/resources/default/Select.stat
```diff
@@ -62,7 +62,7 @@ val lastToken = owner.body.tokens.filter {
   case _: Whitespace | _: Comment => false
   case _ => true
 } // edge case, if body is empty expire on arrow.
-.lastOption.getOrElse(arrow)
+  .lastOption.getOrElse(arrow)
```
poslegm: IMHO previous behavior was prettier 🤔 
@tanishiking @olafurpg what do you think?

tanishiking: I think it's acceptable to add an indent here 👍 , and it's more consistent if scalafmt adds an indent before methods for any objects (even if it's Block).

FYI [prettier add an indent in this case](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEA3TBeTYAHSk0wA8lMBycu6gGhLIE8rqXPGSBfTAen6Y0ACwgBXADYATTAHc4mAIbTZS0gEso0hFmiYYIxQDMIkyRDlaA5pkla4AfhIA6cgAoAlCAYgIABxgNdGRQJQAncMsABQiENGQQJWwIDWkfEAAjcKUwAGs4GABlf1ybZBhw8ThfERgAW0kAdRENeDRSsDgi+LaNbDaWRLA0BN8tNDhwmGic63qlZGMlSUnfACs0cgAhHPzCoqV6uAAZByWVtZBN8iKbSTgARXEIeAvVmpBS8MnwxMylJk4JIMv5wloYE00oZkAAOAAMvjBEEmTRy-kSYLgv2wcAyAEcXvBZgEEiglGgALRQOBwHTpXzhOCEjRM2ZKeaLJDLD6+Sb1DQVKqfND3J5EvHcy6fGCAqHSGFIABMvkqSg09ig1gAwhB6gtElBoHjfOJJgAVQFknlXbDVACS2l0RTA4MCAEFtEUYCwHu9JjweEA)

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1344,8 +1344,12 @@ class Router(formatOps: FormatOps) {
       case FormatToken(_, c: T.Comment, between) =>
         Seq(Split(newlines2Modification(between), 0))
       // Commented out code should stay to the left
-      case FormatToken(c: T.Comment, _, between) if c.syntax.startsWith("//") =>
-        Seq(Split(Newline, 0))
+      case FormatToken(c: T.Comment, right, _) if c.syntax.startsWith("//") =>
```
tanishiking: Could you add some comments that explain the behavior with some code examples? So that people who read this code can easily understand the meaning.

scalafmt-tests/src/test/resources/test/ContinuationIndent.stat
```diff
@@ -37,3 +37,15 @@ class ExtendTest
     with D
     with E =>
 }
+<<< #1334 1
```
tanishiking: It might be better to move these tests to `test/resources/default/indent.stat` or something because this behavior is nothing to do with `continuationIndent`?

tanishiking: it'll be helpful if you name the test with more meaningful description (you can include #1334 to point the issue number of course :) )

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: will do

scalafmt-tests/src/test/resources/test/ContinuationIndent.stat
kitbellew: no problem. perhaps i misunderstand the meaning of continuation; i thought that since we are indenting because of the `.something` which continues from previous line, it should go in this file.

kitbellew: that was an oversight, thanks for noticing.

scalafmt-tests/src/test/resources/default/Select.stat
```diff
@@ -277,3 +277,78 @@ object a {
...
+<<< #1334 very long unbroken select chain
+class Foo {
+  val vv = v.aaaaaaaaaa.bbbbbbbbbb.cccccccccc.dddddddddd.eeeeeeeeee.ffffffffff.gggggggggg.hhhhhhhhhh.iiiiiiiiii
+}
+>>>
+class Foo {
+  val vv =
+    v.aaaaaaaaaa.bbbbbbbbbb.cccccccccc.dddddddddd.eeeeeeeeee.ffffffffff.gggggggggg.hhhhhhhhhh.iiiiiiiiii
+}
+<<< #1334 very long broken select chain
```
olafurpg: This change looks unrelated to comments, the current behavior of scalafmt is to format everything in a single line.

The current behavior of Scalafmt is to never break on select nodes `a.b` unless `b` is an "applied" node like `a.b()` or `a.b[T]` or `a.b { ... }`. This is an intentional decision for better or worse, and we're going to change it then we'll need careful performance testing to ensure the search state doesn't explode in cases that formatted fine before.

scalafmt-tests/src/test/resources/default/Select.stat
olafurpg: I like the new behavior.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1340,6 +1340,16 @@ class Router(formatOps: FormatOps) {
           Split(Space, 0, policy = singleLine),
           Split(Newline, 1).withPolicy(penalizeNewlineByNesting(cond, arrow))
         )
+      case FormatToken(_, dot @ T.Dot(), _) =>
+        // allow keeping a newline (and indent) within a select chain
+        val noNewLine = !style.activeForEdition_2019_11 ||
+          newlines == 0 || rightOwner.isNot[Term.Select]
+        val newlinePenalty = if (noNewLine) 0 else 10 + treeDepth(rightOwner)
+        Seq(
```
olafurpg: FWIW, doubling the search state here is likely to introduce new "search state exploded" errors for programs that formatted fine before.

Issue #1043, opened 2017-09-11
Feature: a one-liner version of newlines.alwaysBeforeElseAfterCurlyIf=true


kitbellew:
This template is a guideline, not a strict requirement.
- **Version**: 1.2.0
- **Integration**: CLI
- **Configuration**:
```
style = default
maxColumn = 100
danglingParentheses = true

align = none
align.tokens = []
align.openParenCallSite = false

newlines.alwaysBeforeTopLevelStatements = true
newlines.sometimesBeforeColonInMethodReturnType = false
newlines.alwaysBeforeCurlyBraceLambdaParams = true
newlines.afterImplicitKWInVerticalMultiline = true
newlines.beforeImplicitKWInVerticalMultiline = true

runner.fatalWarnings = true
runner.optimizer.forceConfigStyleOnOffset = 50
runner.optimizer.forceConfigStyleMinArgCount = 5

rewrite.rules = [ExpandImportSelectors, PreferCurlyFors]
```

## Steps

I ran scalafmt like this:

```bash
scalafmt
```

## Problem

Input is
```scala
      if ('0' <= ch && ch <= '9') { ch - '0' }
      else if ('A' <= ch && ch <= 'F') { ch - 'A' + 10 }
      else if ('a' <= ch && ch <= 'f') { ch - 'a' + 10 }
      else { -1 }
```

Scalafmt formats code like this:
```scala
      if ('0' <= ch && ch <= '9') { ch - '0' } else if ('A' <= ch && ch <= 'F') { ch - 'A' + 10 } else if ('a' <= ch && ch <= 'f') {
        ch - 'a' + 10
      } else { -1 }
```

## Expectation

I would like the formatted output to look either exactly like the input, or like this:
```scala
      if ('0' <= ch && ch <= '9') {
        ch - '0'
      } else if ('A' <= ch && ch <= 'F') {
        ch - 'A' + 10
      } else if ('a' <= ch && ch <= 'f') {
        ch - 'a' + 10
      } else {
        -1
      }
```
but definitely not the way it was chosen.

olafurpg:
Thanks for reporting! This is a duplicate of https://github.com/scalameta/scalafmt/issues/271, I agree the formatted output is not good.

Note that when given
```scala
    if ('0' <= ch && ch <= '9') {
        ch - '0'
      } else if ('A' <= ch && ch <= 'F') {
        ch - 'A' + 10
      } else if ('a' <= ch && ch <= 'f') {
[...]

PR #1560, opened 2019-11-26
Router: preserve single-line formatting for if/try

### LINKED ISSUE #271: } else
Original:

``` scala
                 @inline
                 def getOr0(): Int =
-                  if (in.hasRemaining) { bytesRead += 1; in.get() }
-                  else 0 // 0 is not a valid next byte
+                  if (in.hasRemaining) { bytesRead += 1; in.get() } else 0 // 0 is not a valid next byte
```

kitbellew:
Fixes #271 #1002 #1043.

Follow-on to #1554.

The diffs on `scala-repos` (obtained by running 2.3.0-rc2 and this build independently on the original, never before formatted base code, and then comparing): https://github.com/kitbellew/scala-repos/commit/73a76cb8dc9e97be06497ece131b0131586ffaaa?w=1

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -9,6 +9,7 @@ import org.scalafmt.util._
 
 import scala.collection.mutable
 import scala.language.implicitConversions
+import scala.meta.classifiers.Classifier
 import scala.meta.tokens.{Token, Tokens}
 import scala.meta.tokens.{Token => T}
 import scala.meta.{
@@ -210,34 +211,43 @@ class Router(formatOps: FormatOps) {
               }
             }
 
-        val skipSingleLineBlock =
-          startsLambda ||
-            newlines > 0 ||
-            leftOwner.parent.exists {
-              case _: Term.If | _: Term.Try | _: Term.TryWithHandler => true
-              case _ => false
-            }
+        def getSingleLineDecisionPre2019Nov = leftOwner.parent match {
+          case Some(_: Term.If | _: Term.Try | _: Term.TryWithHandler) => null
+          case _ => Policy.emptyPf
+        }
+        def getSingleLineDecisionFor2019Nov = {
+          type Classifiers = Seq[Classifier[Token, _]]
+          def classifiersByParent: Classifiers = leftOwner.parent match {
+            case Some(_: Term.If) => Seq(T.KwElse.classifier)
+            case Some(_: Term.Try | _: Term.TryWithHandler) =>
+              Seq(T.KwCatch.classifier, T.KwFinally.classifier)
+            case _ => Seq.empty
+          }
+          val classifiers: Classifiers = leftOwner match {
+            // for catch with case, we should go up only one level
+            case _: Term.Try if rightOwner.is[Case] =>
+              Seq(T.KwFinally.classifier)
+            case _ if !style.activeForEdition_2020_01 => Seq.empty
+            case _ => classifiersByParent
+          }
+
+          val breakSingleLineAfterClose = classifiers.nonEmpty && {
+            val afterClose = tokens(close).right
+            classifiers.exists(_(afterClose))
+          }
+          if (!breakSingleLineAfterClose) Policy.emptyPf
+          else decideNewlinesOnlyAfterClose(close)
+        }
+        def getSingleLineDecision: Policy.Pf =
+          if (newlines > 0) null
+          else if (style.activeForEdition_2019_11)
+            getSingleLineDecisionFor2019Nov
+          else
+            getSingleLineDecisionPre2019Nov
 
         // null if skipping
-        val singleLineDecision: Policy.Pf =
-          if (skipSingleLineBlock) {
-            null
-          } else if (!style.activeForEdition_2019_11) {
-            Policy.emptyPf
-          } else {
-            def isCatch = leftOwner match {
-              // for catch with case, we should go up only one level
-              case _: Term.Try if rightOwner.is[Case] => true
-              case _ => false
-            }
-
-            val afterClose = tokens(close).right
-            val breakSingleLineAfterClose =
-              if (isCatch) afterClose.is[T.KwFinally]
-              else false
-            if (!breakSingleLineAfterClose) Policy.emptyPf
-            else decideNewlinesOnlyAfterClose(close)
-          }
+        val singleLineDecision =
+          if (startsLambda) null else getSingleLineDecision
 
         val spaceMod = xmlSpace(leftOwner)
 
--- a/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
+++ b/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
@@ -385,9 +385,7 @@ object mixed {
 >>>
 {
   if (cond) foo
-  else {
-    val chars = s.toCharArray
-  }
+  else { val chars = s.toCharArray }
   val prefix =
     "sbt.paths." + projectName + "." + uncapitalize(config.name) + "."
 }
--- a/scalafmt-tests/src/test/resources/default/Advanced.stat
+++ b/scalafmt-tests/src/test/resources/default/Advanced.stat
@@ -88,9 +88,8 @@ val createAsStat = if (semantics.asInstanceOfs == Unchecked) {
 def processOptions(): Unit = {
   if (option.startsWith("relSourceMap:")) {
     val uriStr = option.stripPrefix("relSourceMap:")
-    try {
-      relSourceMap = Some(new URI(uriStr))
-    } catch {
+    try { relSourceMap = Some(new URI(uriStr)) }
+    catch {
       case e: URISyntaxException => error(s"$uriStr is not a valid URI")
     }
   } else {
--- a/scalafmt-tests/src/test/resources/default/Try.stat
+++ b/scalafmt-tests/src/test/resources/default/Try.stat
@@ -62,33 +62,26 @@ object Foo {
 }
 >>>
 object Foo {
-  try {
-    foo()
-  } finally {
-    bar()
-  }
+  try { foo() }
+  finally { bar() }
 }
 <<< try/catch/finally curly no single line blocks
 object Foo {
-  try { foo()  }
+   try{foo()}
   catch { case bar => baz }
-  finally {  qux()  }
+ finally {qux()  }
 }
 >>>
 object Foo {
-  try {
-    foo()
-  } catch { case bar => baz }
-  finally {
-    qux()
-  }
+  try { foo() }
+  catch { case bar => baz }
+  finally { qux() }
 }
 <<< try/catch with handler
 try { foo()  } catch handler
 >>>
-try {
-  foo()
-} catch handler
+try { foo() }
+catch handler
 <<< #350 assignment: short single line
 val template: String = try source.mkString finally source.close()
 >>>
--- a/scalafmt-tests/src/test/resources/unit/If.stat
+++ b/scalafmt-tests/src/test/resources/unit/If.stat
@@ -132,17 +132,31 @@ else println(1)
 if (true) return
 else println(1)
 <<< #1043
-      if ('0' <= ch && ch <= '9') { ch - '0' }
-      else if ('A' <= ch && ch <= 'F') { ch - 'A' + 10 }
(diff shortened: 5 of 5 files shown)


olafurpg:
@kitbellew I just kickstarted a v2.3.0-RC1 release with the current master. I'm eager to try out your recent contributions!

I have sent you a collaborator invitation so that you can merge PRs, restart the CI and cut new releases. We can do another release after this PR or whenever you want to use a recent change. Feel free to send me an email if you have any questions.

kitbellew:
> I have sent you a collaborator invitation so that you can merge PRs, restart the CI and cut new releases. We can do another release after this PR or whenever you want to use a recent change. Feel free to send me an email if you have any questions.

@olafurpg thank you, i am honoured.

> @kitbellew I just kickstarted a v2.3.0-RC1 release with the current master. I'm eager to try out your recent contributions!

let me know if you find anything strange.

i have a few things i am still working on, both on the published list of issues and some that have been distracting my opinionated team and introducing formatting-related comments in code reviews. i am trying to eradicate those and even get to the point where no matter what input formatting you present, the result will be the same (although, given the number of checks `newlinesBetween`, that's not likely to be).

kitbellew:
> The CI failure looks legitimate
> ...
> The tests run scalafmt twice on the input code to assert that the second format produces the same output as the first format.

I think that there are two separate rules involved that keep playing tennis with each other between different runs. Will try to fix, obviously.

About the idempotency check: do you have any idea why all of the builds succeeded except one? However, I did run it on `scala-repos` and noticed plenty of the same in one runs. I wonder if the formatter gets "tired" and does different things at different times.

I saw, in the test setup code, that there's a call to a `newTest` method which is invoked once per test, even though the formatter is invoked twice. What's the significance?

poslegm:
Did I understand correctly that idempotency violated for default behavior on master? It can lead to flaky CI after upgrade. Can we cut v2.3.0 release after fix?

kitbellew:
> Did I understand correctly that idempotency violated for default behavior on master? It can lead to flaky CI after upgrade. Can we cut v2.3.0 release after fix?

The problem is with this commit, which is still open, not with the commits which have been merged. However, @olafurpg was planning to build a release candidate, and I will try to use it on a few other repositories we have, to make sure that's still true.

poslegm:
I was confused by [this comment](https://github.com/scalameta/scalafmt/pull/1554#issuecomment-558975154)

> running this merged code against the previously checked scala-repos introduces a couple of minor changes, so there's some flipping around.

kitbellew:
> I was confused by [this comment](https://github.com/scalameta/scalafmt/pull/1554#issuecomment-558975154)
> 
> > running this merged code against the previously checked scala-repos introduces a couple of minor changes, so there's some flipping around.

of course, i understand now. that comment referred to running the #1554 build multiple times and producing exactly two lines of change, those were fixed in #1562.

poslegm:
Awesome!

kitbellew:
@olafurpg @poslegm @tanishiking the tests no longer fail, nor do the `scala-repos` (see link in PR comment above) contain any unusual stuff. PTAL.

poslegm:
Thank you for your work!

Sadly, I'm not sure how useful these changes are 😞 

For example:

Looks like `else` keyword relates to `while`. Original code is strange too, but formatting made it even more strange.

Screen Shot 2019-11-28 at 8 05 49 PM

A lot of programmers use `if` with curly braces for simpler line addition into conditions block. With origin code I can simply copy-paste `log.debug(x)` before returning `None`. With formatted one I need to first add newline after curly brace.
Screen Shot 2019-11-28 at 8 06 17 PM

Code lines is cheap and I don't want formatter to force me to save them.

It may be reasonable to keep one-liners in curly braces as is, but I can't understand profit of squashing bodies.

Also I prefer current behavior
```scala
// before
object a {
if (a) { println(x) }
else { println(x) }
}
// after
object a {
  if (a) {
    println(x)
  } else {
    println(x)
  }
}
```

because of code base unification. There is only one way to write `if-else` conditions and all code from all developers follows them.

kitbellew:
> For example:
> 
> Looks like `else` keyword relates to `while`. Original code is strange too, but formatting made it even more strange.

i have noticed this, too, will try to propose a fix.

> Screen Shot 2019-11-28 at 8 05 49 PM

> It may be reasonable to keep one-liners in curly braces as is, but I can't understand profit of squashing bodies.

this commit is not squishing bodies. as i mentioned in the description, the way i generated the diff is by running the 2.3.0-rc2 build and this one separately on *non-formatted* code. so what you are seeing is the difference between 2.3.0-rc2 *expanding* the body, and this new one *not doing it*. the key here is *preserve single-line*.

poslegm:
> so what you are seeing is the difference between 2.3.0-rc2 expanding the body, and this new one not doing it.

Oh, sorry, it makes a big difference :+1:

kitbellew:
> P.S. I really appreciate you submitting changes to the scala-repos. This greatly simplifies the review 🙏

of course, glad to do it. reviewers assume most of the responsibility for the merged code, hence the authors should simplify that task as much as possible.

kitbellew:
one question: i am actually in agreement that there should be one way to do something within a given codebase, and since there are diverging opinions, would it make sense to add a new flag? that way, either all blocks (including single-case catch statements) are expanded k&r style regardless of size, or all short blocks are inlined even if they had been expanded.

my general preference is to stop code review comments of the "add a newline here, scalafmt will format differently" kind. ideally, we would process as if there were no newlines to begin with (except for `//`), with implicit semicolons between statements, and then see what happens. perhaps we need a flag for that :)

olafurpg:
> my general preference is to stop code review comments of the "add a newline here, scalafmt will format differently" kind. ideally, we would process as if there were no newlines to begin with (except for //), with implicit semicolons between statements, and then see what happens. perhaps we need a flag for that :)

I agree this is a desirable goal (at least in theory) and this is how I originally wanted Scalafmt to work when I started out (with zero config!). It became clear to me with time that a clear majority of users wanted some manual control over line breaks in situations like here:

```scala
val x = 42
// vs
val x =
  42

case x => println(x)
// vs
case x =>
  println(x)

if (x) 42
// vs
if (x)
  42

...
```

The primary reason manual newlines are supported in these situations is to allow people to make a sequence of case/val/if statements consistent. For example, without optional manual line breaking, you could be forced to use the following inconsistent output:

```scala
// maxColumn |
val x = 42
val y =
  4222222
val z = 41

// maxColumn           |
case 21 => println(42)
case 21000 =>
  println(x)
case 31 => println(y)
...
```

Ideally, I'd love if Scalafmt did not support manual line breaks (remove all usage of `newlinesBetween`) while still preserving consistency between a sequence of statements. However, to implement that we would need a more sophisticated formatting algorithm than the one currently used by Scalafmt.

olafurpg:
I'm fine with introducing a flag that tries to reduce the usage of `newlinesBetween` as much as possible. It could be called something like `avoidManualNewlines` (open for suggestions, name up for debate).

@kitbellew I'm interested to see how far you can take it. When trying to get rid of `newlinesBetween`, I struggled the most with infix operators (`Term.ApplyInfix`). It was quite tricky to get large conditions to format nicely such as 
```scala
// maxColumn                          |
message.endsWith("bar") && (message.length < 50 || message.startsWith("foo")
```

olafurpg:
FWIW, me and my former colleague at the Scala Center managed to get quite far with a clean rewrite of Scalafmt that used "Wadler's pretty printer". See https://github.com/scalacenter/scala-syntax

I believe the "traverse AST and print syntax" is fundamentally a better approach compared to the token-based approach currrently used in Scalafmt (`Router` and `BestFirstSearch`). The tree printer approach makes it easier to ignore the original formatting of the source file. However, the challenge with tree printers is to preserve formatting of trivia like comments.

kitbellew:
> The primary reason manual newlines are supported in these situations is to allow people to make a sequence of case/val/if statements consistent. For example, without optional manual line breaking, you could be forced to use the following inconsistent output:
> 
> ```scala
> // maxColumn |
> val x = 42
> val y =
>   4222222
> val z = 41

generally, the bar for standard formatting is supporting or at least not interfering with readability. this example is quite readable to me :)

however, an `ocd` flag could be introduced, but then you'd have to be able to consider this code not a line at a time but as a block. and what constitutes a block? another area where some creative manipulation could be introduced.

also, people could be encouraged to rename variables and methods, to fit within the line length, and that would again change the formatting, defeating the purpose. oh well.

kitbellew:
> FWIW, me and my former colleague at the Scala Center managed to get quite far with a clean rewrite of Scalafmt that used "Wadler's pretty printer". See https://github.com/scalacenter/scala-syntax
> 
> I believe the "traverse AST and print syntax" is fundamentally a better approach compared to the token-based approach currrently used in Scalafmt (`Router` and `BestFirstSearch`). The tree printer approach makes it easier to ignore the original formatting of the source file. However, the challenge with tree printers is to preserve formatting of trivia like comments.

why or where did you stop? what's the plan for that alternative?

kitbellew:
finally, coming down to earth, what are your thoughts on this specific commit? i can't generate an obvious set of diffs in `scala-repos`; building on top of a 2.3.0-rc2 run (which had already expanded the one-liners) results in no change, of course, and the diffs i included in the description don't really make it clear that the left side is 2.3.0-rc2 and not the original code on which the formatting was applied for the first time.

poslegm:
I like the idea of ​​hiding these changes behind the configuration flag

kitbellew:
> I looked through the scala-repos diff and I like a lot of the changes. I don't believe we can enable this option by default because it's quite a drastic change.

@olafurpg i guess i am not explaining the source of diffs very well, as this commit *doesn't* produce much in terms of change. let me create `scala-repos` diffs for you so you can take a look.

this is what i did:
* i checked out a835854b9206aa1e5af8f939c35205462e7588f6, your *second* oldest `scala-repos` commit, where you are adding the code, but before the third commit which is labelled "Format EVERYTHING" (I don't know what you used for format, though, because the "First scalafmt run" commit is checked in later)
* i cherry-picked the commit 78fdb28ed0b8b6fd6dfac8e9d7d8ea03d57ccf57 which adds scalafmt configuration
* then i run scalafmt twice, both times on top of that unformatted code (and NOT on top of each other); once using the v2.3.0-rc2 that's in the master branch, and another time using this commit, creating two separate branches (v230 and 1560)
* then in the v230 branch i copied everything from the 1560 and created a comparison commit (via `git checkout v230; git checkout 1560 .; git commit -a`)

the differences you are seeing are the result of this:

if source was
```
if(a){b}
else{
c
}
```
then v230 formatted it as
```
if (a ) {
  b
} else {
  c
}
```
and 1560 formatted as
```
if (a) { b }
else {
  c
}
```

> I'm willing to make an exception in introducing a new option such as `avoidManualNewlines=true` (false by default) with the goal of reducing usage of manual control over the formatting layout. This goal aligns with my original aspirations for Scalafmt and I'm open to consider enabling it by default in a future edition.
> 
> What do you think?

as i said, this is commit is actually abiding by the input newline hints, but i do have some additional code which actually tries to do it, so i am all for it.

kitbellew:
I just ran the build of this commit *on top of* `scala-repos` from v230, and there were no changes.

kitbellew:
> I like the idea of ​​hiding these changes behind the configuration flag

ok, let me propose a change to add a flag here.

olafurpg:
@kitbellew does this change go against your goal of removing usage of `newlinesBetween`? Previously, users were forced into multiline blocks regardless of the original formattinig while after this PR users have the choice to keep the block on a single line 🤔

kitbellew:
> @kitbellew does this change go against your goal of removing usage of `newlinesBetween`? Previously, users were forced into multiline blocks regardless of the original formattinig while after this PR users have the choice to keep the block on a single line

I'm working on that other change, it covers many cases already but I'd like to avoid piecemeal approach and see if I can cover all. I'm also certain that we'll have a discussion about that. I'm planning to introduce 4 options, would like your opinion:

* flag not specified, produces current mixed approach of sometimes ignoring input line breaks and sometimes following them;
* option to keep line breaks as specified;
* two options to *try* and ignore line breaks, one for fewer lines and another for shorter lines; both are affected by config style, select chains bin pack and other similar options, not sure if they should supercede then)

Meanwhile, most of my team likes a more compact, vertically compressed look, and the comments in code review frequently suggest removing braces to avoid expansion. (Hence was my desire to rewrite the redundant braces logic.) So this change, illogically, will actually lead to fewer formatting-related comments. 😀

kitbellew:
@olafurpg rebased and merged, after realizing you had approved it earlier; thank you.

PR #1561, opened 2019-11-26
Setup GitHub Actions to generate native images

olafurpg:
This PR adds a new CI job to build a GraalVM native image for Linux and macOS. The `scalafmt` binary can be downloaded from the GitHub Actions UI under the "Artifacts" tab

Screenshot 2019-11-27 at 10 57 45


olafurpg:
For releases, I'm planning to start with manually uploading the zip files. I suspect we don't do stable release too often so a manual approach should be fine for now.

scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -89,9 +89,21 @@ object Cli {
...
       case None => ScalafmtCoreRunner
-      case Some(v) if v == Versions.version =>
-        ScalafmtCoreRunner
-      case _ => ScalafmtDynamicRunner
+      case Some(v) =>
+        if (v == Versions.version) ScalafmtCoreRunner
+        else {
+          val isNativeImage =
+            "true" == System.getProperty("scalafmt.native-image", "false")
+          if (isNativeImage) {
```
poslegm: I like it 👍

olafurpg: This doesn't seem to work yet, I'll try to fix it before merging.

olafurpg: OK, should be fixed now!
Screenshot 2019-11-27 at 14 51 02

PR #1562, opened 2019-11-27
Router: fix split rule after catch in try

kitbellew:
Instead of maintaining exceptions to an otherwise mandated line break,
let's specify explicitly when this could happen. That is, rather than
break always except when followed by a parenthesis, we should break
only if followed by the finally clause.

Follow-on to #1554.


kitbellew:
@olafurpg @poslegm @tanishiking PTAL, a minor bugfix.

PR #1563, opened 2019-11-27
RedundantBraces: lambda braces to single-arg apply, remove arou…

### LINKED ISSUE #1147: New RedundantBraces is too aggressive

### LINKED ISSUE #1237: [Request] Setting to rewrite block-argument-inside-parens to block-argument
This is a feature request:

I'd like to be able to rewrite code such that

```scala
foo.bar(param => { body })
```
Is rewritten to
```scala
foo.bar { param => body }
```

With obvious support for multiline expressions, etc. This comes up a lot with `map` and `flatMap`

kitbellew:
Fixes #1147
Fixes #1237

### DIFF
--- a/scalafmt-tests/src/test/resources/rewrite/RedundantBraces-ParenLambdas.stat
+++ b/scalafmt-tests/src/test/resources/rewrite/RedundantBraces-ParenLambdas.stat
@@ -1,0 +1,113 @@
+rewrite.rules = [RedundantBraces]
+rewrite.redundantBraces.methodBodies = true
+rewrite.redundantBraces.maxLines = 1
+<<< #1237 1 ctor is untouched
+object a {
+  val a = new b(c => { d })
+}
+>>>
+object a {
+  val a = new b(c => { d })
+}
+<<< #1237 2 apply single-stat with comments
+object a {
+  val a = b(c => {
+  d // c1
+  } // c2
+  )
+}
+>>>
+object a {
+  val a = b(c => {
+    d // c1
+  } // c2
+  )
+}
+<<< #1237 3 apply multiple single-stat
+object a {
+  val a = b(c => { d }, e => { f })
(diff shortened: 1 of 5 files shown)


scalafmt-tests/src/test/resources/rewrite/RedundantBraces-ParenLambdas.stat
```diff
@@ -0,0 +1,109 @@
+rewrite.rules = [RedundantBraces]
+rewrite.redundantBraces.methodBodies = true
+rewrite.redundantBraces.maxLines = 1
+<<< #1237 1 ctor is untouched
+object a {
+  val a = new b(c => { d })
+}
+>>>
+object a {
+  val a = new b(c => { d })
```
olafurpg: 👍

PR #1564, opened 2019-11-27
Performance improvements

olafurpg:
Incorporates changes from #1386 and makes them compile on 2.11/2.12/2.13. The performance improvements are small but noticeable!

```
[info] Benchmark                          Mode  Cnt     Score     Error  Units
[info] Micro.ExtraLarge.scalafmtAfter     avgt   10   826.260 ± 139.865  ms/op
[info] Micro.ExtraLarge.scalafmtBefore    avgt   10   960.471 ± 157.597  ms/op
```

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -90,7 +90,12 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
     }
     def addOptional(tree: Tree): Unit =
       tree.tokens.headOption.foreach(x => optional += hash(x))
-    def iter(tree: Tree): Unit = {
+
+    import scala.collection.mutable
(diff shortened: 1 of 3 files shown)


olafurpg:
The commit changing `Input.VirtualFile` to `Input.File` (https://github.com/scalameta/scalafmt/commit/ddb5bd831fff1afc1458f9160f934be148a48eec) was not included since it should no longer be necessary with the next release of Scalameta (see https://github.com/scalameta/scalameta/pull/1914)

olafurpg:
Thank you @jrudolph for this contribution!

olafurpg:
Re-ran the benchmark with more iterations and the changes are less significant but still noticeable
```
# after
[info] Micro.ExtraLarge.scalafmt  avgt   30  831.742 ± 49.979  ms/op
# before
[info] Micro.ExtraLarge.scalafmt  avgt   30  901.238 ± 10.708  ms/op
```

jrudolph:
Cool, almost forgot about that :) Thanks for bringing it over the finish line.

PR #1566, opened 2019-11-27
Formatter upgraded to 2.3.0-RC1

poslegm:
I like this changes

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version=2.3.0-RC1
+version=2.3.0-RC2
 project.git = true
 project.excludeFilters = [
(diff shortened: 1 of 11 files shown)


.scalafmt.conf
```diff
@@ -1,4 +1,4 @@
-version=2.0.0-RC5
+version=2.3.0-RC1
```
olafurpg: The native image binaries had a wrong version number that got fixed in https://github.com/scalameta/scalafmt/commit/e35c8273dcc34bd340a51dc45b350985155a75a6

The RC2 release has completed
```suggestion
version=2.3.0-RC2
```

PR #1567, opened 2019-11-28
Router: break between curly and catch, like else

kitbellew:
Use the setting for 'else' after curly 'if' to determine whether to add
newlines before 'catch' and 'finally'.

Those who prefer
```
  if (a) {
    b
  }
  else c
```
over
```
  if (a) {
    b
  } else c
```
likely also prefer
```
  try {
    a
  }
  finally b
```
over
```
  try {
    a
  } finally b
```


kitbellew:
@olafurpg @poslegm @tanishiking PTAL. i didn't generate `scala-repos`, as this behaviour is under an optional flag (and there are unit tests), but let me know if you'd prefer it.

PR #1586, opened 2019-12-04
adding possibility to specify custom repositories for coursier (#1521)

slivkamiro:
I'm kind of lost in tests. I would need help with writing a test for ScalafmtDynamicDownloader, if necessary (I haven't found any).
I'm also struggling with sbt-scalafmt part. I can't find a way how to map resolvers configuration into list of string urls. Any ideas?

### DIFF
--- a/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamicDownloader.scala
+++ b/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamicDownloader.scala
@@ -5,7 +5,6 @@ import java.net.URL
 
 import coursierapi._
 import org.scalafmt.dynamic.ScalafmtDynamicDownloader._
-import org.scalafmt.dynamic.ScalafmtVersion
 import org.scalafmt.dynamic.ScalafmtVersion.InvalidVersionException
 
 import scala.collection.JavaConverters._
@@ -14,6 +13,7 @@ import scala.util.Try
 
 class ScalafmtDynamicDownloader(
     downloadProgressWriter: OutputStreamWriter,
+    customRepositories: List[Repository],
     ttl: Option[Duration] = None
 ) {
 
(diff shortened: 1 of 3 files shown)


scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamicDownloader.scala
```diff
@@ -91,7 +91,7 @@ class ScalafmtDynamicDownloader(
         "https://oss.sonatype.org/content/repositories/snapshots"
       ),
       MavenRepository.of("https://oss.sonatype.org/content/repositories/public")
-    )
+    ) ++ customRepositories
```
tomasherman: i think custom repos should either override the defaults, or have higher priority - one usecase of this (ours) is that oss repositories are disabled and must go through our own artifactory and it causes timeouts when trying to download stuff from OSS ... therefore it makes sense to try custom repositories first and fallback on defaults

slivkamiro: timeouts are good point, will move it in front then

sideeffffect:
@slivkamiro thank you so much for this!
I know this is just `scalafmt` proper, but are you perhaps considering implementing support for this in `sbt-scalafmt` plugin? It would be great, if the `-Dsbt.repository.config=/.../repositories -Dsbt.override.build.repos=true` overrides were respected.

scalafmt-interfaces/src/main/java/org/scalafmt/interfaces/Scalafmt.java
```diff
@@ -74,6 +73,11 @@
      */
     Scalafmt withReporter(ScalafmtReporter reporter);
 
+    /**
+     * Use this repositories to resolve dependencies.
+     */
+    Scalafmt withRepositories(String ... repositories);
```
olafurpg: Let's make it clear that these must be Maven repositories
```suggestion
    Scalafmt withMavenRepositories(String ... repositories);
```

slivkamiro: no problem, I will amend the commit if you don't mind

slivkamiro:
> @slivkamiro thank you so much for this!
> I know this is just `scalafmt` proper, but are you perhaps considering implementing support for this in `sbt-scalafmt` plugin? It would be great, if the `-Dsbt.repository.config=/.../repositories -Dsbt.override.build.repos=true` overrides were respected.

Yes, see https://github.com/scalameta/sbt-scalafmt/pull/73

PR #1588, opened 2019-12-05
RedundantBraces: fix how we find {} in func body

### LINKED ISSUE #1583: [Bug] scalafmt 2.3.0-RC2 throws error when trying to format my code
- **Integration**: IntelliJ
- **Configuration**:
```
version = 2.3.0-RC2
docstrings = ScalaDoc
maxColumn = 100

rewrite.rules = [
  AvoidInfix        
  RedundantBraces   
  RedundantParens   
  SortImports 
  PreferCurlyFors 
]

rewrite.redundantBraces.maxLines = 1
trailingCommas = preserve
```

## Steps

Given code like this:
```scala
package fmt

import java.time.LocalDateTime

object Test {

    val profiles = Seq(
      Profile("TanVn", Some(LocalDateTime.now())),
      Profile("TanVn2",Some(LocalDateTime.now().minusDays(10)))
    )
    val patitions = profiles.partition( p => p.createdAt match {
      case Some(value) => value.getYear < 2019
      case None =>false
    })
  }

}

case class Profile(
    name: String,
    createdAt: Option[LocalDateTime]
)


```

```bash
scalafmt
```

## Problem

```scala
error: { expected but case found
```

kitbellew:
Fixes #1583.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -104,11 +104,15 @@ case object RedundantBraces extends Rewrite {
         if f.tokens.last.is[Token.RightBrace] && getTermLineSpan(f) > 0 =>
       val rbrace = f.tokens.last
       val lbrace = ctx.matchingParens(TokenOps.hash(rbrace))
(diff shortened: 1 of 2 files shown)


kitbellew:
@olafurpg @poslegm @tanishiking bug fix.

kitbellew:
@olafurpg tag pushed!

Issue #1593, opened 2019-12-07
Excess new line in the chained package clauses


iRevive:
- **Version**: "org.scalameta" % "sbt-scalafmt" % "2.2.1"
- **Integration**: IntelliJ
- **Configuration**:
```
version = 2.3.1
newlines.alwaysBeforeTopLevelStatements = true
```

## Steps

Given code like this:
```scala
package a.b.c
package d

import e.f
package object g {}
```

When I run scalafmt like this:

```bash
sbt scalafmt
```

## Problem

Scalafmt formats code like this:
```scala
package a.b.c

package d

import e.f
package object g {}
```

## Expectation

I would like the formatted output to look like this:
```scala
package a.b.c
package d

import e.f
package object g {}
```

## Note

Everything works fine until the parent package is not multilevel.

PR #1594, opened 2019-12-07
Do not add a new line in multilevel chained package clause

iRevive:
This PR fixes https://github.com/scalameta/scalafmt/issues/1593

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -259,24 +259,35 @@ class FormatWriter(formatOps: FormatOps) {
     }
     initStyle.newlines.alwaysBeforeTopLevelStatements && {
       val formatToken = toks(i).formatToken
+
       def checkPackage: Option[Boolean] =
         if (!initStyle.activeForEdition_2019_11) None
         else
           Some(owners(formatToken.left))
             .collect { case term: Term.Name => term.parent }
             .flatten
-            .collect { case pkg: Pkg => pkg.stats.headOption }
+            .collect {
+              // package a
+              case pkg: Pkg =>
+                pkg.stats.headOption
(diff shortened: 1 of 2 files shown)


scalafmt-tests/src/test/resources/newlines/alwaysBeforeTopLevelStatements.source
```diff
@@ -36,3 +36,16 @@ package b {
...
+package f
+import g.h.i.j
+package k {}
+>>>
+package a.b.c
+package d.e
+package f
+
+import g.h.i.j
+package k {}
```
kitbellew: should there be an empty line here as well? perhaps it was this way before.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -259,24 +259,35 @@ class FormatWriter(formatOps: FormatOps) {
...
             .collect { case term: Term.Name => term.parent }
             .flatten
-            .collect { case pkg: Pkg => pkg.stats.headOption }
+            .collect {
+              // package a
+              case pkg: Pkg =>
+                pkg.stats.headOption
+
+              // package a.b.c
+              case select: Term.Select =>
```
kitbellew: would this pattern match `package a.b.c {` as well? there might an `Apply` there somewhere.

kitbellew:
@iRevive thank you for the fix, and for adding the test.

@olafurpg the commit is good to go.

scalafmt-tests/src/test/resources/newlines/alwaysBeforeTopLevelStatements.source
iRevive: Currently, scalafmt does not change formatting between import and class and keeps an empty line (if it present) as is.

poslegm:
Thank you!

PR #1600, opened 2019-12-14
Router: dangle only if breaking in multi-line lambda

### LINKED ISSUE #1592: Version 2.3.1 does not converge.
- **Version**: 2.3.1
- **Integration**: IntelliJ
- **Configuration**:
```
version = "2.3.1"
maxColumn = 120 
trailingCommas = always
  until
  to
  by
  eq
  ne
  "should.*"
  "contain.*"
  "must.*"
  in
  ignore
  be
  taggedAs
  thrownBy
  synchronized
  have
  when
  size
  only
  noneOf
  oneElementOf
  noElementsOf
  atLeastOneElementOf
  atMostOneElementOf
  allElementsOf
  inOrderElementsOf
  theSameElementsAs
  // these are new
  throws
  returns
  satisfy
]
```

## Steps

Given code like this:
```scala
object Foo {
  def xxxxxxxxxxxxxxxxxxx(struct1: Xxxxxxxxxxx, struct2: Xxxxxxxxxxx): Seq[Xxxxxxxxxxx] = {
    (1, 2) match {
      case (1, 2) =>
        argMap1.keys.toSet.union(argMap2.keys.toSet).toSeq.flatMap { key =>
          if (argOp1 == argOp2) {
            xxxxxxxxxxxxxxxxxxx(argMap1(key), argMap2(key))
              .map(d => d.copy(path = Seq(op, Yyyyyyyyyyyy(key)) ++ d.path))
          }
        }
    }
  }
}
```

Sorry for the big snippet, but I couldn't seem to repro if I minimized further. Something very finicky is going on.

## Problem

Scalafmt (from the command line) will keep swapping the formatting back and forth between the above and
```scala
object Foo {
  def xxxxxxxxxxxxxxxxxxx(struct1: Xxxxxxxxxxx, struct2: Xxxxxxxxxxx): Seq[Xxxxxxxxxxx] = {
    (1, 2) match {
      case (1, 2) =>
        argMap1.keys.toSet.union(argMap2.keys.toSet).toSeq.flatMap { key =>
          if (argOp1 == argOp2) {
            xxxxxxxxxxxxxxxxxxx(argMap1(key), argMap2(key)).map(d => d.copy(path = Seq(op, Yyyyyyyyyyyy(key)) ++ d.path),
            )
          }
        }
    }
  }
}
```

## Workaround

Of course you can force some newlines to prevent the problem, but we have scalafmtCheckAll on in CI, and so this is not acceptable because CI will never pass.

kitbellew:
Fixes #1592

NB: There are *no diffs* on `scala-repos` for this PR.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Policy.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Policy.scala
@@ -17,19 +17,19 @@ case class Policy(
     isSingleLine: Boolean = false
 )(implicit val line: sourcecode.Line) {
 
-  def merge(
-      other: Policy.Pf,
-      newExpire: Int
-  ): Policy =
-    Policy(f.orElse(other), newExpire)
+  def orElse(other: Policy.Pf, minExpire: Int = 0): Policy =
+    if (other == Policy.emptyPf) this
+    else
+      copy(
+        f = if (f == Policy.emptyPf) other else f.orElse(other),
+        expire = math.max(minExpire, expire)
+      )
 
-  def merge(other: Policy, newExpire: Int): Policy =
-    merge(other.f, newExpire)
+  def orElse(other: Policy): Policy =
+    orElse(other.f, other.expire)
 
-  def merge(other: Policy): Policy = {
-    val newExpire = expire.max(other.expire)
-    merge(other, newExpire)
-  }
+  def orElse(other: Option[Policy]): Policy =
+    other.fold(this)(orElse)
 
   def andThen(other: Policy): Policy = {
     if (this.f == Policy.emptyPf) other
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Policy.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Policy.scala
@@ -11,14 +11,14 @@ import org.scalafmt.Error.NoopDefaultPolicyApplied
   * @param expire The latest token offset.
   */
 case class Policy(
-    f: PartialFunction[Decision, Decision],
+    f: Policy.Pf,
     expire: Int,
     noDequeue: Boolean = false,
     isSingleLine: Boolean = false
 )(implicit val line: sourcecode.Line) {
 
   def merge(
-      other: PartialFunction[Decision, Decision],
+      other: Policy.Pf,
       newExpire: Int
   ): Policy =
     Policy(f.orElse(other), newExpire)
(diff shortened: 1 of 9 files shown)


kitbellew:

poslegm:
I don't understand the reason to break line which fits the `maxColumn` condition 😕. Can you please explain motivation?

kitbellew:
> Thank you for tracking down this issue! That was more work than I expected 😅

The logic is quite simple: either the lambda fully fits on one line (penalty 0), or we induce a break after the `=>` (also penalty 0). This occasionally leads to cases when scalafmt finds a way to keep the lambda on one line, by breaking earlier. I could put a higher penalty on keeping lambda on one line, but that seems illogical.

> 
> Looking at the scala-repos diff, I'm not sure I like the formatting changes
> 
> Screenshot 2019-12-16 at 11 41 09

Length is 84, even without the indent. Infix breaks are disabled, so breaking on `(` is the only place. I don't know, however, why `_ =>` has a break before it.

> Screenshot 2019-12-16 at 11 41 02

While I personally prefer this formatting, it also seemed consistent with the existing code which seems to penalize newlines deeper into an apply chain.

> Screenshot 2019-12-16 at 11 40 50

Line is too long.

> Screenshot 2019-12-16 at 11 40 32
> 
> Screenshot 2019-12-16 at 11 40 29
> 
> Could you motivate why it's desirable to break on the dot `.` instead of inside the lambda?

Let's see.
* having `a.b(c => d)` format on one line is desirable, hence the `NoSplit` with `SingleLineBlock`
* having a break after `c =>` was the whole original point of the exercise, hence `NoSplit` with a break after `=>`
* if `c` is too long and we'll exceed `maxColumn` on `=>`, we need the `Newline` split as well but its penalty has to be higher than the one before it; it already tries to penalize all newlines inside  so it could potentially format the lambda on one line after breaking; (it uses penalty 10; I tried with 1 and 100, and 10 leads to fewer lines of `scala-repos` diffs as one possible metric to optimize).

In order to accomplish the option of breaking followed by a single-line as a better option to not breaking with a single line, we'd need another split. In the past, you weren't too keen on that.

kitbellew:
> I don't understand the reason to break line which fits the `maxColumn` condition 😕. Can you please explain motivation?

Миша, can you point me to an example where a break is introduced when the line fits? So far, all examples were of lines being longer than 80.

poslegm:
Sorry, I just didn't realized that maxColumn has default value 80

kitbellew:
@olafurpg @tanishiking @poslegm reworked the commits (several purely cosmetic changes, with the last one actually implementing the bugfix). There are no changes in `scala-repos` with this solution.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Policy.scala
```diff
@@ -59,11 +59,14 @@ case class Policy(
 }
 
 object Policy {
-  val IdentityPolicy: PartialFunction[Decision, Decision] = {
+
+  type Pf = PartialFunction[Decision, Decision]
```
poslegm: 👍

PR #1604, opened 2019-12-15
FormatOps: consider comment for RHS optimal token

kitbellew:
We are also moving the optimal break point to before a close parenthesis
if a comment follows it and doesn't fit.

Since we don't break just before a comment, it occasionally led to very
long lines. While this change can't guarantee the lines won't be longer
than maxColumn either, they will be shorter because we will induce a
break before the close parenthesis:

instead of
```
  a(b) // comment
```
we'll format as
```
  a(
    b
  ) // comment
```
An unexpectedly related (through failing idempotency checks) but somewhat different problem is with breaking on assign in apply; it's follow-on change to the feature introduced 2.3, and modifies
```
  a(b = c) // comment
```
to (if the original form exceeds `maxColumn`)
```
  a(b =
    c
  ) // comment
```
or (if the previous form exceeds `maxColumn`)
```
  a(
    b = c
  ) // comment
```

`scala-repos` diffs:
* try break before a comment in apply: https://github.com/kitbellew/scala-repos/commit/83c792bf41b187013d552d8ef4b817d5950409a3?w=1
* reduce break penalty for assign: https://github.com/kitbellew/scala-repos/commit/f38073e70fdbd1be2f98074e20762585e8b356d8?w=1
* break on assign in apply if dangle: https://github.com/kitbellew/scala-repos/commit/ba5d78b262419e9c8cc8a03b724461496efa15e6?w=1
* newline split for assign in apply: https://github.com/kitbellew/scala-repos/commit/d576f9ffaea4ced6f3ab40c9ac1bfa4dadc5348c?w=1

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -776,7 +776,7 @@ class Router(formatOps: FormatOps) {
 
         val keepConfigStyleSplit =
           style.optIn.configStyleArguments && newlines != 0
-        val splitForAssignToken =
+        val splitsForAssign =
           if (defnSite || isBracket || keepConfigStyleSplit) None
           else
             getAssignAtSingleArgCallSite(leftOwner).map { assign =>
@@ -784,9 +784,15 @@ class Router(formatOps: FormatOps) {
                 case b: Term.Block => b.tokens.head
                 case _ => assign.tokens.find(_.is[T.Equals]).get
               }
-              Split(NoSplit, 1 + nestedPenalty + lhsPenalty)
-                .withPolicy(newlinePolicy.andThen(SingleLineBlock(assignToken)))
-                .withOptimalToken(assignToken)
+              val noSplitCost = 1 + nestedPenalty + lhsPenalty
+              Seq(
+                Split(NoSplit, noSplitCost)
+                  .withOptimalToken(assignToken)
+                  .withPolicy(
+                    newlinePolicy
+                      .andThen(SingleLineBlock(assignToken))
+                  )
+              )
             }
 
         val noSplitPolicy =
@@ -802,7 +808,7 @@ class Router(formatOps: FormatOps) {
             newlineModification,
             (1 + nestedPenalty + lhsPenalty) * bracketMultiplier,
             policy = newlinePolicy.andThen(singleLine(4)),
-            ignoreIf = args.length > 1 || isTuple || splitForAssignToken.isDefined
+            ignoreIf = args.length > 1 || isTuple || splitsForAssign.isDefined
           ).withOptimalToken(expirationToken)
             .withIndent(indent, close, Right),
           Split(
@@ -821,7 +827,7 @@ class Router(formatOps: FormatOps) {
             ignoreIf = singleArgument || isTuple
           ).withOptimalToken(expirationToken)
             .withIndent(indent, close, Right)
-        ) ++ splitForAssignToken.toSeq
+        ) ++ splitsForAssign.getOrElse(Seq.empty)
 
       // Closing def site ): ReturnType
       case FormatToken(left, T.Colon(), _)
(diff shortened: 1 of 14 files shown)


kitbellew:

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -709,26 +705,31 @@ class Router(formatOps: FormatOps) {
...
 
-        val charactersInside = (close.start - open.end) - 2
-
         val defnSite = isDefnSite(leftOwner)
         val expirationToken: Token =
           if (defnSite && !isBracket) defnSiteLastToken(open, leftOwner)
           else rhsOptimalToken(leftTok2tok(close))
+        val newlineExpirationTokenOption =
+          Some(expirationToken)
+            .filterNot(_.is[T.Comment] && style.activeForEdition_2019_11)
```
kitbellew: I had an edition guard here, but perhaps it makes sense to move it into the `rhsOptimalToken` method, plus I had forgotten that we released already and the edition might need to be incremented.

N.B. I know there was a discussion earlier about the difference between editions and releases but I am not yet too clear on that... if a release forces one to change the edition, they seem to be just two different ways of accomplishing the same.

kitbellew:
@olafurpg @tanishiking PTAL.

not sure what the code review protocol is, I assumed there are no self-approvals, even with an incremental fix as easy as this.

kitbellew:
@tanishiking thank you for approving. i would like to hold off merging for now, i have been building a few other changes on top and would like to split a few commits for clarity. will upload them shortly.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -764,8 +764,28 @@ class Router(formatOps: FormatOps) {
                 case b: Term.Block => b.tokens.head
                 case _ => assign.tokens.find(_.is[T.Equals]).get
               }
+              val assignIndent =
```
kitbellew: this block is useful but unrelated to RHS optimal token, i am splitting off. it actually turned out to be a can of worms, somewhat similar to #1600 in size.

kitbellew:
@tanishiking @poslegm @olafurpg PTAL.

Compared to the  previous version approved by @tanishiking, some refactoring code was split into separate commits, and the logic including comments in optimal token was split from the changes affecting the assign handling. As usual, all tests are added in the beginning and then progressively explain the change.

Also, `scala-repos` diffs are available in the PR description for each commit.

poslegm:
I was confused by [this](https://github.com/kitbellew/scala-repos/commit/83c792bf41b187013d552d8ef4b817d5950409a3?w=1#diff-187d6e2cd7648489ef25af310462e06dR589)
Screen Shot 2019-12-24 at 5 28 55 PM

I expected something like

```scala
closeServerSideAndWaitForClientReadable(
  fullClose = false
) // send EOF (fin) from the server side
```

IMHO named argument formatting should not be equal to lambda parameters formatting:
```scala
copy(timeout =
  SomeMaxFiniteDuration
) // we map the Infinite duration to a special marker,

map(timeout =>
  timeout + 1
)
``` 

But it's an open question.

kitbellew:
> I was confused by [this](https://github.com/kitbellew/scala-repos/commit/83c792bf41b187013d552d8ef4b817d5950409a3?w=1#diff-187d6e2cd7648489ef25af310462e06dR589)
> Screen Shot 2019-12-24 at 5 28 55 PM
> 
> I expected something like
> 
> ```scala
> closeServerSideAndWaitForClientReadable(
>   fullClose = false
> ) // send EOF (fin) from the server side
> ```

Of course, that is also part of this PR and fixed in the very next commit, here's the relevant diff: https://github.com/kitbellew/scala-repos/commit/ba5d78b262419e9c8cc8a03b724461496efa15e6#diff-187d6e2cd7648489ef25af310462e06d

kitbellew:
No code change, added a few more tests, one of which was initially failing with "bug in scalafmt".

kitbellew:
@poslegm are there any other outstanding issues? is it ready to merge?

@olafurpg @tanishiking any additional comments? I added the edition and the explanation (as @olafurpg requested), and @tanishiking approved prior to my commit-splitting.

kitbellew:
> Everything is ok 👍

@poslegm thank you, still working on a rebase, will update later.

kitbellew:
@olafurpg @poslegm thank you, successfully rebased and merged.

Issue #1607, opened 2019-12-24
Fix issues missed or introduced in 2.3


kitbellew:
* incomplete handling of assign in apply (see https://github.com/scalameta/scalafmt/pull/1551/commits/ab21ffa6b56dd5982a520760eff6a6cf5e3262cf)
  * split after assign in apply, split before comment #1604 (merged)
* incomplete handling of lambda in apply (see https://github.com/scalameta/scalafmt/pull/1551/commits/8abca92c8b032ca72ad200fda74cc496478c841c)
  * dangle only after break #1600 (merged)
  * various bugfixes #1612 (merged)
* various changes to avoid idempotency issues
  * infix after assign #1608 (merged)
* cosmetic changes

PR #1608, opened 2019-12-24
FormatOps: allow break after assign in infixSplit

kitbellew:
This fix improves some formatting and also removes some idempotency issues while fixing the larger assign break.

`scala-repos` diffs:
* break after assign in infixSplit: https://github.com/kitbellew/scala-repos/commit/a318836fe35b44f62892d23c3aff6ba9bb0057a2?w=1
* handle split after "while ()": https://github.com/kitbellew/scala-repos/commit/bed9ada263f79577063aa60880579ae81f1c6ca3?w=1


kitbellew:
@tanishiking @olafurpg @poslegm PTAL.

#1604 was becoming too large, so I created a tracking issue #1607 and started with this change first.

Issue #431, opened 2016-09-10
Really long list, caveats inside


RatanRSur:
I understand that for many situations, since arguments are usually not the same length (not just a single char as in this case) a newline for every argument would definitely be the prettiest solution. I just wanted to report this because it's a situation where the default behavior seems a bit pathological and bin packing looks a lot better.

Original:

``` scala
  val secret: List[Bit] = List(0,0,1,1,1,0,1,0,1,1,1,0,0,1,1,0,1,0,0,1,1,0,1,0,1,1,0,0,1,1,1,1,1,0,1,0,1,1,0,0,0,0,1,0,1,1,1,0,0,1,0,0,1,0,0,0,1,0,0,0,1,0,1)
```

Scalafmt:

``` scala
  val secret: List[Bit] = List(0,
                               0,
                               1,
                               1,
                               1,
                               0,
                               1,
                               0,
                               1,
                               1,
                               1,
                               0,
                               0,
                               1,
                               1,
                               0,
                               1,
                               0,
                               0,
                               1,
                               1,
                               0,
                               1,
                               0,
                               1,
                               1,
                               0,
                               0,
                               1,
                               1,
                               1,
                               1,
                               1,
                               0,
                               1,
                               0,
                               1,
                               1,
                               0,
                               0,
                               0,
                               0,
                               1,
                               0,
                               1,
                               1,
                               1,
                               0,
                               0,
                               1,
                               0,
                               0,
                               1,
                               0,
                               0,
                               0,
                               1,
                               0,
                               0,
                               0,
                               1,
                               0,
                               1)
```

I would like the input to look like this:

``` scala
  val secret: List[Bit] = List(0, 0, 1, 1, 1, 0, 1, 0, 1, 1, 1, 0, 0, 1, 1, 0,
                               1, 0, 0, 1, 1, 0, 1, 0, 1, 1, 0, 0, 1, 1, 1, 1,
                               1, 0, 1, 0, 1, 1, 0, 0, 0, 0, 1, 0, 1, 1, 1, 0,
                               0, 1, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 1) 
```

olafurpg:
I agree this is unfortunate output. See #347 for related discussion. The scala.js style would format your example like this

```scala
object foo {
  def main(args: Array[String]) {
[...]

PR #1614, opened 2020-01-01
Router bugfix: keep single-line comment split

kitbellew:
Even if we are bin-packing, we shouldn't fold single-line comments which
were previously detached. First, the general rule is breaks for comments
are normally maintained; second, even for bin-packing, comments would
frequently be used to separate groups within literal lists, and removing
the break is undesirable. This was also not requested in #431, the issue
which was fixed with the introduction of this behaviour.


kitbellew:

For some context: I am working on simplifying the logic of handling single-line comments (something @olafurpg commented on here: https://github.com/scalameta/scalafmt/blob/b779741ed30af56ec9aba3b33016f222de995abd/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala#L1569), but this specific bin-pack behaviour is inconsistent with how single-line comments are generally handled, and I can't design a general approach that would work here as well.

kitbellew:
@olafurpg would be good to hear your opinion. code and example is simple, so the question is only whether we would like to do it, and as a bugfix, not via an edition. (this change doesn't modify `scala-repos`, obviously.)

kitbellew:
@tanishiking thank you. also, didn't realize opening an issue was desirable.

PR #1615, opened 2020-01-02
Various refactoring and simplification commits

kitbellew:
No behaviour is modified, formatting is as before.


kitbellew:
@olafurpg @tanishiking @poslegm PTAL, let me know whether this is called for.

kitbellew:
@olafurpg thanks, merging

PR #1616, opened 2020-01-03
Some performance optimizations for FormatToken

kitbellew:
They seem to bring about 5-10% speedup while running on `scala-repos`.

P.S. `bin/run_benchmarks.sh` doesn't seem to work, because `sbt test` keeps failing.


kitbellew:

PR #1617, opened 2020-01-03
add in a bit of documentation about Metals

### LINKED ISSUE #1502: Write a documentation about how to use scalafmt from metals
[metals](https://scalameta.org/metals/) is a Scala language server based on scalameta, and we can format code with Scalafmt from metals pretty easily http://scalameta.org/metals/blog/2019/01/24/tin.html#formatting-with-scalafmt

It would be nice if we have documentation about metals at an [installation section](https://scalameta.org/scalafmt/docs/installation.html)

ckipp01:
This will close #1502 

Not a ton to add, but I think it's great to include a little note about the automatic support that Metals has for scalafmt. Let me know if you'd like me to add anything else!

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -60,6 +60,17 @@ Scalafmt formatter:
 
 It is not possible to reset this setting for all existing projects.
 
+## Metals
+
+[Metals](https://scalameta.org/metals/) automatically uses Scalafmt to respond to formatting
+requests from the editor, according to the configuration defined in `.scalafmt.conf`.
+
+In most editors, if you there is no `.scalafmt.conf`, upon receiving the first format request Metals
+will create the `.scalafmt.conf` file for you.
+
+You can find more information on triggering a format request for individual editors in their
+respective parts of the [Metals site](https://scalameta.org/metals/docs/editors/overview.html).
+
 ## sbt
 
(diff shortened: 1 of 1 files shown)


ckipp01:
> This is a great idea! What do you think about also adding custom headers for VS Code, Vim, Emacs, Sublime Text and Eclipse linking to the new Metals header?

Are you referring to the installation section? Like including each editor and linking them to the new Metals header?

olafurpg:
Yes

ckipp01:
> Yes

Sounds good. I'll add these in.

ckipp01:
So I was just going to work on this but realized that some of the info is pretty repetitive to that which is already in the Metals documentation. At least for example in the Vim section, I already have this logged in the readme and the website. I want to avoid just duplicating it all over the place. Do you think it's ok to merge in as is just with a note about Metals and a link back to the Metals site? If not, feel free to close this as I probably won't get to going through all the editors and documenting the specific stuff on formatting for a while.

olafurpg:
I'm sorry, I didn't mean to write full installation instructions for each editor. I'll follow up with a PR adding a small header for each editor linking directly to the Metals documentation without any more details :)

PR #1626, opened 2020-01-09
FormatOps: check "new anonymous" in infix indent

### LINKED ISSUE #1547: Increased indentation after migration to 2.+
- **Version**: 2.2.2
- **Integration**: IntelliJ
- **Sbt-plugin**: 2.2.1
- **Sbt**: 1.3.3
- **Configuration**:
```
version=2.2.2
encoding = "UTF-8"
maxColumn = 120
lineEndings = unix
align = most
align.tokens = [
  {code = "<-"},
  {code = "=>", owner = "Case"},
  {code = "%", owner = "Term.ApplyInfix"},
  {code = "%%", owner = "Term.ApplyInfix"}
]
rewrite.rules = [RedundantBraces, RedundantParens, SortImports, SortModifiers, PreferCurlyFors]
rewrite.sortModifiers.order = ["override", "private", "protected", "implicit", "final", "sealed", "abstract", "lazy"]
newlines.sometimesBeforeColonInMethodReturnType = false
unindentTopLevelOperators = true
danglingParentheses = true
```

## Steps

Given code like this:
```scala
private trait A {
  val test = "test "
}

behavior of "test case"
it should "test something" in new A {
  println(test)
}
```

```bash
sbt test:scalafmt
```

## Problem

```scala
private trait A {
  val test = "test "
}

behavior of "test case"
it should "test something" in new A {
      println(test)
    }
```

## Expectation

```scala
private trait A {
  val test = "test "
}

behavior of "test case"
it should "test something" in new A {
  println(test)
}
```

## Workaround

Revert scalafmt to `version=1.5.1` without any configuration changes from above.
or
this seems to be related to https://github.com/scalameta/scalafmt/issues/1439 because commenting out `unindentTopLevelOperators` fixes the problem.

kitbellew:
Fixes #1547. Follow-on to #1440.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -494,7 +494,8 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
     if (style.verticalAlignMultilineOperators) {
       if (formatToken.left.text == "=") 2 else 0
     } else if (style.unindentTopLevelOperators &&
-      rhsArgs.headOption.forall(_.isNot[Term.Block]) &&
+      rhsArgs.headOption
+        .forall(x => x.isNot[Term.Block] && x.isNot[Term.NewAnonymous]) &&
       !isTopLevelInfixApplication(owner) &&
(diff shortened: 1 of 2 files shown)


kitbellew:

PR #1629, opened 2020-01-11
Add headers for each invidiual editor that Metals supports.

olafurpg:
Users who don't know what Metals is may be looking for editor-specific
headers.

cc/ @ckipp01

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -60,16 +60,48 @@ Scalafmt formatter:
 
 It is not possible to reset this setting for all existing projects.
 
+## VS Code
+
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -313,23 +345,6 @@ Mill have scalafmt support built-in:
 
 - [scalafmt module](http://www.lihaoyi.com/mill/page/configuring-mill.html#reformatting-your-code)
 
-## Vim
```
olafurpg: I don't know if this installation method is still working so  removed it in favor of Metals.

PR #1635, opened 2020-01-17
RedundantBraces: fix various bugs with general expressions

### LINKED ISSUE #1631: Scalafmt Removes Braces Incorrectly
Rf: https://youtrack.jetbrains.com/issue/SCL-16861

I originally raised the above issue with the Intellij team but I think now it is a scalafmt issue. Please see above link for the actual bug details.

This is my scalafmt config:

```
addSbtPlugin("org.scalameta"     % "sbt-scalafmt"    % "latest.integration")
```

```
version=2.0.0-RC5
style = defaultWithAlign
maxColumn = 120
lineEndings = unix
importSelectors = singleLine

project {
  git = true
}

align = most

align {
  tokens = [ {code = "=>", owner = "Case|Type.Arg.ByName"}, "=", "<-", "->", "%", "%%", "should", "shouldBe", "must" ]
  arrowEnumeratorGenerator = true
  openParenCallSite = false
  openParenDefnSite = false
}

binPack {
  parentConstructors = false
}

continuationIndent {
  callSite = 2
  defnSite = 2
}

newlines {
  penalizeSingleSelectMultiArgList = false
  sometimesBeforeColonInMethodReturnType = true
}

rewrite {
  rules = [RedundantBraces, RedundantParens, AsciiSortImports]
  redundantBraces {
    maxLines = 100
    includeUnitMethods = true
    stringInterpolation = true
    generalExpressions = true
  }
}

spaces {
  inImportCurlyBraces = false
  beforeContextBoundColon = false
}
```

### LINKED ISSUE #1633: Redundant braces in general expressions: various bugs
* non-value generating statements #1631 
* Incorrect newlines in infix
```
  a ~> {
    b
  } ~> c
```
is rewritten as
```
  a ~>
    b
  ~> c
```
instead of
```
  a ~>
    b ~> c
```
* Infix lambda calls:
```
  a map { _.toString }
```
to
```
  a map _.toString
```
* lambda expressions:
```
  val a = if (b) { (c: Int) => d } else { (c: Int) => e }
```
to
```
  val a = if (b) (c: Int) => d else (c: Int) => e
```

kitbellew:
* If the block is part of an infix expression, rewrite only if it also contains an infix expression
* do not rewrite infix block if its closing brace is the first non-whitespace character on the line
* don't apply the special if-within-if logic when the nested expression is not `if`
* change how and when we remove newlines after a brace is removed

Fixes #1631.
Fixes #1633.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -121,8 +121,8 @@ case object RedundantBraces extends Rewrite {
     // a single-stat lambda with braces can be converted to one without braces,
     // but the reverse conversion isn't always possible
     case fun @ Term.Function(_, body)
-        if fun.tokens.last.is[Token.RightBrace]
-          && exactlyOneStatement(body) && isLineSpanOk(body) =>
+        if fun.tokens.last.is[Token.RightBrace] &&
+          isSingleStatLineSpanOk(body) =>
       val rbrace = fun.tokens.last
       val lbrace = ctx.matchingParens(TokenOps.hash(rbrace))
       if (lbrace.start <= body.tokens.head.start) {
@@ -167,23 +167,12 @@ case object RedundantBraces extends Rewrite {
       }
     }
 
-  private def exactlyOneStatement(b: Term.Block): Boolean =
-    b.stats.lengthCompare(1) == 0
-
-  private def exactlyOneStatement(t: Term): Boolean = t match {
-    case b: Term.Block => exactlyOneStatement(b)
-    case _ => true
-  }
-
   private def removeBlock(b: Term.Block)(implicit ctx: RewriteCtx): Boolean = {
     b.parent.exists {
 
       case p: Case =>
         settings.generalExpressions && {
-          (p.body eq b) ||
-          exactlyOneStatement(b) &&
-          blockSizeIsOk(b) &&
-          !retainSingleStatBlock(b)
+          (p.body eq b) || shouldRemoveSingleStatBlock(b)
         }
 
       case _: Term.Apply =>
@@ -195,15 +184,12 @@ case object RedundantBraces extends Rewrite {
       case d: Defn.Def =>
         def disqualifiedByUnit =
           !settings.includeUnitMethods && d.decltpe.exists(_.syntax == "Unit")
-        def innerOk =
-          b.stats.head match {
-            case _: Term.Function | _: Defn => false
-            case _ => true
-          }
+        def innerOk(s: Stat) = s match {
+          case _: Term.Function | _: Defn => false
+          case _ => true
+        }
         settings.methodBodies &&
-        exactlyOneStatement(b) &&
-        blockSizeIsOk(b) &&
-        innerOk &&
+        getSingleStatIfLineSpanOk(b).exists(innerOk) &&
         !isProcedureSyntax(d) &&
         !disqualifiedByUnit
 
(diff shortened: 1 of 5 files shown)


kitbellew:

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenTraverser.scala
```diff
@@ -42,25 +42,45 @@ class TokenTraverser(tokens: Tokens) {
     }
   }
 
+  def findAfter(
```
tanishiking: Can you add some explanation about the behavior of `predicate` with scaladoc?: it's a bit difficult to catch the meaning of this method for me.

- If it returns None: skip and keep searching.
- Some(false): stop searching.
- Some(true): return that token.

scalafmt-tests/src/test/resources/rewrite/RedundantBraces.stat
```diff
@@ -237,3 +237,310 @@ object a {
...
+object a {
+  b getOrElse c
+}
+<<< #1633 3.5: function block in a val statement
+object a {
+  b map { _.select }
+}
+>>>
+object a {
+  b map _.select
```
tanishiking: This doesn't compile.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -129,37 +129,49 @@ case object RedundantBraces extends Rewrite {
...
+             * follows a non-whitespace character on the same line as we don't
+             * break lines around infix expressions.
+             * we shouldn't join with the previous line (which might also end
+             * in a comment), and if we keep the break before the right brace
+             * we are removing, that will likely invalidate the expression. */
+            ctx.tokenTraverser.findBefore(close)(isLFSkipWhitespace).isEmpty
+          } else {
+            removeLFToAvoidEmptyLine(close)
+            true
+          }
```
tanishiking: So this tries to fix.

```
  a ~> {
    b
  } ~> c
is rewritten as

  a ~>
    b
  ~> c
instead of

  a ~>
    b ~> c
```

and


```
Infix lambda calls:
  a map { _.toString }
to

  a map _.toString
```

from https://github.com/scalameta/scalafmt/issues/1633, and https://github.com/scalameta/scalafmt/issues/1631, right?

---

I think this leads to invalid formatting result: for example

if we format the following code:

```scala
val x = xs map { _.toString }
```

to

```scala
val x = xs map _.toString
```

which doesn't compile.

Maybe we should stop remove braces for `Term.ApplyInfix` like `Term.Apply` doesn't https://github.com/adampauls/scalafmt/blob/7a3ace0a8209008ad13916e19a0919e425ed43f0/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala#L201-L205 what do you think about this?

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -129,37 +129,49 @@ case object RedundantBraces extends Rewrite {
...
-  private def removeTrailingLF(
-      bodyEnd: Position,
-      close: Token
-  )(implicit builder: PatchBuilder, ctx: RewriteCtx): Unit =
-    if (close.pos.startLine != bodyEnd.endLine) {
-      import ctx.tokenTraverser._
-      val next = nextToken(close)
-      if (next.is[LF])
-        builder += TokenPatch.Remove(next)
+  private def isLFSkipWhitespace(token: Token): Option[Boolean] =
```
tanishiking: Can we add brief scaladoc that explains that this method is used with the `findAfter/findBefore`? because as long as this method isn't passed to the predicate of `findAfter/Before`, this method doesn't make sense, and it's a bit confusing for me.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
kitbellew: i was trying to remove as little functionality as possible. i am leaning towards what looks like a reasonably safe option of basically allowing only another infix with a parent infix, using the precedence logic that was there before. thoughts?

scalafmt-tests/src/test/resources/rewrite/RedundantBraces.stat
kitbellew: fixed. allowing only infix with infix.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -220,35 +236,42 @@ case object RedundantBraces extends Rewrite {
   }
 
   /** Some blocks look redundant but aren't */
-  private def retainSingleStatBlock(b: Term.Block): Boolean =
+  private def retainSingleStatBlock(b: Term.Block): Boolean = {
+    val stat = b.stats.head
```
tanishiking: Though I know we normally don't pass `Block` whose stats is empty to this function, can we use `headOption` here?

kitbellew: let me try. not sure what the function should return if `stats` is empty or more than one.

kitbellew: @tanishiking PTAL: https://github.com/scalameta/scalafmt/pull/1635/commits/614d9671895d791c748f6a06f88525d015a1a4a6

Issue #1605, opened 2019-12-17
Allow setting newlines at EOF to false.


bruno-medeiros:
I don't really care that some old fashioned Unix tools required newlines at end of file, I'd like a setting that would remove them please.

poslegm:
I think it will be reasonable to have config for it

olafurpg:
Thank you for reporting! I’m on the fence about introducing a new configuration option for this feature. Is there a reason the newline at the end of the file is causing issues?

bruno-medeiros:
It is not causing any issues per se. It is just for aesthetics: ie, one less scroll line, and also having the editor cursor going after the final '}' in the file when one presses Ctrl+End or so. (as opposed to the line below the `}`). I understand these are subjective preferences :)

olafurpg:
Closing as wontfix since newlines at the end-of-file is a widely used convention in the programming community.

PR #1639, opened 2020-01-19
Allow setting newlines at EOF to false (#1605)

fijolekProjects:
This PR adds ability to configure newlines at the end of file.
At first I tried to add test in `scalafmt-tests` module, but then I've found out that method `HasTests.trimmed` always adds newline to each test case. I guess that changing this behaviour based on `ScalafmtConfig.newlines.alwaysAtTheEndOfFile` would be confusing, so I've added regular test in `ScalafmtTest` instead.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -454,6 +454,29 @@ else {
 }
 ```
 
+### `newlines.alwaysAtTheEndOfFile`
+
+```scala mdoc:defaults
+newlines.alwaysAtTheEndOfFile
+```
+
+```scala mdoc:scalafmt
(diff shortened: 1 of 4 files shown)


fijolekProjects:
Sorry about that. It had "good first time contribution" label, so I've assumed it's ready to implement. If it's controversial then maybe these two (#1605, #1639) should be closed?

olafurpg:
@fijolekProjects I'm sorry for the confusion. I removed the label and closed the original issue.

poslegm:
I'm sorry for the label: I marked issue as «good first contribution» because of easy implementation. But you are right and it is bad label for controversial issues.

PR #1646, opened 2020-01-25
Strict filters for exclude files in project (#1645)

### LINKED ISSUE #1645: Scalafmt Fails to Format Scala Sources in Play Project
- **Version**: 2.3.2
- **Integration**: Intellij / VS Code
- **Configuration**:
```
version=2.3.2
style = defaultWithAlign
maxColumn = 120
lineEndings = unix
importSelectors = singleLine

project {
  git = true
}

align = most

align {
  tokens = [ {code = "=>", owner = "Case|Type.Arg.ByName"}, "=", "<-", "->", "%", "%%", "should", "shouldBe", "must" ]
  arrowEnumeratorGenerator = true
  openParenCallSite = false
  openParenDefnSite = false
}

binPack {
  parentConstructors = false
}

continuationIndent {
  callSite = 2
  defnSite = 2
}

newlines {
  penalizeSingleSelectMultiArgList = false
  sometimesBeforeColonInMethodReturnType = true
}

rewrite {
  rules = [RedundantBraces, RedundantParens, AsciiSortImports]
  redundantBraces {
    maxLines = 100
    includeUnitMethods = true
    stringInterpolation = true
  }
}

spaces {
  beforeContextBoundColon = true
  inImportCurlyBraces = false
  beforeContextBoundColon = false
}

project.excludeFilters = [
  ".*\\.sc"
]
```

## Steps

Given code like this:
```scala
ORIGINAL CODE
```

```bash
scalafmt
```

## Problem

Scalafmt formats does not format this code accordingly:
```scala
case class Test(
            f : Int,
    d : Int
)
```

## Expectation

```scala
case class Test(
    f : Int,
    d : Int
)
```

## Notes
You can download this repo: https://github.com/hmrc/cgt-property-disposals-frontend to reproduce the bug. Create any case class in the models folder with incorrect format, open a sbt session, and compile the sources. Even though scalamfmtOnCompile is set to true and you can see in the logs that it has formatted some files, it fails to format the case class you have created.

poslegm:
Fixes #1645 

Description of problem in the related issue. Or may be just set `strict = true` for _all_ `FilterMatcher`'s exclude filters?

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
@@ -2,6 +2,7 @@ package org.scalafmt.config
 
 import metaconfig._
 import org.scalafmt.util.OsSpecific
+import FilterMatcher.mkRegexp
 
 case class ProjectFiles(
     git: Boolean = false,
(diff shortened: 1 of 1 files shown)


olafurpg:
A workaround for #1645 in the meantime is to add a `$` at the end of the regexp.

poslegm:
`PathMatcher` would be a good replacement. So we need to:
1. Implement `FilterGlobMatcher` as an alternative for [FilterMatcher](https://github.com/scalameta/scalafmt/blob/c92153e777984db6d69ec359dd1b1115bd2199d6/scalafmt-core/shared/src/main/scala/org/scalafmt/config/FilterMatcher.scala#L7)
2. Replace `exclude`/`include` in the documentation with `excludeGlobs`/`includeGlobs`
3. Keep old `exclude` and `include` in the config for compatibility purposes.

Sounds good!

PR #1647, opened 2020-01-25
State/FormatWriter: various optimizations and simplifications

kitbellew:
A few of the changes appear to speed up the formatting slightly (approx 5%). Additionally, they will allow in the future to add generic handling of comments in the code, similar to whitespace.

No formatting changes are introduced.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -287,7 +287,7 @@ class FormatWriter(formatOps: FormatOps) {
           case TrailingCommas.never
               if left.is[T.Comma] && rightIsCloseDelim &&
                 !tok.right.is[T.Comment] && isNewline =>
-            sb.deleteCharAt(sb.length - 1)
+            sb.setLength(sb.length - 1)
             sb.append(whitespace)
 
           // foo(
@@ -318,7 +318,7 @@ class FormatWriter(formatOps: FormatOps) {
           case _
               if left.is[T.Comma] && rightIsCloseDelim &&
                 !tok.right.is[T.Comment] && !isNewline =>
-            sb.deleteCharAt(sb.length - 1)
+            sb.setLength(sb.length - 1)
 
           case _ => sb.append(whitespace)
         }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -22,13 +22,11 @@ class FormatWriter(formatOps: FormatOps) {
     val sb = new StringBuilder()
     var lastState = State.start // used to calculate start of formatToken.right.
     val locations = getFormatLocations(tokens.arr, splits, debug = false)
-    val tokenAligns = locations.tokenAligns
 
     locations.iterate.foreach { entry =>
       val location = entry.curr
       val state = location.state
       val formatToken = location.formatToken
-      val whitespace = entry.getWhitespace
 
       formatToken.left match {
         case c: T.Comment =>
@@ -61,13 +59,7 @@ class FormatWriter(formatOps: FormatOps) {
           sb.append(rewrittenToken)
       }
 
-      handleTrailingCommasAndWhitespace(
-        formatToken,
-        state,
-        sb,
-        whitespace,
-        tokenAligns
-      )
+      entry.formatWhitespace(sb)
 
       lastState = state
     }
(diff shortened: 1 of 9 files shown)


kitbellew:

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -365,207 +422,93 @@ class FormatWriter(formatOps: FormatOps) {
...
 
 object FormatWriter {
 
   case class FormatLocation(
       formatToken: FormatToken,
-      split: Split,
       state: State
   )
+
+  private val indentations: Array[String] = {
```
poslegm: There is `scala.collection.immutable.ArraySeq` — immutable collection backed by Array. It's more safe to use it

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -365,207 +422,93 @@ class FormatWriter(formatOps: FormatOps) {
...
+    val arr = new Array[String](64)
+    arr(0) = ""
+    (1 until arr.length).foreach(idx => arr(idx) = " " + arr(idx - 1))
+    arr
+  }
+
+  private def getIndentation(len: Int): String =
```
poslegm: Can you please write a comment with explanation of this trick? I didn't immediately realize that it is string concatenation cache

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -125,101 +119,164 @@ class FormatWriter(formatOps: FormatOps) {
...
+      Iterator.range(0, locations.length).map(new Entry(_))
+
+    private def getAlign(tok: Token, alignOffset: Int = 0): Int =
+      tokenAligns.get(hash(tok)).fold(0)(_ + alignOffset)
+
+    class Entry(val i: Int) {
+      val curr = locations(i)
+      val previous = locations(math.max(i - 1, 0))
+
+      @inline def tok = curr.formatToken
```
poslegm: I don't sure that inline optimization enabled for scalafmt builds

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/State.scala
```diff
@@ -126,11 +93,66 @@ object State {
...
+    formatOff = false
+  )
+
+  // this is not best state, it's higher priority for search
+  object Ordering extends Ordering[State] {
+    override def compare(x: State, y: State): Int = compareAt(x, y, 0)
+  }
+
+  private val comparisons: Seq[(State, State) => Int] = Seq(
+    compareCost,
```
poslegm: Also you can encode it without `IndexOutOfBounds` risk

```scala
sealed trait Comparison {
  def compare(s1: State, s2: State): Int
  def next: Option[Comparison]
}
case object Cost extends Comparison {
  override def compare(s1: State, s2: State) = ???
  override def next = Some(SplitsLength)
}
case object SplitsLength extends Comparison {
  override def compare(s1: State, s2: State) = ???
  override def next = Some(SplitOrigin)
}
case object SplitOrigin extends Comparison {
  override def compare(s1: State, s2: State) = ???
  override def next = None
}
```

poslegm:
It's hard to review, I might miss something important :с

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: will do

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: will do

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/State.scala
kitbellew: if you feel strongly, i can encode. however, this approach is a bit harder to maintain (reordering or adding a new comparison is not very easy), and is probably a little slower.

i could replace `i < comparisons.length` with the slightly slower
```
  (i + 1) match {
    case comparisons.length => 0
    case next => <tail recursive call>
  }
```
which would be equivalent to `next match { case Some(_) => ...; case None => ... }` in your case.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: i understand. well, in case it is enabled in the future, should i keep these annotations?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: will do

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/State.scala
kitbellew: if you feel strongly, i can encode. however, this approach is a bit harder to maintain (reordering or adding a new comparison is not very easy), and is probably a little slower.

i could replace `i < comparisons.length` with the slightly slower
```
  (i + 1) match {
    case comparisons.length => 0
    case next => <tail recursive call>
  }
```
which would be equivalent to `next match { case Some(_) => ...; case None => ... }` needed in your case.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: i understand. well, in case it is enabled in the future, should i keep these annotations?

kitbellew:
> It's hard to review, I might miss something important :с

please let me know how to make it easier.

i usually review one commit at a time, with the whitespace removed. with this approach, the only commit that is harder to review is the refactor titled "FormatWriter: handle whitespace in FormatLocations" (it moves a block virtually unchanged).

i could split the PR into several but unlike gerrit, github doesn't allow putting PRs on top of each other and allowing gradual merging.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: turns out `ArraySeq` was not available in 2.11. will use `ArrayBuffer.view`, looks about the same.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
poslegm: Yes, of course. Just noticing that now it will not be inlined

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
poslegm: :+1:

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/State.scala
poslegm: My point is ADT encoding protect you from index out of bounds errors in runtime after comparison rules modification. So it simplifying maintaing in this sense. But you are right, there is a trade-off with adding new rules in "linked list".

Anyway it is not a blocker for this PR

poslegm:
Review is sophisticated because I am not in context of your perfomance research 😄 I will try to re-read it by commits when I have some free time. 

I think that it is a precious contribution and PR can be merged now, because there is no formatting changes. 5% perfomance enhancement would be advantageous for large codebases!

kitbellew:
> Review is sophisticated because I am not in context of your perfomance research 😄 I will try to re-read it by commits when I have some free time.
> 
> I think that it is a precious contribution and PR can be merged now, because there is no formatting changes. 5% perfomance enhancement would be advantageous for large codebases!

Thank you! :) If you do get around to reviewing it later and see something, let me know, easy to revise.

PR #1649, opened 2020-01-26
SortModifiers: Use .syntax on tokens, not tree

### LINKED ISSUE #1506: Maintain manual line breaks within concatenated annotation
- **Version**: 2.0.1
- **Integration**: IntelliJ
- **Configuration**:
```
version = "2.0.1"
align = none
align.tokens = []
optIn.configStyleArguments = false
maxColumn = 120
rewrite.rules = [PreferCurlyFors, AsciiSortImports, SortModifiers, RedundantBraces, RedundantParens]
rewrite.redundantBraces.methodBodies = false
  "`sealed`"
  "`abstract`"
  "`override`"
  "`private`"
  "`protected`"
  "`implicit`"
  "`final`"
  "`lazy`"
]
project.git = true
```

## Steps

Given code like this:
```scala
  @annotation(
    "Here is a really long string, which needs to be wrapped." +
      " scalafmt forces this concatenated string onto a single line" +
      " but it would be preferable for it to behave like it does for the string below."
  )
  val SomeLongString = "Here is a really long string, which needs to be wrapped." +
    " This split string will stay wrapped properly, but the same string in the annotation" +
    " won't and is forced to be on a single line."
```

Run format in IntelliJ with it configured to use scalafmt

## Problem

```scala
  @annotation(
    "Here is a really long string, which needs to be wrapped." + " scalafmt forces this concatenated string onto a single line" + " but it would be preferable for it to behave like it does for the string below."
  )
  val SomeLongString = "Here is a really long string, which needs to be wrapped." +
    " This split string will stay wrapped properly, but the same string in the annotation" +
    " won't and is forced to be on a single line."
```

## Expectation

[Unchanged within the annotation]

## Workaround

None known. I just have super long strings within my long annotations.

kitbellew:
Whether a bug or a feature, but for Mod.Annotation, syntax loses all NL tokens. So instead of tree.syntax, let's use tree.tokens.syntax which doesn't have that problem.

Fixes #1506.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/SortModifiers.scala
@@ -48,9 +48,12 @@ object SortModifiers extends Rewrite {
 
       ctx.addPatchSet(sortedMods.zip(sanitized).flatMap {
         case (next, old) =>
-          val removeOld = old.tokens.map(TokenPatch.Remove)
-          val addNext = TokenPatch.AddRight(old.tokens.head, next.syntax)
-          removeOld :+ addNext
+          if (next eq old) Seq.empty
+          else {
+            val removeOld = old.tokens.tail.map(TokenPatch.Remove)
+            val addNext = TokenPatch.Replace(old.tokens.head, next.syntax)
+            removeOld :+ addNext
+          }
       }: _*)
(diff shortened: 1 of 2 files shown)


kitbellew:
@olafurpg @tanishiking @poslegm PTAL. Perhaps the right place to fix is `scalameta` but I am not really familiar with that repository.

PR #1651, opened 2020-01-28
Policy: always pass sourcecode.Line implicitly

kitbellew:
Without it, errors are reported with the wrong line (albeit without the
source file anyway).


kitbellew:

PR #1654, opened 2020-01-28
Scalafmt bugfix: restore .format signatures while passing filename

### LINKED ISSUE #1628: 2.3.3-RC1 always crashes with "classpath is corrupted"
I tried upgrading to 2.3.3-RC1 and I always get

 ```[error] scalafmt version 2.3.3-RC1 classpath is corrupted```

I've tried running `sbt clean update` and that did not help. This issue is not present with 2.3.2 so I do not believe it is my setup.

Thanks for the great tool, looking forward to using 2.3.3!

My configuration is:

```
version = "2.3.3-RC1"
maxColumn = 120 
trailingCommas = always
  until
  to
  by
  eq
  ne
  "should.*"
  "contain.*"
  "must.*"
  in
  ignore
  be
  taggedAs
  thrownBy
  synchronized
  have
  when
  size
  only
  noneOf
  oneElementOf
  noElementsOf
  atLeastOneElementOf
  atMostOneElementOf
  allElementsOf
  inOrderElementsOf
  theSameElementsAs
  // these are new
  throws
  returns
  satisfy
]
```

kitbellew:
Create a new, package-private method `.formatCode` with the implementation based on the existing .format.

Make sure not to modify the existing interfaces of `.format` overloads, as they are externally expected (via reflection).

The bug was introduced in #1608 and removed the 3-parameter version of the `.format` overload as "unnecessary".

Fixes #1628.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
@@ -24,6 +24,7 @@ object Scalafmt {
   private val WindowsLineEnding = "\r\n"
   private val UnixLineEnding = "\n"
 
+  // XXX: don't modify signature, scalafmt-dynamic expects it via reflection
   /**
     * Format Scala code using scalafmt.
     *
@@ -43,6 +44,15 @@ object Scalafmt {
       style: ScalafmtConfig,
       range: Set[Range],
       filename: String
+  ): Formatted = {
+    formatCode(code, style, range, filename)
+  }
+
+  private[scalafmt] def formatCode(
+      code: String,
+      style: ScalafmtConfig = ScalafmtConfig.default,
+      range: Set[Range] = Set.empty,
+      filename: String = "<input>"
   ): Formatted = {
     try {
       val runner = style.runner
@@ -82,12 +92,13 @@ object Scalafmt {
     }
   }
 
+  // XXX: don't modify signature, scalafmt-dynamic expects it via reflection
   def format(
       code: String,
       style: ScalafmtConfig = ScalafmtConfig.default,
       range: Set[Range] = Set.empty[Range]
   ): Formatted = {
-    format(code, style, range, "<input>")
+    formatCode(code, style, range)
   }
 
   def parseHoconConfig(configString: String): Configured[ScalafmtConfig] =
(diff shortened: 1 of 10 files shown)


kitbellew:

tanishiking:
Thank you for the quick fix! I'll take a look this evening (UTC+09:00).

scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
```diff
@@ -82,6 +92,15 @@ object Scalafmt {
     }
   }
 
+  // XXX: don't modify signature, scalafmt-dynamic expects it via reflection
+  def format(
```
olafurpg: This looks like a new signature, did it exist in an older version of Scalafmt? Can you elaborate on how this fixes #1628

kitbellew: @olafurpg per our standard "i screwed up" practice, i reverted the original commit that breaks the contract (that's commit 1) and then added a second one which mainly restores the functionality but without touching the signatures of the two .format methods, you can probably see it by reviewing via the commits tab rather than the changes tab.

@tanishiking posted a stack trace yesterday which shows the failure to be in trying to find .format with the three-parameter signature, via reflection: https://github.com/scalameta/scalafmt/issues/1628#issuecomment-579372180

While I, unfortunately, don't have a way to reproduce the problem, this is most likely the only reason for #1628.

tanishiking: This regression had been introduced in https://github.com/scalameta/scalafmt/pull/1608 (sorry I missed this change), and this PR basically revert the root cause of this regression.

Basically, it's probably better to write a detailed description, so that even people (who don't follow the context) can understand the change :)

olafurpg: Thanks for the explanation! I haven't followed all the recent changes closely. It's very important that the `Scalafmt.scala` file does not have binary breaking changes. The `scalafmt-dynamic` interface is an attempt to provide a richer public API that is binary stable, but build tools already interface with the `Scalafmt` object so we need to keep supporting it

PR #1661, opened 2020-02-01
Minor improvements to split, decision and policy logic

kitbellew:
* Policy: return just the splits in the partial func
  * the point of a policy is to narrow down the set of splits, there's no need to duplicate the logic of building up an object which contains the input format token and can then be fed into another policy
* Policy: make sure to extend expire in andThen
  * combining two policies would occasionally fail since the expiration point of the second one was ignored
* Modification: optimize/cache newlines computation
  * minor performance


kitbellew:

PR #1660, opened 2020-01-31
Router: allow single-line curly lambdas

kitbellew:
The rule applies only if `newlines.afterCurlyLambda` is either `never`, or `preserve` without a line break, and abides by `newlines.alwaysBeforeCurlyBraceLambdaParams`.

Also, make sure to clarify the other space rule for lambda to break on arrow before breaking on the closing brace, so that it doesn't end up with a lone break at the end.

This is a prerequisite to adding the rewrite rule requested in #1027. Also, fixes #965.

For `alwaysBeforeCurlyBraceLambdaParams = false`:
```
  // new
  a map { x => x }
  // old
  a map { x =>
    x
  }
```
For `alwaysBeforeCurlyBraceLambdaParams = true`:
```
  // new
  a map {
    x => x
  }
  // old
  a map {
    x =>
      x
  }
```

`scala-repos` diffs:
  * forbid multiline toks in single-line block: https://github.com/kitbellew/scala-repos/commit/2daf6cae8c63eeaef01ac32f2461d1a7d6c2540b?w=1
  * allow single-line curly lambdas: https://github.com/kitbellew/scala-repos/commit/2f5c04590797971b908c9908866cb32c91ca9d41?w=1

PR #1663, opened 2020-02-02
RedundantBraces: use () for {} one-line lambdas

kitbellew:
Output a single-line { a => b } as (a => b), as mentioned in #1027.

`scala-repos` diff: https://github.com/kitbellew/scala-repos/commit/1ba9dbff2ae0c1f977191b64725d3c6f9e180fd4?w=1

NB: We can't easily do the opposite, formatting multi-line paren lambdas using braces, it can sometimes lead to incorrect code:
```
  a(x =>
    1
    + 2)
```
is the same as
```
  a { x =>
    1 +
    2
  }
```
but not
```
a { x =>
  1
  +2
}
```

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -2,10 +2,12 @@ package org.scalafmt.internal
 
 import java.util.regex.Pattern
 
+import org.scalafmt.rewrite.RedundantBraces
 import org.scalafmt.util.TreeOps
 
 import scala.annotation.tailrec
 import scala.collection.IndexedSeq
+import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 import scala.meta.tokens.Token
 import scala.meta.tokens.{Token => T}
@@ -59,9 +61,9 @@ class FormatWriter(formatOps: FormatOps) {
         case c: T.Constant.Double =>
           sb.append(initStyle.literals.double.process(c.syntax))
         case token =>
+          val syntax = Option(location.replace).getOrElse(token.syntax)
           val rewrittenToken =
-            formatOps.initStyle.rewriteTokens
-              .getOrElse(token.syntax, token.syntax)
+            formatOps.initStyle.rewriteTokens.getOrElse(syntax, syntax)
           sb.append(rewrittenToken)
       }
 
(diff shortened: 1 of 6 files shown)


kitbellew:

kitbellew:
@olafurpg @tanishiking @poslegm if you have some time to review, that would be ideal. it's a logical follow-up to #1660 which will be released in 2.4.0 and could be part of it.

tanishiking:
Thanks! I'll take a look at this next weekend.

kitbellew:
> Thanks! I'll take a look at this next weekend.

@tanishiking thank you, looking forward to it.

kitbellew:
> I like this changes! It's really desired feature for scalafmt.
> 
> It's worth making sure that there are `.stat` tests about **not** replacing braces when `RedundantBraces` is off.

added a test to confirm this. also added a test to verify that nested blocks of braces are correctly rewritten.

kitbellew:
> This is an impressive diff! I personally don't use `RedundantBraces` but with this change I might be tempted to give it a spin :)

good to know! we are all about removing the artistic urge of fellow scala practitioners :)

adampauls:
@kitbellew I wanted to follow up on this PR. I don't quite understand the difficulty with 

```scala
a(x =>
    1
    + 2)
```

You can always remove the line breaks, switch to curlies, and format. Presumably the difficulty is that you want to preserve the precise line breaks, inline with how scalafmt typically operates? My team would certainly be happy if their were an optional rewrite that didn't preserve line breaks in this case, but also, would it not be simple to do the transformation from parens to curlies iff the expression would stay the same? Or is that condition harder to check than I thought?

I'd be happy to take a stab at this if you can give me some pointers.

kitbellew:
> @kitbellew I wanted to follow up on this PR. I don't quite understand the difficulty with
> 
> ```scala
> a(x =>
>     1
>     + 2)
> ```
> 
> You can always remove the line breaks, switch to curlies, and format. Presumably the difficulty is that you want to preserve the precise line breaks in line with scalafmt's general philosophy?

@olafurpg tried, in vain, to handle line breaks around infix operations and ultimately gave up; so, in this case, that's part of it.

i think that if we can come with a way to modify infix line breaks in a way where an infix operator ends a line rather than starts it, changing parens to curlies should end up with an equivalent expression.

but as it stands right now, there's no simple fix.

adampauls:
I don't follow. I think in paren context, you can always safely remove newlines? The issue is that there might be blocks inside where newlines can't be safely removed like
```scala
a(x => {1
 + 2
} + 3)
```
Is that right? This feels like it should not be that hard to deal with, but I'm sure I'm oversimplifying. Any chance you could point to me to the part of the code I might want to look, if such simple pointers exist?

kitbellew:
@adampauls the approach i have taken before with braces-to-parens was *not* to modify how `scalafmt` decides to split the code (meaning, whether to use a space or newline or nothing) but *after* the splits have been determined, see if the end result has no newlines between `{` and `}` (and you can see the code in this PR).

for the opposite direction, i would do the same, and if there are post-formatting newlines between `(` and `)`, see if you can replace them with braces. the complexity:
* correctness of the resulting code is the biggest question
* potential line overflow: switching to a brace adds two characters (from `(` to ` { `) on the first line; nothing one can do about that, as the whitespace splitting is decided by then
* identifying whether the closing `)` is dangling, because the closing `}` must be made so

PR #1665, opened 2020-02-02
RedundantParens: rewrite single-arg apply of block

### LINKED ISSUE #1321: Feature request: Rewrite rule for "function argument with parens and braces"
I'd like scalafix to be able to detect when a single-argument parameter group where the parameter is a function (things like `map`, `flatMap`, etc) have redundant parens around function braces.

```
//before
foo.map(x => { blah(x) })
//after
foo.map { x => blah(x) }
```

and
```
//before
foo.map({x => blah(x) })
//after
foo.map { x => blah(x) }
```

kitbellew:
Fixes #1321.

Rewrites `a({ b })` and `(a { b })` as `a { b }` (both are considered a single-arg `Apply` of a `Block`).

`scala-repos` diffs: https://github.com/kitbellew/scala-repos/commit/90e2f8b23ea114c5e15f0c7e8b9b1b2833f58121?w=1


kitbellew:

PR #1666, opened 2020-02-02
Router: treat single-arg lambda of init like apply

kitbellew:
Just like #1551, we need to format ctor/init consistently.

`scala-repos` diffs: https://github.com/kitbellew/scala-repos/commit/e5a3562989d7f02f4c348a0b730cc74ca9a1dfa4?w=1


kitbellew:

PR #1669, opened 2020-02-03
Trailing commas: add a rewrite rule, remove first

### LINKED ISSUE #1667: Non-idempotent formatting in 2.4.0-RC1
- **Configuration**:
```
version = "2.4.0-RC1"
maxColumn = 120 
trailingCommas = always
  until
  to
  by
  eq
  ne
  "should.*"
  "contain.*"
  "must.*"
  in
  ignore
  be
  taggedAs
  thrownBy
  synchronized
  have
  when
  size
  only
  noneOf
  oneElementOf
  noElementsOf
  atLeastOneElementOf
  atMostOneElementOf
  allElementsOf
  inOrderElementsOf
  theSameElementsAs
  // these are new
  throws
  returns
  satisfy
]
```

## Steps

Given code like this:
```scala
object Foo {
  def foo(): Unit = {
    aaaaaaaaaaaaaaa(c, ddddddd).map { zzzzzzzzzz =>
      val ccccccccccccccccc = zzzzzzzzzz.bbbbbbbbbbbbbbbbbb(EEEEE.xxxxxxxxxxx, Schema.String).toOption.flatMap {
        case c: Seq[String] =>
          xxxxxxxxxxxxxxxxxxxxxxxx(c, (v: String) => s"""\\"$v\\"""")(Schema.String).map(s =>
            Yyyyyyyyyyyyyyyy("SSSSSSS", s),
          )
      }
    }.get
  }
}
```

(Sorry for the mess, it was hard to minimize further).

`sbt scalafmtCheckAll` does not succeed after `sbt scalamftAll` is run.

## Workaround

I can introduce newlines to get around the problem, but we don't want to have `scalafmtCheckAll` in CI if it's not idempotent.

## Notes

Even though the 
```scala
 xxxxxxxxxxxxxxxxxxxxxxxx(c, (v: String) => s"""\\"$v\\"""")(Schema.String)
            .map(s => Yyyyyyyyyyyyyyyy("SSSSSSS", s))
```

snippet is the one that gets changed back and forth by scalafmt, the `.get` seemed to be necessary for reproduction, so I think there's something going on in an interaction between the two.

kitbellew:
In order to avoid idempotence issues triggered by modified commas at the final stage of formatting, let's remove them in the beginning to have a clean format, and append as necessary.

This rule is only applied if the TrailingCommas configuration takes one of the mutating settings, i.e. always or never.

Fixes #1667.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -218,68 +218,48 @@ class FormatWriter(formatOps: FormatOps) {
 
         import org.scalafmt.config.TrailingCommas
 
-        val owner = owners(tok.right)
-        if (!runner.dialect.allowTrailingCommas ||
-          !TreeOps.isImporterOrDefnOrCallSite(owner)) {
+        /* If we are mutating trailing commas ('always' or 'never'), we should
+         * have removed them first in RewriteTrailingCommas; now we simply need
+         * to append them in case of 'always', but only when dangling */
+        val appendComma = runner.dialect.allowTrailingCommas &&
+          initStyle.trailingCommas == TrailingCommas.always &&
+          !tok.left.is[T.Comma] && !tok.left.is[T.Comment] && {
+          def owner = owners(tok.right)
+
+          val (skip, nextNonCommentTok) = nextNonCommentWithCount(tok)
+          val right = nextNonCommentTok.right
+          def isNewline =
+            Seq(curr, locations(math.min(i + skip, locations.length - 1)))
+              .exists(_.state.split.modification.isNewline)
+
+          // Scala syntax allows commas before right braces in weird places,
+          // like constructor bodies:
+          // def this() = {
+          //   this(1),
+          // }
+          // This code simply ignores those commas because it does not
+          // consider them "trailing" commas. It does not remove them
+          // in the TrailingCommas.never branch, nor does it
+          // try to add them in the TrainingCommas.always branch.
+
+          // skip empty parens/braces/brackets
+          def rightIsCloseDelim = right match {
+            case _: T.RightBrace =>
+              !tok.left.is[T.LeftBrace] && owner.is[Importer]
+            case _: T.RightParen =>
+              !tok.left.is[T.LeftParen] && TreeOps.isDefnOrCallSite(owner)
+            case _: T.RightBracket =>
+              !tok.left.is[T.LeftBracket] && TreeOps.isDefnOrCallSite(owner)
+            case _ => false
+          }
+          rightIsCloseDelim && isNewline
+        }
+
+        if (appendComma)
+          sb.append(',').append(getWhitespace(-1))
+        else
           sb.append(getWhitespace(0))
-          return
-        }
-        val isImport = owner.isInstanceOf[Importer]
-
-        val left = tok.left
-        val (skip, nextNonCommentTok) = nextNonCommentWithCount(tok)
-        val right = nextNonCommentTok.right
-        val isNewline =
-          Seq(curr, locations(math.min(i + skip, locations.length - 1)))
-            .exists(_.state.split.modification.isNewline)
-
-        // Scala syntax allows commas before right braces in weird places,
-        // like constructor bodies:
-        // def this() = {
-        //   this(1),
-        // }
-        // This code simply ignores those commas because it does not
-        // consider them "trailing" commas. It does not remove them
-        // in the TrailingCommas.never branch, nor does it
-        // try to add them in the TrainingCommas.always branch.
-        lazy val rightIsCloseDelim = right
-          .is[CloseParenOrBracket] || (right.is[T.RightBrace] && isImport)
-
-        initStyle.trailingCommas match {
-          // foo(
-          //   a,
-          //   b
-          // )
-          //
-          // Insert a comma after b
-          case TrailingCommas.always
-              if !left.is[T.Comma] &&
-                !left.is[T.Comment] &&
-                !left.is[T.LeftParen] && // skip empty parentheses
-                rightIsCloseDelim && isNewline =>
-            sb.append(',').append(getWhitespace(-1))
-
-          // foo(
-          //   a,
-          //   b,
-          // )
-          //
-          // Remove the comma after b
-          case TrailingCommas.never
-              if left.is[T.Comma] && rightIsCloseDelim && isNewline =>
-            sb.setLength(sb.length - 1)
-            sb.append(getWhitespace(1))
-
-          // foo(a, b,)
-          //
-          // Remove the comma after b
-          case _ if left.is[T.Comma] && rightIsCloseDelim && !isNewline =>
-            sb.setLength(sb.length - 1)
-
-          case _ => sb.append(getWhitespace(0))
-        }
       }
-
     }
   }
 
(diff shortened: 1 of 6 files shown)


kitbellew:
@olafurpg @tanishiking @poslegm @adampauls PTAL

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -218,68 +218,47 @@ class FormatWriter(formatOps: FormatOps) {
...
-          // Remove the comma after b
-          case _ if left.is[T.Comma] && rightIsCloseDelim && !isNewline =>
-            sb.setLength(sb.length - 1)
-
-          case _ => sb.append(getWhitespace(0))
+        val appendComma = runner.dialect.allowTrailingCommas &&
+          initStyle.trailingCommas == TrailingCommas.always && {
+          tok.left match {
+            case _: T.Comma | _: T.Comment | _: T.LeftParen | _: T.LeftBrace =>
+              // skip empty parens/braces
```
tanishiking: This comment means:

if the left token is `{` or `(`, that means the parentheses/braces is empty, and if we add a comma there, the result would be something like

```scala
foo (,)
```

that's why we skip adding commas for empty parens/braces.

---

It's probably better to move this comment like this:

```scala
            case _: T.LeftParen | _: T.LeftBrace | // skip empty parens/braces
                 _: T.Comma | _: T.Comment =>
```

 So that people can easily understand we are skipping empty parens/braces by not adding commas if the left token is `(` or `{`.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RewriteTrailingCommas.scala
```diff
@@ -0,0 +1,86 @@
+package org.scalafmt.rewrite
+
+import scala.meta._
+import scala.meta.classifiers.Classifier
+import scala.meta.tokens.Token
+
+import org.scalafmt.util.Whitespace
+
+case object RewriteTrailingCommas extends Rewrite {
```
tanishiking: Can we rename to `RemoveTrailingCommas`? (It looks like this rule just remove trailing commas)
Or do we have any plans to add some rules that add a comma later?

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RewriteTrailingCommas.scala
```diff
@@ -0,0 +1,86 @@
...
+      case t: Type.Function => afterArgs(t.params)
+      case t: Ctor.Primary => afterEachArgs(t.paramss)
+
+      case t: Term.Apply => afterArgs(t.args)
+      case t: Type.Apply => afterArgs(t.args)
+      case t: Pat.Extract => afterArgs(t.args)
+      case t: Pat.Tuple => afterArgs(t.args)
+      case t: Term.Tuple => afterArgs(t.args)
+      case t: Term.ApplyType => afterArgs(t.targs)
+      case t: Init => afterEachArgs(t.argss)
```
tanishiking: I'm not 100% sure that these patterns exhaustively cover all patterns, but looking at the test cases, it's ok I guess :+1:

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Rewrite.scala
```diff
@@ -85,7 +86,12 @@ object Rewrite {
   }
 
   def apply(input: Input, style: ScalafmtConfig): Input = {
-    val rewrites = style.rewrite.rules
+    val trailingCommaRewrite =
+      if (!style.runner.dialect.allowTrailingCommas ||
+        style.trailingCommas == TrailingCommas.preserve) Seq.empty
```
tanishiking: If the `RewriteTrailingCommas` does only remove trailing commas, can we populate `RewriteTralingComma` only when `style.trailingCommas === TrailingCommas.never` ?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: will clarify :) in fact, original code only checked for '(', so i am improvising here.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Rewrite.scala
kitbellew: unfortunately, we can't. the whole point is to remove idempotency issues, and that requires feeding the same exact code (ignoring whitespace, that is) into the formatter in successive iterations.

so, we have two options:
* in rewrite, remove unless `preserve`; then add in format if `always` and has newline, or
* in rewrite, remove for `never`, add for `always`; then remove in format if `always` and no newline.

the first option (the one i implemented) is simpler code but could potentially lead to `maxColumn+1` lines.

the second is a bit more code, probably less confusing in the rewrite step (don't need to explain why we are removing in case of `always`) and will not overflow lines.

@tanishiking can you indicate your preference, please?

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RewriteTrailingCommas.scala
kitbellew: i would keep the name, regardless of which of the two options we go with (see previous comment, in `Rewrite.scala`), for two reasons:
* it's consistent with other `RewriteXXX` classes
* while it might only remove in reality, it's applied to all mutating cases (`always` and `never`).

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RewriteTrailingCommas.scala
kitbellew: the original code (in `FormatWriter`) checked for `isDefnOrCallSite`, so I took all patterns from `isDefnSite` and `isCallSite` :)

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Rewrite.scala
kitbellew: ah, one complication with the second option for `always` case which i just discovered.

since scala doesn't allow one-line `a(b, c,)` (unlike python), we must append the comma *and the newline* (if missing), which might mess with the config style logic.

oh well. i'd stay with option 1, then.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: done

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Rewrite.scala
kitbellew: next patch explains why we only remove, but the name (for now) is kept the same. i do not feel too strongly, let me know if after my explanation you'd still prefer to call it `RemoveTrailingCommas`. it will be invoked for `always` as well, though.

tanishiking: Ah, I got it! As long as there's some explanation about this, I think we don't need to rename it :)
Can we add some comment on `val appendComma = ...` side that we have to drop trailing comma beforehand, to avoid idempotency case?

Others are LGTM:+1:

kitbellew: will do shortly, thank you

kitbellew: @tanishiking done. explanations added, regarding idempotence and appending commas in formatter. PTAL.

PR #1673, opened 2020-02-04
ScalafmtConfig bugfix: read align tokens as seq, not set

### LINKED ISSUE #1309: Scalafmt not idempotent for successive formats from SBT
## Setup

- **Scalafmt version**: 1.6.0-RC4 
- **integration**: SBT (`sbt-scalafmt` SBT plugin)

Assuming that the following configuration is correct:
```
style = defaultWithAlign
align.tokens.add = ["="]
```

Having the following minimal file to be formatted:
```
package playground

case class Human(name: String, age: Int)

object MinimalFile {
  val obj = Human(
    name = "John",
    age = 20
  )
}
```

I expect it to be formatted by Scalafmt like so:
```
package playground

case class Human(name: String, age: Int)

object MinimalFile {
  val obj = Human(
    name = "John",
    age  = 20
  )
}
```

Notice that there has been a single space added after the `age` token, so that equals sign (`=`) is aligned within the case class arguments, namely:
```
name = "John",
age  = 20
```

## Problem

Scalafmt correctly formats the file, as expected, after `scalafmt` command is called from SBT (using `sbt-scalafmt` plugin in the project). The problem is that if I restart SBT and execute `scalafmt` again, the file is brought back to its original version, with `=` sign not aligned within `Human` case class arguments, namely:
```
val obj = Human(
  name = "John",
  age = 20
)
```

The file formatting oscillates between unformatted (`=` not aligned) and formatted version, for every following SBT restart and `scalafmt` command call. File remains unchanged when calling `scalafmt` multiple times, but without restarting SBT.

## Details

I thought that it's related to some SBT cache problem, as it has been mentioned a few times in other issues reported on non-idempotency in scalafmt, but actually, it seems that it's all about current Scalafmt logic.

`ScalafmtConfig` class contains alignment rules as `align` parameter ([source code](https://github.com/scalameta/scalafmt/blob/master/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala#L120)). For the setup mentioned above, `align.tokens` for `ScalafmtConfig` evaluates to:
```
tokens = {HashSet$HashTrieSet@10865} "HashSet$HashTrieSet" size = 15
 0 = {AlignToken@11230} "AlignToken(←,Enumerator.Generator)"
 1 = {AlignToken@11231} "AlignToken(=,.*)"
 2 = {AlignToken@11232} "AlignToken(%,Term.ApplyInfix)"
 3 = {AlignToken@11233} "AlignToken({,Template)"
 4 = {AlignToken@11234} "AlignToken(⇒,Case)"
 5 = {AlignToken@11235} "AlignToken(extends,Defn.(Class|Trait|Object))"
 6 = {AlignToken@11236} "AlignToken(→,Term.ApplyInfix)"
 7 = {AlignToken@11237} "AlignToken(=,(Enumerator.Val|Defn.(Va(l|r)|Def|Type)))"
 8 = {AlignToken@11238} "AlignToken(<-,Enumerator.Generator)"
 9 = {AlignToken@11239} "AlignToken(//,.*)"
 10 = {AlignToken@11240} "AlignToken(->,Term.ApplyInfix)"
 11 = {AlignToken@11241} "AlignToken(%%,Term.ApplyInfix)"
 12 = {AlignToken@11242} "AlignToken(=>,Case)"
 13 = {AlignToken@11243} "AlignToken(},Template)"
 14 = {AlignToken@11244} "AlignToken(%%%,Term.ApplyInfix)"
```

As you can see, `=` alignment token appears twice:
```
1 = {AlignToken@11231} "AlignToken(=,.*)"
(...)
7 = {AlignToken@11237} "AlignToken(=,(Enumerator.Val|Defn.(Va(l|r)|Def|Type)))"
```

Then, when `alignMap` is evaluated ([source code](https://github.com/scalameta/scalafmt/blob/master/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala#L161)), we call:
```
lazy val alignMap: Map[String, Regex] =
  align.tokens.map(x => x.code -> x.owner.r).toMap
```

Because we convert all of the tokens to map, we end up having either `AlignToken(=,.*)` in it (then the file is formatted properly) or `AlignToken(=,(Enumerator.Val|Defn.(Va(l|r)|Def|Type)))` (in this case the formatting from our custom Scalafmt config is not applied).

As `alignMap` is what Scalafmt uses to actually apply the given formatting, we end up having non-idempotent `scalafmt` behavior for every SBT restart.

kitbellew:
That way, we'll preserve ordering and the mapping.
Fixes #1309.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
@@ -58,7 +58,7 @@ import metaconfig.generic.Surface
 case class Align(
     openParenCallSite: Boolean = false,
     openParenDefnSite: Boolean = false,
-    tokens: Set[AlignToken] = Set(AlignToken.caseArrow),
+    tokens: Seq[AlignToken] = Seq(AlignToken.caseArrow),
     arrowEnumeratorGenerator: Boolean = false,
     ifWhileOpenParen: Boolean = true,
     tokenCategory: Map[String, String] = Map(),
@@ -74,7 +74,7 @@ case class Align(
     )
 ) {
   implicit val reader: ConfDecoder[Align] = generic.deriveDecoder(this).noTypos
-  implicit val alignReader: ConfDecoder[Set[AlignToken]] =
+  implicit val alignReader: ConfDecoder[Seq[AlignToken]] =
     ScalafmtConfig.alignTokenReader(tokens)
 }
 
@@ -84,7 +84,7 @@ object Align {
   val none: Align = Align(
     openParenCallSite = false,
     openParenDefnSite = false,
-    tokens = Set.empty,
+    tokens = Seq.empty,
     ifWhileOpenParen = false,
     tokenCategory = Map.empty,
     treeCategory = Map.empty
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/AlignToken.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/AlignToken.scala
@@ -29,7 +29,7 @@ object AlignToken {
       case els => fallbackAlign.reader.read(els)
     }
 
-  val default = Set(
+  val default = Seq(
     caseArrow,
     AlignToken("extends", "Defn.(Class|Trait|Object)"),
     AlignToken("//", ".*"),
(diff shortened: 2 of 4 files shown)


kitbellew:

PR #1675, opened 2020-02-05
Scalafmt is supported by format-all for emacs

sideshowcoder:
A common formatting extension supporting a wide varity of languages.


poslegm:
Thank you! Can you please notify that it's externally maintained integration? Like for [maven plugin](https://scalameta.org/scalafmt/docs/installation.html#maven)

PR #1677, opened 2020-02-06
Rewrites: traverse tree once, applying all rules

kitbellew:
Also, turn rewrites into sessions with the context instance.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Rewrite.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Rewrite.scala
@@ -5,6 +5,7 @@ import scala.collection.mutable
 import metaconfig.ConfCodec
 import scala.meta._
 import scala.meta.tokens.Token.LF
+import scala.meta.transversers.SimpleTraverser
 
 import org.scalafmt.config.ReaderUtil
 import org.scalafmt.config.ScalafmtConfig
@@ -66,7 +67,11 @@ case class RewriteCtx(
 }
 
 abstract class Rewrite {
-  def rewrite(implicit ctx: RewriteCtx): Unit
+  def create(implicit ctx: RewriteCtx): RewriteSession
+}
+
+abstract class RewriteSession(implicit ctx: RewriteCtx) {
+  def rewrite(tree: Tree): Unit
 }
 
 object Rewrite {
@@ -124,7 +129,14 @@ object Rewrite {
       style.runner.dialect(input).parse(style.runner.parser) match {
         case Parsed.Success(ast) =>
           val ctx = RewriteCtx(style, ast)
-          rewrites.foreach(_.rewrite(ctx))
+          val rewriteSessions = rewrites.map(_.create(ctx)).toList
+          val traverser = new SimpleTraverser {
+            override def apply(tree: Tree): Unit = {
+              rewriteSessions.foreach(_.rewrite(tree))
+              super.apply(tree)
+            }
+          }
+          traverser(ast)
           val out = ctx.applyPatches
           input match {
             case Input.File(path, _) =>
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/RewriteSettings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/RewriteSettings.scala
@@ -2,7 +2,7 @@ package org.scalafmt.config
 
 import metaconfig._
 import org.scalafmt.Error.InvalidScalafmtConfiguration
-import org.scalafmt.rewrite.{AvoidInfix, Rewrite}
+import org.scalafmt.rewrite.Rewrite
 
 case class RewriteSettings(
     rules: Seq[Rewrite] = Nil,
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/AvoidInfix.scala
@@ -11,7 +11,15 @@ import scala.meta.tokens.Token.{
 }
 import scala.meta._
 
-case object AvoidInfix extends Rewrite {
+object AvoidInfix extends Rewrite {
+  override def create(implicit ctx: RewriteCtx): RewriteSession =
+    new AvoidInfix
+}
+
+class AvoidInfix(implicit ctx: RewriteCtx) extends RewriteSession {
+
+  private val matcher = ctx.style.rewrite.neverInfix.toMatcher
+
   // In a perfect world, we could just use
   // Tree.transform {
   //   case t: Term.ApplyInfix => Term.Apply(Term.Select(t.lhs, t.op), t.args)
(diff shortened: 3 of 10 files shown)


kitbellew:

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Rewrite.scala
```diff
@@ -66,7 +67,11 @@ case class RewriteCtx(
 }
 
 abstract class Rewrite {
-  def rewrite(implicit ctx: RewriteCtx): Unit
+  def create(implicit ctx: RewriteCtx): RewriteSession
+}
+
+abstract class RewriteSession(implicit ctx: RewriteCtx) {
```
poslegm: What is purpose to make `ctx` implicit?

kitbellew: less code to write, don't need to pass it around. perhaps this is also the result of our general rule that it's a good idea to use implicits for unique instances of a less common type. if you prefer, can make it explicit easily, please let me know.

olafurpg:
This is an interesting change! Do you have any numbers to motivate this optimization?

kitbellew:
> This is an interesting change! Do you have any numbers to motivate this optimization?

@olafurpg i usually compare run times on `scala-repos` when the run results in no changes (so no time is spent on showing diffs). this change seemed to produce a minor, barely perceptible speedup (i used all rewrites except AvoidInfix) but this approach is quite random and not very convincing.

can you help me with setting up of your benchmarking tools? i tried `./bin/run-benchmarks.sh` but it asks to run `sbt test` which fails with `(dynamic / Test / executeTests) java.lang.NoClassDefFoundError: org/scalatest/FunSuiteLike`. i am not very familiar with `sbt` (we use maven) so i don't know how to fix it.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Rewrite.scala
poslegm: I prefer explicit constructors but it doesn't really matter

olafurpg:
The following command will run a benchmark on a single file with ~4k lines of code with some rewrites enabled, 10 warmup iterations and 10 benchmark iterations.

```
benchmarks/jmh:run -i 10 -wi 10 -f1 -t1 org.scalafmt.benchmarks.Micro.ExtraLarge.scalafmt_rewrite
```

You can tweak the `scalafmt_rewrite` method to enable more rules.

olafurpg:
It prints something like this for me
```
[info] # Run progress: 0.00% complete, ETA 00:00:20
[info] # Fork: 1 of 1
[info] # Warmup Iteration   1: 3349.428 ms/op
[info] # Warmup Iteration   2: 1385.353 ms/op
[info] # Warmup Iteration   3: 1323.863 ms/op
[info] # Warmup Iteration   4: 1314.509 ms/op
[info] # Warmup Iteration   5: 1180.080 ms/op
[info] # Warmup Iteration   6: 1119.079 ms/op
[info] # Warmup Iteration   7: 1079.098 ms/op
[info] # Warmup Iteration   8: 1043.373 ms/op
[info] # Warmup Iteration   9: 1031.648 ms/op
[info] # Warmup Iteration  10: 1026.872 ms/op
[info] Iteration   1: 958.440 ms/op
[info] Iteration   2: 955.910 ms/op
[info] Iteration   3: 924.383 ms/op
[info] Iteration   4: 984.063 ms/op
[info] Iteration   5: 1022.945 ms/op
[info] Iteration   6: 989.306 ms/op
[info] Iteration   7: 983.143 ms/op
[info] Iteration   8: 983.382 ms/op
[info] Iteration   9: 965.896 ms/op
[info] Iteration  10: 1016.920 ms/op
[info] Result "org.scalafmt.benchmarks.Micro.ExtraLarge.scalafmt_rewrite":
[info]   978.439 ±(99.9%) 44.063 ms/op [Average]
[info]   (min, avg, max) = (924.383, 978.439, 1022.945), stdev = 29.145
[info]   CI (99.9%): [934.376, 1022.502] (assumes normal distribution)
[info] # Run complete. Total time: 00:00:32
[info] REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
[info] why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
[info] experiments, perform baseline and negative tests that provide experimental control, make sure
[info] the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
[info] Do not assume the numbers tell you what you want them to tell.
[info] Benchmark                          Mode  Cnt    Score    Error  Units
[info] Micro.ExtraLarge.scalafmt_rewrite  avgt   10  978.439 ± 44.063  ms/op
```

kitbellew:
@olafurpg ok, so the results of the benchmark. i set in `FormatBenchmark.scala` the following:
```
  rules = Seq(SortImports, RedundantBraces, PreferCurlyFors, SortModifiers)
```

then ran your command through `sbt` before and after this commit.
* before:
```
[info]   880.639 ±(99.9%) 49.038 ms/op [Average]
[info]   (min, avg, max) = (824.018, 880.639, 941.453), stdev = 32.436
[info]   CI (99.9%): [831.601, 929.677] (assumes normal distribution)
```
* after:
```
[info]   844.943 ±(99.9%) 27.216 ms/op [Average]
[info]   (min, avg, max) = (818.603, 844.943, 869.071), stdev = 18.002
[info]   CI (99.9%): [817.726, 872.159] (assumes normal distribution)
```

then I ran again, in the opposite order:
* after:
```
[info]   741.733 ±(99.9%) 29.887 ms/op [Average]
[info]   (min, avg, max) = (716.586, 741.733, 775.446), stdev = 19.768
[info]   CI (99.9%): [711.846, 771.620] (assumes normal distribution)
```
* before:
```
[info]   761.939 ±(99.9%) 34.677 ms/op [Average]
[info]   (min, avg, max) = (737.170, 761.939, 816.624), stdev = 22.937
[info]   CI (99.9%): [727.262, 796.616] (assumes normal distribution)
```

_some_ improvement (3-5%), although nothing to write home about.

kitbellew:
@olafurpg @poslegm what do you think, should we proceed or abandon? the performance improvement is slight but positive in both runs.

olafurpg:
I’m torn. There is a tradeoff since the new design is more complicated IMO. It’s easy to explain “every rule traverses the entire tree”, and I’m not sure if the speed ups are significant enough to justify the introduced complexity. What do you think?

I haven’t worked on these rules much for a while so I’m ok going ahead with this change if you want to.

kitbellew:
@olafurpg i don't feel too strongly but surprised about the complexity. traverse tree, rewrite each node using a few rules... doesn't sound much different :-)

kitbellew:
honestly, i feel bad. i didn't think this would be too controversial. perhaps we should see what @poslegm thinks.

kitbellew:
Lovely! Thank you, guys.

PR #1679, opened 2020-02-06
RewriteTrailingCommas: trivial, remove duplicate

kitbellew:
One of the patterns was present in both isDefnSite and isCallSite (not
sure why, and which of them should keep the pattern is a separate issue)
so it ended up twice in this class. Remove the second.


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RewriteTrailingCommas.scala
```diff
@@ -36,7 +36,6 @@ case object RewriteTrailingCommas extends Rewrite {
       case t: Ctor.Primary => afterEachArgs(t.paramss)
 
       case t: Term.Apply => afterArgs(t.args)
-      case t: Type.Apply => afterArgs(t.args)
```
kitbellew: the other copy shows up above, line 31. github doesn't allow commenting on that line :)

PR #1681, opened 2020-02-07
ScalaTest -> MUnit

poslegm:
Migration to [MUnit](https://scalameta.org/munit/blog/2020/02/01/hello-world.html).

ScalaTest is present in transitive deps still. We need to update scalameta-testkit after 4.3.2 will be published in Maven Central


scalafmt-tests/src/test/scala/org/scalafmt/ToHoconSuite.scala
```diff
@@ -1,10 +1,6 @@
...
-import metaconfig.Conf
-import metaconfig.ConfEncoder
-import org.scalafmt.config.Config
-import org.scalafmt.util.DiffAssertions
-import org.scalatest.funsuite.AnyFunSuite
+import munit.FunSuite
 
-class ConfDiffSuite extends AnyFunSuite with DiffAssertions {}
-class ToHoconSuite extends AnyFunSuite with DiffAssertions {}
+class ConfDiffSuite extends FunSuite {}
```
poslegm: WTF? Is this classes necessary?

scalafmt-dynamic/src/test/scala/tests/DynamicSuite.scala
```diff
@@ -419,7 +418,7 @@ class DynamicSuite extends AnyFunSuite with DiffAssertions {
     val config = configOpt.get
     assert(config.hasRewriteRules)
     val configWithoutRewrites = config.withoutRewriteRules
-    assert(config !== configWithoutRewrites)
+    assert(config != configWithoutRewrites)
```
olafurpg: There is an assertNotEquals combinator.

scalafmt-tests/src/test/scala-2.12/org/scalafmt/ScalafmtProps.scala
```diff
@@ -33,7 +34,14 @@ class ScalafmtProps extends FormatAssertions {
           case Formatted.Success(formatted) =>
             assertFormatPreservesAst[Source](code, formatted)
             val formattedSecondTime = Scalafmt.format(formatted, config).get
-            assertNoDiff(formattedSecondTime, formatted, "Idempotency")
+            Diffs.assertNoDiff(formattedSecondTime, formatted, diff => {
```
olafurpg: I would ideally like to avoid using internal APIs, what is missing from the public api?

scalafmt-tests/src/test/scala/org/scalafmt/ToHoconSuite.scala
olafurpg: Is there any issue if we remove these?

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
```diff
@@ -1,31 +1,27 @@
...
 import org.scalafmt.internal.FormatWriter
-import org.scalatest.funsuite.AnyFunSuiteLike
 
-trait HasTests extends AnyFunSuiteLike with FormatAssertions {
+import scala.collection.mutable
+import scala.meta.Tree
+import scala.meta.parsers.Parse
+import scala.util.Try
+
+trait HasTests extends FormatAssertions {
```
olafurpg: You can btw mix FunSuite by adding a self parameter “{this: FunSuite => ...”

olafurpg: This means you can only implement this trait from a subclass or FunSuite

scalafmt-tests/src/test/scala/org/scalafmt/ToHoconSuite.scala
poslegm: I think that no. I guess they was created as TODO-markers but were never finished

scalafmt-tests/src/test/scala-2.12/org/scalafmt/ScalafmtProps.scala
poslegm: We need diff report [here](https://github.com/scalameta/scalafmt/pull/1681/files/0231ae14725abf8f68e36ecc3d8abcbbcd688e1a#diff-44d568678882dc62a2d6ff89d804a8eeR77). I will try to extract it from `FailException`'s [message](https://github.com/scalameta/munit/blob/master/munit/shared/src/main/scala/munit/Assertions.scala#L55)

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
poslegm: Reasonable constraint

scalafmt-tests/src/test/scala/org/scalafmt/ToHoconSuite.scala
olafurpg: I think those test suites got moved upstream into the metaconfig repo. We can remove them here.

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
poslegm: No, we can do it now, because of `ManualTests` and `UnitTests` [objects](https://github.com/poslegm/scalafmt/blob/a90ca60f778c48fd1fd13c312500c5a4b2f16a3c/scalafmt-tests/src/test/scala/org/scalafmt/ManualTests.scala#L8) which extends this trait.

So, we need to small redesign this part of code before introducing `this: FunSuite =>` here. I think it's not necessarily now 😄

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
```diff
@@ -457,15 +454,15 @@ trait CliTestBehavior { this: AbstractCliTest =>
       val conf = Cli.getConfig(args, opts)
       Cli.run(conf.get)
 
-      assertNoDiff(root / "scalatex.scalatex", unformatted)
-      assertNoDiff(root / "sbt.sbtfile", sbtOriginal)
+      assertNoDiff(dir2string(root / "scalatex.scalatex"), unformatted)
```
kitbellew: what does `dir2string` do?

this file has a few similar changes (like `size` to `length`). are they related to munit change or just useful cleanup?

scalafmt-tests/src/test/scala-2.12/org/scalafmt/ScalafmtProps.scala
poslegm: done

scalafmt-dynamic/src/test/scala/tests/DynamicSuite.scala
poslegm: done

scalafmt-tests/src/test/scala/org/scalafmt/ToHoconSuite.scala
poslegm: done

scalafmt-tests/src/test/scala-2.12/org/scalafmt/ScalafmtProps.scala
```diff
@@ -33,7 +34,14 @@ class ScalafmtProps extends FormatAssertions {
...
+              throw DiffFailure(
+                "Idempotency",
+                formatted,
+                formattedSecondTime,
+                diff
+              )
+            }, "Idempotency")
```
kitbellew: optional: should we at some point rename it to "idempotencE"? https://en.wikipedia.org/wiki/Idempotence

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
```diff
@@ -153,37 +149,6 @@ trait HasTests extends AnyFunSuiteLike with FormatAssertions {
 
   def ignore(t: DiffTest): Boolean = false
 
-  def runTest(run: (DiffTest, Parse[_ <: Tree]) => Unit)(t: DiffTest): Unit = {
```
kitbellew: looks like `CanRunTests` has now become what `HasTests` used to be, and `HasTests` has smaller role. is this related to munit change or just useful refactor?

scalafmt-tests/src/test/scala-2.12/org/scalafmt/ScalafmtProps.scala
```diff
@@ -34,14 +34,16 @@ class ScalafmtProps extends FunSuite with FormatAssertions {
...
             val formattedSecondTime = Scalafmt.format(formatted, config).get
-            Diffs.assertNoDiff(formattedSecondTime, formatted, diff => {
-              throw DiffFailure(
-                "Idempotency",
-                formatted,
-                formattedSecondTime,
-                diff
-              )
-            }, "Idempotency")
+            try assertNoDiff(formattedSecondTime, formatted, "Idempotency")
```
kitbellew: looks like munit is not a third-party package. would it make sense to display the diff as part of `assertNoDiff` and make the change in munit?

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
```diff
@@ -767,8 +764,8 @@ trait CliTestBehavior { this: AbstractCliTest =>
 }
 
 class CliTest extends AbstractCliTest with CliTestBehavior {
-  testsFor(testCli("1.6.0-RC4")) // test for runDynamic
-  testsFor(testCli(Versions.version)) // test for runScalafmt
+  testCli("1.6.0-RC4") // test for runDynamic
```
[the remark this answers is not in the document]
poslegm: Looks like [nothing](https://github.com/scalatest/scalatest/blob/3.1.x/scalatest/src/main/scala/org/scalatest/funsuite/AnyFunSuiteLike.scala#L239-L258)

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
poslegm: > what does dir2string do?

[Reads](https://github.com/poslegm/scalafmt/blob/5ac6a99f179e79b5f7be80ad9b84ef91dc5b6e88/scalafmt-tests/src/test/scala/org/scalafmt/cli/FileTestOps.scala#L36) files from directory. (One file in this case)

Old `assertNoDiff` implementation from scalafmt did it itself, but MUnit doesn't know this custom logic.

All of changes except size -> length is MUnit related.

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
poslegm: MUnit related. It's about `FunSuite` inheritance constraints.

scalafmt-tests/src/test/scala-2.12/org/scalafmt/ScalafmtProps.scala
poslegm: It will be good! This is just work around api limitations.

scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
```diff
@@ -2,39 +2,44 @@ package org.scalafmt.util
...
-    val absFile = AbsoluteFile.fromPath(f.toString).get
-    val ops = new GitOpsImpl(absFile)
-    init(ops)
-    // initial commit is needed
-    val initF = touch("initialfile")(ops)
-    add(initF)(ops)
-    commit(ops)
-    try withFixture(test.toNoArgTest(ops))
-    finally deleteTree(f)
+  private val ops: Fixture[GitOpsImpl] = new Fixture[GitOpsImpl]("gitOpsImpl") {
```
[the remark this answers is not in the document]
poslegm: ops created once on test suite initialization. [More about MUnit fixtures](https://scalameta.org/munit/docs/fixtures.html#reusable-test-local-fixtures)

scalafmt-tests/src/test/scala-2.12/org/scalafmt/ScalafmtProps.scala
poslegm: done

poslegm:
@kitbellew 

> why munit is more appropriate than scalatest for this repo

1. As you have already noticed, this is a migration from third-party library to scalameta project
2. I liked MUnit concept and wanted to try it on real project. It was a small stress-test for the new library which revealed some issues

scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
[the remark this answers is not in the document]
poslegm: Yes, good catch. Rewritten

kitbellew:
@poslegm perhaps i am doing something wrong, but Intellij no longer runs any tests when i invoke `FormatTests`. just says `Process finished with exit code 0`.

how is it actually working? there are no `test(...) { ... }` blocks in `FormatTests`, not sure how `munit` figures out what to run. Tried changing
```
  tests
    .sortWith(bySpecThenName)
    .withFilter(testShouldRun)
    .foreach(runTest(run))
```
to
```
  tests
    .sortWith(bySpecThenName)
    .foreach { t =>
      Console.err.println(t.fullName)
      if (testShouldRun(t))
        super.test(t.fullName) { runTest(run)(t) }
    }
```
but nothing was displayed on the console.

help? :)

poslegm:
@kitbellew 

I've never launched FormatTests from IDEA 😕
sbt command `tests/testOnly *FormatTests` works as before. Tests execution located [here](https://github.com/scalameta/scalafmt/blob/2e13b7ce4e08aaa28c90bfa0a86a5aaf91937cce/scalafmt-tests/src/test/scala/org/scalafmt/util/CanRunTests.scala#L18).

If `FormatTests` have worked in IDEA with ScalaTest, I will put some time to debugging.

kitbellew:
> If `FormatTests` have worked in IDEA with ScalaTest, I will put some time to debugging.

yes, definitely worked :) that's how i use it. if you could fix it, i'd really appreciate it.

@olafurpg says `munit` should work in Intellij because it looks like `junit` but perhaps something else `junit`-compatible is missing in our `FormatTests`.

olafurpg:
While we resolve the integration issues with IntelliJ we may want to revert this PR. No pressure from me to use MUnit in Scalafmt.

poslegm:
Ok, I found the problem:
https://github.com/scalameta/scalafmt/blob/2e13b7ce4e08aaa28c90bfa0a86a5aaf91937cce/scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala#L44-L45

When running `FormatTests` with sbt absolute path for tests directory is `/Users/poslegm/Documents/scalafmt/scalafmt-tests/src/test/resources`.
When running with IDEA it is 
`/Users/poslegm/Documents/scalafmt/scalafmt-tests/scalafmt-tests/src/test/resources` (incorrect path).

poslegm:
Fixed for me: https://github.com/scalameta/scalafmt/pull/1692

kitbellew:
> Fixed for me: #1692

thank you, i tried it, and it works *better* than before. however, it looks like it's running only a subset of tests and then stops. not sure if that's the expected behaviour.

kitbellew:
@olafurpg is there any way to navigate, in intellij, to the location of the failing test? i tried adding munit.Location with the `.stat` file and appropriate line to various places after @poslegm's commits, but nothing "clickable" appears (except the location of the `FailException` instantiation).

poslegm:
@kitbellew https://github.com/scalameta/munit/issues/48 

I will revert tests back to ScalaTest tomorrow. There is two annoying IDEA issues which interfere with work and need to be resolved before complete migration. I will try to deal with them when I have free time.

olafurpg:
MUnit knows the location of failed tests (we can even make it work for *.stat files) but I haven’t figured out yet how to communicate it to IntelliJ  (for example via JUnit test annotations)

PR #1686, opened 2020-02-09
Self formatting with 2.4.0-RC2

poslegm:
Just applied last RC to source code

Is it missing from release notes that «Format single-line curly lambdas using parens» works only with `RedundantBraces` rule?

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version=2.3.2
+version=2.4.0-RC2
 project.git = true
 project.excludeFilters = [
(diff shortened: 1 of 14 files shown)


kitbellew:
> Is it missing from release notes that «Format single-line curly lambdas using parens» works only with `RedundantBraces` rule?

I will add, thanks for noticing.

poslegm:
> some benchmark file is still to be formatted. latent idempotence problem?

No, I just incorrectly runs the formatter. Everything is good now

PR #1688, opened 2020-02-09
TreeOps: find a parent, avoid getting all parents

kitbellew:


kitbellew:

PR #1689, opened 2020-02-09
Router: extract simple T.Dot patterns before chain

kitbellew:
Let's extract the simple Dot patterns from a more complex select chain rule, as all the run-up ultimately becomes unnecessary and makes it a little less readable.


kitbellew:

kitbellew:
> It would be helpful to add some description to PR, even if the change is small enough :)

thank you, done.

PR #1691, opened 2020-02-09
Policy: add apply() with partial func in 2nd group

kitbellew:
This makes instantiating policies with the partial function look a bit
cleaner.


kitbellew:

PR #1692, opened 2020-02-09
Resource folder path resolution from classloader

poslegm:
I think it's more reliable way to get resources. Should fix issue https://github.com/scalameta/scalafmt/pull/1681#issuecomment-583871622

### DIFF
--- a/scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
+++ b/scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
@@ -41,8 +41,7 @@ trait HasTests extends FormatAssertions {
     }
   )
   lazy val debugResults = mutable.ArrayBuilder.make[Result]
-  val testDir =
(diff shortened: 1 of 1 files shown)


poslegm:
It's correct behaviour, because there is 1270 format tests. IDEA renders it wrong: https://github.com/scalameta/munit/issues/47

Ok, we can revert MUnit migration if it interferes with work on scalafmt. IDEA issues looks confusing.

kitbellew:
@poslegm if the issue is simply that intellij renders it wrong, i don't care. just wanted to be sure it runs all tests, so that my continued messing with the codebase doesn't introduce bugs accidentally. :)

olafurpg:
I don’t want the MUnit migration to regress the experience of testing via IntelliJ so please feel free to revert back to ScalaTest. Since MUnit is just a JUnit runner I think the IntelliJ testing UX can be amazingly good but it’s not there yet.

PR #1696, opened 2020-02-10
VerticalMultiline: check config style

### LINKED ISSUE #1362: Conflict between new line before / after implicit and verticalMultilineAtDefinitionSite?
- **Version**: 2.0.0-RC4
- **Integration**: CLI
- **Configuration**:
```
maxColumn = 80
danglingParentheses = true

```

## Steps

Given code like this:
```scala
case class Foo(
  a: Int,
  b: Boolean,
  c: String,
  d: String,
  e: String,
  f: String,
  g: String
) {
  def run[F[_]](
  )(implicit
    F: Monad[F]
  ): F[Resolution] =
    ???
}
```

## Problem

```scala
case class Foo(
  a: Int,
  b: Boolean,
  c: String,
  d: String,
  e: String,
  f: String,
  g: String) {
  def run[F[_]](
  )(implicit
    F: Monad[F]
  ): F[Resolution] =
    ???
}
```

## Expectation

I would like
```scala
  f: String,
  g: String
) {
```
```scala
  f: String,
  g: String) {
```

## Workaround

Setting
```
```
```scala
  )(implicit
```
to
```scala
  )(
    implicit
```

### LINKED ISSUE #1539: verticalMultiline disables optIn.configStyleArguments

kitbellew:
If `optin.configStyleArguments` is set, incorporate the setting into the vertical multiline logic.

For the purposes of this integration, we'll assume that config style means there's a line break after the opening parenthesis (without requiring dangling, or a line break before the closing one), and `optIn.configStyleArguments` is set.

Fixes #1539.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -986,7 +986,7 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
       if (isBracket) close // If we can fit the type params, make it so
       else lastParen // If we can fit all in one block, make it so
 
-    val maxArity = valueParamsOwner match {
+    def maxArity = valueParamsOwner match {
       case d: Decl.Def if d.paramss.nonEmpty => d.paramss.map(_.size).max
       case d: Defn.Def if d.paramss.nonEmpty => d.paramss.map(_.size).max
       case m: Defn.Macro if m.paramss.nonEmpty => m.paramss.map(_.size).max
@@ -995,12 +995,16 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
       case _ => 0
     }
 
+    def configStyle =
+      style.optIn.configStyleArguments &&
+        style.activeForEdition_2020_03 && ft.newlinesBetween != 0
+
     def belowArityThreshold =
       maxArity < style.verticalMultiline.arityThreshold
 
     Seq(
       Split(Space(style.spaces.inParentheses), 0)
-        .onlyIf(isBracket || belowArityThreshold)
+        .onlyIf(isBracket || !configStyle && belowArityThreshold)
         .withPolicy(SingleLineBlock(singleLineExpire)),
       Split(Newline, 1) // Otherwise split vertically
         .withIndent(firstIndent, close, Right)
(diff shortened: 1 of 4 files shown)


kitbellew:
@olafurpg @tanishiking @poslegm here's an attempt to combine vertical multiline with other options. Let me know if you find the result acceptable. I haven't been able to find a formal definition of config style, so the result might be incorrect.

kitbellew:
> I don't know how to make it easy yet (enumeration instead of boolean flags may be?). So, I am OK with it, but it would be ideal to simplify this.

there are a couple of things we can do with existing `verticalMultiline`-specific parameters: 
* move the `excludeDanglingParens` configuration parameter into regular `DanglingParentheses`, combining with `danglingParentheses.defnSite` (and possibly implement the difference in logic)
* move `newlineBeforeImplicitKW` and `newlineAfterImplicitKW` into regular `newlines` block and implement the non-vertical-multiline logic for them
* get rid of the `newlineAfterOpenParen` in favor of `configStyleArguments`, although i still don't know clearly what *config style* means

poslegm:
Sorry for long delay

> move the excludeDanglingParens configuration parameter into regular DanglingParentheses, combining with danglingParentheses.defnSite (and possibly implement the difference in logic)

good!

> move newlineBeforeImplicitKW and newlineAfterImplicitKW into regular newlines block and implement the non-vertical-multiline logic for them

good!

> get rid of the newlineAfterOpenParen in favor of configStyleArguments, although i still don't know clearly what config style means

Hmm, not sure about it

kitbellew:
> > move newlineBeforeImplicitKW and newlineAfterImplicitKW into regular newlines block and implement the non-vertical-multiline logic for them
> 
> good!
 #1749

kitbellew:
> > move the excludeDanglingParens configuration parameter into regular DanglingParentheses, combining with danglingParentheses.defnSite (and possibly implement the difference in logic)
> 
> good!
#1761

kitbellew:
@poslegm uploaded the remaining modification, after moving parts into #1749 and #1761.

PR #1697, opened 2020-02-11
MUnit reverted to ScalaTest

poslegm:
Until https://github.com/scalameta/munit/issues/47, https://github.com/scalameta/munit/issues/48 resolution


poslegm:
I think that https://github.com/scalameta/scalafmt/pull/1692 is valuable change anyway, so it doesn't need to be reverted

PR #1698, opened 2020-02-11
Tests: determine test's location, provide to Intellij

kitbellew:
While this approach will not make the test case itself clickable, any failures (which Intellij's runner will mark with `ScalaTestFailureLocation` and then parse to provide a clickable `... at (<link>)`) will have a link which will open the test file on the right line.


kitbellew:

PR #1701, opened 2020-02-12
A couple of bugfixes in handling of first arg after "("

### LINKED ISSUE #1703: scalafmt crashes with head of empty list error on anonymous SAM
- **Configuration**:
```
version = "2.4.0-RC2"
maxColumn = 120 
trailingCommas = always
  until
  to
  by
  eq
  ne
  "should.*"
  "contain.*"
  "must.*"
  in
  ignore
  be
  taggedAs
  thrownBy
  synchronized
  have
  when
  size
  only
  noneOf
  oneElementOf
  noElementsOf
  atLeastOneElementOf
  atMostOneElementOf
  allElementsOf
  inOrderElementsOf
  theSameElementsAs
  // these are new
  throws
  returns
  satisfy
]


```

## Steps

Given code like this:
```scala
object Foo {
  def foo(): Unit = ()
  
  def run(): Thread = {
    new Thread(() => foo())
  }
}
```

```bash
scalafmt
```

## Problem

Scalafmt formats produces the error 
```
head of empty list: [...]/Foo.scala
```

## Workaround

This formats fine: 

```scala
object Foo {
  def foo(): Unit = ()

  def run(): Thread = {
    val r: Runnable = () => foo()
    new Thread(r)
  }
}
```

kitbellew:
Fixes #1703 

`scala-repos` diffs:
* Router bugfix: split, indent comments consistently https://github.com/kitbellew/scala-repos/commit/93a939860e153783b1e3e6d79005d0ef2381a386
* TreeOps bugfix: find Init param group correctly https://github.com/kitbellew/scala-repos/commit/68cb5977aadcfc68466ebfff49e4e81e966f4987?w=1
* Router bugfix: consistently allow break on "(/**/" https://github.com/kitbellew/scala-repos/commit/3fd4a1a7241282b43aed2615b5383bc0f4588511?w=1

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatToken.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatToken.scala
@@ -1,5 +1,6 @@
 package org.scalafmt.internal
 
+import scala.meta.Tree
 import scala.meta.tokens.Token
 
 import org.scalafmt.util.TokenOps._
@@ -43,7 +44,9 @@ object FormatToken {
     */
   case class Meta(
       between: Array[Token],
-      idx: Int
+      idx: Int,
+      leftOwner: Tree,
+      rightOwner: Tree
   )
 
 }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
@@ -57,9 +57,9 @@ class BestFirstSearch(
     noOptimizations.contains(token.left)
   }
 
-  def getLeftLeft(curr: State): Token = {
-    tokens(Math.max(0, curr.depth - 1)).left
-  }
+  @inline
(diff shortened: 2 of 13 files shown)


kitbellew:

kitbellew:
@olafurpg @tanishiking @poslegm would appreciate your insights. i would prefer not to release v2.4.0 without these fixes (follow-on to changes included in this release). thanks!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatToken.scala
```diff
@@ -43,7 +44,9 @@ object FormatToken {
     */
   case class Meta(
       between: Array[Token],
-      idx: Int
+      idx: Int,
+      lowner: Tree,
+      rowner: Tree
```
tanishiking: What about `leftOwner`, `rightOwner` instead?
Personally, I prefer explicit names in Scala.

kitbellew: no problem, will do.

PR #1705, opened 2020-02-15
Trailing commas bugfix: can't have trailing comma in `preserve` without a break

kitbellew:
Follow-on to #1669


kitbellew:
@olafurpg @tanishiking @poslegm PTAL. this is another bugfix for a change which would be included in v.2.4.0, would prefer not to release v2.4.0 without it.

PR #1710, opened 2020-02-15
TreeOps: structure the details of defns or calls

kitbellew:
Prepare for a number of fixes which require a better understanding of a definition.

Also, we can vastly simplify the RewriteTrailingCommas logic immediately, and will be able to improve the FormatOps.getApplyArgs logic later, allowing us to tuck implicit in multi-arg calls or apply trailing commas to calls with a given arity.

No changes to tests or `scala-repos`.


kitbellew:

PR #1722, opened 2020-02-18
FormatOps: in one-arg-per-line, handle lbrace case

### LINKED ISSUE #1599: Unable to format file due to bug in Scalafmt
- **Version**: 2.3.2
- **Integration**: cli
- **Configuration**: default
```
version = 2.3.1
```

## Steps

Given code like this:
```scala
object A {
  def foo() = {
    x.map(_.y(
      abcd,
      { case ((abcdefghij, aswbasdfaw), asfda) => aswdf},
      { case (abcdefghij, sadfasdass) => (asdfa.sadvfs(abcdefghij).get, asdfasdfasfdasda.asdfas(asdfasdaas).get) },
      foo
    ))
  }
}
```

```bash
scalafmt
```

## Problem

```scala
Caused by: org.scalafmt.Error$PreciseIncomplete: minimized.scala:6: error: Unable to format file due to bug in scalafmt
      { case (abcdefghij, sadfasdass) => (asdfa.sadvfs(abcdefghij).get, asdfasdfasfdasda.asdfas(asdfasdaas).get) },
                                                                                                           ^
```

## Expectation

I would like scalafmt to format the file successfully.

kitbellew:
The `scalafmt` error described in #1599 turns out to be mainly due to the rule which formats multiple method arguments one per line. Router would attempt to format `, {` combination as a newline followed by the entire block on a single line, or a space without additional rules instead.

Two problems with this approach:
* because of some aspect of `BestFirstSearch` algorithm that I don't yet fully understand, having these two possible splits led to the search reaching a dead end before it attempted the space split 
  * the expectation was that having additional splits could only lead to a *search state exploded* error, not this)
  * removing the newline split (leaving the space split only) solved the problem
* also, in cases when the space split was chosen, the result wasn't guaranteed to have one argument per line

Instead:
* force a newline for regular or partial functions
* for all other blocks, modify the space rule to force breaks after `{` and before `}`.

Fixes #1599.

`scala-repos` diffs: https://github.com/kitbellew/scala-repos/commit/e2042d04c44c988e50189ff1575e0f68fbeef445?w=1

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -389,10 +389,14 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
       case Decision(t @ FormatToken(_: T.Comma, right, _), splits)
           if owner == t.meta.leftOwner &&
             // TODO(olafur) what the right { decides to be single line?
-            !right.is[T.LeftBrace] &&
             // If comment is bound to comma, see unit/Comment.
             (!right.is[T.Comment] || t.newlinesBetween != 0) =>
-        splits.filter(_.modification.isNewline)
+        if (!right.is[T.LeftBrace])
+          splits.filter(_.modification.isNewline)
+        else if (!style.activeForEdition_2020_03)
+          splits
+        else
+          SplitTag.OneArgPerLine.activateOnly(splits)
     }
   }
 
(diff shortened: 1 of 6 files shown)


kitbellew:
@olafurpg would appreciate some insight on the behaviour of `BestFirstSearch` algorithm. why would it stop searching when there are two splits for a given token, but not when one of them is removed? i had been under impression that more splits might potentially lead to explosion of search state, not to pruning of the search tree.

kitbellew:

PR #1723, opened 2020-02-18
RedundantBraces: add parensForOneLineApply flag

kitbellew:
By default, it's enabled in edition 2020-01, just as before, but can be
disabled if necessary, as has been requested by several users.


kitbellew:

PR #1728, opened 2020-02-21
Various cleanup and refactoring changes in BestFirstSearch

kitbellew:
Removes some unused code or unnecessary complexity. Doesn't change formatting.

Precursor to the algorithm fix which addresses a host of "unable to format due to bug in scalafmt" issues, to be submitted after.


kitbellew:

PR #1729, opened 2020-02-21
BestFirstSearch: don't give up too soon

### LINKED ISSUE #448: Unable to format due to bug in scalafmt
``` scala
<<< #805
object Foo {
  code.collect {
    case d: Defn.Def if (d match {
      case _ => false


    }) &&
      d=>
  }
}
>>>
object Foo {
  try proc.waitFor()
  catch { case e: Throwable => run() }
}
```

### LINKED ISSUE #1462: Unable to format file due to bug in scalafmt
- **Version**: 2.0.0
- **Integration**: sbt-scalafmt 2.0.2
- **Configuration**:
```
version = 2.0.0-RC8

align = none
align.arrowEnumeratorGenerator = false
align.tokens = []
binPack.literalArgumentLists = true
danglingParentheses = true
docstrings = ScalaDoc
encoding = "UTF-8"
lineEndings = unix
maxColumn = 110
newlines.alwaysBeforeElseAfterCurlyIf = false
optIn.breakChainOnFirstMethodDot = true
project.git = true
rewrite.redundantBraces.methodBodies = false
rewrite.rules = [ExpandImportSelectors, PreferCurlyFors, RedundantBraces, RedundantParens]
verticalMultiline.arityThreshold = 3
verticalMultiline.excludeDanglingParens = []
verticalMultiline.newlineAfterOpenParen = true
verticalMultiline.newlineBeforeImplicitKW = true

```

## Steps

Given code like this:
```scala
object Test {

  def create(
    somethingInfo: foo.T = Something("my-something", Some("test-something"), "something", Some(Format("something", "something")),
      Some(something.Something("something", "something", "something", "something")))
  ): Unit = {
    w(foo)
  }

}

```

```bash
sbt scalafmtAll
```

## Problem

Scalafmt fails with:
```scala
[error] /Users/[...]/ScalafmtBug.scala:6: error: Unable to format file due to bug in scalafmt
[error]   ): Unit = {
[error]             ^: [...]/ScalafmtBug.scala
[error] (Test / scalafmt) /Users/[...]/ScalafmtBug.scala:6: error: Unable to format file due to bug in scalafmt
[error]   ): Unit = {
[error]             ^: [...]/ScalafmtBug.scala
[error] Total time: 0 s, completed Jul 23, 2019 2:55:45 PM

```

## Expectation

That this can be formatted.

## Notes

Sorry for a bad issue title, I don't know exactly what causes this as it seems to be a combination of things.

I had a hard time making a minimal reproduction as a lot of seemingly innocuous changes caused formatting to succeed.

### LINKED ISSUE #1485: Unable to format file due to bug in scalafmt when configStyleArguments=false
- **Version**: 2.0.0
- **Integration**: IntelliJ, sbt
- **Configuration**:
```
version = 2.0.0
style = IntelliJ
maxColumn = 120
docstrings = ScalaDoc
```

## Steps

error log `Exception in thread "main" org.scalafmt.Error$PreciseIncomplete: <input>:16: error: Unable to format file due to bug in scalafmt
        )` on save or run `scalafmt` in sbt.
        
origin code:
```scala
class A {
  def foo: B =
    a.b() { c =>
        val d = c.add(
          D[E](
            1, {
              case conditionNameShouldLongEnough =>
                foo1(objectNameShouldLongEnough, objectNameShouldLongEnough)
                0
              case conditionNameShouldLongEnough =>
                foo1(objectNameShouldLongEnough, objectNameShouldLongEnough)
                1
            }
          )
        )
      }
      .f("f")
}
```

Here is my demo to call internal api of scalafmt, it is proved that this bug is introduced by configuring `optIn.configStyleArguments` to `false`. style `Intellij` include this configuration, so using config file I paste above also has this issue.
```scala
import org.scalafmt.Scalafmt
import org.scalafmt.config.ScalafmtConfig.default
import org.scalafmt.config.{ContinuationIndent, Docstrings, ScalafmtConfig}

object Test extends App {
  val code =
    """
      |class A {
      |  def foo: B =
      |    a.b() { c =>
      |        val d = c.add(
      |          D[E](
      |            1, {
      |              case conditionNameShouldLongEnough =>
      |                foo1(objectNameShouldLongEnough, objectNameShouldLongEnough)
      |                0
      |              case conditionNameShouldLongEnough =>
      |                foo1(objectNameShouldLongEnough, objectNameShouldLongEnough)
      |                1
      |            }
      |          )
      |        )
      |      }
      |      .f("f")
      |}
    """.stripMargin
  println(code)
  val config = ScalafmtConfig.default.copy(
    version = "2.0.0",
    docstrings = Docstrings.ScalaDoc,
    maxColumn = 120,
    continuationIndent = ContinuationIndent(2, 2),
    optIn = default.optIn.copy(
      configStyleArguments = false
    ))

  val formattedCode = Scalafmt.format(code,style = config,range = Set.empty).get
  println(formattedCode)
}
```

### LINKED ISSUE #1599: Unable to format file due to bug in Scalafmt

### LINKED ISSUE #1717: `Unable to format file due to bug in scalafmt` with 2.4.0 & 2.4.1
- **Version**: 2.4.0, 2.4.1
- **Integration**: sbt-scalafmt
- **Configuration**:
```
version = 2.4.1
maxColumn = 100
```

## Steps

Given code like this:
```scala
class Bar {
  def foo(request: Request): RpcResult[Response] = {
    repository
      .flatMap { campaigns =>
        request match {
          case Nil => connection.pure(Right(toResponse(campaigns, Map.empty)))
          case _ =>
            getProfileEmails(request.securityContext, profileIds).to[ConnectionIO].map {
              profileEmails =>
                Right(toResponse(campaigns, profileEmails))
            }
        }
      }.transact(managerTransactor)
  }
}
```

```bash
sbt scalafmtAll
```

## Problem

Scalafmt fails:
```
Bar.scala:12: error: Unable to format file due to bug in scalafmt
```

## Expectation

I would like the file to be reformatted

## Workaround

Stick to 2.3.2

kitbellew:
Currently, the algorithm would clear the search priority queue when the top candidate points to a new statement. Sometimes this leads to losing a viable path and the dreaded "Unable to format due to bug" error.

Rather than clearing the queue, let's maintain several generations of priority queues and create a new generation on a new statement, thus continuing to process subsequent states just like the old algorithm did but with the previously discarded states available as backup.

Apply this modification to top-level invocations only, so that nested calls will still terminate early; thus the backup candidates are really processed at the very end of the existing logic.

Fixes #448. Fixes #1462. Fixes #1485. Fixes #1599. Fixes #1717.

NB: `scala-repos` diffs, looks like they contain files previously unformatted by `scalafmt`: https://github.com/kitbellew/scala-repos/commit/0876f01369b34584e5efa04cdc29b88a93ff1a1d?w=1

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
@@ -18,7 +18,7 @@ import org.scalafmt.util.TreeOps
 /**
   * Implements best first search to find optimal formatting.
   */
-class BestFirstSearch(
+private class BestFirstSearch private (
     val formatOps: FormatOps,
     range: Set[Range],
     formatWriter: FormatWriter
@@ -180,8 +180,6 @@ class BestFirstSearch(
             .foreach(Q.enqueue(_))
         else if (escapeInPathologicalCases &&
           visits(curr.depth) > maxVisitsPerToken) {
-          Q.clear()
-          best.clear()
           runner.event(CompleteFormat(explored, deepestYet))
           throw SearchStateExploded(
             deepestYet,
@@ -268,3 +266,14 @@ class BestFirstSearch(
 }
 
 case class SearchResult(state: State, reachedEOF: Boolean)
+
+object BestFirstSearch {
+
+  def apply(
+      formatOps: FormatOps,
+      range: Set[Range],
+      formatWriter: FormatWriter
+  ): SearchResult =
+    new BestFirstSearch(formatOps, range, formatWriter).getBestPath
+
+}
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
@@ -133,11 +133,18 @@ private class BestFirstSearch private (
       depth: Int = 0,
       maxCost: Int = Integer.MAX_VALUE
   ): State = {
-    val Q = new mutable.PriorityQueue[State]()
-    var result: State = null
+    def newGeneration = new mutable.PriorityQueue[State]()
+    var Q = newGeneration
+    var generations: List[mutable.PriorityQueue[State]] = Nil
+    def addGeneration =
+      if (Q.nonEmpty) {
+        generations = Q :: generations
+        Q = newGeneration
+      }
     Q += start
+
     // TODO(olafur) this while loop is waaaaaaaaaaaaay tooo big.
-    while (Q.nonEmpty) {
+    while (true) {
       val curr = Q.dequeue()
       explored += 1
       runner.event(Explored(explored, depth, Q.size))
@@ -152,9 +159,10 @@ private class BestFirstSearch private (
       if (null == splitToken ||
         splitToken.left.start >= stop.start &&
         splitToken.left.start < splitToken.left.end) {
-        result = curr
-        Q.clear()
-      } else if (shouldEnterState(curr)) {
+        return curr
+      }
+
+      if (shouldEnterState(curr)) {
         val style = styleMap.at(splitToken)
         if (curr.depth > deepestYet.depth) {
           deepestYet = curr
@@ -168,7 +176,8 @@ private class BestFirstSearch private (
             dequeueOnNewStatements &&
             dequeueSpots.contains(tokenHash) &&
             (depth > 0 || !isInsideNoOptZone(splitToken)))
-            Q.clear()
+            if (depth == 0) addGeneration
+            else Q.clear()
         }
 
         val blockClose = shouldRecurseOnBlock(splitToken, stop, style)
@@ -230,8 +239,18 @@ private class BestFirstSearch private (
           }
         }
       }
+
+      if (Q.isEmpty) {
+        if (generations.isEmpty)
+          return null
+
+        Q = generations.head
+        generations = generations.tail
+      }
     }
-    result
+
+    // unreachable
+    null
   }
 
   def getBestPath: SearchResult = {
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
@@ -105,12 +105,6 @@ private class BestFirstSearch private (
     state.column << 8 | state.indentation
   }
 
-  def hasReachedEof(state: State, stop: Token): Boolean =
-    state.depth >= tokens.length || {
-      val token = tokens(state.depth)
-      token.left.start >= stop.start && token.left.end > token.left.start
-    }
-
   val memo = mutable.Map.empty[(Int, StateHash), State]
 
   def shortestPathMemo(start: State, stop: Token, depth: Int, maxCost: Int)(
@@ -122,7 +116,7 @@ private class BestFirstSearch private (
     else {
       // Only update state if it reached stop.
       val nextState = shortestPath(start, stop, depth, maxCost)
-      if (tokens(nextState.depth).left != stop) None
+      if (null == nextState) None
       else {
         memo.update(key, nextState)
         Some(nextState)
@@ -140,7 +134,7 @@ private class BestFirstSearch private (
       maxCost: Int = Integer.MAX_VALUE
   ): State = {
     val Q = new mutable.PriorityQueue[State]()
-    var result = start
+    var result: State = null
     Q += start
     // TODO(olafur) this while loop is waaaaaaaaaaaaay tooo big.
     while (Q.nonEmpty) {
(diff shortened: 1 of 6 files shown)


kitbellew:
@olafurpg the change is simple enough but likely requires your level of understanding of the existing algorithm to review, PTAL.

@tanishiking @poslegm fyi.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
```diff
@@ -139,12 +139,13 @@ class BestFirstSearch(
       depth: Int = 0,
       maxCost: Int = Integer.MAX_VALUE
   ): State = {
-    val Q = new mutable.PriorityQueue[State]()
+    val Q = new mutable.PriorityQueue[(Int, State)]()
```
olafurpg: I just noticed the commit here https://github.com/scalameta/scalafmt/commit/2cd7a731259df9bfbb76d3b3cc4ed1f5d24f076e#diff-0944a127ca40d4a12020bbe4e69623cb

The ordering comparison method is quite performance sensitive so I think it's better to try and avoid redundant allocations and implement the ordering logic on `Integer.compare(...)` like it was done before.

We might also want to encode the generation number as a field on `State` and update the ordering logic.

kitbellew: > I just noticed the commit here [2cd7a73#diff-0944a127ca40d4a12020bbe4e69623cb](https://github.com/scalameta/scalafmt/commit/2cd7a731259df9bfbb76d3b3cc4ed1f5d24f076e#diff-0944a127ca40d4a12020bbe4e69623cb)
> 
> The ordering comparison method is quite performance sensitive so I think it's better to try and avoid redundant allocations and implement the ordering logic on `Integer.compare(...)` like it was done before.

i will try to figure out how to profile the code, but I'm fairly certain that using `Integer.valueOf` (which turned primitives into boxed objects) must have had a cost.
> 
> We might also want to encode the generation number as a field on `State` and update the ordering logic.

let's try.

p.s. one unrelated question, regarding performance. there's a memo lookup there, I'm guessing the dp part of the algorithm, on the column/indentation of the intermediate solution. has anyone tried to observe that if a given solution works for some starting column but has some slack on the right before it hits maxColumn, it can be shifted without recomputing? is that even true?

olafurpg: `Integer.compare` doesn't box primitives
```java
    public static int compare(int x, int y) {
```

> has anyone tried to observe that if a given solution works for some starting column but has some slack on the right before it hits maxColumn, it can be shifted without recomputing? is that even true?

Nobody has looked into that from what I know

kitbellew: > `Integer.compare` doesn't box primitives
> 
> ```java
>     public static int compare(int x, int y) {
> ```

of course! but this does (https://github.com/scalameta/scalafmt/commit/2cd7a731259df9bfbb76d3b3cc4ed1f5d24f076e#diff-0944a127ca40d4a12020bbe4e69623cb):
```
  val costCompare = Integer.valueOf(-this.cost).compareTo(-that.cost)

  // from Integer.java
    public static Integer valueOf(int i) {
        if (i >= IntegerCache.low && i <= IntegerCache.high)
            return IntegerCache.cache[i + (-IntegerCache.low)];
        return new Integer(i);
    }
```

i will look into performance of `State` comparisons later, but for now rather than putting `gen` into `State`, what about just implementing it as a list of priority queues instead? allocating a new queue and comparing values in it is likely cheaper than putting a value in `State` and comparing all `State`s.

olafurpg: Aha, yes indeed :)

There's no good reason to do `Integer.valueOf(primitive).compareTo(..)`

olafurpg: > what about just implementing it as a list of priority queues instead? 

Sounds good to me 👍

Issue #1632, opened 2020-01-13
Migrate from Travis CI to Github Actions


poslegm:
We already have native images pipeline in Github Actions. It might make sense to move the rest jobs to Actions for unified CI system.

olafurpg:
This is a good idea. In Metals we still run the publish jobs from Travis because it’s already working and migrating requires updating secrets. I’m fine if we do the same partial migration in Scalafmt, only move test jobs and keep release jobs on Travis

PR #1732, opened 2020-02-22
CI moved from Travis to GitHub Actions

poslegm:
Fixes https://github.com/scalameta/scalafmt/issues/1632


.github/workflows/ci.yml
```diff
@@ -27,3 +27,27 @@ jobs:
...
+        os: [macOS-latest, ubuntu-latest]
+        include:
+          - os: macOS-latest
+            url: https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-20.0.0/graalvm-ce-java11-darwin-amd64-20.0.0.tar.gz
+            artifact: scalafmt-macos
+          - os: ubuntu-latest
+            url: https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-20.0.0/graalvm-ce-java11-linux-amd64-20.0.0.tar.gz
+            artifact: scalafmt-linux
+    steps:
+      - uses: actions/checkout@v1
```
kitbellew: in another commit, you use v2 and v6, respectively. not sure if it's important.

.github/workflows/ci.yml
```diff
@@ -20,9 +20,8 @@ jobs:
           java-version: ${{ matrix.java }}
       - run: 
           # for GitOps tests
-          git config --global user.email "scalafmt@scalameta.org"
-          git config --global user.name "scalafmt"
-          sbt ci-test
+          git config --global user.email "scalafmt@scalameta.org" && git config --global user.name "scalafmt"
```
kitbellew: Would it work instead:
```
GIT_AUTHOR_NAME=... GIT_AUTHOR_EMAIL=... sbt ci-test
```

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: olafurpg/setup-scala@v2
+      - uses: olafurpg/setup-scala@v6
```
kitbellew: do you need to change v2 to v6 below, in formatting?

.github/workflows/ci.yml
poslegm: Good catch, it's copy-pasted config with old version

.github/workflows/ci.yml
kitbellew: looks like it didn't. the previous solution is good.

.github/workflows/ci.yml
```diff
@@ -1,9 +1,36 @@
...
   pull_request:
+  
 jobs:
-  build:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [openjdk@1.8, openjdk@1.11]
+        scala: ["2.11", "2.12", "2.13"]
```
tanishiking: Can we run cross-platform tests (including windows, ubuntu, and macOS)?

Since Github actions's concurrent limit is 20 https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits, it might be better to run cross scala version tests by `sbt +ci-test` instead?

olafurpg: I agree with not having a cross build per scala version.

olafurpg: It would also be nice to run tests on windows (no need for macos) but it can be done separately since it requires a lot of work sometimes.

.github/workflows/ci.yml
olafurpg: Environment variables don’t transfer between runs unless you explicitly export them by printing some special syntax.

.github/workflows/ci.yml
poslegm: `sbt +ci-test` will run tests only on 2.13 because of subprojects configuration. Now we have 9 parallel jobs, so there is no need to tweak it for concurrency limit (yet).

> I agree with not having a cross build per scala version.

Why? Sometimes cross build tests is useful. With their help I found several bugs on PR stage.

olafurpg: The `ci-test` command can be tweaked to run `+tests/test` instead.

https://github.com/scalameta/scalafmt/blob/2957451116d15276bad4a16c460b93ef76202079/build.sbt#L59

> Why? Sometimes cross build tests is useful. With their help I found several bugs on PR stage.

The CI should absolutely run tests for all Scala versions, my comment was rather "there is no need to have a custom CI job per Scala version". It's fine to run 2.11/2.12/2.13 tests in the same job.

olafurpg: For example, in Metals we run `+cross/test` in the same CI job that runs ~600 test cases against seven different Scala versions and we haven't felt the need to split it into multiple jobs yet. One nice thing we do however is to encode the Scala version number in the test name so that it's easier to see what Scala version the test failed on.

poslegm: Done

PR #1735, opened 2020-02-23
Test debug: create a debug instance per test

kitbellew:
Also, print debug after each test, not after all.


kitbellew:

PR #1737, opened 2020-02-24
Split: switch ignoreIf to tag, allows more control

kitbellew:
Rather than enabling a split using a simple boolean flag, let's allow a split to be enabled in some cases. This will allow us to set up some splits which will not be used unless some policy in a specific context decided to use them (for instance, splits for ", {" that would be triggered only if `OneArgPerLine` policy is used). See #1722, will be used in that PR.

Also, make a few minor cosmetic changes.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
@@ -75,23 +75,27 @@ case class Split(
     else if (isActive) copy(tag = tag)
     else throw new UnsupportedOperationException("Multiple tags unsupported")
 
-  def withOptimalToken(token: Option[Token]): Split = token match {
-    case Some(token) => withOptimalToken(token)
-    case _ => this
+  def withOptimalTokenOpt(
+      token: => Option[Token],
+      killOnFail: Boolean = false
+  ): Split =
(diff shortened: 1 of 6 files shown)


kitbellew:

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
```diff
@@ -55,60 +55,92 @@ case class Split(
...
-    Split(NoSplit, 0, ignoreIf = true)
+    Split(NoSplit, 0, tag = SplitTag.Ignored)
+
+  def onlyTaggedFor(tag: SplitTag.Value)(splits: Seq[Split]): Seq[Split] =
+    splits.flatMap { s =>
+      if (s.isTaggedFor(tag)) Some(s.copy(tag = SplitTag.Active)) else None
+    }
+}
 
+object SplitTag extends Enumeration {
```
poslegm: My only issue is `scala.Enumeration` usage

```scala
sealed trait SplitTag

object SplitTag {
  case object Ignored extends SplitTag
  case object Active extends SplitTag
}
```
sealed hierarchies is a little bit verbose, but a simple and reliable syntax.

kitbellew: will do, thank you.

PR #1740, opened 2020-02-24
FormatOps: use TreeMap for exclude range checking

kitbellew:


kitbellew:
benchmark shows minimal improvement, let's skip for now.

Issue #1410, opened 2019-04-25
trailing commas "always" does not respect function arity


russwyte:
This template is a guideline, not a strict requirement.
- **Version**: 2.0.0-RC6
- **Integration**: IntelliJ
- **Configuration**:
```
version = 2.0.0-RC6

align = true

maxColumn = 120

rewrite.rules = [RedundantBraces]

rewrite.redundantBraces.stringInterpolation = true

newlines.alwaysBeforeMultilineDef = true

newlines.alwaysBeforeTopLevelStatements = true

trailingCommas = always

```

## Steps

Given code like this:
```scala
   case Scrolled(key) =>
      resultFor(key) { rts =>
        rts.viewport.fold(noChange) { v =>
          val l = v.top(rts)
          effectOnly(
            Effect.action(
              ActionBatch(
                UpdateReaderState(rts.copy(scrollY = v.scrollY, location = l)),
                FindPageNumber(key)
              )
            )
          )
        }
      }
```

When I run scalafmt like this:

```bash
scalafmt
```

## Problem

Scalafmt formats code like this:
```scala
    case Scrolled(key) =>
      resultFor(key) { rts =>
        rts.viewport.fold(noChange) { v =>
          val l = v.top(rts)
          effectOnly(
            Effect.action(
              ActionBatch(
                UpdateReaderState(rts.copy(scrollY = v.scrollY, location = l)),
                FindPageNumber(key),
              ),
            ),
          )
        }
      }
```

## Expectation

I would like the formatted output to look like this:
```scala
    case Scrolled(key) =>
      resultFor(key) { rts =>
        rts.viewport.fold(noChange) { v =>
          val l = v.top(rts)
          effectOnly(
            Effect.action(
              ActionBatch(
                UpdateReaderState(rts.copy(scrollY = v.scrollY, location = l)),
                FindPageNumber(key),
              )
            )
          )
        }
      }
```

## Workaround

I haven't found any workaround.
## Notes

It seems like trailing commas "always" configuration should respect function/method arity - otherwise it looks like every function/ethod is a var-arg function/method.

russwyte:
In the example above effectOnly and Effect.action can only ever take one argument each - it makes no sense to have comma there. That could be very confusing to developers looking at the source code - even if it does compile. It seems we should only use trailing commas where we can actually pass var-args.

russwyte:
[...]

PR #1745, opened 2020-02-26
FormatOps: find the exact param group for each "("

kitbellew:
Rather than returning the concatenation of all parameter groups, return the specific one.

With that, we can now definitively determine the number of arguments in the specific parameter group, rather than the total number across all groups as we did before. That is, in
```
      def f(a: Int)(b: String, c: Boolean): Unit
```
we used to return 3 in getApplyArgs for both `(`, but would now return 1 and 2, respectively.

`scala-repos` diffs: https://github.com/kitbellew/scala-repos/commit/68a90df4556bfa220ac81d57b301e1aecad35ed2?w=1


kitbellew:
@olafurpg @tanishiking @poslegm PTAL. this is needed to address #1362 and #1410.

PR #1746, opened 2020-02-26
ScalafmtConfig: rename parameters when loading

### LINKED ISSUE #1362: Conflict between new line before / after implicit and verticalMultilineAtDefinitionSite?

kitbellew:
This way, we don't need to maintain multiple parameters in different places.

Also, rename vertical multiline params. We'll reuse them to bring consistency with non-vertical-multiline paths.

Will be used in #1696, and to fix #1362.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -502,6 +502,34 @@ else {
 }
 ```
 
+### `newlines.beforeImplicitKW`
+
+```scala mdoc:defaults
+newlines.beforeImplicitKW
+```
+
+```scala mdoc:scalafmt
+maxColumn = 60
+verticalMultiline.atDefnSite = true
+newlines.beforeImplicitKW = true
+---
+def format(code: String, age: Int)(implicit ev: Parser, c: Context): String
+```
+
+### `newlines.afterImplicitKW`
+
+```scala mdoc:defaults
+newlines.afterImplicitKW
+```
+
+```scala mdoc:scalafmt
+maxColumn = 60
+verticalMultiline.atDefnSite = true
+newlines.afterImplicitKW = true
+---
+def format(code: String, age: Int)(implicit ev: Parser, c: Context): String
+```
+
 ## Rewrite Rules
 
 To enable a rewrite rule, add it to the config like this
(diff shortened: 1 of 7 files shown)


kitbellew:

perhaps `metaconfig` is ultimately a better place for this, this implementation is not generic enough for that, and should likely include annotations at nested class level and possibly even expansions of the type used in `danglingParentheses`.

olafurpg:
I’m not sure I understand the change here. I believe it’s possible to evolve the configuration without custom transformations like in this PR. The rule of thumb I try to follow is to deprecate fields and then remove them after a couple years. I can’t remember exactly but I think metaconfig has a option to report deprecation warning even on successful config parsing.

olafurpg:
I would recommend avoiding this custom rename step since it makes it harder to reason about how the configuration is parsed. Can you elaborate on the need for this change?

kitbellew:
@olafurpg based on the discussion with @poslegm in #1696, the idea is to move the `implicitKW` parameters from `verticalMultiline` into the `newlines` section and implement the logic for non-vertical-multiline case.

unfortunately, these parameters were themselves moved to vertical multiline section earlier, which would mean we'd have 3 versions of the same parameter, plus the code which tries to read all of them and figure out what to do.

this change attempts to allow renaming parameters across sections (unlike the `@ExtraName` intra-section annotation), and also to ensure that only one of them is set, to simplify reading of the values. so this proposed change is a generalization of the `@ExtraName` annotation.

what would you recommend in this case? what to do about those 3 config values?

poslegm:
What about just deprecation and 

```scala
val afterImplicitKW = newlines.afterImplicitKW || verticalMultiline.newlineAfterImplicitKW || newlines.afterImplicitKWInVerticalMultiline
```

in the `ScalafmtConfig`?

kitbellew:
> What about just deprecation and
> 
> ```scala
> val afterImplicitKW = newlines.afterImplicitKW || verticalMultiline.newlineAfterImplicitKW || newlines.afterImplicitKWInVerticalMultiline
> ```
> 
> in the `ScalafmtConfig`?

sure, can do that.

PR #1749, opened 2020-02-28
Router: handle implicit in non-vertical-multiline

### LINKED ISSUE #1362: Conflict between new line before / after implicit and verticalMultilineAtDefinitionSite?

kitbellew:
This will format "implicit" on the same line as the opening paren, if possible. It will also break after the "implicit" unless all parameters fit on one line.

Fixes #1362.

`scala-repos` diffs: https://github.com/kitbellew/scala-repos/commit/eea2e89051a744b4c13c6e844aaf54ebaa12d64b?w=1

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -502,6 +502,68 @@ else {
 }
 ```
 
+### Newlines around `implicit` parameter list modifier
+
+#### Before
+
+> If set, forces newline before `implicit`. Otherwise, newline
+> can still be added if the keyword would overflow the line.
+
+```scala mdoc:defaults
+newlines.beforeImplicitParamListModifier
+```
+```scala mdoc:scalafmt
+maxColumn = 60
+newlines.beforeImplicitParamListModifier = true
+---
+def format(code: String, age: Int)(implicit ev: Parser, c: Context): String
+```
+
+#### After
+
+> If set, forces newline after `implicit`. Otherwise, newline can
+> still be added unless `before` is true, or the entire implicit
+> parameter list fits on a line, or config style is false.
+
+```scala mdoc:defaults
+newlines.afterImplicitParamListModifier
+```
+```scala mdoc:scalafmt
+maxColumn = 60
+newlines.afterImplicitParamListModifier = true
+---
+def format(code: String, age: Int)(implicit ev: Parser, c: Context): String
+```
+
+#### Both before and after
+```scala mdoc:scalafmt
+maxColumn = 60
+newlines.afterImplicitParamListModifier = true
+newlines.beforeImplicitParamListModifier = true
+---
+def format(code: String, age: Int)(implicit ev: Parser, c: Context): String
+```
+
+#### Neither before nor after
+```scala mdoc:scalafmt
+maxColumn = 60
+newlines.afterImplicitParamListModifier = false
+newlines.beforeImplicitParamListModifier = false
+---
+def format(code: String, age: Int)(implicit ev: Parser, c: Context): String
+```
+
+#### With `optIn.configStyleArguments`
+```scala mdoc:scalafmt
+maxColumn = 60
+optIn.configStyleArguments = true
+newlines.afterImplicitParamListModifier = true
+---
+def format(code: String, age: Int)(
+   implicit ev: Parser, c: Context
+): String
+```
+
 ## Rewrite Rules
 
 To enable a rewrite rule, add it to the config like this
@@ -711,16 +773,35 @@ verticalMultiline.newlineAfterOpenParen = true
 def other(a: String, b: String)(c: String, d: String) = a + b + c
 ```
 
-### `verticalMultiline.newlineBeforeImplicitKW`
+### Vertical multiline with `implicit` parameter lists
 
-```scala mdoc:defaults
-verticalMultiline.newlineBeforeImplicitKW
-```
+> Also see the general section on
+> [implicit parameter lists](#newlines-around-implicit-parameter-list-modifier).
 
+#### Before only
 ```scala mdoc:scalafmt
 maxColumn = 60
 verticalMultiline.atDefnSite = true
-verticalMultiline.newlineBeforeImplicitKW = true
+newlines.beforeImplicitParamListModifier = true
+---
+def format(code: String, age: Int)(implicit ev: Parser, c: Context): String
+```
+
+#### After only
+```scala mdoc:scalafmt
+maxColumn = 60
+verticalMultiline.atDefnSite = true
+newlines.afterImplicitParamListModifier = true
+---
+def format(code: String, age: Int)(implicit ev: Parser, c: Context): String
+```
+
+#### Before and after
+```scala mdoc:scalafmt
+maxColumn = 60
+verticalMultiline.atDefnSite = true
+newlines.beforeImplicitParamListModifier = true
+newlines.afterImplicitParamListModifier = true
 ---
 def format(code: String, age: Int)(implicit ev: Parser, c: Context): String
 ```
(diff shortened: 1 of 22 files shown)


kitbellew:

poslegm:
Is implicit formatting like

```scala
def marshal[T: ToEntityMarshaller](value: T)(implicit
    ec: ExecutionContext,
    mat: Materializer): HttpEntity.Strict = {}
```
the new _default_ behavior?

And are your sure that next release is a good time to remove a lot of deprecated options? We can wait a lot with this breaking changes (for 3.0.0 may be).

kitbellew:
> Is implicit formatting like
> 
> ```scala
> def marshal[T: ToEntityMarshaller](value: T)(implicit
>     ec: ExecutionContext,
>     mat: Materializer): HttpEntity.Strict = {}
> ```
> 
> the new _default_ behavior?

should we make non-vertical-multiline defaults the same as for vertical multiline?

> And are your sure that next release is a good time to remove a lot of deprecated options? We can wait a lot with this breaking changes (for 3.0.0 may be).

i don't feel strongly either way, although there's something to be said for making the logic a bit simpler. @olafurpg mentioned that he waits "a couple of years" before deprecating, and 1.6.0 was about two years ago.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/DanglingParentheses.scala
```diff
@@ -3,32 +3,37 @@ package org.scalafmt.config
 import metaconfig._
 
 case class DanglingParentheses(
-    callSite: Boolean,
-    defnSite: Boolean
+    callSite: Option[Boolean] = None,
```
poslegm: I see only `!style.danglingParentheses.defnSite.isFalse` and `!cfg.defnSite.isFalse` usage which means "true or None" i.e. default value in this PR. Did I miss something?

And in follow up of https://github.com/scalameta/scalafmt/pull/1696 discussion: when the user sees the default value `danglingParentheses = true` in the documentation, he will not explicitly set it in the settings. Maybe it makes sense to not introduce third state of config (`None`) and just its implement behavior for `true` state?

poslegm:
> i don't feel strongly either way, although there's something to be said for making the logic a bit simpler. @olafurpg mentioned that he waits "a couple of years" before deprecating, and 1.6.0 was about two years ago.

Ok. scalafmt doesn't follows semver completely, so it is not bad thing.

poslegm:
> should we make non-vertical-multiline defaults the same as for vertical multiline?

Don't understand your question 😅 What is "non-vertical-multiline defaults"? All settings exclude `VerticalMultiline`?

I just wanted to clarify whether `implicit` keyword will stay with open paren by default

poslegm:
In resume:

- I am OK with new `implicit` keyword formatting 👍 It looks nice, thank you!
- I am OK with two years old deprecations cleanup
- I am not sure about `Option[Boolean]` configs. What do @olafurpg and @tanishiking think about it?

kitbellew:
@poslegm thank you, eagle eye! let me review the change once again, I'll add a few more tests.
* i will remove the Option change for now, it was mostly a preparation for the #1696 change and that can be done later
* on the default, i was trying to keep it the same as the current vertical multiline and apparently... didn't
* one thing that is still a bit tricky is interaction with config style; if it's set or defaulted to true while the "before implicit" flag is set or defaulted to false, we have a conflict (one requires a newline and the other a nosplit); we should either decide which of the two settings has a preference, or define attached implicit followed by a newline as still config style. thoughts on the preference?

scalafmt-core/shared/src/main/scala/org/scalafmt/config/DanglingParentheses.scala
kitbellew: Done. The `Option` is removed.

kitbellew:
> I just wanted to clarify whether `implicit` keyword will stay with open paren by default

yes. since we are not using `Option`, i used the same defaults as in the vertical multiline section, setting both flags to `false`.

kitbellew:
@poslegm PTAL

kitbellew:
> I prefer `newlines.beforeImplicitParamListModifier` priority over `optIn.configStyleArguments`. Because config style is a common concept and `newlines` is a specific setting. It seems clear to me that the user set config style for arguments and **overrides** it by `beforeImplicitParamListModifier `
> 
> Thank you for this well-documented change! I am OK with it, but I would like to hear the opinion of other maintainers.

@olafurpg @tanishiking your thoughts on this change? it expands the newline parameters before and after the `implicit` modifier of a parameter list to apply to all cases, not just vertical multiline.

olafurpg:
What is the migration guide for users who enabled `newlines.beforeImplicitKWInVerticalMultiline = true`?

Upgrading from 2.4.2 to 2.5.1 introduced this error message
```
Invalid field: beforeImplicitKWInVerticalMultiline. ...
```

I think we should try to never remove configuration options since it's very difficult for users to figure out the migration path. The 2.5.x release notes did mention `beforeImplicitKWInVerticalMultiline`, it took me a while to track down this PR.

kitbellew:
> What is the migration guide for users who enabled `newlines.beforeImplicitKWInVerticalMultiline = true`?
> 
> Upgrading from 2.4.2 to 2.5.1 introduced this error message
> 
> ```
> Invalid field: beforeImplicitKWInVerticalMultiline. ...
> ```
> 
> I think we should try to never remove configuration options since it's very difficult for users to figure out the migration path. The 2.5.x release notes did mention `beforeImplicitKWInVerticalMultiline`, it took me a while to track down this PR.

I think it's one of those which have been deprecated a while ago. I will add to release notes for 2.5.0, except people apparently don't read them.

olafurpg:
I didn't read the deprecation annotation on the removed field 🙈 Scalafmt 2.4.2 didn't report warnings when using this field so I didn't notice it was deprecated. It's good that the new version reports deprecation warnings.

> I will add to release notes for 2.5.0, except people apparently don't read them.

No worries, please ignore this. Sorry for the noise!

rulle-io:
Sorry, I cannot find the aforementioned release notes about *newlines.beforeImplicitKWInVerticalMultiline* removal even if I explicitely search for them. @kitbellew Can you, please, attach a link to this page here.

rulle-io:
Have to go for a code diff search. Found [this:](https://github.com/scalameta/scalafmt/pull/1749/files#diff-5f5955ed2f0111c6cd6d8c12c7b5d38cc1dfbad858c80034de952cc6a2e6a2d8L107-L110).
```
@deprecated("Use VerticalMultiline.newlineAfterImplicitKW instead")
afterImplicitKWInVerticalMultiline: Boolean = false,
@deprecated("Use VerticalMultiline.newlineBeforeImplicitKW instead")
beforeImplicitKWInVerticalMultiline: Boolean = false,
```

PR #1750, opened 2020-02-28
RedundantBraces: don't rewrite to parens if nested typed func

kitbellew:
Fixes #1748.

PR #1751, opened 2020-02-28
CI tuning

poslegm:
I think that problems with OOM at CI steps starts after squashing all cross-build steps into one `sbt +tests/test`.

We can still it in one job, but in different sbt launches.


poslegm:
Because of problems with https://github.com/scalameta/scalafmt/pull/1750

cc @kitbellew

Issue #1627, opened 2020-01-09
Add an option to format without regard to input line breaks


kitbellew:
To avoid frequent discussions in code review requesting formatting changes in the code to induce `scalafmt` to produce different output, I would like to introduce an option to format input and ignore input newlines (except adjacent to comments).

Propose 4 options:

* flag not specified, produces current mixed approach of sometimes ignoring input line breaks and sometimes following them;
* option to keep line breaks as specified;
* two options to try and ignore line breaks, one for fewer lines and another for shorter lines; both are affected by config style, select chains bin pack and other similar options, not sure if they should supercede then)

As this requires a lot of changes, I'd like to consider this option as work-in-progress until all changes have been merged, and larger community had time to comment. It will not be subject to edition guards until the logic has been finalized and declared completed.

NB: Follow-on to #1560

kitbellew:
@olafurpg @tanishiking @poslegm need recommendation on this proposal.
* is the approach meaningful to you in the first place?
* do the options proposed make sense?
* making the changes is complex and will take some time, they will likely benefit from people trying out and commenting; I would like to add the options with a (temporary) disclaimer, then work on a sequence of PRs, possibly over a couple of releases until the formatting changes have been completed; thoughts about that?

poslegm:
> is the approach meaningful to you in the first place?

_Personally_ I don't need this changes. And I never heard feedback about scalafmt from my teammates with something like it. But right now I'm only speaking for myself, not for the best solution for everyone.

> do the options proposed make sense?

I strongly prefer forcing users to multiline blocks every time. So my only wish for potential changes is the configuration flag to keep current behavior.

> making the changes is complex and will take some time, they will likely benefit from people trying out and commenting; I would like to add the options with a (temporary) disclaimer, then work on a sequence of PRs, possibly over a couple of releases until the formatting changes have been completed; thoughts about that?

I am completely OK with splitting big changes into couple of small releases

ekrich:
@kitbellew I am mostly aligning with the sentiments of @olafurpg in the series of comments in #1560

My first impression of `scalafmt` was just that I didn't want to worry about formatting at all and this is what I found when working on Scala Native. If you look a the Google Java rules, they are really draconian. I have seen similar for C++ with the order of `private`, `public` etc. 

Scala gives a better approach with nested functions, and order can make the code more readable and tools can alphabetize the methods, functions etc. so they don't need to be that way in the actual code. This is a bit off base but may help with my philosophy. 

[...]

PR #1752, opened 2020-02-29
Router: handle if with/without braces consistently

### LINKED ISSUE #1747: Inconsistent formatting of if/else
Version: 2.4.2

```
val a = 5

// 1
val ok1 = if (a > 10) Some(a) else None

// 2
val ok2 = if (a > 10) {
  Some(a)
} else {
  None
}

// 3
val ok3 = if (a > 10)
  Some(a)
else
  None

// 4
val ok4 = if (a > 10) Some(a) else {
  None
}

// examples 1-2 are unaffected

// example 3 is re-written as:
val ko3 =
  if (a > 10)
    Some(a)
  else
    None

// example 4 is re-written as:
val ko4 =
  if (a > 10) Some(a)
  else {
    None
  }

```

It seems that the formatting of examples 3 and 4 is not consistent with example 2; The "if" is pushed to the next line for some reason.

kitbellew:
To do that, allow space only if 'else' is empty or the entire expression fits on the line.

Fixes #1747.

`scala-repos` diffs: https://github.com/kitbellew/scala-repos/commit/3670a380d2efa282b8432c8f1ab45ca49efa2991?w=1

p.s. old diffs: https://github.com/kitbellew/scala-repos/commit/dc650f97f6666e1120da601e2c98ad2233d86c96?w=1

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1090,22 +1090,23 @@ class Router(formatOps: FormatOps) {
             }
             val jsNative = isJsNative(right)
             val noNewline = jsNative
-            val spacePolicy: Policy = rhs match {
-              case _: Term.If => twoBranches
-              case _: Term.ForYield => twoBranches
+            val (spacePolicy, spaceIndent) = rhs match {
+              case t: Term.If if style.activeForEdition_2020_03 =>
+                SingleLineBlock(t.cond.tokens.last) -> t.elsep.tokens.nonEmpty
+              case _: Term.If => twoBranches -> false
+              case _: Term.ForYield => twoBranches -> false
               case _: Term.Try | _: Term.TryWithHandler
                   if style.activeForEdition_2019_11 && !noNewline =>
-                null // we force newlines in try/catch/finally
-              case _ => NoPolicy
+                (null, false) // we force newlines in try/catch/finally
+              case _ => NoPolicy -> false
             }
-            val noSpace = null == spacePolicy ||
+            val noSpace = null == spacePolicy || isSingleLineComment(right) ||
               (!jsNative && newlines > 0 && leftOwner.isInstanceOf[Defn])
-            val spaceIndent = if (isSingleLineComment(right)) 2 else 0
             Seq(
               Split(Space, 0, policy = spacePolicy)
                 .notIf(noSpace)
                 .withOptimalToken(optimal)
-                .withIndent(spaceIndent, expire, Left),
+                .withIndent(if (spaceIndent) 2 else 0, expire, Left),
               Split(Newline, 1 + penalty)
                 .notIf(noNewline)
                 .withIndent(2, expire, Left)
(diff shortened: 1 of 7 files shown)


kitbellew:

kitbellew:
> Thank you, I like it! But I would rather wait for the rest maintainers opinion because of large diffs in scala-repos

@olafurpg @tanishiking your thoughts on this change? the diffs in `scala-repos` change
```
  val a = if (b) {
    c
  } else d
```
to
```
  val a =
    if (b) {
      c
    } else d
```
for consistency with existing
```
  val a =
    if (b)
      c
    else d
```

tanishiking:
Thanks! I'll take a look this evening :)

kitbellew:
@tanishiking @poslegm PTAL, updated code with Rikito's suggestion, uploaded new diffs (mostly whitespace for indent, and some non-brace `if`s looking more like brace `if`s).

poslegm:
I don't sure that additional indent is the best solution for the consistency problem. 

Especially I don't like this:
```scala
val x = if (x > max) "a"
  else "b"
```

Personally I prefer first version of @kitbellew's changes because it perfectly standardizes if/else formatting. But because of large diff in real-world codebases we may be forced to keep current "liberal" behavior 😞

kitbellew:
> Especially I don't like this:
> 
> ```scala
> val x = if (x > max) "a"
>   else "b"
> ```

Neither do I (including the brace version). Continuing without a break logically means that there's only one expression on the right, which works with a `def` returning a `Unit`, a `match`, or a `case` guard; but it reads poorly in assignment which requires both `then` and `else` (i.e., two expressions).

Abandon? Add a parameter similar to `newlines.alwaysBeforeMultilineDef` or `alwaysBeforeElseAfterCurlyIf`, whichever one looks more appropriate? Generally, I was planning to handle such cases in my #1627 which I will start submitting soon anyway.

poslegm:
It really looks like a good case for `newlines.*` configuration. I am ok with new setting for this.

kitbellew:
> It really looks like a good case for `newlines.*` configuration. I am ok with new setting for this.

should we simply reuse `alwaysBeforeMultilineDef` with an edition guard? this setting seems quite similar in nature, especially if we interpret `Def` as `Defn` (which covers `val`, `var` etc.). we've done it before with `alwaysBeforeElseAfterCurlyIf` as applied to `try-finally`.

poslegm:
> should we simply reuse alwaysBeforeMultilineDef with an edition guard?
sounds reasonable

paulkore:
Hi all, I'm glad to see that progress is being made on this issue!

@tanishiking , to answer your question from before, I agree with you: 
Placing the "if" on the same line as the val assignment seems to be standard practice in languages where "if/else" has a return value.

However, I'm not sure about the indentation. 
Based on the codebases that I've worked in, it seems more popular to keep the same indentation in if/else, regardless of whether it's used procedurally or in an assignment:

```
// procedural
if (a > 10) {
  println("A")
} else {
  println("B")
}

// assignment
val x = if (a > 10) {
  Some(a)
} else {
  None
}
```

tanishiking:
Thank you so much for modifying @kitbellew and your comment @paulkore 

IMO, I think we don't have to fix this "inconsistency" between if/else with and without braces. (I mean it's best to maintain the status quo).

Though @paulkore mentioned that there's an inconsistency between those formatting results, I think we don't have to honor the consistency about the indent/newline between them, because (as far as I know) a lot of scala developers usually use different indentation/newline styles for if/else statement with and without braces.

```scala
// without braces
val ok3 =
  if (a > 10) Some(a)
  else None

// with braces
val ok3 = if (a > 10) {
  Some(a)
} else {
  None
}
```

That's why I believe it's better not to change the current behavior.
It might be better to hear more opinions from other scala developers, I guess.

---

It's ok with adding an option but in that case, I think it's better to keep the default behavior.

PR #1759, opened 2020-03-01
Alignment by right-side expression

### LINKED ISSUE #1753: Align = to the right in for comprehension
I couldn't find any config option to change the aligning of `=` in for comprehension.
I want it to be aligned to the right instead of to the left, so that such code

```scala
for {
  _ <- f1
  someValue <- f2
  _ = doLogs
 _ <- f3
} yield ()
```
becomes:
```scala
for {
  _         <- f1
  someValue <- f2
  _          = doLogs
  _         <- f3
} yield ()
```
instead of:
```scala
for {
  _         <- f1
  someValue <- f2
  _         = doLogs
  _         <- f3
} yield ()
```

poslegm:
Fixes #1753 

Current formatting:
```scala
libraryDependencies ++= Seq(
  "org.scalacheck"  %% "scalacheck" % scalacheckV,
  "io.get-coursier" % "interface"   % "0.0.17"
)
```

There is no need to align separator symbols. The alignment of _expressions_ is much more important for pleasant code reading

New formatting:

```scala
libraryDependencies ++= Seq(
  "org.scalacheck" %% "scalacheck" % scalacheckV,
  "io.get-coursier" % "interface"  % "0.0.17"
)
```

scala-repos diff with `align=most`: https://github.com/poslegm/scala-repos/commit/659f12339ac13d02b00c697a36aeb811a53a6d6c

What do you think: should it be hidden above `edition`?

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -20,7 +20,7 @@ import scala.meta.{Importer, Mod, Pkg, Term, Tree}
 /**
   * Produces formatted output from sequence of splits.
   */
-class FormatWriter(formatOps: FormatOps)(implicit style: ScalafmtConfig) {
+class FormatWriter(formatOps: FormatOps) {
   import FormatWriter._
   import formatOps._
 
@@ -626,9 +626,8 @@ class FormatWriter(formatOps: FormatOps)(implicit style: ScalafmtConfig) {
               alignmentUnits.foreach { info =>
                 import info._
                 val tokenLengthGap =
-                  if (style.activeForEdition_2020_03)
-                    widest.separatorLength - separatorLength
-                  else 0
+                  if (!initStyle.activeForEdition_2020_03) 0
+                  else widest.separatorLength - separatorLength
                 previousSeparatorLengthGaps(lineIndex) = tokenLengthGap
                 finalResult += tokenHash -> (widest.width - width + tokenLengthGap)
               }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -3,6 +3,7 @@ package org.scalafmt.internal
 import java.util.regex.Pattern
 
 import org.scalafmt.rewrite.RedundantBraces
+import org.scalafmt.util.TokenOps.TokenHash
 import org.scalafmt.util.TreeOps
 
 import scala.annotation.tailrec
@@ -596,18 +597,22 @@ class FormatWriter(formatOps: FormatOps) {
           if (matches == 0 || doubleNewline || !locationIter.hasNext) {
             var column = 0
             val columns = minMatches
+            // TODO document
+            val separatorLengthGaps = Array.fill(block.length)(0)
             while (column < columns) {
-              val blockWithWidth = block.map { line =>
-                val location = line(column)
-                val previousWidth =
-                  if (column == 0) 0 else line(column - 1).shift
-                val key = location.shift - previousWidth
-                key -> hash(location.formatToken.left)
-              }
-              val maxWidth = blockWithWidth.map(_._1).max
-              blockWithWidth.foreach {
-                case (width, tokenHash) =>
-                  finalResult += tokenHash -> (maxWidth - width)
+              val alignmentUnits =
+                prepareAlignmentInfo(block, separatorLengthGaps, column)
+
+              val maxWidth = alignmentUnits.map(_.width).max
+
+              val maxWidthTokenLength =
+                alignmentUnits.maxBy(_.width).separatorLength
+
+              alignmentUnits.foreach { info =>
+                import info._
+                val tokenLengthGap = maxWidthTokenLength - separatorLength
+                separatorLengthGaps(lineIndex) += tokenLengthGap
+                finalResult += tokenHash -> (maxWidth - width + tokenLengthGap)
               }
               column += 1
             }
@@ -624,6 +629,25 @@ class FormatWriter(formatOps: FormatOps) {
     }
   }
 
+  private def prepareAlignmentInfo(
+      block: Vector[Array[FormatLocation]],
+      separatorLengthGaps: Array[Int],
+      column: Int
+  ): Vector[AlignmentUnit] =
+    block.zipWithIndex.map {
+      case (line, i) =>
+        val location = line(column)
+        val previousWidth =
+          if (column == 0) 0 else line(column - 1).shift
+        val key = location.shift - previousWidth + separatorLengthGaps(i)
+        AlignmentUnit(
+          key,
+          hash(location.formatToken.left),
+          location.formatToken.right.text.length,
+          i
+        )
+    }
+
 }
 
 object FormatWriter {
@@ -633,6 +657,29 @@ object FormatWriter {
       state: State,
       shift: Int = 0,
       replace: String = null
+  )
+
+  /**
+    * Alignment information extracted from FormatToken. Used only when align!=none.
+    * For example:
+    * ```
+    * libraryDependencies ++= Seq(
+    *   "io.get-coursier" % "interface" % "0.0.17",
+    *   "org.scalacheck" %% "scalacheck" % scalacheckV
+    * )
+    * ```
+    *
+    * `"io.get-coursier" % "interface" % "0.0.17"`
+    *  |<--------------->|      => width
+    *  hash("io.get-coursier")  => tokenHash
+    *  length(%)                => separatorLength
+    *  line number in block (1) => lineIndex
+    * */
+  case class AlignmentUnit(
+      width: Int,
+      tokenHash: TokenHash,
+      separatorLength: Int,
+      lineIndex: Int
   )
 
   // cache indentations to some level
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -617,7 +617,7 @@ class FormatWriter(formatOps: FormatOps)(implicit style: ScalafmtConfig) {
               * )
               * ```
               * */
-            val previousSeparatorLengthGaps = Array.fill(block.length)(0)
+            val previousSeparatorLengthGaps = new Array[Int](block.length)
             while (column < columns) {
               val alignmentUnits =
                 prepareAlignmentInfo(block, previousSeparatorLengthGaps, column)
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -2,6 +2,7 @@ package org.scalafmt.internal
 
 import java.util.regex.Pattern
 
+import org.scalafmt.config.ScalafmtConfig
 import org.scalafmt.rewrite.RedundantBraces
 import org.scalafmt.util.TokenOps.TokenHash
 import org.scalafmt.util.TreeOps
@@ -19,7 +20,7 @@ import scala.meta.{Importer, Mod, Pkg, Term, Tree}
 /**
   * Produces formatted output from sequence of splits.
   */
-class FormatWriter(formatOps: FormatOps) {
+class FormatWriter(formatOps: FormatOps)(implicit style: ScalafmtConfig) {
   import FormatWriter._
   import formatOps._
 
@@ -624,7 +625,10 @@ class FormatWriter(formatOps: FormatOps) {
               val widest = alignmentUnits.maxBy(_.width)
               alignmentUnits.foreach { info =>
                 import info._
-                val tokenLengthGap = widest.separatorLength - separatorLength
+                val tokenLengthGap =
+                  if (style.activeForEdition_2020_03)
+                    widest.separatorLength - separatorLength
+                  else 0
                 previousSeparatorLengthGaps(lineIndex) = tokenLengthGap
                 finalResult += tokenHash -> (widest.width - width + tokenLengthGap)
               }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -572,36 +572,10 @@ class FormatWriter(formatOps: FormatOps) {
           if (!ok || floc.formatToken.leftHasNewline) columnShift = 0
           columnShift += floc.shift
           if (ok)
-            if (isCandidate(floc)) {
-
-              /**
-                * Separator length gap needed to align blocks with different token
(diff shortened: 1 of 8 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -594,18 +595,42 @@ class FormatWriter(formatOps: FormatOps) {
...
+              *
+              * With gaps considering:
+              * ```
+              * libraryDependencies ++= Seq(
+              *   "org.scalacheck" %% "scalacheck" % scalacheckV,
+              *   "io.get-coursier" % "interface"  % "0.0.17"
+              * )
+              * ```
+              * */
+            val previousSeparatorLengthGaps = Array.fill(block.length)(0)
```
kitbellew: is it the same as `new Array[Int](block.length)`?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -594,18 +595,42 @@ class FormatWriter(formatOps: FormatOps) {
...
-                  finalResult += tokenHash -> (maxWidth - width)
+              val alignmentUnits =
+                prepareAlignmentInfo(block, previousSeparatorLengthGaps, column)
+
+              val maxWidth = alignmentUnits.map(_.width).max
+
+              val maxWidthTokenLength =
+                alignmentUnits.maxBy(_.width).separatorLength
+
+              alignmentUnits.foreach { info =>
```
kitbellew: why not `alignmentUnits.zipWithIndex.foreach` and remove the `lineIndex` field below?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -594,18 +595,42 @@ class FormatWriter(formatOps: FormatOps) {
...
-                key -> hash(location.formatToken.left)
-              }
-              val maxWidth = blockWithWidth.map(_._1).max
-              blockWithWidth.foreach {
-                case (width, tokenHash) =>
-                  finalResult += tokenHash -> (maxWidth - width)
+              val alignmentUnits =
+                prepareAlignmentInfo(block, previousSeparatorLengthGaps, column)
+
+              val maxWidth = alignmentUnits.map(_.width).max
```
kitbellew: why not, in one pass, get
```
  val widest = alignmentUnits.maxBy(_.width)
```
and then, below, use `widest.width` and `widest.separatorLength`?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -622,6 +647,28 @@ class FormatWriter(formatOps: FormatOps) {
     }
   }
 
+  private def prepareAlignmentInfo(
+      block: Vector[Array[FormatLocation]],
+      separatorLengthGaps: Array[Int],
+      column: Int
+  ): Vector[AlignmentUnit] =
+    block.zipWithIndex.map {
```
kitbellew: why not `block.zip(separatorLengthGaps` and remove the `lineIndex` field below?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -622,6 +647,28 @@ class FormatWriter(formatOps: FormatOps) {
...
+      case (line, i) =>
+        val location = line(column)
+        val previousWidth =
+          if (column == 0) 0 else line(column - 1).shift
+        val key = location.shift - previousWidth + separatorLengthGaps(i)
+        val separatorLength =
+          if (location.formatToken.right.is[Token.Comment]) 0
+          else location.formatToken.right.text.length
```
kitbellew: is `text` better than (or different from) `syntax` (which is what i've seen used commonly)?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -622,6 +647,28 @@ class FormatWriter(formatOps: FormatOps) {
     }
   }
 
+  private def prepareAlignmentInfo(
+      block: Vector[Array[FormatLocation]],
+      separatorLengthGaps: Array[Int],
+      column: Int
+  ): Vector[AlignmentUnit] =
```
kitbellew: very, very optional: why are we using `Vector` when we don't do index lookups? http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#lists-vs-vectors

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -594,18 +595,42 @@ class FormatWriter(formatOps: FormatOps) {
           if (matches == 0 || doubleNewline || !locationIter.hasNext) {
             var column = 0
             val columns = minMatches
+
+            /**
+              * Separator length gap needed for align blocks with different token
+              * lengths by expression names, not tokens itself.
```
kitbellew: tokens *themselves*

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -594,18 +595,42 @@ class FormatWriter(formatOps: FormatOps) {
           if (matches == 0 || doubleNewline || !locationIter.hasNext) {
             var column = 0
             val columns = minMatches
+
+            /**
+              * Separator length gap needed for align blocks with different token
```
kitbellew: "*to* align" or "for *aligning*"

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -594,18 +595,42 @@ class FormatWriter(formatOps: FormatOps) {
...
+              * lengths by expression names, not tokens itself.
+              *
+              * Without gaps considering:
```
kitbellew: without considering gaps

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -594,18 +595,42 @@ class FormatWriter(formatOps: FormatOps) {
...
+              * ```
+              * libraryDependencies ++= Seq(
+              *   "org.scalacheck"  %% "scalacheck" % scalacheckV,
+              *   "io.get-coursier" % "interface"   % "0.0.17"
+              * )
+              * ```
+              *
+              * With gaps considering:
```
kitbellew: maybe: taking gaps into account

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
poslegm: Yes

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
poslegm: Because I want protection against further code changes which may broke `alignmentUnits` order. `lineIndex` field always contain actual line number in origin array

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
poslegm: Good catch!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
poslegm: Similar motivation https://github.com/scalameta/scalafmt/pull/1759#discussion_r386346282

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
poslegm: `text` comes from [Position](https://github.com/scalameta/scalameta/blob/master/scalameta/inputs/shared/src/main/scala/scala/meta/inputs/Position.scala#L20)

`syntax` is a `text` with dialect check. `syntax` would be more correct to use here. Replaced.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
poslegm: Because `Vector` used here before 🤷‍♂ 

May be it comes because of append operation [here](https://github.com/scalameta/scalafmt/blob/master/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala#L585)

poslegm:
@kitbellew Thank you for review and grammar check!

> if the earlier approach is a bug, you don't need an edition guard; otherwise, probably do

Without documented specification it's hard to judge is it bug or no. 

> somewhat unrelated: does this, perhaps, address the problem someone reported on gitter: https://gitter.im/scalameta/scalafmt?at=5e52f9dc4880f07ed1f4a1b3

No, it's another thing.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -19,7 +20,7 @@ import scala.meta.{Importer, Mod, Pkg, Term, Tree}
 /**
   * Produces formatted output from sequence of splits.
   */
-class FormatWriter(formatOps: FormatOps) {
+class FormatWriter(formatOps: FormatOps)(implicit style: ScalafmtConfig) {
```
kitbellew: you don't need this. `formatOps` contains the configuration, via `initStyle`, and it's even imported directly two lines below.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -624,7 +625,10 @@ class FormatWriter(formatOps: FormatOps) {
               val widest = alignmentUnits.maxBy(_.width)
               alignmentUnits.foreach { info =>
                 import info._
-                val tokenLengthGap = widest.separatorLength - separatorLength
+                val tokenLengthGap =
+                  if (style.activeForEdition_2020_03)
```
kitbellew: oh, is that all the change from previous behaviour? :)

optional OCD version:
```
  if (!initStyle.activeForEdition_2020_03) 0
  else widest.separatorLength - separatorLength
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
poslegm: wow, thank you

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
poslegm: Yes, only gaps to compensate tokens length difference :)

poslegm:
> i am probably a picky reviewer

It's better than miss a bug or smelly code :)

> can you hold off merging for a second, please? i wanted to test something.

Yes, of course

kitbellew:
> > i am probably a picky reviewer
> 
> It's better than miss a bug or smelly code :)
> 
> > can you hold off merging for a second, please? i wanted to test something.
> 
> Yes, of course

i was looking into the `FormatWriter` code which changes braces into parentheses, because it uses alignment information (and the `shift` field specifically). since you don't modify `shift`, i am not clear whether it will continue working, and i didn't a good-enough test for that.

so i tried modifying `shift` directly, and it looks like this *small* patch works as well, and possibly continues working with the braces-to-parens rewrite:
```
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -570,10 +570,15 @@ class FormatWriter(formatOps: FormatOps) {
           if (!ok || floc.formatToken.leftHasNewline) columnShift = 0
           columnShift += floc.shift
           if (ok)
-            if (isCandidate(floc))
+            if (isCandidate(floc)) {
+              val gapWidth = floc.formatToken.right match {
+                case _: T.Comment => 0
+                case t => t.syntax.length
+              }
               columnCandidates += floc.copy(shift =
-                floc.state.prev.column + columnShift
+                floc.state.prev.column + columnShift + gapWidth
               )
+            }
           ok
         }
         while (shouldContinue(locationIter.next()) && locationIter.hasNext) {}
```

poslegm:
hmm, it's good simplification. I will try how it works with braces -> parens config enabled

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -594,18 +622,15 @@ class FormatWriter(formatOps: FormatOps) {
           if (matches == 0 || doubleNewline || !locationIter.hasNext) {
             var column = 0
             val columns = minMatches
+
             while (column < columns) {
-              val blockWithWidth = block.map { line =>
```
kitbellew: do we need any of the changes here and below, after the `shift` adjustment above?

poslegm:
@kitbellew take a look at the new [scala-repos diff](https://github.com/poslegm/scala-repos/commit/9cf1a8cd49eadb65a217b28663bf606f6a556739) please. It seems that `shift` patch not working good as is.

I'll try to test origin implementation with `RedundantBraces` config

kitbellew:
> @kitbellew take a look at the new [scala-repos diff](https://github.com/poslegm/scala-repos/commit/9cf1a8cd49eadb65a217b28663bf606f6a556739) please. It seems that `shift` patch not working good as is.
> 
> I'll try to test origin implementation with `RedundantBraces` config

thank you for checking. i wonder if this whole `shift` logic is broken. and thank you for checking the braces to parens part!

poslegm:
@kitbellew tested origin implementation with 
```
align=most
```

There is a large [diff](https://github.com/poslegm/scala-repos/commit/171be1c2b98a319a56b03ceb8c3ad7b5f33a8bdc) but all of it is idempotence issues reproduces with v2.4.2 😢 

I will revert commit with `shift` batch and merge origin implementation because it works stable on scala-repos. Also it seems more accurate way because gaps logic localized in alignment algorithm and don't touch `shift` which used in other places.

kitbellew:
> @kitbellew tested origin implementation with
> 
> ```
> align=most
> rewrite.rules = [RedundantBraces]
> ```
> 
> There is a large [diff](https://github.com/poslegm/scala-repos/commit/171be1c2b98a319a56b03ceb8c3ad7b5f33a8bdc) but all of it is idempotence issues reproducible with v2.4.2 😢
> 
> I will revert commit with `shift` batch and merge origin implementation because it works stable on scala-repos. Also it seems more accurate way because gaps logic localized in alignment algorithm and don't touch `shift` which used in other places.

all right, sorry i had you chase this problem.

PR #1761, opened 2020-03-02
DanglingParentheses: move `exclude` list from VerticalMultiline

kitbellew:
With the default of empty, this doesn't change any existing formatting. However, just like with `VerticalMultiline`, it will allow users to reduce effect of `danglingParenthese.defnSite = true`.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
@@ -26,6 +26,7 @@ import scala.meta.tokens.Tokens
 import scala.reflect.ClassTag
 import scala.reflect.classTag
 import org.scalafmt.Error
+import org.scalafmt.config.{DanglingParentheses, ScalafmtConfig}
 
 /**
   * Stateless helper functions on [[scala.meta.Tree]].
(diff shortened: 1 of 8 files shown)


kitbellew:
@olafurpg @tanishiking @poslegm PTAL. Second PR leading up to #1696 which will fix #1539.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -586,4 +587,26 @@ object TreeOps {
...
+          style.danglingParentheses.exclude
+      excludeList.nonEmpty && {
+        val exclude = tree match {
+          case _: Ctor.Primary | _: Defn.Class =>
+            DanglingParentheses.Exclude.`class`
+          case _: Defn.Trait => DanglingParentheses.Exclude.`trait`
+          case _: Defn.Def => DanglingParentheses.Exclude.`def`
+          case _ => null
+        }
+        null != exclude && excludeList.contains(exclude)
```
poslegm: `excludeList.contains(null) == false`

kitbellew: if you don't mind, I'd leave it as-is. the intent is more explicit and doesn't require a list scan for something definitely not there

poslegm: 👌

PR #1762, opened 2020-03-03
Fixed AST changes after formatting variant types with symbolic name

### LINKED ISSUE #1568: variance definitions and symbolic type parameter
Default behavior resulted in resolved this issue #98 receives this code

```scala
trait Foo[+ ->[_, _]]
```

and transforms that into
```scala
trait Foo[+->[_, _]]
```

Which have syntactical significant difference.

Scalafmt should keep at least one space between variance declaration and symbolic type parameter name.

poslegm:
Fixes #1568

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1368,9 +1368,8 @@ class Router(formatOps: FormatOps) {
       // Type variance
       case tok @ FormatToken(T.Ident(_), T.Ident(_) | T.Underscore(), _)
           if isTypeVariant(leftOwner) =>
-        Seq(
-          Split(NoSplit, 0)
-        )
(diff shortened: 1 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1368,9 +1368,8 @@ class Router(formatOps: FormatOps) {
       // Type variance
       case tok @ FormatToken(T.Ident(_), T.Ident(_) | T.Underscore(), _)
           if isTypeVariant(leftOwner) =>
-        Seq(
-          Split(NoSplit, 0)
-        )
+        if (isSymbolicIdent(tok.right)) Seq(Split(Space, 0))
```
kitbellew: Optional: i created a shortcut for this:

`Space(isSymbolicIdent(...))`...

poslegm: I forgot about it, thank you

PR #1763, opened 2020-03-04
Router: remove currently unused comma split

kitbellew:
Follow-on to #1722, which introduced splits which always take precedence when used.

No `scala-repos` diffs.


kitbellew:

PR #1765, opened 2020-03-04
#1627 [01]: ScalafmtConfig: flag to ignore input line breaks

kitbellew:
The flag can be:

* _classic_ (unspecified): consistent with the current mixed behaviour
* _keep_: will try to preserve existing line breaks
* _fold_: removes some line breaks, for a more compact look
* _unfold_: inserts line breaks as appropriate, expanding

Also, add empty test suites for each value of this new flag. We will be adding test cases and the code using the flag in the next commits.

Towards #1627.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Newlines.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Newlines.scala
@@ -95,8 +95,15 @@ import metaconfig.generic.Surface
   *         ...
   *       }
   *   }}}
+  * @param source
+  *   Controls how line breaks in the input source are handled
+  *   If `classic` (default), the old mixed behaviour will be used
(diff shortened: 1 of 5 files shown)


kitbellew:

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Newlines.scala
```diff
@@ -95,8 +95,15 @@ import metaconfig.generic.Surface
   *         ...
   *       }
   *   }}}
+  * @param source
+  *   Controls how line breaks in the input source are handled
+  *   If `classic` (default), the old mixed behaviour will be used
```
poslegm: Users don't know what is "old behaviour" 😄 
In the next steps we must remember to rewrite this more detailed

PR #1766, opened 2020-03-06
Router bugfix: extend line break past infix for any block

kitbellew:
This approach is better than applying a higher cost to blocks abutting an infix. Fixes a number of line-too-long problems and also avoid line breaks when they are not necessary.

`scala-repos` diffs: https://github.com/kitbellew/scala-repos/commit/787fbfa747792b9cf56034bee8f41e3909ac3362?w=1


kitbellew:

Issue #1712, opened 2020-02-16
optIn.blankLineBeforeDocstring is backwards


gm2211:
This template is a guideline, not a strict requirement.
- **Version**: 2.4.0
- **Configuration**: optIn.blankLineBeforeDocstring
```
version = 2.4.0
optIn.blankLineBeforeDocstring = false
```

## Steps

Given code like this:
```scala
trait BackendClient {
  /** Returns the endpoint for authentication. */
  def auth(): AuthClient
}
```

When I run scalafmt via intellij, I get

## Problem

Scalafmt formats code like this:
```scala
trait BackendClient {

  /** Returns the endpoint for authentication. */
  def auth(): AuthClient
}
```

## Expectation

I would like the formatted output to look like this:
```scala
trait BackendClient {
  /** Returns the endpoint for authentication. */
  def auth(): AuthClient
}
```

## Workaround

I've found that by setting `optIn.blankLineBeforeDocstring = true` I get what I want, but not what I expect.

## Notes

This was also called out in a comment in the original PR https://github.com/scalameta/scalafmt/pull/1101/files#r162779035 - the comment was acknowledged, but it seems the feature was shipped flipped anyway, flipping the docs https://github.com/scalameta/scalafmt/pull/1101/commits/6328acadd3488ded7f64f937061a6d81bf9699dd instead of either renaming the flag (e.g. `optIn.allowNoBlankLineBeforeDocstring`) or setting it to true by default.

PR #1768, opened 2020-03-06
Fixed contradiction between blankLineBeforeDocstring name and value

poslegm:
Fixes https://github.com/scalameta/scalafmt/issues/1712

Because we can't just flip setting flag without breaking changes, I introduced new setting with similar name but reversed behavior (defaults maintain the same formatting). Old setting was deprecated.


scalafmt-core/shared/src/main/scala/org/scalafmt/config/OptIn.scala
```diff
@@ -76,10 +76,31 @@ case class OptIn(
...
+    * Setting behavior and name was mixed. After deprecation and right naming
+    * we need to:
+    * if `forceBlankLineBeforeDocstring` (new name) has default value (true)
+    *   fallback to `blankLineBeforeDocstring` (old config) which may be
+    *   configured in .scalafmt.conf
+    * if `forceBlankLineBeforeDocstring` configured to non-default value
+    *   don't look at the old name
+    * */
+  lazy val forceNewlineBeforeDocstringSummary: Boolean =
+    if (forceBlankLineBeforeDocstring) !blankLineBeforeDocstring
```
kitbellew: is it the same as:

```
forceBlankLineBeforeDocstring && !blankLineBeforeDocstring
```

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
```diff
@@ -59,8 +59,9 @@ object TokenOps {
     }
 
   def forceDocstringBlankLine(token: Token, style: ScalafmtConfig): Boolean = {
-    if (style.optIn.blankLineBeforeDocstring) false
-    else token.is[Comment] && token.syntax.startsWith("/**")
+    style.optIn.forceNewlineBeforeDocstringSummary &&
```
kitbellew: why not inline the logic here, as the only place we use it?

```
(style.optIn.forceBlankLineBeforeDocstring ||
  !style.optIn.blankLineBeforeDocstring) &&
  token.is[Comment] && token.syntax.startsWith("/**")
```

poslegm: Because I prefer to encapsulate deprecated flags handling in the `config` package and don't propagate it to the formatting algorithm

PR #1770, opened 2020-03-07
Fetch tags on CI

poslegm:
It may be a source of incorrect website rendering

Screen Shot 2020-03-07 at 9 00 50 PM


poslegm:
Only empirically after website deploy

PR #1771, opened 2020-03-07
BestFirstSearch: check optimal token not to overflow

kitbellew:
Follow the post-optimal states on the same line while we get exactly one active split to use; if the line overflows, don't mark optimal as found. This way, we are more likely to fit within the column limit and select a better solution.

Also, now that BestFirstSearch logic dealing with optimal tokens has been improved, the suboptimal special case handling of infix expressions around blocks is no longer needed.

`scala-repos` diffs:
* BestFirstSearch: check optimal tok not to overflow: https://github.com/kitbellew/scala-repos/commit/fd8b18e8d4638c73724a343b343a9e9c83151c00?w=1
* Router: always use optimal token within infix op: https://github.com/kitbellew/scala-repos/commit/5a8132dc0f796690190079d2b8be6fd5d01398c1?w=1

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
@@ -1,5 +1,6 @@
 package org.scalafmt.internal
 
+import scala.annotation.tailrec
 import scala.collection.mutable
 import scala.meta.tokens.Token
 
@@ -201,14 +202,20 @@ private class BestFirstSearch private (
                     nextState.split.cost == 0 =>
                 val nextNextState =
                   shortestPath(nextState, token, depth + 1, maxCost = 0)
-                if (null != nextNextState) {
-                  optimalNotFound = false
-//                  logger.elem(split, splitToken, formatWriter.mkString(nextNextState.splits), tokens(nextNextState.splits.length))
-                  Q.enqueue(nextNextState)
+                val furtherState =
+                  if (null != nextNextState && style.activeForEdition_2020_03)
+                    traverseSameLine(nextNextState, depth)
+                  else nextNextState
+                if (null != furtherState) {
+                  if (furtherState.column > style.maxColumn)
+                    Q.enqueue(nextNextState)
+                  else {
+                    optimalNotFound = false
+                    Q.enqueue(furtherState)
+                  }
                 } else if (!killOnFail &&
                   nextState.cost - curr.cost <= maxCost) {
                   // TODO(olafur) DRY. This solution can still be optimal.
-//                  logger.elem(split, splitToken)
                   Q.enqueue(nextState)
                 } else { // else kill branch
                   if (updateBest) best.remove(curr.depth)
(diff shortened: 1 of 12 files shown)


kitbellew:
@olafurpg @tanishiking @poslegm PTAL. another PR to combat line overflow, mostly around infix expressions.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
```diff
@@ -215,14 +202,20 @@ private class BestFirstSearch private (
                     nextState.split.cost == 0 =>
                 val nextNextState =
                   shortestPath(nextState, token, depth + 1, maxCost = 0)
-                if (null != nextNextState) {
-                  optimalNotFound = false
-//                  logger.elem(split, splitToken, formatWriter.mkString(nextNextState.splits), tokens(nextNextState.splits.length))
-                  Q.enqueue(nextNextState)
+                val furtherState =
+                  if (null == nextNextState) nextNextState
+                  else if (!style.activeForEdition_2020_03) nextNextState
```
poslegm:
```suggestion
                  if (null == nextNextState || !style.activeForEdition_2020_03) nextNextState
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
```diff
@@ -251,6 +244,50 @@ private class BestFirstSearch private (
     null
   }
 
+  private def getActiveSplits(state: State, ft: FormatToken): Seq[Split] = {
+    val splits =
+      if (state.formatOff) List(provided(ft))
+      else if (ft.inside(range)) routes(state.depth)
+      else List(provided(ft))
```
poslegm:
```suggestion
      if (ft.inside(range)) routes(state.depth)
      else List(provided(ft))
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
```diff
@@ -251,6 +244,50 @@ private class BestFirstSearch private (
...
+  @tailrec
+  private def traverseSingleSplitPath(state: State, depth: Int): State =
+    if (state.depth >= tokens.length) state
+    else {
+      trackState(state, depth, 0)
+
+      val splitToken = tokens(state.depth)
+      val activeSplits = getActiveSplits(state, splitToken)
+
+      if (activeSplits.length != 1)
```
poslegm:
```suggestion
      if (activeSplits.lengthCompare(1) != 0)
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
```diff
@@ -251,6 +244,50 @@ private class BestFirstSearch private (
...
+        if (activeSplits.length != 0) state else null // dead end
```
poslegm:
```suggestion
        if (activeSplits.nonEmpty) state else null // dead end
```

PR #1772, opened 2020-03-08
#1627 [02]: Router: ignore line breaks for blocks

### LINKED ISSUE #271: } else

kitbellew:
While previously we were trying to keep blocks without line breaks if that's how they were originally input, now we actively attempt to take a multi-line representation and fold it into a single line (fold), or take a single-line representation and unfold it into multiple lines (unfold). In the [keep] case, we'll keep the line break if it was there before.

Fixes #271 #1002 #1043
Helps with #1627

`scala-repos` diffs:
* _classic_: none
* _keep_: https://github.com/kitbellew/scala-repos/commit/51c4d1e3ad2cdbcdce2ba29aea78cb1209d25f94?w=1
* _fold_: https://github.com/kitbellew/scala-repos/commit/242d091577f73a3e3f3e74c29458e76c79145def?w=1
* _unfold_: https://github.com/kitbellew/scala-repos/commit/0a882e3102bf94c134d5d0e07904076ee0cb2add?w=1

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -216,9 +216,9 @@ class Router(formatOps: FormatOps) {
               .andThen(decideNewlinesOnlyAfterToken(arrowOptimal))
           }
 
-        def getSingleLineDecisionPre2019Nov = leftOwner.parent match {
-          case Some(_: Term.If | _: Term.Try | _: Term.TryWithHandler) => null
-          case _ => Policy.emptyPf
+        def getSingleLineDecisionPre2019NovOpt = leftOwner.parent match {
+          case Some(_: Term.If | _: Term.Try | _: Term.TryWithHandler) => None
+          case _ => Some(Policy.emptyPf)
         }
         def getSingleLineDecisionFor2019Nov = {
           type Classifiers = Seq[Classifier[Token, _]]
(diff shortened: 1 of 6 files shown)


kitbellew:
@olafurpg @tanishiking @poslegm PTAL, 2nd of many #1627 PRs.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -243,21 +243,35 @@ class Router(formatOps: FormatOps) {
...
           if (newlines > 0) null
           else if (style.activeForEdition_2019_11)
             getSingleLineDecisionFor2019Nov
           else
             getSingleLineDecisionPre2019Nov
 
+        def getSingleLineLambdaDecision: Policy.Pf = {
+          val ok = !style.newlines.alwaysBeforeCurlyBraceLambdaParams &&
+            getSpaceAndNewlineAfterCurlyLambda(newlines)._1
+          if (ok) getSingleLineDecisionFor2019Nov else null
```
poslegm: I am concerned about an increasing number of nulls here. It may be error-prone in the future maintaining. Does `Option` really beats perfomance in this case?

kitbellew: probably not. i can try to see if we can change to Option instead, the problem earlier was that Split takes a Policy, so i had to do `policyOpt.orNull` anyway.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -276,7 +290,8 @@ class Router(formatOps: FormatOps) {
           Split(Space, 0)
             .notIf(
               style.newlines.alwaysBeforeCurlyBraceLambdaParams ||
-                isSelfAnnotation || lambdaPolicy == null
+                isSelfAnnotation || lambdaPolicy == null ||
+                (style.newlines.source eq Newlines.keep) && newlines != 0
```
kitbellew: > 1. Why `keep` adds newline before lambda argument?

this is the line. if newline is present, the Space split is disabled. (of course, it's up to us to define what `keep` does.)

ah, one more thing which I should have mentioned: the `scala-repos` diffs are not generated by running one version of scalafmt and then another (because the next one can't undo what the first one did). instead, both ran on the same original, pre-scalafmt commit, and the diff is the comparison between results. that's how a space split becomes a newline in `keep`.

Issue #1424, opened 2019-05-24
Multiline comments have * alignment changed with "format: off"


huonw:
- **Version**: 0.5
- **Integration**: sbt
- **Configuration**:
```
version=2.0.0-RC5
```

## Steps

Given code like this:
```scala
// format: off
/*
        *
*

*/

/**
 * Javadoc style
 *
 */
// format: on
```

When I run scalafmt like this:

```bash
sbt scalafmt 
```

## Problem

Scalafmt formats code like this:
```scala
// format: off
/*
 *
 *

 */

/**
  * Javadoc style
  *
  */
// format: on
```

Notice the indentation has changed, including removing the "javadoc" alignment.

## Expectation

I would like the formatted output to look like this (same as it was originally):
```scala
// format: off
/*
        *
*

*/


/**
 * Javadoc style
 *
 */
// format: on
```

## Workaround

If it's a whole file that needs to have this sort of formatting preserved, `project.excludeFilters = [ "path/to/file" ]` seems to stop the formatting entirely for that file, including of comments.

kitbellew:
@huonw is it still a problem?

huonw:
Sorry, I've changed projects and am no longer using Scala. If the reproducer still reproduces, this is likely still a problem for that project; if it does not (the output matches the expectation), then it is not a problem.

PR #1773, opened 2020-03-09
"// format: off" enabled for comments, literals, etc.

poslegm:
Fixes https://github.com/scalameta/scalafmt/issues/1424

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -73,7 +73,7 @@ class FormatWriter(formatOps: FormatOps) {
           sb.append(rewrittenToken)
       }
 
-      entry.formatWhitespace(sb)
+      entry.formatWhitespace(sb, state.prev.formatOff)
     }
 
     sb.toString()
@@ -320,7 +320,7 @@ class FormatWriter(formatOps: FormatOps) {
         }
       }
 
-      def formatWhitespace(sb: StringBuilder): Unit = {
+      def formatWhitespace(sb: StringBuilder, formatOff: Boolean): Unit = {
 
         import org.scalafmt.config.TrailingCommas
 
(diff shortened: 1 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -71,6 +72,11 @@ class FormatWriter(formatOps: FormatOps) {
       }
 
       entry.formatWhitespace(sb)
```
kitbellew: just curious: will this introduce any (unwanted) alignment here if `format:off` had been used?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -28,12 +27,14 @@ class FormatWriter(formatOps: FormatOps) {
...
     val locations = getFormatLocations(state, debug = false)
 
+    var formatOff = false
     locations.iterate.foreach { entry =>
       val location = entry.curr
       val state = location.state
       val formatToken = location.formatToken
 
       formatToken.left match {
+        case token if formatOff => sb.append(token.syntax)
```
kitbellew: optional:

```
  case token if state.prev.formatOff => sb.append(...)
```

and get rid of `formatOff` variable.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
poslegm: Good catch, fixed in https://github.com/scalameta/scalafmt/pull/1773/commits/1589507b3df085fa80eb5ac28210eb9fe4307406

kitbellew: Sorry, didn't occur to me before: what about multiple spaces in the original code? Perhaps we should just add the contents of FormatToken.between?

poslegm: Added test for multiple spaces https://github.com/scalameta/scalafmt/pull/1773/commits/3185a373616ac507bb0fe4254370307b09379493 Or did you mean other cases?

PR #1774, opened 2020-03-09
#1627 [03]: Router: newlines.source=unfold for semicolons

kitbellew:
For semicolons, we can `unfold` by adding a line break. In the `fold` case, we can try to remove a line break but that doesn't really look that good, plus semicolons shouldn't generally be used anyway.

`scala-repos` diffs:
* _unfold_: https://github.com/kitbellew/scala-repos/commit/2ffa7f60a0aaf4443ceb5af8d407112d827af87d?w=1
* others: unchanged

Helps with #1627


kitbellew:

PR #1776, opened 2020-03-09
Router: modify indent of case clauses consistently

### LINKED ISSUE #1244: Wrong indentation of closing paren if there are newlines and parentheses between "case" and "=>"
- **Version**: 1.6.0-RC4
- **Integration**: sbt
- **Configuration**:
```
# default
```

## Steps

Given code like this:
```scala
x match {
  case Foo(
      a,
      b
  ) =>
    ???
}
```

```bash
$ sbt scalafmt
```

## Problem

```scala
x match {
  case Foo(
      a,
      b
      ) =>
    ???
}
```

## Expectation

```scala
x match {
  case Foo(
      a,
      b
  ) =>
    ???
}
```

## Workaround

There are no workarounds as far as I know.

kitbellew:
Fixes #1244. `scala-repos` diffs: https://github.com/kitbellew/scala-repos/commit/2877c2cc11ac8f6678023498766614669dfefa44

The existing logic was suppressing case pattern indentation which led to incorrect formatting relative to the end of the pattern.

Specifically, here's a problematic pattern:
```
    case pattern(
        argument // indent 1: explicitly suppressed
        ) // closing: clause indentation
          infix another( // indent 2: increased to distinguish from "if"
            ...
          ) if somethingLong
          .methodCalledBySomethingLong =>
      body
```

First, we need to restore indent 1. Second, we need to address indent 2 and there are several options:

* we choose NOT to keep it different from `if`, as `if` can be easily determined as the last line at that level of indentation; that way, we don't add extra indentation for infix, closing parens etc., so it looks like this:
```
    case pattern(
          argument // indent 1: added
        )
        infix another( // indent 2: removed
          ...
        ) if somethingLong
          .methodCalledBySomethingLong =>
      body
```
* if we wanted to keep it different from the `if` guard, we would then have to increase indentation of closing paren, to align with indent 2, which in turn means increasing indent 1, which leads to a bit too much indentation; also, since `if` is frequently tucked after the closing paren on the previous line but could be split, it might lead to an odd formatting like this (and doesn't really emphasize `if` anyway, unless we require a line break before `if`):
```
    case pattern(
            argument // indent 1: added twice
          ) // extra indent
          infix another( // indent 2: unchanged
            ...
          ) if somethingLong
          .methodCalledBySomethingLong =>
      body
```
* finally, an option that was being discussed in the ticket looks a bit like this:
```
    case pattern(
      argument
    )
    infix another(
      ...
    ) if somethingLong
          .methodCalledBySomethingLong =>
      body
```
which is still quite hard to read, as it doesn't really emphasize the `case` keyword, which is an even harder problem when there's no body and another case clause right after:
```
    case pattern1(
      argument
    )
    infix another(
      ...
    ) if guard =>
    case pattern2(
      argument
    )
    infix another(
      ...
    ) if guard =>
```

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
@@ -609,4 +609,11 @@ object TreeOps {
       }
     }
 
+  def isChildOfCaseClause(tree: Tree): Boolean =
+    findTreeWithParent(tree) {
+      case t: Case => Some(tree ne t.body)
+      case _: Pat => None
+      case _ => Some(false)
+    }.isDefined
+
 }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -582,6 +582,8 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
         .isDefined)) 0
     else if (!isNewline &&
       !isSingleLineComment(formatToken.right)) 0
+    else if (style.activeForEdition_2020_03 &&
+      isChildOfCaseClause(owner)) 0
     else 2
   }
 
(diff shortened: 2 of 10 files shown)


kitbellew:
@olafurpg @poslegm PTAL

@tanishiking this addresses the issue you reported.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -609,4 +609,11 @@ object TreeOps {
       }
     }
 
+  def isChildOfCaseClause(tree: Tree): Boolean =
+    findTreeWithParent(tree) {
+      case t: Case => Some(tree ne t.body)
```
poslegm: What is reason to not use just `==`?

poslegm:
P.S. there is a lot formatting changes already. May be cut 2.5.0-RC1 after this PR merging?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
kitbellew: i don't really know how `==` is implemented for trees but i'd like to check that it's *the same exact* instance here rather than something which looks the same.

kitbellew:
> P.S. there is a lot formatting changes already. May be cut 2.5.0-RC1 after this PR merging?

can we get through the `fold/unfold` changes i have so far before we do so? i have 12 additional changes already prepared, have just been submitting them one at a time, to make reviewing easier. would you prefer it if i sent them all at once, in one PR?

poslegm:
Yes, of course. Submit it as you prefer :)

(But as I know scala-stewars don't sell PRs with RC releases. So we can cut it often enough)

PR #1778, opened 2020-03-11
#1627 [04]: Router: ignore line breaks in if-else and bodies

kitbellew:
Modify breaks around control expressions (`if-else`, `for`, `while` etc.) and their bodies.

Helps with #1627.

`scala-repos` diffs:
* `classic` and `keep`: unchanged
* `fold`: https://github.com/kitbellew/scala-repos/commit/50420e514f66910d77ed8a549c56ee212350a329?w=1
* `unfold`: https://github.com/kitbellew/scala-repos/commit/f3e2e427ab5b3b5f2ecf29f4d84679bd0061b1e2?w=1

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1314,7 +1314,7 @@ class Router(formatOps: FormatOps) {
             .withPolicy(SingleLineBlock(expire)),
           Split(Space, 1).withPolicy(breakOnlyBeforeElse)
         )
-      case FormatToken(close: T.RightParen, _, _) if (leftOwner match {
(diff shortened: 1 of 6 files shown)


kitbellew:

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1359,10 +1376,13 @@ class Router(formatOps: FormatOps) {
             case x => throw new UnexpectedTree[Term.If](x)
           }) =>
         val expire = leftOwner.asInstanceOf[Term.If].elsep.tokens.last
+        val noSpace = style.newlines.source match {
```
poslegm: 3-times repeated `noSpace` computation? May be extract it to a function?

kitbellew: done!

PR #1779, opened 2020-03-12
#1627 [05]: Router: ignore line breaks in def/val assignments

kitbellew:
Helps with #1627.

For fold/unfold cases, handle 'if' with/without braces consistently.

Also, use alwaysBeforeMultilineDef consistently for `def`; for some reason, it was not applied when def body ended in a closing `}` or used config-style parens.

`scala-repos` diffs:
* Router: use alwaysBeforeMultilineDef consistently: https://github.com/kitbellew/scala-repos/commit/21a7ff6cfbd0ef6faeb5e34f085feb75154f1e34?w=1
* Router: ignore line breaks in def/val assignments
  * _classic_: unchanged
  * _keep_: https://github.com/kitbellew/scala-repos/commit/c6f7305073e37113589c38f5abc5098f3273663e?w=1
  * _fold_: https://github.com/kitbellew/scala-repos/commit/4761e37e253bab868958ab62ccd92bbdc141989f?w=1
  * _unfold_: https://github.com/kitbellew/scala-repos/commit/459ab1872163c84013c8b0a8aecb703c8f1a3884?w=1

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1627,11 +1627,35 @@ class Router(formatOps: FormatOps) {
         else SingleLineBlock(expire, exclude = exclude)
       Seq(Split(Space, 0, policy = spacePolicy))
     } else {
-      val spacePolicy =
-        if (ft.hasBreak) null
-        else if (style.newlines.alwaysBeforeMultilineDef)
-          SingleLineBlock(expire, exclude = exclude)
-        else Policy.NoPolicy
+      val spacePolicy = style.newlines.source match {
+        case Newlines.classic =>
+          if (ft.hasBreak) null
+          else if (style.newlines.alwaysBeforeMultilineDef)
+            SingleLineBlock(expire, exclude = exclude)
+          else Policy.NoPolicy
+
+        case Newlines.keep =>
+          if (ft.hasBreak) null
+          else if (style.newlines.alwaysBeforeMultilineDef)
+            SingleLineBlock(expire)
+          else Policy.NoPolicy
+
+        case Newlines.unfold => SingleLineBlock(expire)
+
+        case Newlines.fold if style.newlines.alwaysBeforeMultilineDef =>
+          SingleLineBlock(expire)
+        case Newlines.fold =>
+          body match {
+            case _: Term.Try | _: Term.TryWithHandler =>
+              SingleLineBlock(expire)
+            case t: Term.If =>
+              if (t.elsep.tokens.isEmpty)
+                SingleLineBlock(t.cond.tokens.last)
+              else
+                SingleLineBlock(expire)
+            case _ => SingleLineBlock(expire, exclude = exclude)
+          }
+      }
       Seq(
         Split(Space, 0, policy = spacePolicy).notIf(spacePolicy == null),
         Split(Newline, 1).withIndent(2, expire, Left)
@@ -1665,7 +1689,7 @@ class Router(formatOps: FormatOps) {
 
     def baseSpaceSplit =
       Split(Space, 0).notIf(isSingleLineComment(ft.right))
-    def twoBranches =
+    def twoBranches = util.Left(
       baseSpaceSplit
         .withOptimalToken(optimal)
         .withPolicy {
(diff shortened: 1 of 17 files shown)


kitbellew:

scalafmt-tests/src/test/resources/default/Lambda.stat
```diff
@@ -115,12 +115,13 @@ def foo(x: Int) = x match {
     (y: Int) => -y
   }
 >>>
-def foo(x: Int) = x match {
-  case 1 =>
-    (y: Int) => y
-  case _ =>
-    (y: Int) => -y
-}
+def foo(x: Int) =
```
poslegm: Don't understand why `x match {` was not left on its line (like in previous test with single line lambda)

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1103,26 +1131,68 @@ class Router(formatOps: FormatOps) {
...
+                      case Decision(ft, s)
+                          if s.isEmpty && ft.right.is[Token.KwYield] =>
+                        Seq(Split(Space, 0))
+                    }
+                    SingleLineBlock(expire).andThen(noBreakOnYield)
+                  case _: Term.Try | _: Term.TryWithHandler =>
+                    null // we force newlines in try/catch/finally
+                  case _ =>
+                    val ko = okNewline && leftOwner.is[Defn]
+                    if (ko) SingleLineBlock(expire) else NoPolicy
```
poslegm: `ko`?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1103,26 +1131,68 @@ class Router(formatOps: FormatOps) {
...
-            val noNewline = jsNative
-            val spacePolicy: Policy = rhs match {
-              case _: Term.If => twoBranches
-              case _: Term.ForYield => twoBranches
-              case _: Term.Try | _: Term.TryWithHandler
-                  if style.activeForEdition_2019_11 && !noNewline =>
-                null // we force newlines in try/catch/finally
-              case _ => NoPolicy
+            val okNewline = !isJsNative(right)
+            val spacePolicy: Policy = style.newlines.source match {
```
poslegm: It would be good to extract `newlines.source` machinery to another function. Main pattern matching block in `Router` is giant enough :c

scalafmt-tests/src/test/resources/default/Lambda.stat
kitbellew: this is because `newlines.alwaysBeforeMultilineDef` is `true` by default but was ignored in some cases (including this one). one of the commits in this PR fixes this inconsistency, so people can simply set the flag to `false` if this behaviour is not desired.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: will do, makes sense. will update the PR shortly.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: hmm... you're right, illogical.

scalafmt-tests/src/test/resources/default/Lambda.stat
poslegm: But why [this](https://github.com/scalameta/scalafmt/pull/1779/files#diff-a0f91ca7765a8af4af56f16fbffd361eR100-R109) test still correct?

Should it be
```scala
def foo(x: Int) =
  x match {
    case 1 => (y: Int) => y
    case _ => (y: Int) => -y
  }
```
?

kitbellew: i think it's because that test is marked as `SKIP`

poslegm: :smile: reasonable

Sorry for false alarm

kitbellew: nothing to be sorry about; in fact, i really appreciate that you review my changes carefully, the end result is superior, two heads are better than one!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: done

Issue #1534, opened 2019-10-18
Different configuration for scala and sbt files


limansky:
Hi,

 As a user I'd like to have ability to set different rules for different files. For example I'd lke to have alignment for `%` and `%%` in build.sbt, but I definitely don't want to align by `%` in my Scala code. For example, this doesn't make any sense:

```
   def baseTimeList(fromTime: Long, toTime: Long, table: Table) = {
     val startBaseTime = fromTime - (fromTime % table.rowTimeSpan)
     val stopBaseTime = toTime - (toTime      % table.rowTimeSpan)
     startBaseTime to stopBaseTime by table.rowTimeSpan
   }
```

I have following `aling.tokens` value:

```
align.tokens = [
  {code = "=>", owner = "Case"},
  {code = "%", owner = "Term.ApplyInfix"},
  {code = "%%", owner = "Term.ApplyInfix"}
]
```

poslegm:
I've already thought about it. It's really the right feature for sbt configs alignment.

poslegm:
Current workaround: just override scalafmt config inside file like [here]( https://github.com/scalameta/scalafmt/blob/master/project/Dependencies.scala#L5)

poslegm:
@limansky please let us know if config overriding is not enough for this issue solution.

poslegm:
Configs overriding documented on top of "Configuration" section: https://scalameta.org/scalafmt/docs/configuration.html

limansky:
Hi  @poslegm. I've added the line to build.sbt:

```
// scalafmt: { alignTokens = [ {code = "%", owner = "Term.ApplyInfix"}, {code = "%%", owner = "Term.ApplyInfix"} ] }
```

It seems like scalafmt ignore it.

poslegm:
What about 
```
// scalafmt: { align.tokens = [ {code = "%", owner = "Term.ApplyInfix"}, {code = "%%", owner = "Term.ApplyInfix"} ] }
```
?

limansky:
Works like a charm! Thanks!

The only thing, I don't like in this solution is that if I have multiple sbt files, I have to add this line to each of them.

limansky:
Another issue I observed (not related to this issue directly) is that scalafmt organize group name, artifact name and version in columns, but doesn't do it for the scope, e.g:

```
    libraryDependencies ++= Seq(
      "joda-time"      % "joda-time"   % versions.joda,
[...]

PR #1781, opened 2020-03-13
Document config overriding functionality

poslegm:
Scalafmt has powerful config overriding feature but nobody knows about it :(

Helps with #1534

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -13,17 +13,20 @@ align = more    // For pretty alignment.
 maxColumn = 100 // For my wide 30" display.
 ```
 
-There is a possibility to override scalafmt config for a specific file with
+There is a possibility to override scalafmt config for a specific code with
 `// scalafmt: {}` comment:
 
 ```scala mdoc:scalafmt
 ---
-// inside build.sbt
-// scalafmt: { align = most }
+// scalafmt: { align = most, danglingParentheses = false }
 libraryDependencies ++= Seq(
   "org.scalameta" %% "scalameta" % scalametaV,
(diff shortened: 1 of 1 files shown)


poslegm:
Is there reasons to hide it?

docs/configuration.md
```diff
@@ -13,6 +13,19 @@ align = more    // For pretty alignment.
 maxColumn = 100 // For my wide 30" display.
 ```
 
+There is a possibility to override scalafmt config for a specific file with
+`// scalafmt: {}` comment:
+
+```scala mdoc:scalafmt
+---
+// inside build.sbt
+// scalafmt: { align = most }
```
kitbellew: does it accept multiple parameters?

docs/configuration.md
```diff
@@ -13,6 +13,19 @@ align = more    // For pretty alignment.
 maxColumn = 100 // For my wide 30" display.
 ```
 
+There is a possibility to override scalafmt config for a specific file with
```
kitbellew: it's not really file, it's group of lines. i think you can use another comment further down, to modify again.

unfortunately, there's no `off` setting; if you want to restore, you need to set it explicitly.

docs/configuration.md
poslegm: yes, added another param into doc

docs/configuration.md
poslegm: specified it

olafurpg: It should be fairly easy to implement some kind of `off` setting to revert to the original settings

PR #1784, opened 2020-03-15
All tests enabled on CI (and fixed)

poslegm:
Based on https://github.com/scalameta/scalameta/pull/1995#issuecomment-599019837

1. Enabled all tests on CI
2. Resolved `scalameta-testkit` update problems in `DynamicSuite`
3. CLI tests moved from `scalafmt-tests` to `scalafmt-cli` (because I want to depend `scalafmt-dynamic` on `scalafmt-tests` for code reusing, but there were cyclic deps)


scalafmt-dynamic/src/test/scala/tests/DynamicSuite.scala
```diff
@@ -413,7 +413,7 @@ class DynamicSuite extends AnyFunSuite with DiffAssertions {
     f.assertFormat()
     val configOpt = f.dynamic.configsCache
       .getFromCache(f.config)
-      .flatMap(_.toOption.map(_._1))
+      .flatMap(_.fold(_ => None, Some(_)).map(_._1))
```
kitbellew: is this different?

build.sbt
```diff
@@ -223,9 +225,7 @@ lazy val tests = project
       scalatest.value
     )
   )
-  .dependsOn(
-    cli
-  )
+  .dependsOn(coreJVM)
```
kitbellew: is this equivalent?

scalafmt-dynamic/src/test/scala/tests/DynamicSuite.scala
poslegm: It's the same things but Scala 2.11's `Either` hasn't `toOption` method

build.sbt
poslegm: No, because `cli` dependent on `dynamic`

scalafmt-dynamic/src/test/scala/tests/DynamicSuite.scala
kitbellew: what about slightly simpler

```
val configOpt = f.dynamic.configsCache
      .getFromCache(f.config)
      .flatMap(_.fold(_ => None, Some(_._1)))
```

or even
```
val configOpt = f.dynamic.configsCache.getFromCache(f.config).collect {
  case Right((x, _)) => x
}
```

poslegm: Yes, it looks much better

olafurpg:
The convention I try to follow in my sbt builds today is to have a `tests` project that contains all test suites in the `Test` config and utility methods in the `Compile` config. I avoid spreading test suites across different modules because

*  you quickly hit on cyclic dependencies and need to start using `% Test` dependencies, which are an "advanced" sbt feature I think it is better to avoid if possible
* it's slow to run `sbt test` when you have a large number of modules (especially in JS/JVM cross-builds)
* it's faster and easier to discover test suites using tab completion in the sbt shell when they all live in the same project

poslegm:
Sounds reasonable. The problem was that tests already exist in different modules. Created issue to work on it later.

kitbellew:
@olafurpg @poslegm found a limitation with this approach. tests for the dynamic modules are passing today but attempts to upgrade sbt-scalafmt failed with classloader errors. turns out that the tests module depends on all modules hence all classes are available.

must move dynamic tests back into the dynamic module. not sure what build.sbt changes need to be made...

PR #1795, opened 2020-03-20
Preserve newlines in for comprehension generators and values 

### LINKED ISSUE #1256: Preserve newlines in for comprehension generators and values
Since newlines are preserved after `=` in `val`s and `def`s it seems reasonable to do the same inside `for`s and preserve newlines after `<-` as well as `=`:

Original:

    for {
      myLongBooleanValue <- 
        some.very.very.very.very.very.very.long.option.map(_ => true)
      myResult = 
        if (myLongBooleanValue) compute_one_thing
        else compute_another_thing
    } yield myResult

Scalafmt:

    for {
      myLongBooleanValue <- some.very.very.very.very.very.very.long.option
        .map(_ => true)
      myResult = if (myLongBooleanValue) compute_one_thing
      else compute_another_thing
    } yield myResult

The original version seems more readable to me and It would be nice if such inputs were not reformatted.

Using scalafmt 1.6.0-RC3, `style = default`

poslegm:
Fixes #1256

Before:
```scala
for {
  myLongBooleanValue <- some.very.very.very.very.very.very.long.option
    .map(_ => true)
  myResult = if (myLongBooleanValue) compute_one_thing
  else compute_another_thing
} yield myResult
```
After:
```scala
for {
  myLongBooleanValue <- 
    some.very.very.very.very.very.very.long.option.map(_ => true)
  myResult = 
    if (myLongBooleanValue) compute_one_thing
    else compute_another_thing
} yield myResult
```

Especially I don't like current formatting of long `if-else` in for comprehensions. `else` after split located on the same column with left-side identifiers. It makes code reading difficult.

scala-repos vs master: https://github.com/poslegm/scala-repos/commit/0addbc83430c0bb6d30cfc38daa862fb69741ba8

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1493,8 +1493,32 @@ class Router(formatOps: FormatOps) {
           if (style.align.arrowEnumeratorGenerator) StateColumn
           else Num(0)
         Seq(
-          // Either everything fits in one line or break on =>
-          Split(Space, 0).withIndent(indent, lastToken, After)
+          Split(Space, 0)
+            .onlyIf(newlines == 0)
+            .withIndent(indent, lastToken, After),
+          Split(Newline, 1).withIndent(indent, lastToken, After)
+        )
+      case tok @ FormatToken(arrow @ T.Equals(), _, _)
+          if leftOwner.is[Enumerator.Val] =>
+        Seq(
+          Split(Space, 0).onlyIf(newlines == 0),
+          Split(Newline, 1)
+        )
+      case tok @ FormatToken(_, arrow @ T.LeftArrow(), _)
+          if rightOwner.is[Enumerator.Generator] =>
+        val lastToken = rightOwner.tokens.last
+        fitsOneLineOrBreakOnArrow(
+          lastToken,
+          arrow,
+          Seq(Indent(2, lastToken, After))
+        )
+      case tok @ FormatToken(_, arrow @ T.Equals(), _)
+          if rightOwner.is[Enumerator.Val] =>
+        val lastToken = rightOwner.tokens.last
+        fitsOneLineOrBreakOnArrow(
+          lastToken,
+          arrow,
+          Seq(Indent(2, lastToken, After))
         )
       case FormatToken(T.KwYield(), _, _) if leftOwner.is[Term.ForYield] =>
         if (style.newlines.avoidAfterYield && !rightOwner.is[Term.If]) {
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -962,7 +962,7 @@ class Router(formatOps: FormatOps) {
         val lastToken = leftOwner.tokens.last
         val indent: Length =
           if (shouldBreakAfterArrowInFor && right.is[T.LeftBrace]) Num(-2)
-        else if (right.is[T.LeftBrace]) Num(0)
+          else if (right.is[T.LeftBrace]) Num(0)
           else if (style.align.arrowEnumeratorGenerator) StateColumn
           else Num(0)
         Seq(
@@ -977,7 +977,7 @@ class Router(formatOps: FormatOps) {
         val lastToken = leftOwner.tokens.last
         val indent: Length =
           if (shouldBreakAfterArrowInFor && right.is[T.LeftBrace]) Num(-2)
-        else if (right.is[T.LeftBrace]) Num(0)
+          else if (right.is[T.LeftBrace]) Num(0)
           else if (style.align.arrowEnumeratorGenerator) StateColumn
           else Num(0)
         Seq(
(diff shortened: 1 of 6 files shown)


scalafmt-tests/src/test/resources/align/ArrowEnumeratorGenerator.stat
```diff
@@ -13,4 +14,63 @@ for {
...
+  cond <- if (variable) doSomething
+         else doAnything
+} yield cond
+<<< insert newlines to for if no newlines origin
+for {
+  variable <- record.field.field1.map(_.toString)
+  cond <- if (variable) doSomething else doAnything
+} yield cond
+>>>
+for {
```
poslegm: It's really strange formatting

```scala
  cond <- if (variable) doSomething
         else doAnything
```

Is it by design of `align.arrowEnumeratorGenerator = true`?

(It is not new behavior from this PR, formatting on v2.4.2 is same)

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
```diff
@@ -142,6 +142,12 @@ case class Split(
         case x => copy(indents = Indent(x, expire, when) +: indents)
       }
 
+  def withIndent(indent: Indent[Length]): Split =
```
kitbellew: i'd switch the implementation with the one just above. why deconstruct `Indent` object, only to put back again?

scalafmt-tests/src/test/resources/align/ArrowEnumeratorGenerator.stat
kitbellew: agreed. i think someone complained on gitter about this.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1540,15 +1545,16 @@ class Router(formatOps: FormatOps) {
           Split(Newline, 1)
         )
       case tok @ FormatToken(_, arrow @ T.LeftArrow(), _)
-          if rightOwner.is[Enumerator.Generator] =>
+          if rightOwner
+            .is[Enumerator.Generator] && shouldBreakAfterArrowInFor =>
```
kitbellew: ocd. i'd break after `&&` :)

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -956,6 +956,33 @@ class Router(formatOps: FormatOps) {
             .onlyIf(open.pos.endLine == close.pos.startLine)
             .withSingleLine(close, killOnFail = true)
         ) ++ oneArgPerLineSplits
+
+      case tok @ FormatToken(T.LeftArrow(), right, _)
+          if leftOwner.is[Enumerator.Generator] =>
+        val lastToken = leftOwner.tokens.last
+        val indent: Length =
+          if (style.align.arrowEnumeratorGenerator) StateColumn
+          else if (right.is[T.LeftBrace]) Num(-2)
```
kitbellew: this looks unusual. what does this do?

poslegm: The same as https://github.com/scalameta/scalafmt/blob/master/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala#L376-L378

Formatting without `-2`:
```scala
xs.map {
  case x => {
      println("x")
    }
}

for {
  x <- {
       println("x")
     }
} yield x
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
poslegm: Because I want to keep laziness in argument computation of the above method. Yes, it's a little bit inefficient but I don't know how to
1. keep lazy arguments version
2. add method which receives constructed `Indent`

without code duplication.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
poslegm: reformatted

scalafmt-tests/src/test/resources/align/ArrowEnumeratorGenerator.stat
poslegm: created issue https://github.com/scalameta/scalafmt/issues/1797

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
poslegm: I see your solution. We can merge https://github.com/scalameta/scalafmt/pull/1796 first and then I'll reuse it here

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: ah, thank you, so that's why it was there in that case statement. always puzzled me.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
kitbellew: i had the same exact need a couple of days ago, solved it like this:

https://github.com/scalameta/scalafmt/pull/1796/commits/c05e44f5e54a2d27c6ce2a0d465e652633758641#diff-64605fc3e29e25f92371a46c89a960c7R161

poslegm:
@kitbellew yes, I am going to resolve conflicts and merge it today

PR #1796, opened 2020-03-21
#1627 [10]: Router: ignore line breaks in select chains

kitbellew:
Helps with #1627.

`scala-repos` diffs:
* _fold_: https://github.com/kitbellew/scala-repos/commit/09d4b43104eafdf9784b3397bc3495bd34ef7218
* _unfold_: https://github.com/kitbellew/scala-repos/commit/15d30f82ad4090d9d68f0c1dfd9cb9e8ae954d70
* _keep_: https://github.com/kitbellew/scala-repos/commit/b3267ad3da764df77385ec4a6ce4df2da47a5aad
* _classic_: unchanged

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
@@ -26,7 +26,7 @@ case class Split(
     modification: Modification,
     cost: Int,
     tag: SplitTag = SplitTag.Active,
-    indents: Vector[Indent[Length]] = Vector.empty[Indent[Length]],
+    indents: Seq[Indent[_]] = Seq.empty,
     policy: Policy = NoPolicy,
     optimalAt: Option[OptimalToken] = None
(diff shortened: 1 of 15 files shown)


kitbellew:

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
```diff
@@ -26,7 +26,7 @@ case class Split(
     modification: Modification,
     cost: Int,
     tag: SplitTag = SplitTag.Active,
-    indents: Vector[Indent[Length]] = Vector.empty[Indent[Length]],
+    indents: Seq[Indent[_]] = Seq.empty,
```
poslegm: Why was `Length` eliminated?

kitbellew: Length is the base type for both `Num` and `StateColumn`, so it's not needed.

PR #1806, opened 2020-03-22
#1627 [12]: Router: ignore line breaks in enumerators

kitbellew:
Helps with #1627. `scala-repos` diffs:

* _fold_: https://github.com/kitbellew/scala-repos/commit/496e6c6b0d8ec3e8eb1b8d91f618a42a3a3617e4?w=1
* _unfold_: https://github.com/kitbellew/scala-repos/commit/9043367b87139aded36e6b6c1737f70d3c874275?w=1
* _keep_: https://github.com/kitbellew/scala-repos/commit/52486acaa1e141dbc8805d0c57e542c5f2f4a67f?w=1
* _classic_: unchanged

Also, add a follow-up to changes in #1795, diffs at https://github.com/kitbellew/scala-repos/commit/5e73de35b1e365ee2ee17ba273f344ea6b71f1fc
* the closing paren in `ForYield` was unindented late so there was an extra indentation in some cases
* simplify the delayed arrow logic (in case clauses, the rule applies it on `case`, while the enumerator rules is applied just before arrow, hence it can simply be done right after)

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1333,9 +1333,11 @@ class Router(formatOps: FormatOps) {
         val indent: Length =
           if (style.align.ifWhileOpenParen) StateColumn
           else style.continuationIndent.callSite
+        val oldClassic = !style.activeForEdition_2020_03 &&
+          style.newlines.sourceIs(Newlines.classic)
         Seq(
           Split(NoSplit, 0)
-            .withIndent(indent, close, After)
+            .withIndent(indent, close, if (oldClassic) After else Before)
             .withPolicy(penalizeNewlines)
         )
       case FormatToken(T.KwIf(), _, _) if leftOwner.is[Term.If] =>
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1942,16 +1942,26 @@ class Router(formatOps: FormatOps) {
   private def getSplitsEnumerator(
       ft: FormatToken
   )(implicit style: ScalafmtConfig): Seq[Split] = {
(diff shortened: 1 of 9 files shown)


kitbellew:

scalafmt-tests/src/test/resources/align/ArrowEnumeratorGenerator.stat
```diff
@@ -1,5 +1,5 @@
 align.arrowEnumeratorGenerator = true
-maxColumn = 40
+maxColumn = 38
```
poslegm: Why?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1970,44 +1951,54 @@ class Router(formatOps: FormatOps) {
...
+      Seq(Split(Space.orNL(ft.noBreak), 0).withIndent(2, expire, After))
+    else if (style.newlines.sourceIs(Newlines.keep))
+      Seq(
+        if (ft.noBreak) Split(Space, 0)
+        else Split(Newline, 0).withIndent(2, expire, After)
+      )
+    else {
+      val spaceSplit = style.newlines.source match {
+        case _
+            if ft.right.is[T.LeftBrace] && isEnclosedInMatching(rightOwner) =>
```
poslegm: may be move it to another `else if` block above 1970 line?

scalafmt-tests/src/test/resources/align/ArrowEnumeratorGenerator.stat
kitbellew: Otherwise it prefers break and then all on one line which defeats the purpose of this test.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: The reason I did this was because it uses the same three splits as the others below 1970. Above were the special cases.

PR #1810, opened 2020-03-25
#1627 [14]: Router: single-line Defn lambda for fold only

### LINKED ISSUE #1809: Keep lambda with single definition statement on separate line
- **Version**: 2.4.2
- **Integration**: cli
- **Configuration**: defaults
```
version = 2.4.2

```

## Steps

Given code like this:
```scala
    files.par.foreach { p =>
      val td = mtags.index(p.toLanguage, p)
    }
```

## Problem

```scala
    files.par.foreach { p => val td = mtags.index(p.toLanguage, p) }
```

## Expectation

I would like the formatted output to remain unchanged. The body is a definition, I'm planning to write another statement. 

## Workaround

I've found that by adding an extra dummy statement I get the expected behavior.

kitbellew:
Fixes #1809. Helps with #1627.


kitbellew:

Issue #1829, opened 2020-03-30
Interest in containerizing scalafmt?


bsolomon1124:
`scalafmt` as a standalone CLI tool seems like a good candidate for containerization given its lack of support in package managers.  (See for example: https://github.com/koalaman/shellcheck.)

Here is a working Dockerfile:

```
FROM openjdk:8
ARG SCALAFMT_VERSION=2.4.2
ARG SCALAFMT_INSTALL_LOCATION=/usr/local/bin/scalafmt
RUN set -ex \
    && apt-get update -y \
    && rm -rf /var/lib/apt/lists/* \
    && echo insecure > "$HOME/.curlrc" \
    && curl -Lo coursier https://git.io/coursier-cli \
    && chmod u+x coursier \
    && ./coursier bootstrap "org.scalameta:scalafmt-cli_2.12:$SCALAFMT_VERSION" \
        -r sonatype:snapshots \
        -o "$SCALAFMT_INSTALL_LOCATION" \
        --main org.scalafmt.cli.Cli \
    && rm -f coursier \
    && scalafmt --version
CMD ["/usr/local/bin/scalafmt"]
```

Verification:

```
$ docker container run --rm scalafmt-test:4 scalafmt --version
scalafmt 2.4.2
```

The resulting image is 573MB, which adds an incremental 60MB over openjdk:8.

poslegm:
Related: https://github.com/scalameta/scalafmt/issues/1678

Personally I don't see the need to create an Docker image for scalafmt. 

But if some volunteer want  to:
1. create image
2. publish it to public image registry
3. integrate it into our release process
4. and will support in the future

...

Contributions are welcome!

bsolomon1124:
@poslegm would the maintainers of this project willing to host the images under an organization page such as https://hub.docker.com/u/scalameta/?  (I see that org doesn't exist on Docker Hub currently.)

poslegm:
@olafurpg what about Docker Hub account for scalameta projects?

olafurpg:
I just pushed a docker container that you can run like this
```
❯ docker run scalameta/scalafmt:2.4.2 scalafmt --help
```
Does it work for you?

My biggest concern is keeping this image up-to-date with releases. The JVM binary dynamically downloads the appropriate Scalafmt version so in theory I think we don't have to push a new docker image for every release. Any thoughts?

bsolomon1124:
Looks good @olafurpg!  The pull/run is working for me.

[...]

Issue #1678, opened 2020-02-06
FEATURE REQUEST: provide an officially built docker image 


jeff-knurek:
## Problem

Searching for docker image for `scalafmt` results in a lot of uncertainty of what is available, and which version is applicable to the official releases. 

The installation documentation provides lots of formats to install the app into, https://scalameta.org/scalafmt/docs/installation.html
but when running the tool in CI, there's no easy way to include it into the build process. 

## Expectation
There's an official docker image at 
```
https://hub.docker.com/u/scalameta/scalafmt
```
or any other public docker registry
AND
the information about using it is included in the installation docs: https://scalameta.org/scalafmt/docs/installation.html

## Workaround

--EDIT--
the image from https://github.com/mroth/scalafmt-native actually seems to be the best workaround. 

----
There's a community image from https://github.com/wellcomecollection/dockerfiles/blob/master/scalafmt/Dockerfile
that seems to be "usable". But the last update was in August 2019, and the versioning does not align with https://github.com/scalameta/scalafmt/releases
Plus, that image comes with lots of CVEs.

poslegm:
Done, Docker image available [here](https://hub.docker.com/layers/scalameta/scalafmt/v2.6.4/images/sha256-50708925bd69cf70c99d36b6edfde49a00fa613e6ab77144f347769a80c3da2a?context=explore)

Issue #1828, opened 2020-03-30
Fails with FailedToFormat exception out of the box on Debian slim


bsolomon1124:
Following https://scalameta.org/scalafmt/docs/installation.html#native-image:

Dockerfile:

```
FROM debian:stable-slim
ARG VERSION=2.4.2
ARG INSTALL_LOCATION=/usr/local/bin/scalafmt-native
RUN set -ex \
    && apt-get update -y \
    && apt-get install -y --no-install-recommends curl unzip \
    && echo insecure > "$HOME/.curlrc" \
    && curl https://raw.githubusercontent.com/scalameta/scalafmt/master/bin/install-scalafmt-native.sh | \
        sh -s -- $VERSION $INSTALL_LOCATION
```

Invocation:

```
docker container run --rm ^
    --volume C:\Users\myname\dev\project-name:/project-name ^
    scalafmt-test:1 ^
    sh -c "scalafmt-native project-name/path/to/file.scala"
```

Traceback:

```
root@11615410ad72:/# scalafmt-native /project-name/path/to/file.scala
org.scalafmt.cli.FailedToFormat: /project-name/path/to/file.scala
Caused by: java.lang.NoSuchMethodException: org.scalafmt.internal.FormatWriter$$anon$2.last()
        at java.lang.Class.getMethod(DynamicHub.java:1053)
        at org.scalafmt.internal.FormatWriter.reflMethod$Method1(FormatWriter.scala:564)
        at org.scalafmt.internal.FormatWriter.alignmentTokens(FormatWriter.scala:564)
        at org.scalafmt.internal.FormatWriter$FormatLocations.<init>(FormatWriter.scala:258)
        at org.scalafmt.internal.FormatWriter.getFormatLocations(FormatWriter.scala:154)
        at org.scalafmt.internal.FormatWriter.mkString(FormatWriter.scala:26)
        at org.scalafmt.Scalafmt$.formatCode(Scalafmt.scala:73)
        at org.scalafmt.cli.ScalafmtCoreRunner$.unsafeHandleFile(ScalafmtCoreRunner.scala:87)
        at org.scalafmt.cli.ScalafmtCoreRunner$.handleFile(ScalafmtCoreRunner.scala:66)
        at org.scalafmt.cli.ScalafmtCoreRunner$.$anonfun$run$3(ScalafmtCoreRunner.scala:44)
[...]

Issue #1769, opened 2020-03-07
wrong binary in 2.4.2 release?


ornicar:
https://github.com/scalameta/scalafmt/releases/download/v2.4.2/scalafmt-linux.zip

```
./scalafmt --version
scalafmt 2.4.1+33-b35acafc-SNAPSHOT`
```

And therefore it can't format projects where `version="2.4.2"`:

```
error: invalid Scalafmt version.

This Scalafmt installation has version '2.4.1+33-b35acafc-SNAPSHOT' and the version configured in '/home/thib/lila-fishnet/.scalafmt.conf' is '2.4.2'.
To fix this problem, add the following line to .scalafmt.conf:

version = '2.4.1+33-b35acafc-SNAPSHOT'

NOTE: this error happens only when running a native Scalafmt binary.
Scalafmt automatically installs and invokes the correct version of Scalafmt when running on the JVM.
```

poslegm:
Thank you for reporting. The issue may be in not running CI on tag push. I tried to fix it [here](https://github.com/scalameta/scalafmt/commit/33c33f49fc9ddbcba8db7b11a40dc9339c59c2f2), will look at the next release.

Current workaround is non-native images from [coursier](https://github.com/scalameta/scalafmt/commit/33c33f49fc9ddbcba8db7b11a40dc9339c59c2f2).

lbialy:
I guess much bigger issue is that this binary is completely broken:
```
org.scalafmt.cli.FailedToFormat: /home/lbialy/Projects/x/y/z/src/main/scala/com/whatever/configuration.scala
Caused by: java.lang.NoSuchMethodException: org.scalafmt.internal.FormatWriter$$anon$2.last()
        at java.lang.Class.getMethod(DynamicHub.java:1053)
        at org.scalafmt.internal.FormatWriter.reflMethod$Method1(FormatWriter.scala:564)
        at org.scalafmt.internal.FormatWriter.alignmentTokens(FormatWriter.scala:564)
        at org.scalafmt.internal.FormatWriter$FormatLocations.<init>(FormatWriter.scala:258)
        at org.scalafmt.internal.FormatWriter.getFormatLocations(FormatWriter.scala:154)
        at org.scalafmt.internal.FormatWriter.mkString(FormatWriter.scala:26)
        at org.scalafmt.Scalafmt$.formatCode(Scalafmt.scala:73)
[...]

PR #1831, opened 2020-03-31
WIP: Add working Dockerfile for using scalafmt binary

bsolomon1124:
Closes:

- Issue #1829
- Issue #1678

@olafurpg , I think automating the `docker push` will require setting `DOCKER_USERNAME` (e.g. `scalameta`) and its corresponding `DOCKER_PASSWORD`:

> https://docs.travis-ci.com/user/docker/#pushing-a-docker-image-to-a-registry

A maintainer should be able to add those at https://github.com/scalameta/scalafmt/settings/secrets.

Working examples:

```bash
$ docker container run --rm scalafmt:2.4.2 scalafmt --version
scalafmt 2.4.2

$ cat myfile.scala 
object Bar extends App {
      1
}

$ docker container run -i --rm scalafmt:2.4.2 scalafmt --stdin < myfile.scala 
object Bar extends App {
  1
}
Reformatting...
```


bsolomon1124:
Woops, just saw that the CI has been moved to GitHub Actions.  Will fix.

olafurpg:
Thank you for this contribution! I have configured the DOCKER_USERNAME and DOCKER_PASSWORD secrets in the GitHub Actions UI (not Travis CI). Can you revert the .travis.yml changes and create a file `.github/workflows/release.yml` with the contents and update `...` with the docker instructions
```
name: Release
on:
  push:
    branches: [master]
    tags: ["*"]
jobs:
  docker:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: olafurpg/setup-scala@v7
      - uses: olafurpg/setup-gpg@v2
      - run: git fetch --unshallow
      - name: Publish ${{ github.ref }}
        run: |
          SCALAFMT_VERSION=$(git describe --tags | sed 's/^v//')
          docker login ...
        env:
          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
```

olafurpg:
I guess we can also remove `    branches: [master]` to only trigger this on tags

bsolomon1124:
@olafurpg, I see there are RC tags like `v0.1.0-RC0`. Presumably we'd want to exclude them, right?  I would guess `tags: ["*"]` will just match everything.

poslegm:
Hmm, we publish release candidates exactly the same as the normal releases. So it may be helpful to include it into docker releases too for early adoption of new versions

bsolomon1124:
Regarding the RC, I added logic to push those with their version tag, but if an RC tag is detected, to not co-tag that image with 'latest', since 'latest' is regarded as the latest stable version and is the default if none is specified.  Sound OK?

docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+FROM openjdk:8
+ARG SCALAFMT_VERSION
+ARG SCALA_VERSION=2.12
+ARG SCALAFMT_INSTALL_LOCATION=/usr/local/bin/scalafmt
+RUN set -ex \
+    && apt-get update -y \
+    && rm -rf /var/lib/apt/lists/* \
+    && curl -sSLo coursier https://git.io/coursier-cli \
+    && chmod u+x coursier \
+    && ./coursier bootstrap "org.scalameta:scalafmt-cli_$SCALA_VERSION:$SCALAFMT_VERSION" \
```
olafurpg: This assumes the CLI is already published to Maven Central. It can take 1-2hrs for the sync to complete so I wonder if want to build the binary instead. If the docker image is running on Linux we may as well use the native-image binary from here https://github.com/scalameta/scalafmt/blob/8f0ca529c02e4c9832d7bdc10e148bb50e2a2e3d/.github/workflows/ci.yml#L49

Maybe we can somehow depend on the native image artifact directly in a subsequent job by moving this logic to `ci.yml` 🤔 

Then we can also remove the dependency on the JDK, which could slim down the docker image.

bsolomon1124: Okay, I see ☹️ .  Let me toy around with that.

olafurpg: It's possible to download artifacts from other GH Action jobs using https://github.com/actions/download-artifact

olafurpg: You should be able to test this in the PR by commenting out the `docker logic` and `docker push` commands. We link native images on every PR

bsolomon1124: Ok, need to switch to some other work but will come back to this asap.

One question - can we ensure that `release` is run after `ci`?  Where is that ordering defined?

bsolomon1124:
So as I'm understanding it we basically want to mimic:

```
FROM buildpack-deps:buster-curl
ARG SCALAFMT_VERSION
ARG SCALAFMT_INSTALL_LOCATION=/usr/local/bin/scalafmt
SHELL ["/bin/bash", "-o", "pipefail", "-c"]
RUN set -ex \
    && apt-get update -y \
    && apt-get install -y --no-install-recommends unzip \
    && rm -rf /var/lib/apt/lists/* \
    && curl https://raw.githubusercontent.com/scalameta/scalafmt/master/bin/install-scalafmt-native.sh | \
      sh -s -- $SCALAFMT_VERSION $SCALAFMT_INSTALL_LOCATION
CMD ["$SCALAFMT_INSTALL_LOCATION"]
```

But we can go one step further and just directly copy the built binary artifact into the docker image, replacing the curl/sh entirely.

bsolomon1124:
I think for now that https://github.com/scalameta/scalafmt/issues/1828 is kind of in the way of using the native binary.

olafurpg:
@bsolomon1124 that issue has been fixed on master. Have you confirmed that the scalafmt native-image does not work in a docker linux container?

docker/Dockerfile
olafurpg: You can't ensure that `release` runs after `ci` but you can move the release job into the `ci.yml` file to work around it. Here is an example workflow that does something similar https://github.com/mroth/scalafmt-native/blob/c0932c026050ea87da68337e2345f676f23414c4/.github/workflows/build.yml#L100

bsolomon1124:
Going to take another stab at this one during this week now that Issue #1769 is closed.

bsolomon1124:
@olafurpg it looks like `secrets.DOCKER_USERNAME` and `secrets.DOCKER_PASSWORD` aren't being picked up here.  Do you know why that might be?

docker/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM debian:buster-slim
+ARG SCALAFMT_VERSION
+# TODO: (bsolomon1124) I've tried to use Alpine but gives problems at runtime
```
gurinderu, line 3: I added a musl build here https://github.com/scalameta/scalafmt/pull/2041
Now you can try an alpine image

bin/docker-publish.sh
```diff
@@ -0,0 +1,34 @@
...
+set -e
+
+# CD over to docker/
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+DOCKERDIR="$(dirname "$DIR")/docker"
+cd "$DOCKERDIR" || exit 1
+
+# Extract version, build and push
+SCALAFMT_VERSION=$(git describe --tags | sed 's/^v//')
+echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
```
gurinderu, line 16: I think it is better to use https://github.com/marketplace/actions/build-and-push-docker-images

poslegm:
Superseded by https://github.com/scalameta/scalafmt/pull/2059, docker image available [here](https://hub.docker.com/layers/scalameta/scalafmt/v2.6.4/images/sha256-50708925bd69cf70c99d36b6edfde49a00fa613e6ab77144f347769a80c3da2a?context=explore)

PR #1833, opened 2020-04-01
Fixed newlines on NewAnonymous with empty arguments lambda

### LINKED ISSUE #1281: One expression in multi-line lambda { } gets placed at the first line
- **Version**: Bundled with 2018.2.11 IntelliJ IDEA Scala plugin
- **Integration**: IntelliJ
- **Configuration**:
```
maxColumn = 120
align=more
docstrings = JavaDoc
```

## Steps

Given code like this:
```scala
      fontNames forEach { fn =>
        println(resources.getFont(fn))
      }
```

## Problem

```scala
      fontNames forEach { fn => println(resources.getFont(fn))
      }
```

## Expectation

I expect initial code to remain unchanged after formatting

Is there any way to achieve that?

poslegm:
Fixes #1281 

Currently `() => ...` in `NewAnonymous` always squashes:
```
<<<
val foo = new Bar { () =>
  println("hello")
}
>>>
val foo = new Bar { () => println("hello") }
```

it's little inconsistent with:
```
<<<
val foo = new Bar { x =>
  println("hello")
}
>>>
val foo = new Bar { x =>
  println("hello")
}
```

After this changes newlines would be respected in both cases.

Also added tests for `newlines.afterCurlyLambda=squash` from #1832

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -9,6 +9,7 @@ import org.scalafmt.util._
 
 import scala.collection.mutable
 import scala.language.implicitConversions
+import scala.meta.Term.NewAnonymous
 import scala.meta.classifiers.Classifier
 import scala.meta.tokens.{Token, Tokens}
 import scala.meta.tokens.{Token => T}
@@ -251,9 +252,12 @@ class Router(formatOps: FormatOps) {
           case Newlines.keep if newlines != 0 => None
           case Newlines.unfold => None
           case Newlines.fold =>
+            val isTopLevelBlock =
+              leftOwner.parent.exists(_.parent.isEmpty) ||
+                (leftOwner.is[Template] &&
+                  !leftOwner.parent.exists(_.is[NewAnonymous]))
             // do not fold top-level blocks
-            if (leftOwner.parent.exists(_.parent.isEmpty) ||
-              leftOwner.is[Template]) None
+            if (isTopLevelBlock) None
             else if (lambdaPolicy != null) getSingleLineLambdaDecisionOpt
             else Some(getSingleLineDecisionFor2019Nov)
           // old behaviour
@@ -327,6 +331,16 @@ class Router(formatOps: FormatOps) {
         val singleLineSplit =
           Split(Space, 0)
             .notIf(hasSingleLineComment)
+            // for constructors with empty args lambda
+            // new Foo { () =>
+            //   println("wow")
+            // }
+            .notIf(
+              leftOwner.parent.exists(_.is[Template]) &&
+                leftOwner.parent.exists(_.parent.exists(_.is[NewAnonymous])) &&
+                newlines > 0 &&
+                !style.newlines.sourceIs(Newlines.fold)
+            )
             .withPolicy(SingleLineBlock(endOfFunction))
         def newlineSplit =
           Split(Newline, 1 + nestedApplies(leftOwner))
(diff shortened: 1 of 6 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -251,9 +252,12 @@ class Router(formatOps: FormatOps) {
           case Newlines.keep if newlines != 0 => None
           case Newlines.unfold => None
           case Newlines.fold =>
+            val isTopLevelBlock =
+              leftOwner.parent.exists(_.parent.isEmpty) ||
+                (leftOwner.is[Template] &&
+                  !leftOwner.parent.exists(_.is[NewAnonymous]))
```
kitbellew, line 258: i wanted to exclude from folding cases like:
```
  new A {
    def foo = bar
  }
```
i think there's a limit to what can be put on one line and still be readable, and a class definition seemed like a good cases to disallow folding.

however, your test example of `new Bar { () => f(42) }` looks like something else. perhaps we should check the `lambdaPolicy` first?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -327,6 +331,16 @@ class Router(formatOps: FormatOps) {
...
           Split(Space, 0)
             .notIf(hasSingleLineComment)
+            // for constructors with empty args lambda
+            // new Foo { () =>
+            //   println("wow")
+            // }
+            .notIf(
+              leftOwner.parent.exists(_.is[Template]) &&
+                leftOwner.parent.exists(_.parent.exists(_.is[NewAnonymous])) &&
+                newlines > 0 &&
```
kitbellew, line 341: 1. i think since we are past the arguments here, the earlier rule should have prohibited a space here and forced a newline after `=>`; without this change, doesn't work?

2. we are not allowed to examine `newlines` for `unfold`. generally, `newlines` should only be used with `sourceIgnored`, as `logical(logical(newlines) || sourceIgnored)`

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -9,6 +9,7 @@ import org.scalafmt.util._
 
 import scala.collection.mutable
 import scala.language.implicitConversions
+import scala.meta.Term.NewAnonymous
```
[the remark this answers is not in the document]
olafurpg: I always recommend prefixing with `Term.` and `Type.` for consistency. Some types like `Term.Name` and `Type.Name` conflict with each other.

poslegm:
> why are empty arguments treated differently? perhaps i missed this part.

They have different AST: https://astexplorer.net/#/gist/4dacb87e5a2429311f8b739e6423ef73/b0cb7c62b42b6165d265bb16ad07292acd2b3f0a

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
poslegm: Yes, I missed this case. Thank you, will rework it

kitbellew: @poslegm i looked at this change trying to understand why it's different, and ended up changing a few things to make sure `squash` is applied in this case as well. (also reordered and added some tests, i like adding tests before a change and then see how it modifies their behaviour.)

perhaps you can use some of this:

https://github.com/kitbellew/scalafmt/commits/issue-1281

poslegm: Thank you for participating! Your implementation looks more reliable and consistent. I commited it here https://github.com/scalameta/scalafmt/pull/1833/commits/83a36d372039e17f8c8006140feb101856b491f0. The only problem is that it still folds constructors with `Defn`:

```
<<< 
 new A {
    def foo = bar
  }
>>>
new A { def foo = bar }
```

I changed it here: https://github.com/scalameta/scalafmt/pull/1833/commits/873363531ac1dc541f13c8ed0f4f7733f5e179a2

PR #1834, opened 2020-04-02
TokenOps bugfix: logic for newline on docstring

### LINKED ISSUE #1426: feature request: not to delete empty line between file head block comment and the imports
Require feature as following:

not to delete the empty line betweent block comment end  and the first import,
as I show in **don`t delete me** 

```
/** File Name    :
 *  Author       :
 *  Created Time :
 *  Description  :
 */
-----------------------don`t delete me----------------------------
import org.apache.spark.sql.{Dataset, SparkSession}

object SimpleApp {

}
```

kitbellew:
With `optIn.forceBlankLineBeforeDocstring=true`, in addition to forcing a blank line before docstring, we also mistakenly removed one after it. Fixes #1426.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
@@ -53,16 +53,21 @@ object TokenOps {
   def shouldGet2xNewlines(
       tok: FormatToken
   )(implicit style: ScalafmtConfig): Boolean =
-    !forceDocstringBlankLine(tok.left, style) && {
-      tok.hasBlankLine ||
-      (forceDocstringBlankLine(tok.right, style) && !tok.left.is[Comment])
-    }
+    tok.hasBlankLine || blankLineBeforeDocstring(tok.left, tok.right)
 
-  def forceDocstringBlankLine(token: Token, style: ScalafmtConfig): Boolean = {
+  def blankLineBeforeDocstring(
+      left: Token,
+      right: Token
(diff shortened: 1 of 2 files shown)


kitbellew:
@poslegm PTAL

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
```diff
@@ -53,16 +53,21 @@ object TokenOps {
...
-      (forceDocstringBlankLine(tok.right, style) && !tok.left.is[Comment])
-    }
+    tok.hasBlankLine || blankLineBeforeDocstring(tok.left, tok.right)
+
+  def blankLineBeforeDocstring(
+      left: Token,
+      right: Token
+  )(implicit style: ScalafmtConfig): Boolean =
+    right.is[Token.Comment] &&
+      blankLineBeforeDocstring(left, right.asInstanceOf[Token.Comment])
```
poslegm, line 63: What about

```scala
  def blankLineBeforeDocstring(
      left: Token,
      right: Token
  )(implicit style: ScalafmtConfig): Boolean =
    right match {
      case comment: Token.Comment =>
        style.optIn.forceNewlineBeforeDocstringSummary &&
          !left.is[Token.Comment] &&
          comment.syntax.startsWith("/**")
      case _ => false
    }
```
?
one function instead of two, get rid of `asInstanceOf`

kitbellew: i did it on purpose, i will use it in another commit to change the `Router` match for comments.

PR #1837, opened 2020-04-04
Align strip margin: allow disabling the alignment

kitbellew:
Previously, using `assumeStandardLibraryStripMargin = true` (which means
the `.stripMargin` method was not redefined by the user and thus assigns
no special meaning to indentation preceding the `|` character) resulted
in aligning the pipe character with the end of the opening triple-quote.

Now, one can optionally set `align.stripMargin = false` to use a 2-space 
indentation offset relative to the start of the opening line instead.

```
// With align.stripMargin = true
val x = """
          |Hello world!
          |""".stripMargin
// With align.stripMargin = false
val x = """
  |Hello world!
  |""".stripMargin
```

Also, find the string margin char rather than assuming it's always `|` and inspect any parameter to `.stripMargin`.

### DIFF
--- a/scalafmt-tests/src/test/resources/test/StripMargin.stat
+++ b/scalafmt-tests/src/test/resources/test/StripMargin.stat
@@ -33,10 +33,10 @@ val x = """Formatter changed AST
         .stripMargin('?')
 >>>
 val x = """Formatter changed AST
-          |=====================
-          |$diff
-      ?=====================
-      ?$diff
+      |=====================
+      |$diff
+          ?=====================
+          ?$diff
         """
   .stripMargin('?')
 <<< No align | margin 1, different char
@@ -51,10 +51,10 @@ val x = """Formatter changed AST
         .stripMargin('?')
 >>>
 val x = """Formatter changed AST
-          |=====================
-          |$diff
-      ?=====================
-      ?$diff
+      |=====================
+      |$diff
+          ?=====================
+          ?$diff
         """
   .stripMargin('?')
 <<< Align | margin 2
--- a/scalafmt-tests/src/test/resources/test/StripMargin.stat
+++ b/scalafmt-tests/src/test/resources/test/StripMargin.stat
@@ -1,8 +1,8 @@
 assumeStandardLibraryStripMargin = true
 <<< Align | margin 1
 val x = """Formatter changed AST
-          |=====================
-          |$diff
+      |=====================
+      |$diff
         """
         .stripMargin
 >>>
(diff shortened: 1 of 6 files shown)


kitbellew:

scalafmt-tests/src/test/resources/test/StripMargin.stat
```diff
@@ -1,15 +1,62 @@
 assumeStandardLibraryStripMargin = true
 <<< Align | margin 1
 val x = """Formatter changed AST
-          |=====================
-          |$diff
+      |=====================
```
olafurpg, line 4: This looks like a regression. The assuneStandardLibraryStripMargin setting aligns multilingual strings by default and we should try to keep that behavior.

olafurpg: The alignment is disabled by default since the assumption can produce semantic changes.

kitbellew: @olafurpg this is the *input* to the test (it was already aligned, so i decided to change it to emphasize the result). the output is unchanged. apologies for the confusion.

the only change to default is, `align.stripMargin` is set to false when `align = none` (seemed a natural thing to do). everything else is the same.

kitbellew: > The alignment is disabled by default since the assumption can produce semantic changes.

can you please share an example?

olafurpg: My bad 😅 

> The alignment is disabled by default since the assumption can produce semantic changes.

Users can implement their own `stripMargin` method with custom semantics where the indentation  spaces have significant meaning. A contrived example
```scala
implicit class XtensionStripMargin(string: String) {
  def stripMargin = string.length
}
"""a
  |b
  |""".stripMargin
```

What is the difference between these two configs?

```
align.stripMargin = true
```

and 
```
```

kitbellew: > What is the difference between these two configs?
> 
> ```
> assumeStandardLibraryStripMargin = true
> align.stripMargin = true
> ```
> 
> and
> 
> ```
> assumeStandardLibraryStripMargin = true
> ```

`assumeStandardLibraryStripMargin` allows the rewrite of indentation within the string constant (if it's false, nothing is modified); `align.stripMargin` defines the indentation (aligned or not).

```
// align.stripMargin = true
val a = s"""foo
           | bar""".stripMargin

// align.stripMargin = false
val a = s"""foo
  | bar""".stripMargin
```
it looks like a number of people like some sort of alignment, but we do not. in any form.

olafurpg: I see, thanks for the explanation. It would help if the PR description of changes like these included more details. For example

---
Previously, when using `assumeStandardLibraryStripMargin = true` it was not possible to enable vertical alignment of `|` characters in multiline `.stripMargin` strings using 2-space indentation. The `|` characters always aligned by the opening triplequote `"""`. Now, users can optionally enable `align.stripMargin = false` to align the `|` character at 2-spaces instead of by the opening triplequote.

```scala
// With align.stripMargin = true
val x = """
        |Hello world!
        |""".stripMargin
// With align.stripMargin = false
val x = """
  |Hello world!
  |""".stripMargin
```

kitbellew: will do!

kitbellew:
> ... since the following doesn't work
> 
> ```scala
> align = none
> align.stripMargin = true
> ```

is it because `assumeStandardLibraryStripMargin` is disabled by default?

olafurpg:
That's because the HOCON configuration expands into the following JSON and the second `"align"` key overwrites the first key
```json
{
  "align": "none",
  "align": {
    "stripMargin": true
  }
}
```

kitbellew:
> That's because the HOCON configuration expands into the following JSON and the second `"align"` key overwrites the first key

let me test it again. i thought i tried it and it worked...

kitbellew:
> > That's because the HOCON configuration expands into the following JSON and the second `"align"` key overwrites the first key
> 
> let me test it again. i thought i tried it and it worked...

looks like it works:
```
<<< ONLY pipe char on first line #324
align = none
align.stripMargin = true
===
"""|first
  |second""".stripMargin
>>>
"""|first
   |second""".stripMargin
```
but not
```
<<< ONLY pipe char on first line #324
align.stripMargin = true
align = none
===
"""|first
  |second""".stripMargin
>>>
"""|first
  |second""".stripMargin
```

PR #1839, opened 2020-04-06
Sunset editions

olafurpg:
Editions are a failed experiment. My original motivation to add editions
was to make it possible to make improvements to the default formatting
behavior without introducing a new configuration option in every PR.
This made sense when Scalafmt was evolving slowly, but lately the
project has been moving much faster thanks to the heroic efforts of
Albert Meltzer and other contributors.

I'm happy that Scalafmt is still evolving and I like many of the new
changes. However, the latest improvements don't respect the `edition`
setting, which is understandable because the changes are large.

I don't want to remove editions because that would break user's
configuration files. Instead, this commit just updates the website to
say that editions are no longer used.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -98,28 +98,6 @@ val example2 =
   |""".stripMargin
 ```
 
-## Edition
-
-The `edition` setting allows Scalafmt users to stay on older default settings
-while upgrading to the latest Scalafmt release.
-
-The use-case for configuring the `edition` setting is when you want to upgrade
-to the latest Scalafmt version (for example, to enjoy a bug fix in the parser)
-but need more time to deal with the latest changes in the formatting output.
-
-The end goal for users should be to remove the `edition` setting from
-`.scalafmt.conf` in order to enjoy the latest improvements to the formatting
-output.
-
-The `edition` setting is formatted as `"$year-$month"` and should be interpreted
-as: "use the default settings at that given date". The latest edition is selected by default.
-
-Example:
-
-```scala config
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -411,12 +389,15 @@ insert newlines.
 
 ### `danglingParentheses`
 
-While this parameter is not technically under the `newlines` section, it logically belongs there.
+While this parameter is not technically under the `newlines` section, it
```
olafurpg, line 392: These formatting changes come from Prettier, I have "format on save" enabled in VS Code in the Scalafmt repo. 

Maybe we should enforce prettier on all `*.md` files  in CI 🤔

kitbellew: how would we do that? :)

olafurpg:
> I have been religiously using edition guards everywhere except obvious bugs. However,
it did occur to me that the code is becoming a bit harder to maintain.

I noticed! And I appreciate the efforts to respect editions 🙏 I'm sorry for imposing on you those complications, fixing the Scalafmt router is hard enough on its own 😅

docs/configuration.md
olafurpg: I took a stab here https://github.com/scalameta/scalafmt/pull/1840

kitbellew:
@olafurpg should we remove the 2020-03 edition from the code? it hasn't been released yet.

kitbellew:
> @olafurpg should we remove the 2020-03 edition from the code? it hasn't been released yet.

hmm... i take it back. probably need to keep it, to differentiate changes that happened since 2020-01.

PR #1840, opened 2020-04-06
Enforce Prettier formatting on markdown sources.

olafurpg:
Previously, the CI did not check the formatting of markdown files. Now,
we require that users run Prettier on the markdown sources with the
following command
```
npx prettier@2.0.4 --write docs
```


docs/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ title: Changelog
 
 [Full Changelog](https://github.com/scalameta/scalafmt/compare/v2.4.1...v2.4.2)
 
+Hello
+worl
```
kitbellew: 1. typo
2. should it be here?

kitbellew: never mind, disappeared.

olafurpg: I used this to test that the CI would fail

kitbellew:
> I don't mind using `npm` if that's preferred 🤷‍♂

not by me. i have no preference at all.

p.s. looks like different yarn. people complain of name conflicts.

PR #1843, opened 2020-04-08
Compiler warnings enabled

poslegm:
1. Enabled fatal-warnings
2. Enabled warning on unused imports
3. `CompatParCollections` simplified

I want to enable more automatic checks on CI for code health. Started with warning on unused imports but I am going to enable another warnings step-by-step.

Warnings will work for 2.13 only because it's main Scala version for development. Deprecations disabled because it hard to maintain zero-deprecations with support of cross-build for three Scala versions.

What do you think?


poslegm:
It possible with scalafix: https://github.com/NeQuissimus/sort-imports. But I don't know about Intellij compatibility of this rule.

Anyway scalafix checks on CI is in my plans :)

PR #1846, opened 2020-04-10
FormatOps: don't start a statement on a comment

kitbellew:
This is required to introduce infix formatting (because it occasionally starts a statement). This also simplifies the logic of identifying docstrings.


kitbellew:

Issue #1838, opened 2020-04-06
`trailingCommas = multiple` results in config parsing error


TJC:
- **Version**: 2.4.2 (or is it?)
- **Integration**: sbt plugin sbt-scalafmt version 2.3.2
```
version = 2.4.2
trailingCommas = multiple
```

## Steps

When I run scalafmt like this:

```bash
sbt scalafmtAll
```

## Problem

```
[error] Unknown input 'multiple'. Expected one of: always, never, preserve: /Users/myusername/example/.scalafmt.conf
```

## Expectation

I assumed the documentation here would apply to the latest release of scalafmt.

https://scalameta.org/scalafmt/docs/configuration.html#trailing-commas

As of now, it refers to the option "multiple", that was added in 2716c44d88e3797

## Notes

I suspect it's simply the case that the online documentation is ahead of the general release, but this does result in confusion for users like me.

poslegm:
Sorry for the confusion :(

We will release 2.5.0-RC1 with this setting support soon

PR #1847, opened 2020-04-10
#1627 [15]: Documentation: describe newlines.source

kitbellew:
Helps with #1627.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -428,6 +428,39 @@ insert newlines.
 > You might be interested in the [Vertical Multiline](#vertical-multiline)
 > section.
 
+### `newlines.source`
+
+> Since v2.5.0.
+
+This parameter controls the general approach to line breaks, and whether to take into
+account existing newlines in the source. The default value (if the parameter is not
+specified) is the _classic_, original way. Below are the alternatives.
+
+> These alternatives are EXPERIMENTAL and might change in the future without
+> regard to any `edition` settings, until fully released (and this message deleted).
+
+#### `newlines.source=keep`
+
+This approach attempts to preserve line breaks in the input whenever possible.
+
+#### `newlines.source=fold,unfold`
+
+These two approaches _completely ignore_ existing line breaks, except around comments
+and blank lines (i.e., multiple consecutive newlines).
+
+> Might require increasing runner limits
+> (`runner.optimizer.maxVisitsPerToken`, possibly even `runner.maxStateVisits`),
+> to avoid _SearchStateExploded_ exceptions.
+
+`fold` attempts to remove line breaks whenever possible resulting in a more
+horizontal, or vertically compact look.
+
+`unfold`, on the other hand, is intended for those who prefer a more vertical,
+or horizontally compact look.
+
+Both settings attempt to play nice with other parameters, but some combinations
+are prohibited and will result in an error.
+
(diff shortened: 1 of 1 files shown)


kitbellew:

docs/configuration.md
```diff
@@ -428,6 +428,37 @@ insert newlines.
...
+
+This approach attempts to preserve line breaks in the input whenever possible.
+
+#### `newlines.source=fold,unfold`
+
+These two approaches _completely ignore_ existing line breaks, except around comments
+and blank lines (i.e., multiple consecutive newlines).
+
+> Might require increasing runner limits
+> (`runner.optimizer.maxVisitsPerToken`, possibly even `runner.maxStateVisits`),
```
poslegm, line 450: It will be more convenient to show specific values here, so that users can simply copy them

docs/configuration.md
```diff
@@ -428,6 +428,37 @@ insert newlines.
...
+### `newlines.source`
+
+This parameter controls the general approach to line breaks, and whether to take into
+account existing newlines in the source. The default value (if the parameter is not
+specified) is the _classic_, original way. Below are the alternatives.
+
+> These alternatives are EXPERIMENTAL and might change in the future without
+> regard to any `edition` settings, until fully released (and this message deleted).
+
+#### `newlines.source=keep`
```
poslegm, line 440: Ideally, we should add code examples, but it can be done later

kitbellew: i considered it... since it covers way too many cases, a bit hard. suggestions?

kitbellew:
> Just minor comments. But we need to merge this PR after 2.5.0-RC1 release to avoid confuse like #1838

i can merge next week, just before releasing. but other parameter docs have already been merged; would be better to figure out how https://scalameta.org/scalafmt/docs/configuration.html is sourced, perhaps change it to get the latest release somehow.

docs/configuration.md
poslegm: I would pick at least a few illustrative examples from unit tests

For example:

```
val nested =
  promiseIntercept(
    new Actor {
      def receive = {
        case _ =>
      }
    }
  )(result)

class Sample {
  def mkFunc(): String => Boolean = {
    if(flag) {
      v => true
    } else {
      v => false
    }
  }
}
```

docs/configuration.md
```diff
@@ -428,6 +428,37 @@ insert newlines.
...
+> to avoid _SearchStateExploded_ exceptions.
+
+`fold` attempts to remove line breaks whenever possible resulting in a more
+horizontal, or vertically compact look.
+
+`unfold`, on the other hand, is intended for those who prefer a more vertical,
+or horizontally compact look.
+
+Both settings attempt to play nice with other parameters, but some combinations
+are prohibited and will result in an error.
```
poslegm, line 460: May be link at least conflicts resolution from `ScalafmtConfig`? https://github.com/scalameta/scalafmt/blob/master/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala#L161-L178

poslegm:
> would be better to figure out how https://scalameta.org/scalafmt/docs/configuration.html is sourced

I thought about it. May be create branch `v2.5.0-docs` and merge release-dependent documentation changes into it? And merge `v2.5.0-docs` into master after release.

PR #1848, opened 2020-04-10
#1627 [16]: ScalafmtConfig: refactor checking source errors

kitbellew:
Helps with #1627.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
@@ -158,24 +158,23 @@ case class ScalafmtConfig(
     project: ProjectFiles = ProjectFiles(),
     edition: Edition = Edition.Latest
 ) {
+  val errors = new mutable.ArrayBuffer[String]
+  if (newlines.sourceIgnored) {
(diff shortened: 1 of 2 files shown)


kitbellew:

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -158,24 +158,23 @@ case class ScalafmtConfig(
...
     if (!includeCurlyBraceInSelectChains)
       errors += s"includeCurlyBraceInSelectChains=${includeCurlyBraceInSelectChains}"
-    if (errors.nonEmpty) {
-      val prefix = s"newlines: can't use source=${newlines.source} and ["
-      throw new IllegalArgumentException(errors.mkString(prefix, ",", "]"))
-    }
+  }
+  if (errors.nonEmpty) {
+    val prefix = s"newlines: can't use source=${newlines.source} and ["
+    throw new IllegalArgumentException(errors.mkString(prefix, ",", "]"))
```
poslegm, line 177: Custom `SettingsConflictException` would be nice here (but definitely not necessary now)

kitbellew: ok, will do

PR #1850, opened 2020-04-10
State: clarify how priority is determined

kitbellew:
Closer inspection reveals that there's a peculiar inconsistency in how splits are sorted/prioritized (see scaladoc in Split.scala). Changing sorting to make it consistent produces significant changes in formatted output. For now, let's explain this inconsistency in the code.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
@@ -14,6 +14,19 @@ case class OptimalToken(token: Token, killOnFail: Boolean = false)
   * Consider a split to be an edge in a search graph and [[FormatToken]]
   * are the nodes.
   *
+  * NB: there's a historical inconsistency in how splits are sorted; when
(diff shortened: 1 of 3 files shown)


kitbellew:

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
```diff
@@ -14,6 +14,19 @@ case class OptimalToken(token: Token, killOnFail: Boolean = false)
...
+  * precedence). However, when a search state is added into the priority
+  * queue, preference is given to states with lower cost, further token
+  * and, unlike above, a LATER line defining the split.
+  *
+  * A possible reason for the latter is to give those "secondary" splits
+  * a chance to move through the BestFirstSearch algorithm, as otherwise
+  * a sequence of primary splits might end up as the winning solution
+  * even if it exceeds the maxColumn margins, because a secondary split
+  * was deemed unlikely to win and moved to a backup priority queue.
+  *
```
poslegm, line 29: Interesting, thank you very much for explanation

PR #1852, opened 2020-04-10
Upgrade to the latest Metaconfig

olafurpg:


olafurpg:
The only change is that the new version of Metaconfig doesn't transitively bring in MUnit as a dependency.

PR #1853, opened 2020-04-10
FormatOps: implement recursive findToken method

kitbellew:
Use it for various existing traversal methods.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -166,73 +166,44 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
     }
   }
 
-  @tailrec
-  final def prevNonCommentWithCount(
-      curr: FormatToken,
-      accum: Int = 0
-  ): (Int, FormatToken) = {
-    if (!curr.left.is[T.Comment]) accum -> curr
-    else {
-      val tok = prev(curr)
-      if (tok == curr) accum -> curr
-      else prevNonCommentWithCount(tok, accum + 1)
-    }
-  }
-  def prevNonComment(curr: FormatToken): FormatToken =
-    prevNonCommentWithCount(curr)._2
-
-  @tailrec
   final def nextNonCommentSameLine(curr: FormatToken): FormatToken =
-    if (curr.hasBreak || !curr.right.is[T.Comment]) curr
-    else {
-      val tok = next(curr)
-      if (tok == curr) curr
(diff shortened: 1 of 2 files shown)


kitbellew:
@poslegm PTAL

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1242,4 +1211,29 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
       case _ => (tree, select)
     }
 
+  @tailrec
+  final def findTokenWith[A](f: FormatToken => Option[A])(
+      ft: FormatToken,
+      step: Int,
+      end: Int = Int.MaxValue
```
poslegm, line 1218: Maybe I missed it, but I don't found where `end` parameter passes into function. Looks like it always has default value

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1242,4 +1211,29 @@ class FormatOps(val tree: Tree, val initStyle: ScalafmtConfig) {
...
+          val nextFt = tokens(ft, step)
+          if (nextFt eq ft) Left(ft)
+          else findTokenWith(f)(nextFt, step, end)
+        }
+    }
+  }
+
+  final def findToken(
+      ft: FormatToken,
+      step: Int,
```
poslegm, line 1234: I see that step always 1 or -1.

Maybe introduce adt instead of number constant?

```scala
sealed abstract class Step(raw: Int)
object Step {
  final case object Next extends Step(1)
  final case object Prev extends Step(-1)
}

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
kitbellew: will be used later. to look for something between `(` and `)`.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
kitbellew: let me think about it.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
kitbellew: what about this last patch? i removed end (the search can simply stop there) and changed step to `FormatToken`-hopping lambda.

PR #1855, opened 2020-04-11
#1627 [18]: FormatOps: ignore line breaks in infix expressions

kitbellew:
Add the `newlines.afterInfix` parameter to allow formatting of infix expressions and optionally breaking lines after an infix operation (which appears to be always safe, unlike before it); enable it for `newlines.source=fold,unfold`. The behaviour and specific parameters/values will be defined in a separate PR containing the documentation.

Helps with #1627.

`scala-repos` diffs:
* Infix bugfix: handle Type.ApplyInfix correctly
  * `unfold`: https://github.com/kitbellew/scala-repos/commit/6377998f6d7f4c3b05d7b9dbb8486e5adba8b9a0
  * rest unchanged
* FormatOps: ignore line breaks in infix expressions
  * `fold`: https://github.com/kitbellew/scala-repos/commit/1a055b5d397d1f745b016813ea5c6f611a74ddaf
  * `unfold`: https://github.com/kitbellew/scala-repos/commit/edbae8c98e30637849da986dd281f1d800423887
  * rest: unchanged

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/SplitTag.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/SplitTag.scala
@@ -14,9 +14,7 @@ object SplitTag {
 
   abstract class Custom extends SplitTag {
     override final def activateOnly(splits: Seq[Split]): Seq[Split] =
-      splits.flatMap { s =>
-        if (s.isTaggedFor(this)) Some(s.copy(tag = SplitTag.Active)) else None
-      }
+      splits.map(_.activateFor(this)).filter(_.isActive)
   }
 
   case object Active extends Base
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/SplitTag.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/SplitTag.scala
@@ -22,5 +22,6 @@ object SplitTag {
 
   case object OneArgPerLine extends Custom
   case object SelectChainFirstNL extends Custom
+  case object InfixChainNoNL extends Custom
 
 }
--- a/scalafmt-tests/src/test/resources/rewrite/RedundantBraces.stat
+++ b/scalafmt-tests/src/test/resources/rewrite/RedundantBraces.stat
@@ -1,6 +1,7 @@
 rewrite.rules = [RedundantBraces]
 rewrite.redundantBraces.maxLines = 1
 rewrite.redundantBraces.generalExpressions = true
+newlines.afterInfix = some
 <<< basic
 object a {
   def x(i: Int): Int = {
@@ -321,6 +322,8 @@ object a {
   b.c(d => e => { f; g })
 }
 <<< #1027 2.1: parens to braces no
+newlines.afterInfix = keep
+===
 object a {
   b.c (d =>
     e
@@ -607,9 +610,7 @@ object a {
 }
 >>>
 object a {
-  1 + {
-    2
-  } + 3
+  1 + 2 + 3
 }
 <<< #1633 4.2: infix with literal and no break before
 object a {
@@ -618,8 +619,7 @@ object a {
 }
 >>>
 object a {
-  1 +
-    2 + 3
+  1 + 2 + 3
 }
 <<< #1633 4.3: infix with literal and comment
 object a {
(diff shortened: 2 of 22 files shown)


kitbellew:

This turned out to be the most complex problem, took me almost 3 weeks. I know @olafurpg tried solving this before and eventually gave up; I tried to track down all previous issues relating to infix handling and take them into account. I think the end result is reasonable albeit not always optimal.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/SplitTag.scala
```diff
@@ -14,15 +14,14 @@ object SplitTag {
 
   abstract class Custom extends SplitTag {
     override final def activateOnly(splits: Seq[Split]): Seq[Split] =
-      splits.flatMap { s =>
-        if (s.isTaggedFor(this)) Some(s.copy(tag = SplitTag.Active)) else None
-      }
+      splits.flatMap(Some(_).map(_.activateFor(this)).filter(_.isActive))
```
poslegm, line 17:
```suggestion
      splits.map(_.activateFor(this)).filter(_.isActive)
```

scalafmt-tests/src/test/resources/rewrite/RedundantBraces.stat
```diff
@@ -427,12 +430,12 @@ class Test extends AnyWordSpec {
 >>>
 class Test extends AnyWordSpec {
   "SUT" when {
-    "statement about test" should {
-      "some test" in {
-        val s = 1
-      }
+      "statement about test" should {
```
poslegm, line 433: What is reason if this extra indent?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/SplitTag.scala
kitbellew: done, thank you, very good catch :)

scalafmt-tests/src/test/resources/rewrite/RedundantBraces.stat
kitbellew: what would you prefer:

```
// option 1: everything is indented consistently (that's the one implemented now)
val b = qual1 op1 {
    1 + 2 + 3 + 4 + 5
  } op2
  (111 * 222 * 333 * 444 * 555) op3 {
    "1 / 2 / 3 / 4 / 5"
  }

// option 2: if first op is block, it's not indented
val b = qual1 op1 {
  1 + 2 + 3 + 4 + 5
} op2
  (111 * 222 * 333 * 444 * 555) op3 {
    "1 / 2 / 3 / 4 / 5"
  }

// option 3: blocks are not indented
val b = qual1 op1 {
  1 + 2 + 3 + 4 + 5
} op2
  (111 * 222 * 333 * 444 * 555) op3 {
  "1 / 2 / 3 / 4 / 5"
}

// option 4: treat one op vs multiple differently
val a = qual1 op1 {
  1 + 2 + 3 + 4 + 5
}
val b = qual1 op1 {
    1 + 2 + 3 + 4 + 5
  } op2
  (111 * 222 * 333 * 444 * 555) op3 {
    "1 / 2 / 3 / 4 / 5"
  }
```

kitbellew: @poslegm just a little context: none of these options are ideal, in my opinion, but half of the time spent on this commit was to fiddle with indents, so this is all i can offer :)

* option 1 add special handling for blocks because another rule takes care of them without consideration of the infix formatting; it looks better further down the infix expression but the first indent (frequently, the only) looks unusual -- and caught your attention

* option 2 is inconsistent; plus, if the second infix RHS is also a block, we are back to the same problem as in option 1

* option 3 is what the original infix formatting uses, without any special treatment of blocks and leaving them to the apply rules; *we can stay with that, to reduce the amount of change*

* option 4 is weird as it shifts existing code if one adds another infix op to it

kitbellew: ok, i have decided to add another parameter so people can choose between options 1 and 3. the other two will not be available.

poslegm: Sorry for long answer. I think that option 3 is preferable (don't add special indent) because infix operators with block active actively used in testing frameworks and it 4-space indentation looks weird in case of tests

kitbellew: ok, done, no unusual indents. `scala-repos` diffs (without `?w=1`) have been updated, uploading rebased code shortly. the trick was to create an indent object that can be disabled, because indent needs to be applied only to the first newline.

PR #1857, opened 2020-04-11
Flexible literal formatting

### LINKED ISSUE #1844: Format hexadecimal numbers consistently
This is an enhancement request. Please format hexadecimal numbers throughout the code base.
My personal preference is to make them lowercase like the current e-notation default because lowercase stands out better against the numerals.
```
version = 2.4.2
```

## Steps

Given code like this:
```scala
private final val UINT_MAX = 0xFFFFFFFF
assertEquals(0x7F, buf.get(6))
```

When I run scalafmt.

## suggested format

Scalafmt should format code like this:
```scala
private final val UINT_MAX = 0xffffffff
assertEquals(0x7f, buf.get(6))
```

## Notes
Currently if the hexadecimal uses a suffix for the type like `L` the number format
gets cased along with the suffix case. They should be independent.

Combining e-notation and hexadecimal with types be independent of each other if there are settings to choose the preference. Defaults should be like the following:
```scala
2.028241e31f
1.9958403095347198e292d
0xffffffffL
assertEquals(0x838281807f7e7d7cL, buf.getLong())
```

### LINKED ISSUE #1845: Literal format default changes hexadecimal numbers and potentially e-notation
- **Configuration**:
```
version = 2.4.2
```

## Steps

Examples:
```scala
-      assertEquals(0x797a7b7c7d7e7f80L, longBuf1.get())
+      assertEquals(0x797A7B7C7D7E7F80L, longBuf1.get())

// this is okay but the default L would upper case or if
// you had the upper case D set and wanted lower case e
-    assertD2sEquals("1.0E7", 1.0E7d)
+    assertD2sEquals("1.0E7", 1.0e7d)

-    ((i3 & 0xffff).toLong << 32) | (i4.toLong & 0xffffffffL)
+    ((i3 & 0xffff).toLong << 32) | (i4.toLong & 0xFFFFFFFFL)

-  final val MIN_NORMAL        = 1.17549435E-38f
+  final val MIN_NORMAL        = 1.17549435e-38f
```

## Expectation

Scalafmt should change the type at the end of the numeric number consistently
like `f,d,L` but not change e-notation or hexidecimal number they are attached to.

## Workaround

Use the following non-default settings

```properties
# leave literals alone
literals.long = unchanged
literals.float = unchanged
literals.double = unchanged
```

poslegm:
Fixes #1844 
Fixes #1845 

@ekrich describes reasonable feature requests in issues.

Introduced:

1. `literals.hex` which controls case of hexadecimal part of integer literals
```
0xFFFF // literals.hex=Upper
0xffff // literals.hex=Lower
```
2. `literals.scientific` which controls case of `e` in floating point literals in scientific notation:
```
1.0E-1 // literals.scientific=Upper
1.0e-1 // literals.scientific=Lower
```

This changes will lead to diffs after scalafmt update, so I provide small migration guide about it in release notes.

And PR with microsite documentation coming soon.


scalafmt-core/shared/src/main/scala/org/scalafmt/util/LiteralOps.scala
```diff
@@ -0,0 +1,70 @@
+package org.scalafmt.util
```
kitbellew, line 1: format writer is internal. format ops is there, too. should this be?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/LiteralOps.scala
```diff
@@ -0,0 +1,70 @@
...
+    *
+    * literals.long applies to suffix and literals.hex applies to body
+    * */
+  def prettyPrintInteger(
+      str: String
+  )(implicit style: ScalafmtConfig): String = {
+    val sb = new StringBuilder()
+    val body =
+      if (str.startsWith("0x") || str.startsWith("0X")) {
+        sb.append("0x")
```
kitbellew, line 25: what if someone asks to use 0X instead?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/LiteralOps.scala
```diff
@@ -0,0 +1,70 @@
...
+    * 0x is a hex prefix
+    * FF123 is a body
+    * L is a long suffix
+    *
+    * literals.long applies to suffix and literals.hex applies to body
+    * */
+  def prettyPrintInteger(
+      str: String
+  )(implicit style: ScalafmtConfig): String = {
+    val sb = new StringBuilder()
```
kitbellew, line 22: why not pass the string builder from format writer and append to that directly?

and while we're at it, change the Case.process method to do the same?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/LiteralOps.scala
```diff
@@ -0,0 +1,70 @@
...
+    * literals.scientific applies to body and literals.float/double applies to suffix
+    * */
+  private def prettyPrintFloatingPoint(
+      str: String,
+      suffixUpper: String,
+      suffixLower: String,
+      suffixCase: Case
+  )(implicit style: ScalafmtConfig): String =
+    if (str.endsWith(suffixUpper) || str.endsWith(suffixLower)) {
+      style.literals.scientific.process(str.dropRight(1)) +
```
kitbellew, line 65: are we assuming that the suffix contains one character? in that case, can we pass a Char instead of a String?

poslegm: are we assuming that the suffix contains one character?suffix of floating point literals can be

scalafmt-core/shared/src/main/scala/org/scalafmt/util/LiteralOps.scala
poslegm: But `TokenOps` is `util` :man_shrugging: 

In my mind `LiteralOps` is functions for printing literals independent of formatting context

scalafmt-core/shared/src/main/scala/org/scalafmt/util/LiteralOps.scala
poslegm: This is a place that I doubt. Lowercased `0x` was hardcoded and I keep it unchanged. But may be reasonable to make it obey `literals.hex` setting. What do you think?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/LiteralOps.scala
poslegm: I commited https://github.com/scalameta/scalafmt/pull/1857/commits/d0bc2a1397698042efea9998892699dc6b366516 with respecting `literal.hex` for `0x` prefix, and this function greatly simplified. So there is no more need for `StringBuilder`

scalafmt-core/shared/src/main/scala/org/scalafmt/util/LiteralOps.scala
kitbellew: np, it was just a question

scalafmt-core/shared/src/main/scala/org/scalafmt/util/LiteralOps.scala
kitbellew: sure, let's go with String but shouldn't we either verify its length is 1, or just use dropRight(suffix.length)?

ekrich:
> I'm sorry, i should have been clearer. i personally prefer `0xFF` but someone might, for instance, want `0XffL`, and it's not unreasonable.

I didn't even think about a prefix, I thought people would just want the following but then all these cases are in play?
```
Oxff
OXFF
OXff
OxFF
```
Honestly, the first 2 are the only I have seen in code and was think there would be the following with the first as default.
```
// literals.hex = lower
0xff
// literals.hex = upper
0XFF
// literals.hex = unchanged
```

This would mirror the `literals.long` and friends.

kitbellew:
@poslegm do you have any further changes coming soon? i have one more that i will submit today, and after that i am ready to release 2.5.0-rc1 unless you want to wait.

poslegm:
> do you have any further changes coming soon?

No, documentation only

PR #1858, opened 2020-04-12
Add new rewrite rule:  CamelCaseVariableNaming

mggger:
Different developers have different variable naming styles, the most common is:

1.  SnakeCase  ```val student_name = "xxx" ```
2. CamelCase ```val studentName = "xxx" ```

Add new rewrite rule **CamelCaseVariableNaming** to unify variable naming style, here is a example :

```scala
class A {
  val student_name = "wang"
  val studentAge = 22
}

// after formatting
// rewrite.rules = [CamelCaseVariableNaming]
class A {
  val studentName = "wang"
  val studentAge = 22
}

```

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/CamelCaseVariableNaming.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/CamelCaseVariableNaming.scala
@@ -1,0 +1,39 @@
+package org.scalafmt.rewrite
+
+import scala.meta.{Pat, Term, Tree}
+
+
+object VariableNaming extends Rewrite {
+  override def create(implicit ctx: RewriteCtx): RewriteSession =
+    new VariableNaming
+}
+
+class VariableNaming(implicit ctx: RewriteCtx) extends RewriteSession {
+
+  def isSnakeCaseNaming(str: String): Boolean = {
+    str.contains('_')
+  }
+
+  def toCamelCaseNaming(str: String): String = {
+    str.split('_') match {
+      case Array(head) => head
+      case Array(head, tail @ _*) => head + tail
+          .map(s => s.charAt(0).toUpper + s.slice(1, s.length))
+        .reduce( _ + _)
+    }
+  }
+
+  def rewriteCamelCaseNaming(t: Tree): Unit = {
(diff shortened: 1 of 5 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/CamelCaseVariableNaming.scala
```diff
@@ -0,0 +1,40 @@
...
+import scala.meta.{Pat, Tree}
+
+object CamelCaseVariableNaming extends Rewrite {
+  override def create(implicit ctx: RewriteCtx): RewriteSession =
+    new CamelCaseVariableNaming
+}
+
+class CamelCaseVariableNaming(implicit ctx: RewriteCtx) extends RewriteSession {
+
+  def toCamelCaseNaming(str: String): String = {
```
poslegm, line 12: Variable names in Scala may start with underscore:
```scala
val _abc = 1
```

And this function works wrong on it:
```
@ toCamelCaseNaming("_abc")
res2: String = "Abc"
```

scalafmt-tests/src/test/resources/rewrite/CamelCaseVariableNaming.stat
```diff
@@ -0,0 +1,16 @@
...
+  val studentSex = "male"
+}
+>>>
+class A {
+  val studentName = "wang"
+  val studentAge = 12
+  val studentSex = "male"
+}
+
+
```
poslegm, line 16: Would rewrite rule work correct on provided tests?

```scala
<<< rewrite also identifier usages
val x = 42 + snake_case_val
>>>
val x = 42 + snakeCaseVal
<<< don't rewrite names from external libraries
another.library.ExternalClass.snake_case_method_name(42)
>>>
another.library.ExternalClass.snake_case_method_name(42)
```

mggger:
@poslegm Thanks  for your code reviewing, it would be very helpful for me .

@kitbellew Thanks,  there are a lot of flaws in my implementation

kitbellew:
@mggger perhaps you could use `scalastyle` to check compliance (for instance, http://www.scalastyle.org/rules-1.0.0.html#org_scalastyle_scalariform_FieldNamesChecker)

PR #1859, opened 2020-04-13
TreeOps: use numParents for nestedApplies/Select

kitbellew:


kitbellew:
@olafurpg @poslegm PTAL

PR #1863, opened 2020-04-13
Newlines: implicitParamListModifier to replace two

kitbellew:
As these parameters haven't yet been released, let's consolidate these two into a single multi-valued parameter. There are other, unrelated parameters which also follow the same before/after logic.


kitbellew:
@olafurpg @poslegm PTAL

PR #1864, opened 2020-04-13
Newlines: allow break after top-level statements

### LINKED ISSUE #1653: Adding newlines at end of top level definition
## Feature request
Hi All! This is a feature request that I'd be interested in working on myself, as a first time contribution. I'd love to hear feedback if anyone else cares about this kind of a rewrite rule.

There currently exists a scalafmt rule which adds a newline after top level statements. For consistency's sake, it would be great to have a symmetric additional rule which adds a newline before the closing brace of a top level module (object or class definition).

## Current
Scalafmt currently adds a newline at the beginning of top level statements. For example with the following configuration:

```
version = 2.0.1
```

Given code like this:
```scala
object Foo {
  def someCode: Unit =
    ()
}
```

Scalafmt produces this:
```scala
object Foo {

  def someCode: Unit =
    ()
}
```

## Proposed
For smaller objects particularly, the above rule ends up with asymmetric looking code. So it would be nice to have an optional trailing newline as well.

So given a (hypothetical) configuration like this:
```
version = 2.0.1
newlines.alwaysAfterTopLevelStatements = true // new rule
```

Given this code:
```scala
object Foo {
  def someCode: Unit
}
```

Scalafmt would produce:
```scala
object Foo {

  def someCode: Unit

}
```

kitbellew:
When writing newlines before a token, check whether the previous token ended a multi-line top-level statement and optionally add a double.

Fixes #1653.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -527,36 +527,117 @@ def other(a: String, b: String)(c: String, d: String) = a + b + c
 other(a, b)(c, d)
 ```
 
-### `newlines.alwaysBeforeTopLevelStatements`
+### `newlines.topLevelStatements`
+
+> Since v2.5.0.
 
 ```scala mdoc:defaults
-newlines.alwaysBeforeTopLevelStatements
+newlines.topLevelStatements
 ```
 
 ```scala mdoc:scalafmt
-newlines.alwaysBeforeTopLevelStatements = false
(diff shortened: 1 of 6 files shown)


kitbellew:
@olafurpg @tanishiking @poslegm @nasadorian PTAL

PR #1865, opened 2020-04-13
ScalafmtConfig: allow overriding per file pattern

kitbellew:
For instance, we might want to use a slightly different configuration
for `*.sbt` files, or those under `src/test/scala`.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
@@ -23,6 +23,7 @@ object Scalafmt {
 
   private val WindowsLineEnding = "\r\n"
   private val UnixLineEnding = "\n"
+  private val defaultFilename = "<input>"
 
   // XXX: don't modify signature, scalafmt-dynamic expects it via reflection
   /**
@@ -50,12 +51,15 @@ object Scalafmt {
 
   private[scalafmt] def formatCode(
       code: String,
-      style: ScalafmtConfig = ScalafmtConfig.default,
+      baseStyle: ScalafmtConfig = ScalafmtConfig.default,
       range: Set[Range] = Set.empty,
-      filename: String = "<input>"
+      filename: String = defaultFilename
   ): Formatted = {
+    val style =
+      if (filename == defaultFilename) baseStyle
+      else baseStyle.getConfigFor(filename) // might throw for invalid conf
+    val runner = style.runner
     try {
-      val runner = style.runner
       if (code.matches("\\s*")) Formatted.Success(System.lineSeparator())
       else {
         val isWindows = containsWindowsLineEndings(code)
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
@@ -1,8 +1,11 @@
 package org.scalafmt.config
 
+import java.nio.file
+
 import scala.collection.mutable
 import scala.io.Codec
 import scala.meta.Dialect
+import scala.util.Try
 import scala.util.matching.Regex
 import metaconfig.annotation._
 import metaconfig._
@@ -156,6 +159,7 @@ case class ScalafmtConfig(
     onTestFailure: String = "",
     encoding: Codec = "UTF-8",
     project: ProjectFiles = ProjectFiles(),
+    fileOverride: Conf.Obj = Conf.Obj.empty,
     edition: Edition = Edition.Latest
 ) {
   val errors = new mutable.ArrayBuffer[String]
@@ -191,6 +195,7 @@ case class ScalafmtConfig(
     ScalafmtConfig.alignReader(align.reader)
   lazy val alignMap: Map[String, Regex] =
     align.tokens.map(x => x.code -> x.owner.r).toMap
+  private implicit val confObjReader = ScalafmtConfig.confObjReader
   def reader: ConfDecoder[ScalafmtConfig] =
     generic.deriveDecoder(this).noTypos.noTypos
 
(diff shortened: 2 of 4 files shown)


kitbellew:

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -206,6 +208,19 @@ case class ScalafmtConfig(
     continuationIndent.defnSite
   )
 
+  private lazy val regexFileOverride = fileOverride.values.map {
+    case (pattern, conf) =>
+      val style = ScalafmtConfig.configReader(this).read(conf)
+      pattern.r.pattern -> style
```
olafurpg, line 214: What do you think about using `java.nio.file.PathMatcher` instead of regexes? PathMatcher doesn't require escaping dots (`\.`) and it works uniformly across unix/windows. I've been planning for a while to add a PathMatcher-based alternative to `project.includeFilters`

olafurpg: `PathMatcher.matches()` requires the argument to be of type `java.nio.file.Path`, which might introduce complications, but I think it would be worthwhile to try and make it work

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -156,6 +156,7 @@ case class ScalafmtConfig(
     onTestFailure: String = "",
     encoding: Codec = "UTF-8",
     project: ProjectFiles = ProjectFiles(),
+    fileOverride: Conf.Obj = Conf.Obj.empty,
```
olafurpg, line 159: I think this can have the type `Map[PathMatcher, ScalafmtConfig]`, we should fail fast if there are errors in the overridden settings instead of crashing multiple times for each individual file

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
kitbellew: will try

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
kitbellew: how do i make `ScalafmtConfig` the value right when loading the parent config? the value should be a merge of `this` and the custom _subset_ specified on the right, and before `this` has been fully loaded, couldn't do it. also, it should only be applied to the top-level config, not the subset configs.

i do the merge below.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
kitbellew: ah, wait, how does one instantiate a `PathMatcher`? :) it looks it's a trait. i doubt `metaconfig` does it automatically, right?

olafurpg: Metaconfig doesn't support `PathMatcher` out of the box, although it probably should 

```scala
java.nio.file.FileSystems.getDefault().getPathMatcher("glob:**.txt").matches(java.nio.file.Paths.get("foo.txt"))
```

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
olafurpg: It's a good question, let me try to work it out...

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
kitbellew: good, thank you, done. uploading a new patch shortly.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
kitbellew: for now, i will keep it as-is, with the merge happening only when requested.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -232,6 +252,12 @@ object ScalafmtConfig {
   implicit lazy val codecEncoder: ConfEncoder[Codec] =
     ConfEncoder.StringEncoder.contramap(_.name)
 
+  implicit val confObjReader: ConfDecoder[Conf.Obj] =
```
kitbellew, line 255: @olafurpg just curious why `Conf.Obj` decoder isn't available as part of `metaconfig`.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -206,6 +211,21 @@ case class ScalafmtConfig(
     continuationIndent.defnSite
   )
 
+  private lazy val expandedFileOverride = Try {
+    val fs = file.FileSystems.getDefault
+    fileOverride.values.map {
+      case (pattern, conf) =>
+        val style = ScalafmtConfig.configReader(this).read(conf).get
+        fs.getPathMatcher(pattern) -> style
```
kitbellew, line 219: use `PathMatcher`, as requested

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -206,6 +211,21 @@ case class ScalafmtConfig(
     continuationIndent.defnSite
   )
 
+  private lazy val expandedFileOverride = Try {
```
kitbellew, line 214: Using `Try` as `lazy val` will be re-evaluated if an exception is thrown.

scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
```diff
@@ -50,12 +51,15 @@ object Scalafmt {
...
       code: String,
-      style: ScalafmtConfig = ScalafmtConfig.default,
+      baseStyle: ScalafmtConfig = ScalafmtConfig.default,
       range: Set[Range] = Set.empty,
-      filename: String = "<input>"
+      filename: String = defaultFilename
   ): Formatted = {
+    val style =
+      if (filename == defaultFilename) baseStyle
+      else baseStyle.getConfigFor(filename) // might throw for invalid conf
```
kitbellew, line 60: put this outside of `try`, so that any problems with any config override will result in an exception on the first file.

PR #1867, opened 2020-04-13
literals.* docs

poslegm:
Also for release notes:

> In this release changed default formatting of hex literals and literals in scientific notation. Currently hex digits and `e` symbol will lowercased. More details is in the [documentation](https://scalameta.org/scalafmt/docs/configuration.html#newlines). If you don't want large diff after scalafmt update, you can set
```
literals.hexDigits=unchanged
literals.scientific=unchanged
```


scalafmt-core/shared/src/main/scala/org/scalafmt/config/Literals.scala
```diff
@@ -6,7 +6,7 @@ case class Literals(
     long: Case = Case.Upper,
     float: Case = Case.Lower,
     double: Case = Case.Lower,
-    hexBody: Case = Case.Lower,
+    hexDigits: Case = Case.Lower,
```
kitbellew, line 9: great! :) my mild ocd likes
```
hexPrefix
hexDigits
```

docs/configuration.md
```diff
@@ -1377,6 +1377,91 @@ It will also use different parameters for test suites.
...
+
+Responsible for the case of `Long` literals suffix `L`
+
+```scala mdoc:scalafmt
+literals.long=Upper
+---
+123l
+```
+
+### `literals.float`
```
kitbellew, line 1411: do we need to add the default section for each of these? or is `Unchanged` default everywhere (in which case we should mention that once, at the top)?

poslegm: Yes, defaults will not be superfluous

PR #1869, opened 2020-04-13
Fixed alignment overriding

### LINKED ISSUE #1861: local comment configuration does not work
- **Version**: 2.4.2
- **Integration**: metals
- **Configuration**:
```
version = 2.4.2
```

## Steps

Given code like this:
```scala
object Test {
  // scalafmt: { trailingCommas = always }
  val coreDeps = Seq(
    "org.typelevel" %% "cats-core" % catsV,
    "org.typelevel" %% "cats-effect" % catsEffectV
  )
  // scalafmt: { trailingCommas = never }
}
```

via LSP code formatting (but also via a scalafmt binary)

## Problem

```scala
object Test {
  // scalafmt: { trailingCommas = always }
  val coreDeps = Seq(
    "org.typelevel" %% "cats-core" % catsV,
    "org.typelevel" %% "cats-effect" % catsEffectV
  )
  // scalafmt: { trailingCommas = never }
}
```
(the same code)

## Expectation

```scala
object Test {
  // scalafmt: { trailingCommas = always }
  val coreDeps = Seq(
    "org.typelevel" %% "cats-core" % catsV,
    "org.typelevel" %% "cats-effect" % catsEffectV,
  )
  // scalafmt: { trailingCommas = never }
}
```
(note the trailing comma)

## Workaround

N.A.

## Notes

This does not seem to be related to `trailingCommas` specifically; in fact, the local configuration does not seem to work at all, which is confusing, as there exist tests for this e.g. [here](https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/resources/test/DynamicStyle.stat) at a quick glance.

poslegm:
Fixes #1861 

Fixed for alignment. But global problem is still here:
1. every contributor and code-reviewer can forgot the `styleMap` and look at the `initStyle`;
2. There is no documentation about what configs can't be overridden.

File override looks like more reliable and simple mechanism. What do you think, should we continue to support config overrides for code blocks or deprecate it in favor of file override?

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -559,7 +559,11 @@ class FormatWriter(formatOps: FormatOps) {
   def alignmentTokens(
       locations: Array[FormatLocation]
   ): Map[TokenHash, Int] = {
-    if (initStyle.align.tokens.isEmpty || locations.length != tokens.length)
+    lazy val noAlignTokens = locations.forall { floc =>
+      styleMap.at(floc.formatToken).align.tokens.isEmpty
+    }
+
+    if (locations.length != tokens.length || noAlignTokens)
       Map.empty[TokenHash, Int]
     else {
       var columnShift = 0
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -559,10 +559,8 @@ class FormatWriter(formatOps: FormatOps) {
   def alignmentTokens(
       locations: Array[FormatLocation]
   ): Map[TokenHash, Int] = {
-    lazy val noAlignTokens = locations.forall { floc =>
-      styleMap.at(floc.formatToken).align.tokens.isEmpty
-    }
-
+    lazy val noAlignTokens = initStyle.align.tokens.isEmpty &&
+      styleMap.tok2style.values.forall(_.align.tokens.isEmpty)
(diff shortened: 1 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -559,7 +559,11 @@ class FormatWriter(formatOps: FormatOps) {
   def alignmentTokens(
       locations: Array[FormatLocation]
   ): Map[TokenHash, Int] = {
-    if (initStyle.align.tokens.isEmpty || locations.length != tokens.length)
+    lazy val noAlignTokens = locations.forall { floc =>
```
kitbellew, line 562: 1. i'd just `def` instead of `lazy val`, if the value is used only once.
2. rather than iterating over locations and looking up in `styleMap`, why not iterate over `styleMap.values` directly?

poslegm: It may be simply `val`. `lazy` is here because `noAlignTokens` is second operand of `||` and will not computed if `locations.length != tokens.length` is true. But I doubt that this is a significant optimization, so it can be just `val`.

> why not iterate over styleMap.values directly?

How iterate it here? Size of `styleMap` look bigger than `locations` size

kitbellew: if `styleMap` is larger, then yes. but i thought that `styleMap` was only for token ranges for which any custom overrides were specified, no? so in most cases (since override is rare), the map itself would be empty, leaving just `initStyle`.

poslegm: Yes, you are right

PR #1870, opened 2020-04-14
Router: fix strip margin in interpolated strings

kitbellew:
`align.stripMargin=false` incorrectly handled interpolated strings:
* strip margin character was incorrectly determined (`|` was always assumed);
* strip-margin lines were not indented at all.

As a side effect of this fix, we are also fixing the bug with nested comments which were incorrectly formatted.

Follow-on to #1837.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -5,7 +5,7 @@ import java.util.regex.Pattern
 import org.scalafmt.config.ScalafmtConfig
 import org.scalafmt.rewrite.RedundantBraces
 import org.scalafmt.util.TokenOps._
-import org.scalafmt.util.{TreeOps, LiteralOps}
+import org.scalafmt.util.{LiteralOps, TreeOps}
 
 import scala.annotation.tailrec
 import scala.collection.IndexedSeq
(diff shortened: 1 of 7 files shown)


kitbellew:
@olafurpg @poslegm PTAL

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -102,30 +99,31 @@ class FormatWriter(formatOps: FormatOps) {
...
+          getStripMarginChar(t).map { pipe =>
+            (pipe, () => t.right.syntax.headOption.contains(pipe))
+          }
+        }
+      case _ => None
+    }
+    tupleOpt.fold(text) {
+      case (pipe, isFirstCharPipe) =>
+        val pattern =
+          if (pipe == '|') leadingPipeSpace else getStripMarginPattern(pipe)
```
poslegm, line 122: Just propose to hide `if` inside the function

```scala
private def getStripMarginPattern(pipe: Char) =
  if (pipe == '|') leadingPipeSpace else getStripMarginPattern(pipe)

private val leadingPipeSpace = compileStripMarginPattern('|')

private def compileStripMarginPattern(pipe: Char) = 
  Pattern.compile(raw"(?<=\n) *(?=[$pipe])", Pattern.MULTILINE)
```

kitbellew: will do, thank you

Issue #1291, opened 2018-09-24
Add `--test-and-format` option to return non-zero on formatted files present


ArturGajowy:
`scalafmt file1 file2 file3` will return 0 when it successfully formats all of the files, no matter if any of them changed or not.

It would be useful to have a flag that would modify the exit status so that it's 0 only when no files were changed.

So far, I have to first `scalafmt --test` and then `scalafmt` if the test failed.

olafurpg:
Thanks for reporting! I agree this would be a nice feature, contributions are welcome 👍

kitbellew:
@ArturGajowy @olafurpg is this similar to my early desire for #928?

olafurpg:
Yes. Sounds similar. I still think it’s a good idea.

kitbellew:
> Yes. Sounds similar. I still think it’s a good idea.

so did i :) we still run scalafmt twice in our build...

PR #1872, opened 2020-04-14
CLI: add explicit flag --reportError to exit 1 on malformatted code

### LINKED ISSUE #928: Feature request: --test to optionally modify files before failing
I'd like to be able to use "scalafmt --diff --test" as part of the build to ensure proper formatting for a given new commit, but also to allow the author to include the files formatted by scalafmt in the submission.

Currently, this requires running "scalafmt --diff" separately from "scalafmt --test", which is a workaround but requiring multiple steps -- and multiple passes through the code.

kitbellew:
This will allow returning a non-zero exit code not just for `--test` mode.

Also:
* verify conflicting options aren't used (`--test` or `--check` vs `--list` vs `--in-place`)
* fail fast in core runner when `--check` is used (only dynamic runner used to do it)

Fixes #928 #1291.

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtCoreRunner.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtCoreRunner.scala
@@ -2,6 +2,7 @@ package org.scalafmt.cli
 
 import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
 import java.util.function.UnaryOperator
+import util.control.Breaks
 
 import metaconfig.Configured
 import org.scalafmt.Error.{MisformattedFile, NoMatchingFiles}
@@ -40,17 +41,20 @@ object ScalafmtCoreRunner extends ScalafmtRunner {
         val termDisplay =
           newTermDisplay(options, inputMethods, termDisplayMessage)
         val exitCode = new AtomicReference(ExitCode.Ok)
-        inputMethods.par.foreach { inputMethod =>
-          val code = handleFile(inputMethod, options, scalafmtConf)
-          exitCode.getAndUpdate(new UnaryOperator[ExitCode] {
-            override def apply(t: ExitCode): ExitCode =
-              ExitCode.merge(code, t)
-          })
-          PlatformTokenizerCache.megaCache.clear()
-          termDisplay.taskProgress(
-            termDisplayMessage,
-            counter.incrementAndGet()
-          )
+        Breaks.breakable {
+          inputMethods.par.foreach { inputMethod =>
+            val code = handleFile(inputMethod, options, scalafmtConf)
+            exitCode.getAndUpdate(new UnaryOperator[ExitCode] {
+              override def apply(t: ExitCode): ExitCode =
+                ExitCode.merge(code, t)
+            })
+            if (options.check && !code.isOk) Breaks.break
+            PlatformTokenizerCache.megaCache.clear()
+            termDisplay.taskProgress(
+              termDisplayMessage,
+              counter.incrementAndGet()
+            )
+          }
         }
         termDisplay.completedTask(termDisplayMessage, exitCode.get.isOk)
         termDisplay.stop()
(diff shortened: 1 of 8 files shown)


kitbellew:

scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtCoreRunner.scala
```diff
@@ -67,6 +70,7 @@ object ScalafmtCoreRunner extends ScalafmtRunner {
     catch {
       case MisformattedFile(_, diff) =>
         options.common.err.println(diff)
+        if (options.check) Breaks.break
```
poslegm, line 73: `scalafmt --check` on first failure will breaks here before `ExitCode.TestError` returned to the `exitCode` variable. Then how will `--check` will return exit code 1?

scalafmt-cli/src/main/scala/org/scalafmt/cli/WriteMode.scala
```diff
@@ -8,5 +8,13 @@ package org.scalafmt.cli
   */
 sealed trait WriteMode
 
-case object Override extends WriteMode
-case object Stdout extends WriteMode
+object WriteMode {
```
poslegm, line 11: Instead of `null` by default is's possible to create

`case object Uninitialized extends WriteMode` here.

I believe that work with it will be more controllable and easier than with `null`

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
```diff
@@ -84,7 +84,9 @@ abstract class AbstractCliTest extends AnyFunSuite with DiffAssertions {
       assertExit(exit)
       val obtained = dir2string(input)
       assertNoDiff(obtained, expected)
-      val configTest = Cli.getConfig(Array("--test"), config).get
+      val configTest = Cli
+        .getConfig(Array("--test"), config.copy(writeMode = null))
+        .getOrElse(fail(args.mkString("Can't add --test to: [", " ", "]")))
```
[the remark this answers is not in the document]
kitbellew: this is just a test :) i think i overdid it here, the message can be removed as i am resetting `writeMode` above.

scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtCoreRunner.scala
kitbellew: hmmm... let me see.

scalafmt-cli/src/main/scala/org/scalafmt/cli/WriteMode.scala
kitbellew: sure. the value is actually never used and is replaced by the default `Override` in `CliOptions.auto`.

kitbellew: Switched to `Option` and used `None` here.

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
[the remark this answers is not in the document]
kitbellew: removed the unnecessary `fail`, it will never happen.

scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtCoreRunner.scala
kitbellew: moved this check.

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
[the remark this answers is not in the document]
poslegm: > this is just a test

didn't notice it, sorry

kitbellew:
> The only question is what error message will user get if run scalafmt as `./scalafmt --test --list`?

the exception regarding moving from Test write mode to List write mode.

PR #1873, opened 2020-04-14
RedundantBraces: re-enable general expressions

kitbellew:
Follow-on to #1147.


kitbellew:
@olafurpg @poslegm PTAL

PR #1880, opened 2020-04-15
[docs] Pipe Bash script to Bash, not sh

bsolomon1124:
install-scalafmt-native.sh is a Bash script and,
besides just having a Bash shebang, contains

```
    [ "$(uname)" == "Darwin" ]
```

which is valid Bash, but not valid sh.
(See https://stackoverflow.com/a/2011170/7954504)


poslegm:
Thank you!

PR #1886, opened 2020-04-17
Presets: automatically decode all config builtins

kitbellew:
For each, check the `preset` subkey as well. We should deprecate the top-level presets as they don't allow overrides.

For instance, currently, the pre-2.5.0 behaviour of `align = none` can't be achieved because one of the parameters was added which changed default behaviour. With this change, one can use
```
align.preset = none
align.stripMargin = true
```

this also standardizes that all presets are accessible via `preset` key (including the full-config presets which were historically available via `style`). however, previous ways of using presets are still supported albeit not encouraged.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -9,7 +9,7 @@ using [HOCON](https://github.com/lightbend/config) syntax.
 Here is an example `.scalafmt.conf`:
 
 ```scala config
-align = more    // For pretty alignment.
+align.preset = more    // For pretty alignment.
 maxColumn = 100 // For my wide 30" display.
 ```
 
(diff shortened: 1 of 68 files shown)


poslegm:
Thank you! It's really great improvement of configuration usage. I will take a look on it later.

BTW I am trying to implement another configuration feature: universal append to default configurations (like [align.token.add](https://scalameta.org/scalafmt/docs/configuration.html#aligntokensadd)). You may find this interesting https://github.com/olafurpg/metaconfig/pull/88. And about presets — I think they can also be done on the metaconfig side. But unfortunately, it may not be so easy.

olafurpg:
This is a nice elegant solution 👍

PR #1888, opened 2020-04-18
Router: indent comments in the middle of a select

kitbellew:
This doesn't address the newlines.source=classic case as it follows a vastly different execution path. Helps with #1334.


kitbellew:
@olafurpg @poslegm PTAL

PR #1890, opened 2020-04-21
RedundantBraces: exclude nested partial function

### LINKED ISSUE #1889: Scalafmt fails to parse nested partial function
- **Version**: 2.4.2
- **Integration**: IntelliJ

## Steps

Scalafmt fails to parse code like this. The `implicit` is important. It works without it.
```scala
function { implicit c =>
  {
    case bar => foo
  }
}
```

This also breaks
```scala
function(c => {
  case bar => foo
})
```

## Workaround

Define something before the partial function

```scala
function { implicit c =>
  5

  {
    case bar => foo
  }
}

kitbellew:
Fixes #1889.


kitbellew:
@poslegm PTAL

PR #1897, opened 2020-04-23
RedundantParens: support additional patterns

### LINKED ISSUE #1895: RedundantParens: remove parens around variable
- **Version**: 2.5.0-RC1
- **Configuration**:
```
version = 2.5.0-RC1
rewrite.rules=[RedundantParens]
```

## Steps

Given code like this:
```scala
object a {
  val x = (1)
}
```

```bash
scalafmt
```

## Problem

```scala
object a {
  val x = (1)
}
```

## Expectation

```scala
object a {
  val x = 1
}
```

## Notes

More complex example:
```scala
val z = (insertData *> readDatabase(id))
```

kitbellew:
Literals, names, case clause conditions, val/def body (in this latter case, use the partial function to ensure the body wouldn't be amended twice). Also, remove inner parens.

Fixes #1895.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantParens.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantParens.scala
@@ -18,7 +18,7 @@ class RedundantParens(implicit ctx: RewriteCtx) extends RewriteSession {
 
       case t: Defn.Def => maybeRemove(t.body)
 
-      case t => rewriteFunc.lift(t)
+      case t => rewriteFunc.applyOrElse(t, (x: Tree) => remove(x, true))
     }
 
   private val rewriteFunc: PartialFunction[Tree, Unit] = {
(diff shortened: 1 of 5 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantParens.scala
```diff
@@ -12,43 +12,77 @@ class RedundantParens(implicit ctx: RewriteCtx) extends RewriteSession {
...
-        }
+      case t: Defn.Def => maybeRemove(t.body)
+
+      case t => rewriteFunc.applyOrElse(t, (x: Tree) => remove(x, true))
+    }
+
+  private val rewriteFunc: PartialFunction[Tree, Unit] = {
+    case g: Enumerator.Guard => remove(g.cond)
+
+    case t: Case if t.cond.isDefined => remove(t.cond.get)
```
poslegm, line 27:
```suggestion
    case t: Case => t.cond.foreach(remove)
```

kitbellew: the reason i did it so is to make sure this partial function wouldn't match (elsewhere, if the partial didn't match, i would do something else).

in this specific, we can "pretend" it matched as there's probably nothing else that can be done in case of... well, `case`. let me know your preference.

kitbellew: i tested, no change in behaviour, yours is better (don't like `.get` on `Option`).

PR #1910, opened 2020-04-26
ScalafmtDynamicRunner: read and format only matching files

### LINKED ISSUE #1460: Dynamic runner startup significantly slower - add option to force CoreRunner?
I'm using scalafmt via nailgun, which as you know gives rapid startup and execution times. When executing on a project using a different scalafmt version in `.scalafmt.conf` file than the one behind nailgun, startup times are slower than reformatting a whole commit when the versions match:

```
✔ ~/projects/rchain-bloop [charging_for_deploy_4.13 ↓·7↑·13|✚ 1…12⚑ 38] 
23:17 $ time scalafmt build.sbt --test --debug
Working directory: /home/artur/projects/rchain-bloop
parsed config (v2.0.0)
All files are formatted with scalafmt :)

real    0m0.206s
user    0m0.037s
sys     0m0.008s

✔ ~/projects/rchain-bloop [charging_for_deploy_4.13 ↓·7↑·13|✚ 1…12⚑ 38] 
23:18 $ time scalafmt build.sbt --test --debug
Working directory: /home/artur/projects/rchain-bloop
parsed config (v2.0.0-RC8): /home/artur/projects/rchain-bloop/.scalafmt.conf
All files are formatted with scalafmt :)

real    0m1.911s
user    0m0.029s
sys     0m0.020s
```

It'd be great to pass an option (say: `--require-same-version`) to error whenever there'd be a need to use the dynamic runner. This would prevent the slowdown by forcing me to update.

Without passing any flags, I think a warning should be emitted, saying 'you're running scalafmt 2.0.0, but the config you're using requires a different version, which leads to slower startup and execution'. That would at least be true for scalafmt behind nailgun.

Having just the warning would be better than nothing, but would likely go unnoticed (e.g. IDEA doesn't show git hook output till they fail) or require clumsy scripting in the git hook (scanning for the warning in scalafmt output).

Even better solutions would be:
 - auto-updating and automatically switching the bloop version behind nailgun (maybe maintaining a pool of versions running, if people use different scalafmt versions in different projects and interleave interactions with them...)
 - making the dynamic runner as fast as the core one (likely not possible?)

Having the flag would be more than enough for me though :)

kitbellew:
- `ScalafmtSession`: add an interface for batch format. If we need to format a number of files using the same exact scalafmt configuration, let's create a session.
  - This will also allow us to filter inputs based on the project files which was not available and led to dynamic runner loading and formatting all files, including those not covered by the configuration.
- `ScalafmtDynamicRunner`: pre-filter based on customFiles if specified.
- `ScalafmtDynamicRunner`: try to use `ScalafmtSession` to avoid loading unnecessary files

Fixes #1460.

### DIFF
--- a/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala
+++ b/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala
@@ -27,7 +27,7 @@ final case class ScalafmtDynamic(
     configsCache: ReentrantCache[Path, FormatEval[
       (ScalafmtReflectConfig, FileTime)
     ]]
-) extends Scalafmt {
+) extends ScalafmtSessionFactory {
 
   def this() =
     this(
@@ -241,6 +241,23 @@ final case class ScalafmtDynamic(
 
   override def withMavenRepositories(repositories: String*): Scalafmt =
     copy(repositories = repositories.map(MavenRepository.of).toList)
+
+  override def createSession(config: Path): ScalafmtSession =
+    resolveConfig(config).fold(
+      error => { reportError(config, error); null },
+      config => new MySession(config)
+    )
+
+  private class MySession(cfg: ScalafmtReflectConfig) extends ScalafmtSession {
+    override def format(file: Path, code: String): String =
+      tryFormat(file, code, cfg.fmtReflect, cfg).fold(
+        error => { reportError(file, error); code },
+        formatted => formatted
+      )
+    override def matchesProjectFilters(file: Path): Boolean =
+      cfg.isIncludedInProject(file.toString)
+  }
+
 }
 
 object ScalafmtDynamic {
--- a/scalafmt-interfaces/src/main/java/org/scalafmt/interfaces/ScalafmtSession.java
+++ b/scalafmt-interfaces/src/main/java/org/scalafmt/interfaces/ScalafmtSession.java
@@ -1,0 +1,27 @@
+package org.scalafmt.interfaces;
+
+import java.nio.file.Path;
+
+/**
+ * A session based on a fixed configuration.
+ */
+public interface ScalafmtSession {
+
+    /**
+     * Format a single file with the given configuration.
+     *
+     * @param file relative or absolute path to the file being formatted. Used only for the path
+     *             name, the file does not have to exist on disk.
+     * @param code the text contents to format.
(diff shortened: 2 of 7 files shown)


scalafmt-interfaces/src/main/java/org/scalafmt/interfaces/ScalafmtSession.java
```diff
@@ -0,0 +1,27 @@
+package org.scalafmt.interfaces;
+
+import java.nio.file.Path;
+
+/**
+ * A session based on a fixed configuration.
+ */
+public interface ScalafmtSession {
```
poslegm, line 8: It is in interface module because `ScalafmtSession` should be supported in another integrations (sbt plugin for example)?

scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala
```diff
@@ -241,6 +241,23 @@ final case class ScalafmtDynamic(
 
   override def withMavenRepositories(repositories: String*): Scalafmt =
     copy(repositories = repositories.map(MavenRepository.of).toList)
+
+  override def createSession(config: Path): ScalafmtSession =
+    resolveConfig(config).fold(
+      error => { reportError(config, error); null },
```
poslegm, line 247: May be just throw exeption instead of `null` handling?

kitbellew:
@poslegm since this `Session` interface hasn't yet been released (and making changes later is breaking), should we consider also adding a method there to format an entire batch?

kitbellew:
> @poslegm since this `Session` interface hasn't yet been released (and making changes later is breaking), should we consider also adding a method there to format an entire batch?

@poslegm i think we are mostly ready to release, but before we do that, do you want to implement the batch as above? don't want to change the `ScalafmtSession` interface later :)

poslegm:
I thought the session already allows scalafmt to format multiple files without overhead. What is additional benefit from batch method?

kitbellew:
> I thought the session already allows scalafmt to format multiple files without overhead. What is additional benefit from batch method?

code reuse, really :) nothing else.

poslegm:
then I have no wishes for this :)

kitbellew:
> then I have no wishes for this :)

should we release RC3?

poslegm:
I am OK with it

PR #1914, opened 2020-04-27
Usage of default align token owners

### LINKED ISSUE #1896: 2.5.0-RC1: Alignment is too aggressive, aligns over unrelated expressions
- **Version**: 2.5.0-RC1
- **Configuration**:
```
version = "2.5.0-RC1"
align.tokens.add = ["="]
```

## Steps

Given code like this, formatted with scalafmt `2.4.2`:

```scala
object fmt {
  val header  = JwtHeader(algorithm = Some(JwtAlgorithm.HMD5))
  val content = payload.toString
  val claim   = JwtClaim(content = content)
}
```
## Problem

Scalafmt re-formats this code like this, on `2.5.0-RC1`:
```scala
object fmt {
  val header  = JwtHeader(algorithm = Some(JwtAlgorithm.HMD5))
  val content = payload.toString
  val claim   = JwtClaim(content    = content)
}
```

## Workaround

Setting `align = most` instead of adding `=` makes equals align only for `val _ =`, but not in expressions. However, this _removes_ alignment in "config-style" applications like:
```scala
JwtClaim(
  abc     = cba,
  content = content
)
```
->
```scala
JwtClaim(
  abc = cba,
  content = content
)
```

## Expectation

Expected the output to not change. the `=` on 3rd line aligns over to `=` on first line in completely unrelated expression, which seems like too much.

Note the alignment for `val _ =` is expected to be preserved for multiple vals in a row...

poslegm:
Fixes #1896

Motivation described in the issue. Now users don't need to copy-paste
token owners from sources to get optimal alignment for dedicated tokens.

It will slightly change the value of user configs so need to be described in release notes.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/AlignToken.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/AlignToken.scala
@@ -25,7 +25,9 @@ object AlignToken {
   implicit val DefaultAlignTokenDecoder: ConfDecoder[AlignToken] =
     ConfDecoder.instance[AlignToken] {
       case Conf.Str("caseArrow") => Ok(caseArrow)
-      case Conf.Str(regex) => Ok(AlignToken(regex, ".*"))
+      case Conf.Str(regex) =>
+        val owner = default.find(_.code == regex).fold(".*")(_.owner)
+        Ok(AlignToken(regex, owner))
       case els => fallbackAlign.reader.read(els)
     }
 
--- a/scalafmt-tests/src/test/resources/align/defaultTokenOwners.stat
+++ b/scalafmt-tests/src/test/resources/align/defaultTokenOwners.stat
@@ -1,0 +1,39 @@
+align.tokens.add = ["="]
+<<< don't align independent columns (see #1896)
+object fmt {
+  val header = JwtHeader(algorithm = Some(JwtAlgorithm.HMD5))
+  val content = payload.toString
+  val claim = JwtClaim(content = content)
+}
+>>>
+object fmt {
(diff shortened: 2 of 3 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/AlignToken.scala
```diff
@@ -25,7 +25,9 @@ object AlignToken {
   implicit val DefaultAlignTokenDecoder: ConfDecoder[AlignToken] =
     ConfDecoder.instance[AlignToken] {
       case Conf.Str("caseArrow") => Ok(caseArrow)
-      case Conf.Str(regex) => Ok(AlignToken(regex, ".*"))
+      case Conf.Str(regex) =>
+        val owner = default.find(_.code == regex).fold(".*")(_.owner)
```
kitbellew, line 29: looks good.

it probably doesn't play well with a `preset` until the metaconfig changes are merged, right?
let's say,
```
preset = defaultWithAlign
align.preset = ...
```
will mean we are not starting with `default`.

poslegm: `align.tokens.add` successfully overrides preset:

```
<<< ONLY respect specified owner
preset = defaultWithAlign
align.preset = most
align.tokens.add = [{code = "=", owner = "(Term.Assign|Defn.Val)"}]
===
object fmt {
  val header = JwtHeader(algorithm = Some(JwtAlgorithm.HMD5))
  val content = payload.toString
  val claim = JwtClaim(content = content)
}
>>>
object fmt {
  val header  = JwtHeader(algorithm = Some(JwtAlgorithm.HMD5))
  val content = payload.toString
  val claim   = JwtClaim(content    = content)
}
```

PR #1915, opened 2020-04-27
Pretty format multiline expressions in enumerator with align.arrowEnumeratorGenerator

### LINKED ISSUE #1797: align.arrowEnumeratorGenerator formats if/else weird
- **Version**: 2.4.2
- **Integration**: cli
- **Configuration**:
```
version = 2.4.2
```

## Steps

Given code like this:
```scala
for {
  variable <- record.field.field1.map(_.toString)
  cond <- if (variable) doSomething else doAnything
} yield cond
```

```bash
scalafmt
```

## Problem

```scala
for {
  variable <- record.field.field1
               .map(_.toString)
  cond <- if (variable) doSomething
         else doAnything
} yield cond
```

## Expectation

```scala
for {
  variable <- record.field.field1
                .map(_.toString)
  cond <- if (variable) doSomething
          else doAnything
} yield cond
```

## Notes

Looks like we need just +1 indent on [StateColumn](https://github.com/scalameta/scalafmt/blob/e1bbdfc335b3d5ed24faec5a1bfa4a9b43e90e4f/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Indent.scala#L28) in this case

poslegm:
Fixes #1797

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -2093,7 +2093,34 @@ class Router(formatOps: FormatOps) {
           if (close.isEmpty) Split.ignored else Split(Space, 1)
 
         case Newlines.classic if style.align.arrowEnumeratorGenerator =>
-          Split(Space, 1).withIndent(StateColumn, expire, After)
+          Split(Space, 1)
+            .withIndent(StateColumn, expire, After)
+            /**
+              * This gap is necessary for pretty alignment multiline expressions
+              * on the right-side of enumerator.
+              * Without:
+              * ```
+              * for {
+              *    a <- new Integer {
+              *          value = 1
+              *        }
+              *   x <- if (variable) doSomething
+              *       else doAnything
+              * }
+              * ```
+              *
+              * With:
+              * ```
+              * for {
+              *    a <- new Integer {
+              *           value = 1
+              *         }
+              *   x <- if (variable) doSomething
+              *        else doAnything
+              * }
+              * ```
+              * */
+            .withIndent(Num(1), expire, After)
 
         case Newlines.fold | Newlines.classic =>
           postCommentFT.meta.rightOwner match {
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -2081,7 +2081,9 @@ class Router(formatOps: FormatOps) {
       Seq(Split(Space.orNL(ft.noBreak), 0).withIndent(2, expire, After))
     else if (style.newlines.sourceIs(Newlines.keep))
       Seq(
-        if (ft.noBreak) Split(Space, 0)
+        if (ft.noBreak)
+          Split(Space, 0)
+            .withIndents(arrowEnumeratorGeneratorAlignIndents(expire))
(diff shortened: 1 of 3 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -2093,7 +2093,34 @@ class Router(formatOps: FormatOps) {
           if (close.isEmpty) Split.ignored else Split(Space, 1)
 
         case Newlines.classic if style.align.arrowEnumeratorGenerator =>
-          Split(Space, 1).withIndent(StateColumn, expire, After)
+          Split(Space, 1)
```
kitbellew, line 2096: that's probably my fault for not handling this case before, we might want to consider non-classic cases as well.

should we just add this indent once to line 2114 (old) or 2141 (new) to all cases if this flag is on?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -2093,7 +2093,34 @@ class Router(formatOps: FormatOps) {
...
+              * for {
+              *    a <- new Integer {
+              *           value = 1
+              *         }
+              *   x <- if (variable) doSomething
+              *        else doAnything
+              * }
+              * ```
+              * */
+            .withIndent(Num(1), expire, After)
```
kitbellew, line 2123: is it because `<-` is two-characters long? would it still work with the single-character unicode left arrow?

poslegm: No, it's because of space after arrow. Added test for unicode

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
poslegm: It works with `fold`.

`unfold` always add newline after `<-`. Looks like it doesn't support this alignment type at all. May be mark them as conflicting?

```
for {
  x <-
    new Integer {
      def value = 2
    }
  y <- Int(2)
  cond <-
    if (variable)
      doSomething
    else
      doAnything
} yield x + y
```

`keep` formats code ugly, it wasn't fixed in https://github.com/scalameta/scalafmt/pull/1795 :(

```
for {
  x <- new Integer {
    def value = 2
  }
  y <- Int(2)
  cond <- if (variable) doSomething
  else doAnything
} yield x + y
```

kitbellew: looking at the code, `unfold` doesn't add newline for expressions enclosed in `()` (and all except `keep` do the same for `{}`). don't remember why.

poslegm: unfold with braces:

```
for {
  x <- (
      new Integer {
        def value = 2
      }
  )
  y <- Int(2)
  cond <- (
      if (variable)
        doSomething
      else
        doAnything
  )
} yield x + y
```

I still don't think that the `align.arrowEnumeratorGenerator` is applicable here

poslegm:
`SKIP` because keep currently formats this test like

```
for {
  x <- new Integer {
    def value = 2
  }
  y <- Int(2)
  cond <- if (variable) doSomething
  else doAnything
} yield x + y
```

It should be fixed in forward to #1795

poslegm:
Sorry, it's my bad. Keep works well.

PR #1918, opened 2020-04-28
DanglingParentheses: add ctrlSite flag

### LINKED ISSUE #1190: Add configuration to format multiline conditionals with dangling parenthesis
With `danglingParenthesis = true` I would expect conditionals to be formatted like this -

```scala
if (
  numTries > 0 &&
  oldestCompleteTimestampOpt.isDefined &&
  activityCache.nonEmpty
) {
```

instead of -

```scala
if (numTries > 0 &&
    oldestCompleteTimestampOpt.isDefined &&
    activityCache.nonEmpty) {
```

I'm thinking this should either fall under the `danglingParenthesis` config, or there should be some other config to get this output (maybe there already is?). Anyways, this is very low priority but any guidance would be greatly appreciated. Thank you always!

kitbellew:
Fixes #1190.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/DanglingParentheses.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/DanglingParentheses.scala
@@ -5,6 +5,7 @@ import metaconfig._
 case class DanglingParentheses(
     callSite: Boolean,
     defnSite: Boolean,
+    ctrlSite: Boolean = true,
     exclude: List[DanglingParentheses.Exclude] = Nil
 ) extends Decodable[DanglingParentheses] {
   override protected[config] def baseDecoder =
@@ -14,7 +15,7 @@ case class DanglingParentheses(
 object DanglingParentheses {
 
   private val shortcutTrue = DanglingParentheses(true, true)
-  private val shortcutFalse = DanglingParentheses(false, false)
+  private val shortcutFalse = DanglingParentheses(false, false, false)
 
   val default = shortcutTrue
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -554,6 +554,27 @@ object a {
 }
 ```
 
+#### `danglingParentheses.ctrlSite`
+
+Forces dangling on open/close parens around control structures
+(`if`, `while`, `for`) when line breaks must occur.
+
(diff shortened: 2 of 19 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/DanglingParentheses.scala
```diff
@@ -5,6 +5,7 @@ import metaconfig._
 case class DanglingParentheses(
     callSite: Boolean,
     defnSite: Boolean,
+    kwSite: Boolean = false,
```
olafurpg, line 8: What does "kw" stand for? Can we use a more self-descriptive name?

kitbellew: well, it stands for `keyword`. but since keywords cover a lot more, perhaps it should be `ctrl` (for control flow, which is exactly `if`, `while` and `for`... well, and `case` but it doesn't apply here).

olafurpg:
What would be the implication if we change the default to `align.ifWhileOpenParam = false` ?

I like `ctrlSite` more than `kwSite` (at least, it's more aesthetically pleasing since it's four character long like the other settings 😅 )

kitbellew:
> What would be the implication if we change the default to `align.ifWhileOpenParam = false` ?

1. we need to remove it from `align = some, more` (typical presets)
2. lots of howling on gitter; people are very attached to their defaults.

> I like `ctrlSite` more than `kwSite` (at least, it's more aesthetically pleasing since it's four character long like the other settings 😅 )

i have the same ocd ;)

olafurpg:
>     we need to remove it from align = some, more (typical presets)

My understanding is that people primarily enable those to align on `=`, `extends` and other tokens that repeatedly appear in multiple lines but less so for alignment against a single `(` 

>    lots of howling on gitter; people are very attached to their defaults.

As long as they can override the defaults I think it's OK. I care most about having nice defaults for a beginner starting with Scala.

olafurpg:
I'm happy either way we go, but my personal vote is in favor of enabling this by default

kitbellew:
> I'm happy either way we go, but my personal vote is in favor of enabling this by default

good enough for me.

PR #1919, opened 2020-04-28
ContinuationIndent: add granularity to defnSite

### LINKED ISSUE #1487: Granularity on defn site continuation indent
- **Version**: 2.0.0
- **Integration**: IntelliJ
- **Configuration**:
```
```

## Steps

Given code like this:
```scala
class SomeFoobar[Val](
    field1: List[Foo],
    field2: Option[Baz],
    val field3: FooBar[Val])

def apply[Ctx](
    field1FollowedByLongName: Option[Foo],
    field2FollowedByLongName: Option[Foo],
    field3FollowedByLongName: Option[Foo],
    field4FollowedByLongName: Option[Foo],
    field5FollowedByLongName: Option[Foo],
    field6FollowedByLongName: Option[Foo] = None,
    field7FollowedByLongName: Option[Foo]
)
```

The continuation indent is the same for both method and class defs. We'd like the granularity we had with scalariform. 

Can we split `continuationIndent.defnSite` along method and class lines? 
We wish to set 2 for methods and 4 for classes.

Perhaps to do this, introduce two new option to avoid breaking old versions?

`continuationIndent.classDefnSite`
`continuationIndent.fnDefnSite`

kitbellew:
Fixes #1487.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ContinuationIndent.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ContinuationIndent.scala
@@ -4,6 +4,7 @@ import metaconfig._
 
 /**
   * @param defnSite indentation around class/def
+  * @param ctorSite indentation around class constructor parameters
   * @param callSite indentation around function calls, etc.
   * @param extendSite indentation before `extends`
   * @param withSiteRelativeToExtends additional indentation before `with`
@@ -11,11 +12,18 @@ import metaconfig._
 case class ContinuationIndent(
     callSite: Int = 2,
     defnSite: Int = 4,
+    ctorSite: Option[Int] = None,
     extendSite: Int = 4,
     withSiteRelativeToExtends: Int = 0
 ) {
   implicit val reader: ConfDecoder[ContinuationIndent] =
     generic.deriveDecoder(this).noTypos
+
+  def getDefnSite(tree: meta.Tree): Int =
+    (tree match {
+      case _: meta.Ctor => ctorSite
+      case _ => None
+    }).getOrElse(defnSite)
 }
 
 object ContinuationIndent {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -187,6 +187,23 @@ argument1: Type1 // indented by 4
 ): ReturnType
 ```
 
+### `continuationIndent.ctorSite`
+
+Applies to constructors. Defaults to `continuationIndent.defnSite`.
+
+```scala mdoc:scalafmt
+continuationIndent.ctorSite = 4
+continuationIndent.defnSite = 2
+---
+class A(
+ field1: Type1 // indented by 4
+) {
+ def function2(
+  argument1: Type1 // indented by 2
+ ): ReturnType = None
+}
+```
+
 ### `continuationIndent.extendSite`
 
 ```scala mdoc:defaults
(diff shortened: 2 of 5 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/ContinuationIndent.scala
```diff
@@ -4,18 +4,29 @@ import metaconfig._
...
     extendSite: Int = 4,
     withSiteRelativeToExtends: Int = 0
 ) {
   implicit val reader: ConfDecoder[ContinuationIndent] =
     generic.deriveDecoder(this).noTypos
+
+  def getDefnSite(tree: meta.Tree): Int =
+    (tree match {
+      case _: meta.Ctor => ctorSite
+      case _: meta.Decl => declSite
```
olafurpg, line 27: In what situation do we want to distinguish declarations from definitions?

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ContinuationIndent.scala
```diff
@@ -4,18 +4,29 @@ import metaconfig._
...
+  * @param ctorSite indentation around class constructor parameters
   * @param callSite indentation around function calls, etc.
   * @param extendSite indentation before `extends`
   * @param withSiteRelativeToExtends additional indentation before `with`
   */
 case class ContinuationIndent(
     callSite: Int = 2,
     defnSite: Int = 4,
+    declSite: Option[Int] = None,
+    ctorSite: Option[Int] = None,
```
olafurpg, line 17: I agree this name is most consistent with the other options since it matches the scala.meta.Tree names. Still, I regret using "defnSite/declSie" instead of "definitionSite" and "declarationSite"

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ContinuationIndent.scala
kitbellew: frankly, i have no idea. heard someone mention ability to distinguish abstract and concrete methods visually. also, possibly because i don't want to come back to this later, although that's questionable reason.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ContinuationIndent.scala
kitbellew: we can deprecate `defnSite` or use annotation that `metaconfig` allows for aliasing. or leave it be. what would you do?

olafurpg: We can leave this unchanged for now and then separately give readable names to all options. We can continue to support the old names using metaconfig `@DeprecatedName(..)` https://github.com/scalameta/metaconfig/blob/aaab8feb0d7b22f6ae78dda3f291839537e2006b/metaconfig-core/shared/src/main/scala/metaconfig/annotation/Annotations.scala#L7

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ContinuationIndent.scala
olafurpg: I would avoid introducing the decl/defn separation since it's not self-explanatory to most users what the difference is. The original ticket specifically was about separating class constructors from methods, which seems reasonable

kitbellew: all right, will keep `ctor` only.

PR #1922, opened 2020-04-28
State bugfix: filter new indents the same as old

kitbellew:


scalafmt-tests/src/test/resources/default/Apply.stat
```diff
@@ -721,3 +721,35 @@ align.preset = none
...
+<<< config style with empty arg list 3
+object a {
+  foo(
+    // c1
+  )
+}
+>>>
+object a {
+  foo(
+      // c1
```
poslegm, line 753: Can you please explain motivation for this extra indent?

kitbellew: i think `continuationIndent.callSite` is 4 in this test.

PR #1934, opened 2020-04-30
FormatWriter: optionally support multiline align

### LINKED ISSUE #1504: Case arrows not aligning
I've a question regarding alignment on cases for a match expression. It looks like
the arrows are not lining up with `aligh = some` or `align = more`
- **Version**: 2.0.0
- **Integration**: IntelliJ
- **Configuration**:
```
maxColumn = 100
docstrings = ScalaDoc // default

// Alignment settings
align = some
align.arrowEnumeratorGenerator = false
align.openParenDefnSite = false // default but setting for reminder

newlines.afterCurlyLambda = never

binPack.parentConstructors = true
importSelectors = binPack

verticalMultiline.atDefnSite = false

includeNoParensInSelectChains = false // this is the default value
indentYieldKeyword = true

optIn.blankLineBeforeDocstring = true

trailingCommas = never
lineEndings = unix

rewrite.rules = [SortImports, RedundantBraces, RedundantParens, SortModifiers, PreferCurlyFors]

danglingParentheses = true
```

## Steps

Given code like this:
```scala
  trait Identifier

  case class SuperFooBarIdentifier(identifyingid: String) extends Identifier

  case class TastyBarIdentifier (ider: String, int: Int) extends Identifier

  case class BazBar(identifier: Identifier)

  val stringFoo = thing.identifier match {
    case foo : SuperFooBarIdentifier => "somestring" / "other_string" / foo.identifyingid
    case tastyBar: TastyBarIdentifier => "something" / "other_other" ? ("arg1" -> tastyBar.ider) & ("limit" -> 111)
    case unknown => throw new IllegalArgumentException(s"Not sure at all what kinda class this is: ${unknown.getClass}")
  }
```

```bash
scalafmt
```

## Problem

```scala
val stringFoo = thing.identifier match {
    case foo: SuperFooBarIdentifier => "somestring" / "other_string" / foo.identifyingid
    case tastyBar: TastyBarIdentifier =>
      "something" / "other_other" ? ("arg1" -> tastyBar.ider) & ("limit" -> 111)
    case unknown =>
      throw new IllegalArgumentException(
        s"Not sure at all what kinda class this is: ${unknown.getClass}"
      )
  }
```

## Expectation

```scala
  val stringFoo = thing.identifier match {
    case foo: SuperFooBarIdentifier   => "somestring" / "other_string" / foo.identifyingid
    case tastyBar: TastyBarIdentifier =>
      "something" / "other_other" ? ("arg1" -> tastyBar.ider) & ("limit" -> 111)
    case unknown                      =>
      throw new IllegalArgumentException(
        s"Not sure at all what kinda class this is: ${unknown.getClass}"
      )
  }
```

### LINKED ISSUE #1811: Preserve alignment in for comprehension with multiline
- **Configuration**:
```
version = 2.4.2
align = most
```

## Steps

Given code like this:
```scala
for {
  a   <- Option(1)
  bbb <- Option(2)
  cccccc <- Option {
    3
   }
} yield ()
```

```bash
scalafmt
```

## Problem

```scala
for {
  a   <- Option(1)
  bbb <- Option(2)
  cccccc <- Option {
             3
           }
} yield ()
```

## Expectation

```scala
for {
  a      <- Option(1)
  bbb    <- Option(2)
  cccccc <- Option {
              3
            }
} yield ()
```

### LINKED ISSUE #1841: Fails to align with equals if line is wrapped
## Relevant config
```
version = 2.4.2
style = defaultWithAlign
```

## Steps

Upgrading from `2.3.2` to `2.4.2` gives the following. The wrap is correct as it is greater than the default but wrapping makes the equals not align anymore:
```scala
-  private final val countGPRegisters  = 6
-  private final val countFPRegisters  = 8
-  private final val fpRegisterWords   = 2
-  private final val registerSaveWords = countGPRegisters + countFPRegisters * fpRegisterWords
+  private final val countGPRegisters = 6
+  private final val countFPRegisters = 8
+  private final val fpRegisterWords  = 2
+  private final val registerSaveWords =
+    countGPRegisters + countFPRegisters * fpRegisterWords
```
Also this.
```scala
-  private[this] val pbClass       = Class.forName("java.lang.ProcessBuilder")
-  private[this] val redirectClass = pbClass.getClasses find (_.getSimpleName == "Redirect")
+  private[this] val pbClass = Class.forName("java.lang.ProcessBuilder")
+  private[this] val redirectClass =
+    pbClass.getClasses find (_.getSimpleName == "Redirect")
```

kitbellew:
To support this, also fix a couple of issues:

* include align gap in width:
  * because we sort on the width, we might end up with (width+gap) for the currently widest alignment spot to be less than for another, with a smaller width.
* stop align container search
  * there's a different class of containers (currently, "case"), which would only serve as a container for tokens which are formatted below it, and for tokens on the same line, we would continue the search, so that those tokens would align together with the case arrow.

Fixes #1504.
Fixes #1811.
Fixes #1841.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
@@ -61,6 +61,7 @@ import metaconfig.generic.Surface
   *   of the first line, as usual.
   */
 case class Align(
+    multiline: Boolean = false,
     stripMargin: Boolean = true,
     openParenCallSite: Boolean = false,
     openParenDefnSite: Boolean = false,
@@ -107,6 +108,7 @@ object Align {
   // only for the truest vertical aligners, this setting is open for changes,
   // please open PR addding more stuff to it if you like.
   val most: Align = more.copy(
+    multiline = true,
     arrowEnumeratorGenerator = true,
     tokenCategory = Map(
       "Equals" -> "Assign",
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -175,7 +175,7 @@ class FormatWriter(formatOps: FormatOps) {
 
   class FormatLocations(val locations: Array[FormatLocation]) {
 
-    val tokenAligns: Map[TokenHash, Int] = alignmentTokens(locations)
+    val tokenAligns: Map[TokenHash, Int] = alignmentTokens
 
     def iterate: Iterator[Entry] =
       Iterator.range(0, locations.length).map(new Entry(_))
@@ -372,6 +372,70 @@ class FormatWriter(formatOps: FormatOps) {
       }
 
     }
+
+    /**
+      * Returns how many extra spaces are needed to align tokens, as configured
+      * by `initStyle.align.tokens`.
+      */
+    // TODO(olafur) Refactor implementation to make it maintainable. It's super
+    // imperative and error-prone right now.
+    private def alignmentTokens: Map[TokenHash, Int] = {
+      lazy val noAlignTokens = initStyle.align.tokens.isEmpty &&
+        styleMap.tok2style.values.forall(_.align.tokens.isEmpty)
+      if (locations.length != tokens.length || noAlignTokens)
+        Map.empty[TokenHash, Int]
+      else {
+        var columnShift = 0
+        val finalResult = Map.newBuilder[TokenHash, Int]
+        val block = new AlignBlock
+        val locationIter = new LocationsIterator(locations)
+        while (locationIter.hasNext) {
+          var alignContainer: Tree = null
+          val columnCandidates = IndexedSeq.newBuilder[FormatLocation]
+          def shouldContinue(floc: FormatLocation): Boolean = {
+            val ok = !floc.state.split.modification.isNewline
+            if (!ok || floc.formatToken.leftHasNewline) columnShift = 0
+            columnShift += floc.shift
+            if (ok)
+              getAlignContainer(floc).foreach { container =>
+                if (alignContainer eq null)
+                  alignContainer = container
+                if (alignContainer eq container)
+                  columnCandidates += floc.copy(
+                    alignContainer = container,
+                    alignHashKey = getAlignHashKey(floc),
+                    shift = floc.state.prev.column + columnShift
+                  )
+              }
+            ok
+          }
+          while (shouldContinue(locationIter.next()) && locationIter.hasNext) {}
+          val location = locationIter.last
+          val candidates = columnCandidates.result()
+          val doubleNewline = location.state.split.modification.newlines > 1
+          if (block.nonEmpty) {
+            val prev = block.last
+            val matches = columnMatches(prev, candidates, location.formatToken)
+            if (matches > 0) {
+              // truncate candidates if matches are shorter than both lists
+              val truncate =
+                matches < prev.length && matches < candidates.length
+              if (truncate) {
+                block(block.length - 1) = prev.take(matches)
+                block += candidates.take(matches)
+              } else block += candidates
+            }
+            if (matches == 0 || doubleNewline || !locationIter.hasNext) {
+              flushNonEmptyAlignBlock(block, finalResult)
+            }
+          }
+          if (block.isEmpty && candidates.nonEmpty && !doubleNewline)
+            block += candidates
+        }
+        finalResult.result()
+      }
+    }
+
   }
 
   private def isCloseDelimForTrailingCommasMultiple(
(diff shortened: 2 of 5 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
```diff
@@ -61,6 +61,7 @@ import metaconfig.generic.Surface
   *   of the first line, as usual.
   */
 case class Align(
+    multiline: Boolean = false,
```
poslegm, line 64: Test cases looks great. Does this have to be a separate option? May be just apply it for `align!=none`?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -376,6 +373,155 @@ class FormatWriter(formatOps: FormatOps) {
...
+      else {
+        var columnShift = 0
+        val finalResult = Map.newBuilder[TokenHash, Int]
+
+        // all blocks must be here, to get final flush
+        val blocks = new mutable.HashMap[Tree, AlignBlock]
+        def createBlock(x: Tree) = blocks.getOrElseUpdate(x, new AlignBlock)
+        var prevAlignContainer: Tree = null
+        var prevBlock: AlignBlock =
+          if (initStyle.align.multiline) null else createBlock(null)
```
poslegm, line 397: What is purpose of `createBlock(null)` here? Is it necessary to have map with single empty block on line 456?

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
kitbellew: if we enable this code by default, everyone's code will change and people will complain. this is a potentially massive change.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -376,6 +373,155 @@ class FormatWriter(formatOps: FormatOps) {
...
+            if (doubleNewline || matches == 0 && shouldFlush(alignContainer))
+              flushNonEmptyAlignBlock(block, finalResult)
+          }
+          if (alignContainer ne null) {
+            prevAlignContainer = alignContainer
+            prevBlock = block
+            if (block.isEmpty && !doubleNewline) block += candidates
+          }
+        }
+        blocks.valuesIterator.foreach { x =>
```
kitbellew, line 456: this is the flush i am referring to above, on `createBlock`.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -376,6 +373,155 @@ class FormatWriter(formatOps: FormatOps) {
...
+            if (matches > 0) {
+              // truncate candidates if matches are shorter than both lists
+              val truncate =
+                matches < prev.length && matches < candidates.length
+              if (truncate) {
+                block(block.length - 1) = prev.take(matches)
+                block += candidates.take(matches)
+              } else block += candidates
+            }
+            if (doubleNewline || matches == 0 && shouldFlush(alignContainer))
```
kitbellew, line 447: this line would need another check for non-multiline case (see comments on `createBlock`).

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: i considered that adding a map with a single entry for an entire document is very cheap. otherwise, see two comments below, that code would have to be modified to distinguish between multiline and non-multiline, which increases clutter.

but if you prefer, i can easily change that.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
poslegm: Yes, it's a valid point. But what about true by default?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
poslegm: No, it's just clarification 👍

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
kitbellew: i'd leave it to you and @olafurpg to decide if you want to change defaults :)

i think it would be ok to include it as default in `align.preset = most` as it is advertised in the configuration as likely to be aggressive and also change in the future.

poslegm: True for `most` preset is optimal solution

kitbellew: done. also added documentation.

olafurpg: True for most sounds good.

PR #1943, opened 2020-05-03
FormatWriter: flush alignment blocks on blank line

kitbellew:
This didn't include multiline blocks.

### DIFF
--- a/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
+++ b/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
@@ -776,12 +776,12 @@ object a {
 }
 >>>
 object a {
-  def foo: Int    = {
+  def foo: Int = {
     val x = ???
     42
   }
 
-  def bar: Int    = {
+  def bar: Int = {
     for { x <- "hey" } yield x
     42
   }
--- a/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
+++ b/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
@@ -752,3 +752,81 @@ object a {
   private[this] val redirectClass =
     pbClass.getClasses find (_.getSimpleName == "Redirect")
 }
+<<< multiline with blanks
+align.preset = most
+===
+object a {
+  def foo: Int = {
+    val x = ???
+    42
+  }
+
+  def bar: Int = {
+    for { x <- "hey" } yield x
+    42
+  }
+
+  def fooBig: Int = {
+    42
+  }
+
+  def fooBigger: Int = {
+    42
+  }
+}
+>>>
+object a {
+  def foo: Int    = {
+    val x = ???
+    42
+  }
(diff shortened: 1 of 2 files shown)


scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
```diff
@@ -752,3 +752,81 @@ object a {
...
+  def fooBig: Int = {
+    42
+  }
+  def fooBigger: Int = {
+    42
+  }
+}
+>>>
+object a {
+  def foo: Int       = {
```
poslegm, line 818: Is it possible to flush blocks for this case too? I don't think that anyone wants to align `=` owned by function definition

kitbellew: > Is it possible to flush blocks for this case too? I don't think that anyone wants to align `=` owned by function definition

there's a case in the tests supporting this:
```
  private[this] val pbClass       = Class.forName("java.lang.ProcessBuilder")
  private[this] val redirectClass =
    pbClass.getClasses find (_.getSimpleName == "Redirect")
```

do you want to restrict just `Defn.Def` and not `Defn.Val`? what about `def` instead of `val` in that example:
```
  private[this] def pbClass       = Class.forName("java.lang.ProcessBuilder")
  private[this] def redirectClass =
    pbClass.getClasses find (_.getSimpleName == "Redirect")
```

kitbellew: @poslegm if you don't mind, i will merge this as-is, and then send you a patch for review to try to figure out what to do about `=` in definitions.

poslegm: > do you want to restrict just Defn.Def and not Defn.Val?

I just forgot that scalafmt aligns one-line defs  by `=` already. So, it's OK and consistent to align also multilne 👍

scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
```diff
@@ -752,3 +752,81 @@ object a {
   private[this] val redirectClass =
     pbClass.getClasses find (_.getSimpleName == "Redirect")
 }
+<<< multiline with blanks
+align.preset = most
+===
+object a {
```
ghostdogpr, line 758: For some reason if you add a comment, the problem is coming back. For example:
```scala
  val foo: Int = {
    val x = ???
    42
  }

  val bar: Int = {
    for { x <- "hey" } yield x
    42
  }

  val fooBig: Int    = {
    42
  }

  // a
  val fooBigger: Int = {
    42
  }
```
I added the `// a` and it tried to align with the code above.

kitbellew: ok, will take a look. comment probably determines the alignment container incorrectly.

PR #1957, opened 2020-05-06
When using in pipeline don't print info to stderr

sideshowcoder:
When receiving data via stdIn and printing to stdout don't print info messages to stderr. Stderr output might be used to detect failure, info messages (like "Reformatting..."). The alternative is to use --quiet which also removes valid error messages.

The use-case here is the integration into emacs format-all plugin, which is currently using `--quiet` to silence errors. Ideally it should report the errors on failure, but not show info output on every format.

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala
@@ -54,7 +54,7 @@ object CliOptions {
         out =
           guardPrintStream(parsed.quiet && !parsed.stdIn)(parsed.common.out),
         info = guardPrintStream(
-          parsed.quiet || parsed.writeMode == WriteMode.List
+          parsed.stdIn || parsed.quiet || parsed.writeMode == WriteMode.List
         )(auxOut),
         debug = guardPrintStream(parsed.quiet)(
           if (parsed.debug) auxOut else init.common.debug
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala
@@ -54,7 +54,7 @@ object CliOptions {
         out =
           guardPrintStream(parsed.quiet && !parsed.stdIn)(parsed.common.out),
         info = guardPrintStream(
-          parsed.stdIn || parsed.quiet || parsed.writeMode == WriteMode.List
+          parsed.stdIn || parsed.writeMode == WriteMode.Stdout || parsed.quiet || parsed.writeMode == WriteMode.List
         )(auxOut),
         debug = guardPrintStream(parsed.quiet)(
           if (parsed.debug) auxOut else init.common.debug
--- a/scalafmt-tests/src/test/scala/org/scalafmt/cli/CliOptionsTest.scala
+++ b/scalafmt-tests/src/test/scala/org/scalafmt/cli/CliOptionsTest.scala
@@ -113,11 +113,12 @@ class CliOptionsTest extends AnyFunSuite {
     assert(opt.scalafmtConfig.isNotOk)
   }
 
-  test(
-    "don't write info when part of a pipeline receiving from stdin writing to stdout"
-  ) {
-    val args = Array("--stdin")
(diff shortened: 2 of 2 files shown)


scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala
```diff
@@ -54,7 +54,7 @@ object CliOptions {
         out =
           guardPrintStream(parsed.quiet && !parsed.stdIn)(parsed.common.out),
         info = guardPrintStream(
-          parsed.quiet || parsed.writeMode == WriteMode.List
+          parsed.stdIn || parsed.quiet || parsed.writeMode == WriteMode.List
```
kitbellew, line 57: is it important that one is reading from stdin? what if you are reading from a file but writing to stdout (via `WriteMode.StdOut`?)

sideshowcoder: Good point, as far as I understand it if I consume from `stdin` the result will always be written to `stdout` which is problematic when writing info to stderr as it can break pipelines, but the same is true when consuming from a file and writing to `stdout`. I'll change this.

scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala
```diff
@@ -54,7 +54,7 @@ object CliOptions {
         out =
           guardPrintStream(parsed.quiet && !parsed.stdIn)(parsed.common.out),
         info = guardPrintStream(
-          parsed.quiet || parsed.writeMode == WriteMode.List
+          parsed.stdIn || parsed.writeMode == WriteMode.Stdout || parsed.quiet || parsed.writeMode == WriteMode.List
```
kitbellew, line 57: sorry to repeat myself... why do we need to check for `stdIn`? what if `stdIn` is used with `WriteMode.Test`, wouldn't it make sense to continue writing out info messages?

sideshowcoder: I see what you mean now, sorry about not being clear here. As far as I understand the behaviour: If `—stdin` is passed the result will be written to  stdout, without also passing `—stdout`. Given that the behaviour I would expect is to have info not being written to stderr in this case, as it causes the mentioned issue. 

Tbh I was not aware of using test mode with `—stdin`, so would the recommendation be to use `—stdin` and `—stdout` to get the desired bahaviour for my case? 

To be honest I’m getting the feeling this redirecting depending on the situation is becoming a little complex to follow, and maybe the clearer approach would be to offer a `—no-info` flag to explicitly disable info instead of trying to infer it from the various modes. What do you think?

kitbellew: > To be honest I’m getting the feeling this redirecting depending on the situation is becoming a little complex to follow, and maybe the clearer approach would be to offer a `—no-info` flag to explicitly disable info instead of trying to infer it from the various modes. What do you think?

I think it might be a good analog to the `--no-stderr` flag. If not, I'd just check for `WriteMode.StdOut` as sufficient, and remove `stdIn` check as redundant.

sideshowcoder: `--stdin` does not set `WriteMode.Stdout` it is by default set to `WriteMode.Override`, even so `--stdin` causes the output to be written to stdout. To get consistent behaviour either `--stdin` would need to set `WriteMode.Stdout`, but this would make it order dependent in case someone also sets `--test` what the actual `WriteMode` ends up being. 

I'm not quite sure what the "correct" behaviour would be as far as I can see it:

1. `--stdin` sets `WriteMode.Stdout`
2. `--stdin` requires either the `--test` or `--stdout` flags to be set
3.  Keep as is with the default `--stdin` causing `WriteMode.Override`, but introduce `--no-info` flag

I'm not quite sure what makes most sense for this. I feel like `--stdin` with `WriteMode.Override` being set by default somehow seems wrong, but I'm not sure.

kitbellew: if that's correct, then we are overthinking it, and cli opt parsing should change.

PR #1960, opened 2020-05-07
Documentation: retire Homebrew, point to coursier

kitbellew:
Also, make the coursier examples directly usable via copy-paste, update the installation link and use scala 2.13 in examples.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -215,7 +215,7 @@ Create a standalone executable in `/usr/local/bin/scalafmt` with (sudo if
 necessary):
 
 ```sh
-coursier bootstrap org.scalameta:scalafmt-cli_2.12:@STABLE_VERSION@ \
+coursier bootstrap org.scalameta:scalafmt-cli_2.13:@STABLE_VERSION@ \
   -r sonatype:snapshots --main org.scalafmt.cli.Cli \
   --standalone \
   -o /usr/local/bin/scalafmt
@@ -225,7 +225,7 @@ scalafmt --version # should be @STABLE_VERSION@
 Alternatively you can create a slim 15 KiB bootstrap script with:
 
 ```sh
-coursier bootstrap org.scalameta:scalafmt-cli_2.12:@STABLE_VERSION@ \
+coursier bootstrap org.scalameta:scalafmt-cli_2.13:@STABLE_VERSION@ \
   -r sonatype:snapshots --main org.scalafmt.cli.Cli \
   -o scalafmt
 ./scalafmt --version # should be @STABLE_VERSION@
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -199,36 +199,35 @@ object MyScalafmtPlugin extends AutoPlugin {
 include ".scalafmt-common.conf"
 ```
 
-## CLI
+## CLI / Homebrew
```
olafurpg, line 202: This breaks existing links to the `#cli` anchor. I think it's better to keep the `Homebrew` header unchanged

docs/installation.md
```diff
@@ -307,18 +306,6 @@ faster when formatting a single file from the CLI. The downside to Nailgun is
 that the setup is complicated and the long-running server needs to be restarted
 once in awhile.
 
-### Homebrew
```
olafurpg, line 310: I think we should keep this header and recommend these steps instead

```sh
brew install coursier/formulas/coursier
cs install scalafmt
cs update scalafmt
```

docs/installation.md
kitbellew: changed. (i had checked whether `#cli` was used within the file but you're right, it might be referred to externally.)

docs/installation.md
kitbellew: i was thinking that we should refer to coursier's install page (in case something changes in the future) instead of explicit `brew install ...`.

using coursier to install scalafmt is described in its own section above, so wouldn't it be repetitive? (also, i couldn't find `cs` on my mac, with 2.0.0-RC3-4, not until after upgrading to 2.0.0-RC6-15; 2.0.0-RC3-4 also didn't support `install`, said it was experimental.)

docs/installation.md
```diff
@@ -309,15 +322,8 @@ once in awhile.
...
-
-```sh
-brew install --HEAD olafurpg/scalafmt/scalafmt
-scalafmt --version // should be @STABLE_VERSION@
-
-// to upgrade between releases
-brew upgrade scalafmt
-```
+The recommended way to install the scalafmt command line tool is with
+[Coursier](#coursier), itself available via Homebrew.
```
olafurpg, line 326: I think it would be nice to have copy-paste friendly instructions for Homebrew users

````md
The recommended way to install the scalafmt command line tool is with
[Coursier](#coursier), itself available via Homebrew.

```sh
brew install coursier/formulas/coursier
coursier install scalafmt
scalafmt --help
```

If necessary, make sure to follow the Coursier instructions for updating
`$PATH` so that the `scalafmt` binary becomes available in your terminal.
````

kitbellew: done.

PR #1965, opened 2020-05-12
FormatWriter: allow folding scaladoc to one line

### LINKED ISSUE #1403: Oneline short scaladoc
- **Version**: 2.0.0-RC5
- **Integration**: CLI
- **Configuration**:
```
docstrings = ScalaDoc
```

## Steps

Given code like this:
```scala
/**
  * Main entry-point.
  */
class Main
```

```bash
scalafmt
```

## Problem

```scala
/**
  * Main entry-point.
  */
class Main
```

## Expectation

```scala
/** Main entry-point. */
class Main
```

## Workaround

No workaround (outside of manual formatting, which at least is preserved).

## Notes

The Scaladoc usage documentation (https://docs.scala-lang.org/style/scaladoc.html) expressly mentions this case:

> When only a simple, short description is needed, a one-line format can be used:
> 
> ```scala
> /** Does something very simple */
> def simple: Unit = ()
> ```

I would, personally, like to have access to this behaviour also using `docstrings = JavaDoc`.

I'd be happy to attempt to resolve this, given some guidance.

kitbellew:
Fixes #1403.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Docstrings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Docstrings.scala
@@ -4,6 +4,7 @@ import metaconfig._
 import metaconfig.generic.Surface
 
 case class Docstrings(
+    oneline: Docstrings.Oneline = Docstrings.Oneline.keep,
     style: Option[Docstrings.Style] = Some(Docstrings.ScalaDoc)
 ) {
   import Docstrings._
@@ -35,5 +36,14 @@ object Docstrings {
 
   implicit val reader: ConfCodec[Style] =
     ReaderUtil.oneOf[Style](JavaDoc, ScalaDoc)
+
+  sealed abstract class Oneline
+  object Oneline {
+    case object keep extends Oneline
+    case object fold extends Oneline
+    case object unfold extends Oneline
+    implicit val reader: ConfCodec[Oneline] =
+      ReaderUtil.oneOf[Oneline](keep, fold, unfold)
+  }
 
 }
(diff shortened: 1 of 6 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/Docstrings.scala
```diff
@@ -1,13 +1,48 @@
 package org.scalafmt.config
 
-import metaconfig.ConfCodec
+import metaconfig._
+import metaconfig.generic.Surface
 
-sealed abstract class Docstrings
+case class Docstrings(
+    oneline: Option[Docstrings.Oneline] = None,
```
poslegm, line 7: Does `oneline = None` mean «preserve»? May be just add `case object preserve extends Oneline`?

And what is purpose of replacing `Docstrings.preserve` with `style = None`?

kitbellew: There used to be one option, and docstrings = preserve meant doing not modify formatting of docstrings.

Which means there's no formatting style, and any additional options (like oneline) should be ignored because formatting is disabled.

Ideally, since the entire thing is disabled, it should be null, or None, but since we haven't fixed metaconfig and have to put the decoder inside, it can't be null. So I'm setting style = None instead. But I don't want to allow the user to set docstrings.style = preserve either.

With oneline, i can add the third, default option.

poslegm: Ok, I figured out what is problem with style. In `oneline` third option will be good

kitbellew: done

PR #1970, opened 2020-05-14
FormatWriter: implement wrapping of long comments

### LINKED ISSUE #1234: Support paragraph-style line-wrapping of comments
Allow for line-wrapping of comments. For the time being, only support wrapping everything paragraph-style (i.e. without recognizing embedded markup), or nothing.

## Examples
Given this configuration:
```
maxColumn = 80
comments.wrap = true
```
This input:
```scala
// A really long comment that scalafmt should break up but does not because this feature is not implemented yet
object Test {
  // Another long comment demonstrating that wrapped comments will be aligned sensibly
  def foo: Unit = { () }
}

/* Block-style comments should also be wrapped, because some people like them and we want to be inclusive */
object Other {
  /* And block-style comments should also be indented aligned sensibly, because they're just as good as single-line comments */
  def foo: Unit = { () }
}
```
Produces this output:
```scala
// A really long comment that scalafmt should break up but does not because
// this feature is not implemented yet
object Test {
  // Another long comment demonstrating that wrapped comments will be aligned
  // sensibly
  def foo: Unit = { () }
}

/* Block-style comments should also be wrapped, because some people like them
   and we want to be inclusive */
object Other {
  /* And block-style comments should also be indented aligned sensibly, because
     they're just as good as single-line comments */
  def foo: Unit = { () }
}
```

`comments.wrap` should default to false, which would make the output identical to the input, above.

In the future, options like `comments.formatMarkup = "markdown|scaladoc"`, or `comments.block.wrap = true` could be supported

## Open questions
- Should docstring comments also be wrapped?

kitbellew:
Fixes #1234.

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/InputMethod.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/InputMethod.scala
@@ -76,7 +76,7 @@ object InputMethod {
       original: String,
       revised: String
   ): String = {
-    import collection.JavaConverters._
+    import org.scalafmt.CompatCollections.JavaConverters._
     def jList(string: String) =
       java.util.Collections.list(string.linesIterator.asJavaEnumeration)
     val a = jList(original)
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/InputMethod.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/InputMethod.scala
@@ -78,8 +78,7 @@ object InputMethod {
   ): String = {
     import collection.JavaConverters._
     def jList(string: String) =
-      // Predef.augmentString = work around scala/bug#11125 on JDK 11
-      java.util.Collections.list(augmentString(string).lines.asJavaEnumeration)
+      java.util.Collections.list(string.linesIterator.asJavaEnumeration)
     val a = jList(original)
     val b = jList(revised)
     val diff = difflib.DiffUtils.diff(a, b)
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -34,7 +34,7 @@ class FormatWriter(formatOps: FormatOps) {
   import formatOps._
 
   def mkString(state: State): String = {
-    val sb = new StringBuilder()
+    implicit val sb = new StringBuilder()
     val locations = getFormatLocations(state)
 
     locations.iterate.foreach { entry =>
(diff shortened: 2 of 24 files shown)


kitbellew:
@poslegm could you please review again? the previous variant was a little too simple, broke on comments with nested comments or those with a little bit of structure. the new handles a little better.

scalafmt-cli/src/main/scala/org/scalafmt/cli/InputMethod.scala
```diff
@@ -76,10 +76,10 @@ object InputMethod {
       original: String,
       revised: String
   ): String = {
-    import collection.JavaConverters._
+    import org.scalafmt.CompatCollections.JavaConverters._
     def jList(string: String) =
       // Predef.augmentString = work around scala/bug#11125 on JDK 11
```
poslegm, line 81: This comment can be removed if `augmentString` is not necessary for jdk 11 more

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -383,52 +385,183 @@ class FormatWriter(formatOps: FormatOps) {
...
+          style.comments.wrap match {
+            case Comments.Wrap.no => false
+            case Comments.Wrap.trailing => curr.hasBreakAfter
+            case Comments.Wrap.standalone => breakBefore && curr.hasBreakAfter
+          }
+
+        def formatMultilineComment: Unit = {
+          // don't rewrite comments which contain nested comments
+          if (canRewrite && text.lastIndexOf("/*") == 0) {
+            val sectionIter = new Iterator[Iterator[Iterator[String]]] {
```
poslegm, line 464: I don't sure should be type aliases helpful here but just 3-level iterator looks little bit scary

```scala
type WordsIterator = Iterator[String]
type ParagraphIterator = Iterator[WordsIterator]
type SectionIterator = Iterator[ParagraphIterator]
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -945,6 +1083,14 @@ object FormatWriter {
     trailingSpace.matcher(str).replaceAll("")
   }
 
+  private def splitAsIterator(regex: Pattern)(value: String): Iterator[String] =
+    regex.splitAsStream(value).iterator().asScala
+
+  private val slcDelim = Pattern.compile("\\h++")
```
poslegm, line 1089:
```suggestion
  // "slc" means "single-line comment" and "mlc" means "multi-line comment"
  private val slcDelim = Pattern.compile("\\h++")
```

scalafmt-tests/src/test/resources/unit/Comment.stat
```diff
@@ -201,3 +201,150 @@ object literal extends scala.Dynamic { // scalastyle:ignore
...
+  def a = {}
+}
+>>>
+/* A really long comment that scalafmt
+ * should break up but does not because
+ * this feature is not implemented yet */
+// short comment
+object a {
+  // short comment
+  /* Another long comment demonstrating
```
poslegm, line 221: Personally I prefer it to be formatted as 

```
// Another long comment demonstrating
// that wrapped comments will be
// aligned sensibly
```

but this wrap is very useful too

kitbellew: done. i didn't think this would be the case, as to me each `//` line usually represents a separate comment.

PR #1971, opened 2020-05-14
sbt plugin update

poslegm:


docs/installation.md
```diff
@@ -118,7 +118,7 @@ in the `.scalafmt.conf` configuration file and downloaded dynamically.
 
 ```scala
 // In project/plugins.sbt. Note, does not support sbt 0.13, only sbt 1.x.
-addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.3.2") // "2.3.2" is just sbt plugin version
+addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.0") // "2.4.0" is just sbt plugin version
```
kitbellew, line 121: optional: can `mdoc` variable injection be used somehow, both here and in `plugins.sbt`?

https://scalameta.org/mdoc/docs/why.html#variable-injection

poslegm: Yes, but we'll still need to keep this variable up to date manually. @STABLE_VERSION@ comes from git tag, but plugin version managed by other repository

Issue #1976, opened 2020-05-19
beforeContextBoundColon preset deprecation warning


mikail-khan:
As discussed on the Discord channel, the following is producing the preset deprecation warning:

```
Never$: top-level presets deprecated; use 'preset' subsection

spaces {
  beforeContextBoundColon = true
  inImportCurlyBraces = false
  beforeContextBoundColon = false
}
```

According to @poslegm this is a bug related to `beforeContextBoundColon` reading.

poslegm:
Sorry, I deceived you :(

You have duplicated key `beforeContextBoundColon` with different values. 

```
spaces {
  beforeContextBoundColon = Never
  inImportCurlyBraces = false
}

# or

spaces {
  beforeContextBoundColon = Always
  inImportCurlyBraces = false
}
```

works well.

But error message should be definitely more informative in this case.

mikail-khan:
doh! :facepalm: thanks

mikail-khan:
I took out the duplicate config but I still get the warning. Here is the complete config file:

```
version=2.5.2
maxColumn = 120
lineEndings = unix
importSelectors = singleLine

project {
  git = true
}

preset = defaultWithAlign
align.preset = most

align {
  tokens = [ {code = "=>", owner = "Case|Type.Arg.ByName"}, "=", "<-", "->", "%", "%%", "should", "shouldBe", "must" ]
  arrowEnumeratorGenerator = true
  openParenCallSite = false
  openParenDefnSite = false
}

spaces {
  beforeContextBoundColon = true
  inImportCurlyBraces = false
}

continuationIndent {
  callSite = 2
  defnSite = 2
}

newlines {
  penalizeSingleSelectMultiArgList = false
  sometimesBeforeColonInMethodReturnType = true
}

rewrite {
  rules = [RedundantBraces, RedundantParens, AsciiSortImports]
  redundantBraces {
    maxLines = 100
    includeUnitMethods = true
    stringInterpolation = true
  }
}

project.excludeFilters = [
  ".*\\.sc$"
]

```

poslegm:
```
--  beforeContextBoundColon = true
++  beforeContextBoundColon = Never
```

Please do not close the issue so as not to forget warning message clarifying

kitbellew:
@mikail-khan the correct values for `beforeContextBoundColon` are `Always`, `Never` and `IfMultipleBounds`.

[...]

Issue #1523, opened 2019-10-11
Complete configs documentation


poslegm:
I was a witness and participant of scalafmt adoption into a few large codebases and it always caused the developers to study the scalafmt sources for optimal configuration selection. In my opinion software should not force users to read source code (even if users is programmers).

In my sweet dreams [configuration](https://scalameta.org/scalafmt/docs/configuration.html) page in docs contains examples for _all_ available options from `.scalafmt.conf`.

I propose two things:
1. Step-by-step increase documentation coverage for scalafmt settings
2. Don't merge PRs with new features without documentation

Are there any reasons why we shouldn't do this?

tanishiking:
It must be a good idea to increase the documentation coverage. 👍 

If we add a lot of examples, we may want to decide "recommended / popular" configs and "minor" configs so that users don't need to consider too much about which ones should they configure.

olafurpg:
The undocumented configuration options are intentionally not documented since they are usually not well tested or supported. Maybe we should remove the undocumented settings from the source code rather than adding them to the website 😅

kitbellew:
@poslegm anything still missing in documentation?

poslegm:
I think that all notable features are documented now. So, this issue can be closed.

Documentation quality increased a lot over the past year

PR #1977, opened 2020-05-20
Spaces configuration documented

poslegm:
Helps with https://github.com/scalameta/scalafmt/issues/1976 and https://github.com/scalameta/scalafmt/issues/1523

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1762,18 +1762,6 @@ def method[A : Bound]: B
 def method[A: Bound: Bound2]: B
 ```
 
-### `spaces.afterTripleEquals`
-
-```scala mdoc:defaults
-spaces.afterTripleEquals
-```
-
-```scala mdoc:scalafmt
-spaces.afterTripleEquals=true
----
-foo must ===(123)
-```
-
 ### `spaces.inImportCurlyBraces`
 
 ```scala mdoc:defaults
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -1730,6 +1730,126 @@ will apply. It will also use different parameters for test suites.
...
+
+```scala mdoc:scalafmt
+spaces.beforeContextBoundColon=IfMultipleBounds
+---
+def method[A: Bound]: B
+def method[A : Bound]: B
+def method[A: Bound: Bound2]: B
+```
+
+### `spaces.afterTripleEquals`
```
poslegm, line 1765: I have doubts about the rationality of support special formatting for method not even from standard library.

poslegm: @kitbellew @olafurpg what do you think, is it really necessary?

poslegm: > i don't really know how it came about. some colleagues from the spark world know, perhaps?

scalatest and [cats](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/eq.scala#L12) also uses it.

> if we want to deprecate, then i believe the usual algorithm is to mark deprecated and wait 2 years. :)

Yes, it will be good

poslegm: Removed it from the docs until we decided what would happen to it next

PR #1980, opened 2020-05-25
Router: allow folded `extends` if ctor is folded

kitbellew:
As a follow-up to #1979, allow specifying the entire `extends`/`with` chain on one line using new values of an existing `binPack.parentConstructors` parameter.

The two new options allow specifying the chain on one line if fits, either always or only when the primary constructor also fits on one line. For example:

```
class Foo(a: Int, b: Int)
  extends Bar with Baz
```
vs
```
class Foo(
    a: Int,
    b: Int
) extends Bar
    with Baz
```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1261,7 +1261,8 @@ rewrite.redundantBraces.stringInterpolation = true
 s"user id is ${id}"
 ```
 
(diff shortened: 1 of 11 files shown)


docs/configuration.md
```diff
@@ -1261,7 +1261,8 @@ rewrite.redundantBraces.stringInterpolation = true
 s"user id is ${id}"
 ```
 
-`rewrite.redundantBraces.parensForOneLineApply` is `true` by default for `edition` >= 2020-01. See also [newlines.afterCurlyLambda = squash](http://localhost:3000/scalafmt/docs/configuration.html#newlinesaftercurlylambda)
```
poslegm, line 1264: 🤦

PR #1985, opened 2020-05-27
FormatWriter: fix off-by-one bug in comment wrap

kitbellew:
Also, allow formatting text with additional space indentation after the
asterisk.

### DIFF
--- a/scalafmt-tests/src/test/resources/unit/Comment.stat
+++ b/scalafmt-tests/src/test/resources/unit/Comment.stat
@@ -315,10 +315,11 @@ object a {
 }
 >>>
 object a {
-  def a = {} /* And block-style comments
-   * should also be indented aligned
-   * sensibly, because they're just as
-   * good as single-line comments */
+  def a = {} /* And block-style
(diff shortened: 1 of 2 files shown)


scalafmt-tests/src/test/resources/unit/Comment.stat
```diff
@@ -315,10 +315,11 @@ object a {
 }
```
kitbellew, line 315: [github doesn't allow commenting on lines which didn't change]

this file, at the very top, pretends to have 40 columns as the width. however, that's not true, see:

https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala#L140
https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala#L227

this led to the incorrect implementation as i had created a test which is supposed to put exactly 40 characters on the line, and successfully achieved it. except it should have been 39...

PR #1987, opened 2020-05-28
FormatWriter: format multiline docstrings

### LINKED ISSUE #1387: Newline after scaladoc annotation
Recently I had a requirement to move descriptions for scaladoc annotations to a new line. E.g.:

```
/* Some scaladoc for my class
* @tparam someTParam This is a description for my type-param
* @param someParam This is a description for my param
...
*/
```

Becomes:
```
/* Some scaladoc for my class
* @tparam 
* someTParam This is a description for my type-param
* @param someParam 
* This is a description for my param
...
*/
```

I would love to see this as an officially supported option in scalafmt.

### LINKED ISSUE #1887: Scalafmt Doesn't Wrap Lines in Comments
- **Version**: 2.4.2
- **Integration**: IntelliJ
- **Configuration**: The full configuration can be found [here](https://github.com/luna/enso/blob/de666ab9ae4eb0b69925dfc2b5a5b609cefb1ea1/.scalafmt.conf).

## Steps

Given code like this:
```scala
/** This pass implements dataflow analysis for Enso.
  *
  * Dataflow analysis is the processes of determining the dependencies between program expressions.
  *
  * This pass needs to be run after [[AliasAnalysis]], [[DemandAnalysis]], and
  * [[TailCall]]. It also assumes that all members of [[IR.IRKind.Primitive]]
  * have been removed from the IR by the time it runs.
  */

```

When I run scalafmt by using the "format code" intention in IntelliJ.

## Problem
Scalafmt leaves the comment that overflows 80 characters untouched. 

## Expectation

```scala
/** This pass implements dataflow analysis for Enso.
  *
  * Dataflow analysis is the processes of determining the dependencies between
  * program expressions.
  *
  * This pass needs to be run after [[AliasAnalysis]], [[DemandAnalysis]], and
  * [[TailCall]]. It also assumes that all members of [[IR.IRKind.Primitive]]
  * have been removed from the IR by the time it runs.
  */

```

## Workaround
No workaround has been found beyond manually formatting comments.

kitbellew:
Now that scalameta in 4.3.13 supports parsing of docstrings, let's use that and format.

Fixes #1387.
Fixes #1887.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -138,6 +138,44 @@ docstrings.oneline = unfold
 val a = 1
 ```
 
+#### `docstrings.wrap`
+
+Will parse scaladoc comments and reformat them.
+
+> This functionality is generally limited to
+> [standard scaladoc elements](https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html)
+> and might lead to undesirable results in corner cases;
+> for instance, the scaladoc parser doesn't have proper support of embedded HTML.
+
+```scala mdoc:defaults
+docstrings.wrap
+```
+
+```scala mdoc:scalafmt
+docstrings.wrap = yes
+maxColumn = 30
+---
+/**
+ * @param d the Double to square, meaning multiply by itself
+ * @return the result of squaring d
+ *
+ * Thus
+ * - if [[d]] represents a negative value:
+ *  a. the result will be positive
+ *  a. the value will be {{{d * d}}}
+ *  a. it will be the same as for `-d`
+ * - however, if [[d]] is positive
+ *  - the value will still be {{{d * d}}}
+ *    - i.e., the same as {{{(-d) * (-d)}}}
+ *
+ * In other words:
+ * {{{
+ *    res = d * d
+ *        = (-d) * (-d) }}}
+ */
+def pow2(d: Double): Double
+```
+
 ### `assumeStandardLibraryStripMargin`
 
 This parameter simply says the `.stripMargin` method was not redefined
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Docstrings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Docstrings.scala
@@ -6,6 +6,8 @@ import metaconfig._
   * @param oneline
   *        - if fold, try to fold short docstrings into a single line
   *        - if unfold, unfold a single-line docstring into multiple lines
+  * @param wrap
+  *   if yes, allow reformatting/rewrapping the contents of the docstring
   * @param style
   *        - preserve: keep existing formatting
   *        - Asterisk: format intermediate lines with an asterisk below the
@@ -17,6 +19,7 @@ import metaconfig._
   */
 case class Docstrings(
     oneline: Docstrings.Oneline = Docstrings.Oneline.keep,
+    wrap: Docstrings.Wrap = Docstrings.Wrap.no,
     style: Option[Docstrings.Style] = Some(Docstrings.SpaceAsterisk)
 ) {
   import Docstrings._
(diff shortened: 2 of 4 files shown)


kitbellew:
interesting coincidence on 1x87 pattern, that #1987 is addressing #1887 and #1387.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Docstrings.scala
```diff
@@ -17,6 +19,7 @@ import metaconfig._
   */
 case class Docstrings(
     oneline: Docstrings.Oneline = Docstrings.Oneline.keep,
+    wrap: Boolean = false,
```
olafurpg, line 22: My experience is that Boolean config options usually end up becoming enums. I think it’s worth considering making this option an enum instead.

kitbellew: yep... i was thinking the same, as i had just added the `Comments.wrap` parameter as an enum. good catch.

kitbellew:
> * code formatted with backticks (single line and multiline). Even if it’s not official scaladoc markup syntax I’ve seen it used several places and it renders as expected in some cases.
Added a test for single backquotes.

In order to support triple backquotes, we need to modify the parser in scalameta; however, I am not sure how valid or useful that would be, as Intellij doesn't support them, and they don't seem to be supported in any scaladoc documentation (https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html#markup, http://keithpinson.github.io/Scaladoc-HOWTO/api/docSample/syntax.html)

> * quoted strings (single and double), whether they get line wrapped or not we should document the behavior in the tests.

Added both; there's no special handling for now, they are wrapped as any other space-separated words.

> * URLs followed by a white space and period. Just make sure not to remove the space.

Added; no special handling, space separates the period into a separate word.

> * HTML markup, white space inside HTML is significant so we should consider leaving it unchanged. Otherwise we may want to consider writing a full blown XML pretty printer.

Added a test but I doubt this is what you had in mind. To support, we need to extend the scaladoc parser in scalameta; there's `XmlParser` there already but it's not as simple as just using it.

Ideally, if we are to support html, we'd have to parse `<ul><li>` as list blocks/items, `<p>/<br>` as paragraph break, `<pre>` as code, `<a href="x">y</a>` as `[[x y]]`. Also, somehow figure out which markup is embedded in text and which has to be split into a separate paragraph or at least surrounded with newlines.

I'd address this in a subsequent version, as that's highly non-trivial.

docs/configuration.md
```diff
@@ -138,6 +138,44 @@ docstrings.oneline = unfold
...
 ```
 
+#### `docstrings.wrap`
+
+Will parse scaladoc comments and reformat them.
+
+> This functionality is generally limited to
+> [standard scaladoc elements](https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html)
+> and might lead to undesirable results in corner cases;
+> for instance, the scaladoc parser doesn't have proper support of embedded HTML.
```
kitbellew, line 148: @olafurpg would you like to amend this statement to say it will be treated as normal text?

olafurpg: This looks good 👍

kitbellew: perfect, i will merge.

separately, i will add table support to scalameta and then propagate it here. does this sound reasonable: https://www.scala-lang.org/blog/2018/10/04/scaladoc-tables.html

NihalJain:
Hi even after taking latest release i am still seeing issue https://github.com/scalameta/scalafmt/issues/1387? Do I need to something more? 
Ref scalafmt.conf: https://github.com/apache/hbase-connectors/blob/4b9553c400e3307c224b90238dd1b6f57438284a/dev-support/.scalafmt.conf

kitbellew:
> Hi even after taking latest release i am still seeing issue #1387? Do I need to something more? Ref scalafmt.conf: https://github.com/apache/hbase-connectors/blob/4b9553c400e3307c224b90238dd1b6f57438284a/dev-support/.scalafmt.conf

please submit an issue providing all of the information, including a reproducible example.

NihalJain:
> > Hi even after taking latest release i am still seeing issue #1387? Do I need to something more? Ref scalafmt.conf: https://github.com/apache/hbase-connectors/blob/4b9553c400e3307c224b90238dd1b6f57438284a/dev-support/.scalafmt.conf
> 
> please submit an issue providing all of the information, including a reproducible example.

Thanks for replying. Sure will raise one soon.

PR #1990, opened 2020-05-28
FormatWriter bugfix: comment with empty first line

kitbellew:
It wouldn't process the first delimiter correctly and end up adding an extra asterisk.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -550,7 +550,9 @@ class FormatWriter(formatOps: FormatOps) {
           if (canRewrite && text.lastIndexOf("/*") == 0) {
             val sectionIter = new SectIter {
               private val lineIter = {
-                val contents = text.substring(2, text.length - 2).trim
+                val header = mlcHeader.matcher(text)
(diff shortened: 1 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -1141,7 +1141,7 @@ object FormatWriter {
   // "slc" stands for single-line comment
   private val slcDelim = Pattern.compile("\\h++")
   // "mlc" stands for multi-line comment
-  private val mlcLineDelim = Pattern.compile("\\h*+\n\\h*+[*]*+\\h*+")
+  private val mlcLineDelim = Pattern.compile("(?:\\h*+\n|^)\\h*+[*]*+\\h*+")
```
poslegm, line 1144: Strange failure on CI. May be related to https://bugs.openjdk.java.net/browse/JDK-8189343

kitbellew: how do you find those, @poslegm? you're likely right.

i found a different, slightly roundabout way to make it work.

poslegm: > how do you find those, @poslegm?

just googling java 8 vs java 9 regex difference 🤷

PR #2006, opened 2020-06-06
Scaladoc: support table formatting in FormatWriter

kitbellew:
Also update `scalameta` dependency to 4.3.14 to make this functionality available.


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -688,12 +689,48 @@ class FormatWriter(formatOps: FormatOps) {
...
           item.nested.foreach(formatListBlock(itemIndent))
         }
 
+        private def formatTable(table: Scaladoc.Table): Unit = {
+          val rows = table.row.view :+ table.header
+          val align = table.align
+          val maxCols = rows.map(_.col.length).max
+          val colsRange = 0 until maxCols
+          val maxLengths = colsRange.map { x =>
+            rows.filter(_.col.length > x).map(_.col(x).length).max
```
poslegm, line 704:
```suggestion
            rows.collect { case r if r.col.length > x => r.col(x).length }.max
```

One list traverse instead of two

PR #2021, opened 2020-06-15
Router: allow relative formatting of spliced XML

### LINKED ISSUE #1882: Incorrect indentation of code blocks within XML literals
I'm using [Binding.scala](https://github.com/ThoughtWorksInc/Binding.scala) framework and am having trouble formatting code blocks inside nested HTML literals. 

If I write deeply nested HTML and then include a code block, Scalafmt will not keep the correct indentation of the code block.

- **Version**: 2.5.0-RC1
- **Configuration**:
```
version=2.5.0-RC1
align = none
maxColumn = 120
```

The following examples are simplified and exclude annotations used in real [Binding.scala](https://github.com/ThoughtWorksInc/Binding.scala) code.

All examples are run using this command: `scalafmt --config .scalafmt.conf --stdout Example.scala`

### Example1
**Input:**
```scala
object Example1 {
  def apply() = {
    <div>
    <div>
    <div>
    </div>
    </div>
    </div>
  }
}
```

**Output:**

Example 1 output matches input.

**Notes:**

This examples contains 3 nested HTML elements and illustrates that there is no formatting of HTML by Scalafmt. It possibly could help to format by indenting inner elements based on element nesting level, however this isn't a critical issue, possibly more of a nice to have. It may be necessary for Scalafmt to be aware of element nesting level to correctly indent deeply nested code blocks.

**Expected output if Scalafmt were to format based on nesting level**:
```scala
object Example1 {
  def apply() = {
    <div>
      <div>
        <div>
        </div>
      </div>
    </div>
  }
}
```

### Example2
**Input:**

```scala
object Example2 {
  def apply() = {
    <div>
      <div>
        <div>
          <h1>Heading</h1>
          {
            val x = "stuff"
            <div>{x}</div>
          }
        </div>
      </div>
    </div>
  }
}
```

**Output:**

```scala
object Example2 {
  def apply() = {
    <div>
      <div>
        <div>
          <h1>Heading</h1>
          {
      val x = "stuff"
      <div>{x}</div>
    }
        </div>
      </div>
    </div>
  }
}
```

**Notes:**

This illustrates a nested code block. Scalafmt shifts some of the code back to a smaller indentation. I would expect the original indentation to be kept. 

### Example 3
**Input:**

```scala
object Example3 {
  def apply() = {
    <div>
      <div>
        <div>
          <h1>Heading</h1>
    {
      val x = "stuff"
      <div>{x}</div>
    }
        </div>
      </div>
    </div>
  }
}
```

**Output:**

Example 3 output matches input.

**Notes:**

This example contains a code block which is nested too far left. It illustrates that Scalafmt is happy with this nesting level, as it is at the correct nesting level for a block contained inside the `apply` method. I would expect the code block to be indented further to the right to respect the nesting level of the parent HTML element (to match the Example 2 input).

### Example 4
**Input:**

```scala
object Example4 {
  def apply() = {
    <div>
    {
      <div>
      {
        <div>
          <h1>Heading</h1>
        {
          val x = "stuff"
          <div>{x}</div>
        }
        </div>
      }
      </div>
    }
    </div>
  }
}
```

**Output:**

Example 4 output matches input.

**Notes:**

This illustrates a partial workaround to get the inner most code block to indent correctly, however it results in a lot of extra braces and requires less than ideal indentation of the code block. You have to shift the code block left to make it work. Ideally the code block would be shifted right, aligned with the `<h1>Heading</h1>` element above it (as in Example 2 input).

kitbellew:
It involves finding the indentation used within the xml literal and then applying it to the scala code within. We will use the ability to ignore previous indents as this indentation is not relative but absolute.

Fixes #1882.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/XmlLiterals.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/XmlLiterals.scala
@@ -1,0 +1,15 @@
+package org.scalafmt.config
+
+import metaconfig._
+
+case class XmlLiterals(
+    assumeFormatted: Boolean = false
+) {
+  implicit val decoder: ConfDecoder[XmlLiterals] =
+    generic.deriveDecoder(this).noTypos
+}
+
+object XmlLiterals {
+  implicit val surface: generic.Surface[XmlLiterals] = generic.deriveSurface
+  implicit lazy val encoder: ConfEncoder[XmlLiterals] = generic.deriveEncoder
+}
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
@@ -254,4 +254,14 @@ object TokenOps {
         .filter(_.head.tokens.head.start <= math.max(token.end, other.end))
     }
 
+  def getXmlLastLineIndent(tok: Xml.Part): Option[Int] = {
+    val part = tok.value
+    val afterLastNL = part.lastIndexOf('\n') + 1
+    if (afterLastNL <= 0) None
+    else {
+      val nonWs = part.indexWhere(!_.isWhitespace, afterLastNL)
+      Some((if (nonWs < 0) part.length else nonWs) - afterLastNL)
+    }
+  }
+
 }
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2008,6 +2008,48 @@ literals.float=Lower
 10e-1f
 ```
 
+## XML
+
+Controls formatting of Scala embedded within XML.
+
+### `xmlLiterals.assumeFormatted`
+
+> Since v2.6.0.
+
+If set, formats embedded Scala relative to containing XML, making the assumption
+that XML itself is properly formatted. Otherwise, formatting is relative to the
+outer Scala code which contains the XML literals.
+
+```scala mdoc:defaults
+xmlLiterals.assumeFormatted
+```
+
+```scala mdoc:scalafmt
+maxColumn = 40
+xmlLiterals.assumeFormatted = true
+---
+object Example2 {
+  def apply() = {
+      <foo>
+        <bar>{ (1 + 2 + 3).toString("some long format") }</bar>
+      </foo>
+  }
+}
+```
+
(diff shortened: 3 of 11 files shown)


kitbellew:
@dylanarnold please verify this is what you were expecting.

kitbellew:
@poslegm should we do it under a parameter, or without one?

DylanArnold:
@kitbellew This is exactly what I was expecting. I just tried it out and it works wonderfully. Huge thanks for your hard work on this!! :tada:

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
```diff
@@ -254,4 +254,14 @@ object TokenOps {
         .filter(_.head.tokens.head.start <= math.max(token.end, other.end))
     }
 
+  def getXmlLastLineIndent(tok: Xml.Part): Option[Int] = {
+    val part = tok.value
+    val afterLastNL = part.lastIndexOf('\n') + 1
+    if (afterLastNL <= 0) None
+    else {
+      val nonWs = part.indexWhere(!Character.isWhitespace(_), afterLastNL)
```
poslegm, line 262: nit

```suggestion
      val nonWs = part.indexWhere(!_.isWhitespace, afterLastNL)
```

scalafmt-core/shared/src/main/scala/org/scalafmt/config/XmlLiterals.scala
```diff
@@ -0,0 +1,15 @@
+package org.scalafmt.config
+
+import metaconfig._
+
+case class XmlLiterals(
+    assumeFormatted: Boolean = false
```
poslegm, line 6: may be true by default?

kitbellew: if we keep the flag, i'd set it to false, just like we did with `docstrings.wrap` and `comments.wrap`, they are logically similar, i think.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
kitbellew: thank you, i didn't know that!

PR #2065, opened 2020-07-03
ScalafmtDynamic: throw once if can't resolve config

kitbellew:
Intellij invokes the createSession method directly and doesn't expect the null value we return there, so throw ConfigError instead.

Similarly, in the Runner, change the implementation of MyInstanceSession to verify the config first, so that the config parsing exception is not thrown for every file we are trying to format.

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtDynamicRunner.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtDynamicRunner.scala
@@ -8,6 +8,7 @@ import java.util.function.UnaryOperator
 
 import org.scalafmt.CompatCollections.ParConverters._
 import org.scalafmt.Error.{MisformattedFile, NoMatchingFiles}
+import org.scalafmt.dynamic.ScalafmtDynamicError
 import org.scalafmt.interfaces.Scalafmt
 import org.scalafmt.interfaces.ScalafmtSession
 import org.scalafmt.interfaces.ScalafmtSessionFactory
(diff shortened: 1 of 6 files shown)


scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtDynamicRunner.scala
```diff
@@ -27,12 +31,15 @@ object ScalafmtDynamicRunner extends ScalafmtRunner {
       .withReporter(reporter)
       .withRespectProjectFilters(false)
 
-    val session = scalafmtInstance match {
+    val session = Try(scalafmtInstance match {
```
poslegm, line 34: Why not just try/catch here?

scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtDynamicRunner.scala
```diff
@@ -79,6 +86,8 @@ object ScalafmtDynamicRunner extends ScalafmtRunner {
 
   private final class MyInstanceSession(opts: CliOptions, instance: Scalafmt)
       extends ScalafmtSession {
+    // check config first
+    locally { instance.format(opts.configPath, opts.configPath, "") }
```
poslegm, line 90: I can't understand how `locally` helps here

kitbellew: in that case, i will remove. perhaps i am still getting used to being able to invoke methods in scala which are not part of another method.

scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtDynamicRunner.scala
kitbellew: hmm... good idea! :)

PR #2067, opened 2020-07-04
Router: merge classic select chain rule with other

### LINKED ISSUE #1334: Wired alignment when chain contains comment / no parenthesis call / with parenthesis call

kitbellew:
This will apply the same clean indentation logic as for `newlines.source != classic`.

Requires adding support for multiple activation tags within a split; one was used to apply indent to the first break only, and the new one will be used to reduce cost of a split down the chain after the first break and remove the newline penalization policy consistent with the previous rule; this filtering capability was introduced in #2060.

Fixes #1334. Supersedes #1557.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
@@ -38,8 +38,8 @@ case class OptimalToken(token: Token, killOnFail: Boolean = false) {
 case class Split(
     modExt: ModExt,
     cost: Int,
-    tag: SplitTag = SplitTag.Active,
-    activeTag: SplitTag = SplitTag.Active,
+    neededTags: Set[SplitTag] = Set.empty,
+    activeTags: Set[SplitTag] = Set.empty,
     policy: Policy = NoPolicy,
     optimalAt: Option[OptimalToken] = None
 )(implicit val line: sourcecode.Line) {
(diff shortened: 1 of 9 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Split.scala
```diff
@@ -38,8 +38,8 @@ case class OptimalToken(token: Token, killOnFail: Boolean = false) {
 case class Split(
     modExt: ModExt,
     cost: Int,
-    tag: SplitTag = SplitTag.Active,
-    activeTag: SplitTag = SplitTag.Active,
+    neededTags: Long = SplitTag.Active.mask,
```
poslegm, line 41: Not sure how much better the bit mask is than just `Set[SplitTag]` here and `contains` instead of `(neededTags & splitTag.mask) != 0`

kitbellew: done. i thought i had seen a performance difference earlier but couldn't prove it now.

PR #2077, opened 2020-07-07
Router: don't format #! lines in sbt files

### LINKED ISSUE #1716: Spreading of shebang in *.sc files
- **Version**: 2.3.2
- **Integration**: Metals
- **Configuration**:
```
version = "2.3.2"
```

## Steps

Given code that uses a shebang like the following:
```scala
#!/usr/bin/env amm
```

When I run scalafmt the result is as follows:

```git
#!/ usr / bin / env amm
```

## Problem and expectation

I would expect this line to be unchanged, as this change prevents the script from being executed correctly.

##### search terms
scripts, amm, shebang

kitbellew:
Fixes #1716.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -73,9 +73,21 @@ class Router(formatOps: FormatOps) {
     val newlines = formatToken.newlinesBetween
 
     formatToken match {
-      case FormatToken(_: T.BOF, _, _) =>
+      case FormatToken(_: T.BOF, right, _) =>
+        val policy = right match {
+          case T.Ident(name) // shebang in .sc files
+              if filename.endsWith(".sc") && name.startsWith("#!") =>
+            val nl = findFirst(next(formatToken), Int.MaxValue)(_.hasBreak)
+            nl.fold[Policy](Policy.NoPolicy) { ft =>
+              Policy(ft.left) {
+                case Decision(t, _) if t.noBreak =>
+                  Seq(Split(Space(t.between.nonEmpty), 0))
+              }
+            }
+          case _ => Policy.NoPolicy
+        }
         Seq(
-          Split(NoSplit, 0)
+          Split(NoSplit, 0).withPolicy(policy)
         )
       case FormatToken(_, _: T.EOF, _) =>
         Seq(
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/InputMethod.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/InputMethod.scala
@@ -10,8 +10,6 @@ import org.scalafmt.util.AbsoluteFile
 import org.scalafmt.util.FileOps
 
 sealed abstract class InputMethod {
-  def isSbt: Boolean = filename.endsWith(".sbt")
-  def isSc: Boolean = filename.endsWith(".sc")
   def readInput(options: CliOptions): String
   def filename: String
   def write(formatted: String, original: String, options: CliOptions): ExitCode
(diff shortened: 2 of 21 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -73,9 +73,21 @@ class Router(formatOps: FormatOps) {
     val newlines = formatToken.newlinesBetween
 
     formatToken match {
-      case FormatToken(_: T.BOF, _, _) =>
+      case FormatToken(_: T.BOF, right, _) =>
+        val policy = right match {
+          case T.Ident(name) // shebang in .sbt/.sc files
+              if name.startsWith("#!") &&
+                style.runner.dialect.allowToplevelTerms =>
```
poslegm, line 80: As far as I understand, it works for `runner.dialect = sbt`.

But shebangs much more commonly used in [ammonite](https://ammonite.io/#ScalaScripts) scripts. It may be makes sense to create alias [here](https://github.com/scalameta/scalafmt/blob/master/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala#L98)

```scala
val ScalaScripts = Sbt1
```

kitbellew: the dialect for sbt is not set explicitly, the dynamic runner does it automatically: https://github.com/scalameta/scalafmt/blob/a7bb6920a8ba486cfbff74d056037dfeb36c7e33/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamic.scala#L199

problem is, as the code above shows (and that's just what we fixed in #2075 today), the dialect used for `.sc` files is not a given instance, it's based on the main dialect configured for scala, with `.forSbt` invoked (https://github.com/kitbellew/scalafmt/blob/8be8f7beb3194f2224b9580ec55e2bab909ba8e4/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala#L30)

so, in this method, i don't know the name of the file (to check for `.sc`) and i can't compare dialect to `sbt` or `sbt1` or `sbt0137`. so i check for `allowToplevelTerms` flag as the only thing remaining.

kitbellew: ok, i figured out a way to pass the name of the file to this code. in order to be able to test, had to modify HasTests to take a filename to use and had to invoke `.forSbt` inside the formatter (hence changes to core and dynamic runners to stop doing that). PTAL.

philwalk:
It turns out that `.sc` extension is not reserved for `ammonite` scripts, but is commonly used (in my experience) for any scala script.  Some of my scripts have no extension, which is handy to shadow other system tools.

Any script that is 100% scala following the `hashbang`  will be supported, I hope.

It would be nice if the following very generic and particularly useful hashbang line were ignored:
```sh
<input>:1: error: [dialect scala3] ; expected but @ found
#!/usr/bin/env -S scala @classpathAtfile
```
Thanks for a fabulous tool!

kitbellew:
> It turns out that `.sc` extension is not reserved for `ammonite` scripts, but is commonly used (in my experience) for any scala script. Some of my scripts have no extension, which is handy to shadow other system tools.
> 
> Any script that is 100% scala following the `hashbang` will be supported, I hope.
> 
> It would be nice if the following very generic and particularly useful hashbang line were ignored:
> 
> ```shell
> <input>:1: error: [dialect scala3] ; expected but @ found
> #!/usr/bin/env -S scala @classpathAtfile
> ```
> 
> Thanks for a fabulous tool!

is this a question or request?

philwalk:
> is this a question or request?
I guess both.

Am I right to assume that any shebang lines should be ignored?
If so, then there may be a bug, since parsing of the line results in the `@` being rejected.'

I'm willing to file a bug, if that's what it is.

PR #2083, opened 2020-07-08
Policy: introduce more specific expire strategies

kitbellew:
Currently, policy expiration is applied narrowly (i.e., unlike indents which can expire _before_ or _after_ the expiration token, all policies expire the same way) and incorrectly (they expire not after the specified token, but after the one which follows).

So, add three expiration strategies (before, on, and after) and remove the current after-the-next buggy expiration.

There is **exactly one** diff in `scala-repos` (and it looks *correct*, judging by formatting of similar code before and after the modified block) and one diff in the unit tests (also, looks right), so the change also appears to fix some rare bugs.

From `scala-repos`:
```
   private val _right = new function.Function2[Any, Any, Any]
     with ((Any, Any) ⇒ Any) { def apply(l: Any, r: Any) = r }
-  private val _both =
-    new function.Function2[Any, Any, Any] with ((Any, Any) ⇒ Any) {
-      def apply(l: Any, r: Any) = new akka.japi.Pair(l, r)
-    }
+  private val _both = new function.Function2[Any, Any, Any]
+    with ((Any, Any) ⇒ Any) {
+    def apply(l: Any, r: Any) = new akka.japi.Pair(l, r)
+  }
   private val _none = new function.Function2[Any, Any, NotUsed]
     with ((Any, Any) ⇒ NotUsed) { def apply(l: Any, r: Any) = NotUsed }
```

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Policy.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Policy.scala
@@ -56,21 +56,23 @@ object Policy {
 
   def apply(
       expire: Token,
+      endPolicy: End = End.After,
       noDequeue: Boolean = false
   )(f: Pf)(implicit line: sourcecode.Line): Policy =
-    new ClauseImpl(f, expire.end, noDequeue)
+    new ClauseImpl(f, expire.end, endPolicy, noDequeue)
 
   abstract class Clause(implicit val line: sourcecode.Line) extends Policy {
     val f: Policy.Pf
     val endPos: Int
+    val endPolicy: End
 
     override def toString = {
       val noDeqPrefix = if (noDequeue) "!" else ""
-      s"${line.value}<$endPos${noDeqPrefix}d"
+      s"${line.value}$endPolicy$endPos${noDeqPrefix}d"
     }
 
     override def unexpired(ft: FormatToken): Policy =
-      if (ft.left.end <= endPos) this else NoPolicy
+      if (endPolicy.notExpiredBy(ft, endPos)) this else NoPolicy
 
     override def filter(pred: Clause => Boolean): Policy =
       if (pred(this)) this else NoPolicy
@@ -81,6 +83,7 @@ object Policy {
   private class ClauseImpl(
       val f: Policy.Pf,
       val endPos: Int,
+      val endPolicy: End,
       val noDequeue: Boolean
   )(implicit line: sourcecode.Line)
       extends Clause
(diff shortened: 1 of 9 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Policy.scala
```diff
@@ -99,13 +115,67 @@ object Policy {
...
+      Proxy.apply(policy.filter(pred))(factory)
+
+    override def unexpired(ft: FormatToken): Policy =
+      Proxy.apply(policy.unexpired(ft))(factory)
+
+    override def noDequeue: Boolean = policy.noDequeue
+
+  }
+
+  sealed abstract class End {
```
poslegm, line 160: nit

```suggestion
  sealed trait End {
```

scalafmt-core/shared/src/main/scala/org/scalafmt/util/PolicyOps.scala
```diff
@@ -8,22 +8,20 @@ import org.scalafmt.internal.Policy
 
 object PolicyOps {
 
-  class PenalizeAllNewlines private[PolicyOps] (f: Policy.Pf, endPos: Int)(
-      implicit line: sourcecode.Line
-  ) extends Policy.Clause(f, endPos)
-
-  def penalizeAllNewlines(
+  case class penalizeAllNewlines(
```
poslegm, line 11: nit
```suggestion
  case class PenalizeAllNewlines(
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Policy.scala
kitbellew: why trait? i was under impression that abstract classes are a little more efficient than traits in calling methods.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/PolicyOps.scala
kitbellew: will do. was trying to avoid renaming about 15 invocations in router, as they were all using the method.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Policy.scala
poslegm: Wow, i don't know nothing about it. Can you please share an article or benchmarks about it?

I prefer trait here because usually the only reason to use abstract class in sealed hierarchies is passing parameters to constructor. And here there is no args

kitbellew: https://www.scala-lang.org/blog/2016/07/08/trait-method-performance.html
https://shipilev.net/blog/2015/black-magic-method-dispatch/

will change to trait for now.

PR #2088, opened 2020-07-11
Don't fail test workflow on release

poslegm:
Tests need all tags to be fetched but latest tag already pushed to master on release. For this reason git fetch fails on workflows started with git push --tags

Other jobs works correct already.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: olafurpg/setup-scala@v7
         with:
           java-version: ${{ matrix.java }}
-      - run: git fetch --tags
+      - run: git fetch --tags -f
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - uses: olafurpg/setup-scala@v7
         with:
           java-version: ${{ matrix.java }}
-      - run: git fetch --tags
```
kitbellew, line 22: does it always complain, or only when the tag points to a different sha1?

poslegm: I couldn't find the exact cause. It works well on fork's CI, but fails in the main repo. Perhaps it related to merged commits from forks

PR #2101, opened 2020-07-19
Policy: combine End with position, use with Proxy

kitbellew:


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Policy.scala
```diff
@@ -151,31 +168,42 @@ object Policy {
...
 
     override def unexpired(ft: FormatToken): Policy =
-      Proxy(policy.unexpired(ft))(factory)
+      if (!end.forall(_.notExpiredBy(ft))) NoPolicy
+      else Proxy(policy.unexpired(ft), end)(factory)
 
     override def noDequeue: Boolean = policy.noDequeue
 
-    override def toString: String = s"*($policy)"
+    override def toString: String = s"*($policy)${end.fold("")(_.toString)}"
```
poslegm, line 176:
```suggestion
    override def toString: String = s"*($policy)${end.mkString("")}"
```

PR #2105, opened 2020-07-20
ScalafmtConfig: validate once in the decoder only

### LINKED ISSUE #2098: Error formatting file with scalafmt 2.6.3
- **Version**: 2.6.3
- **Integration**: CLI
- **Configuration**:
```
version=2.6.3

style = defaultWithAlign

maxColumn = 120
project.git = true

lineEndings=preserve
```

## Problem

After upgrading to version 2.6.3 when I run scalafmtAll or scalafmt I get the following error
```scala
[error] can't use: [
[error] 	binPack.unsafeCallSite || binPack.unsafeDefnSite (not implemented)
[error] ]: XYZ.scala
```

## Notes

Seems that this was introduce within the commit 
https://github.com/scalameta/scalafmt/commit/91ae24f2d1b81f19dc3b3b16012b01469553e05d
However `binPack.unsafeCallSite` and `binPack.unsafeDefnSite` are false by default, so as far as I understand this error should not happen. For the record, it even fails when I set them as false in the scalafmt config.

kitbellew:
There are multiple places where a configuration instance is created, and some of them are not really appropriate for validation, such as `StyleMap` where custom configuration copies are created based on comments in the code, or to support `binPack.literalArgumentLists`.

Therefore, let's just validate the initial configuration in the decoder and not the custom adjustments.

Fixes #2098.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/ValidationOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/ValidationOps.scala
@@ -3,19 +3,25 @@ package org.scalafmt.util
 import scala.collection.mutable
 
 object ValidationOps {
-  def assertNonNegative(ns: sourcecode.Text[Int]*): Unit = {
+
+  def addIfNegative(
+      ns: sourcecode.Text[Int]*
+  )(implicit errors: mutable.Buffer[String]): Unit =
     ns.foreach { n =>
       if (n.value < 0)
-        throw new IllegalArgumentException(
-          s"${n.source} must be non-negative, was ${n.value}"
-        )
+        errors += s"${n.source} must be non-negative, was ${n.value}"
     }
-  }
 
   def addIf(
       what: sourcecode.Text[Boolean],
       why: => String = ""
   )(implicit errors: mutable.Buffer[String]): Unit =
-    if (what.value) errors += what.source + why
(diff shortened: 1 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/util/ValidationOps.scala
```diff
@@ -3,19 +3,25 @@ package org.scalafmt.util
...
+        errors += s"${n.source} must be non-negative, was ${n.value}"
     }
-  }
 
   def addIf(
       what: sourcecode.Text[Boolean],
       why: => String = ""
   )(implicit errors: mutable.Buffer[String]): Unit =
-    if (what.value) errors += what.source + why
+    addIfDirect(what.value, what.source + why)
```
poslegm, line 19: Is it a hack around

> sourcecode.Text[T]: when you want to take a value of type T, but also want to get the "source text" of that particular value. Note that if you have multiple statements in a {} block, sourcecode.Text will only capture the source code for the last expression that gets returned. This implicit is slightly experimental; be sure to report any bugs you find!

?

Only last block capture is source of #2098 problem.

poslegm: I see direct usage at 386, sorry

PR #2109, opened 2020-07-21
Edition: remove this feature completely

kitbellew:


blast-hardcheese:
> No description provided.

kitbellew:
> > No description provided.

@blast-hardcheese is there a missing question?

blast-hardcheese:
Are there migration notes somewhere?

I think I understand the motivation, as maintaining this level of complexity forever seems like a thankless job, but deleting `edition = 2019-09` ended up leaving me with ten thousand lines of formatting changes.

kitbellew:
> Are there migration notes somewhere?
> 
> I think I understand the motivation, as maintaining this level of complexity forever seems like a thankless job, but deleting `edition = 2019-09` ended up leaving me with ten thousand lines of formatting changes.

Of course. In addition to maintaining the complexity, the realization was that if editions were implemented correctly (i.e, no formatting changes would be applied with an upgraded version of `scalafmt`), the same result can be accomplished by continuing to use the older version anyway.

blast-hardcheese:
that's also a very fair point. Hopefully this PR can serve as documentation of the decision, I was more surprised than anything to click through the release notes from scala-steward and just find "Yeah, we deleted it".

The rationale makes sense though. Thanks.

PR #2115, opened 2020-07-24
Dedicated setting for redundant braces in if-else (#2038)

### LINKED ISSUE #2038: Add option to prevent rewrite.redundantBraces.generalExpressions from affecting statements
- **Version**: 2.6.1
- **Integration**: IntelliJ
- **Configuration**:
```
version = "2.6.1"
preset = default
```

## Steps

Given code like this, I would like an option that preserves the curly braces for control structures that will always produce unit.
```scala
if (foo) {
  doSideEffect()
}

for(i <- 0 to 5) {
  println(i)
}

while (cond) {
  doThing()
}
```

poslegm:
Fixes #2038

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/RedundantBracesSettings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/RedundantBracesSettings.scala
@@ -8,7 +8,8 @@ case class RedundantBracesSettings(
     maxLines: Int = 100,
     stringInterpolation: Boolean = false,
     parensForOneLineApply: Option[Boolean] = None,
-    generalExpressions: Boolean = true
+    generalExpressions: Boolean = true,
+    ifElseExpressions: Boolean = false
 ) {
   val reader: ConfDecoder[RedundantBracesSettings] =
     generic.deriveDecoder(this).noTypos
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -239,6 +239,9 @@ class RedundantBraces(implicit ctx: RewriteCtx) extends RewriteSession {
       case p: Term.Function if isFunctionWithBraces(p) =>
         okToRemoveAroundFunctionBody(b, true)
 
+      case _: Term.If =>
+        settings.ifElseExpressions && shouldRemoveSingleStatBlock(b)
+
       case _ =>
         settings.generalExpressions && shouldRemoveSingleStatBlock(b)
     }
--- a/scalafmt-tests/src/test/resources/rewrite/RedundantBraces-if.stat
+++ b/scalafmt-tests/src/test/resources/rewrite/RedundantBraces-if.stat
@@ -251,6 +251,7 @@ object a {
 }
 <<< don't rewrite on disabled flag
 rewrite.redundantBraces.ifElseExpressions = false
+rewrite.redundantBraces.generalExpressions = true
 ===
 object a {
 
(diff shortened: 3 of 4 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/RedundantBracesSettings.scala
```diff
@@ -8,7 +8,8 @@ case class RedundantBracesSettings(
     maxLines: Int = 100,
     stringInterpolation: Boolean = false,
     parensForOneLineApply: Option[Boolean] = None,
-    generalExpressions: Boolean = true
+    generalExpressions: Boolean = true,
+    ifElseExpressions: Boolean = false
```
poslegm, line 12: Not sure about default bu removing braces around if statements look rare desire

kitbellew: i think the default is fine.

however, i think one of the requests was to be able to rewrite `if` statements consistently; that is, either rewrite both `if` and `else`, or keep both unchanged (that might mean the flag is not just boolean). i don't know how difficult it would be to implement this logic.

also, in #2038, they want to rewrite assignments but not unit expressions... not suggesting to add it here but simply asking whether it will be simple to do that with the current approach.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -239,6 +239,9 @@ class RedundantBraces(implicit ctx: RewriteCtx) extends RewriteSession {
       case p: Term.Function if isFunctionWithBraces(p) =>
         okToRemoveAroundFunctionBody(b, true)
 
+      case _: Term.If =>
+        settings.ifElseExpressions && shouldRemoveSingleStatBlock(b)
```
kitbellew, line 243: should we apply the same exactly logic to try-finally, or do-while, or for-yield? should we have three separate flags, or one flag (`ctrlExpressions`, similar to `danglingParentheses.ctrlSite`).

or even one super-generic list flag like align tokens which allows configuring matching types?:
```
rewriteTrees {
  include = [Term.If, Case]
  exclude = ...
}
```

scalafmt-tests/src/test/resources/rewrite/RedundantBraces-if.stat
```diff
@@ -249,3 +249,41 @@ object a {
   } else if (d)
     println("d!")
 }
+<<< don't rewrite on disabled flag
+rewrite.redundantBraces.ifElseExpressions = false
```
kitbellew, line 253: should we add `generalExpressions = true` here as well? so that the test verifies that if-else is controlled via the new flag and not the old one...

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
poslegm: Up to this point, I've only seen questions about `if-else` statements. Other expressions formatting satisfies users. So I propose to limit us to just `if-else` flag until we get another feature-requests.

scalafmt-tests/src/test/resources/rewrite/RedundantBraces-if.stat
poslegm: good idea

scalafmt-core/shared/src/main/scala/org/scalafmt/config/RedundantBracesSettings.scala
poslegm: I can't estimate the complexity of consistent and assignment formatting too. At first glance most users will be satisfied with this simple flag

PR #2116, opened 2020-07-24
Install script compatibility with auto-uploaded artifacts

poslegm:
Github action for auto-uploading artifacts to releases can't zip them. So native images from v2.7.0 will be downloaded as is.

Another way is to zip them on CI but it requires not obvious to me yaml machinery

### DIFF
--- a/bin/install-scalafmt-native.sh
+++ b/bin/install-scalafmt-native.sh
@@ -1,22 +1,86 @@
 #!/usr/bin/env bash
 set -eux
 
-VERSION="v${1:-2.3.0-RC2}"
+RAW_VERSION=${1:-2.3.0-RC2}
+VERSION="v$RAW_VERSION"
 INSTALL_LOCATION=${2:-/usr/local/bin/scalafmt-native}
 
 CWD=$(pwd)
 
+# from https://gist.github.com/Ariel-Rodriguez/9e3c2163f4644d7a389759b224bfe7f3
+semver_compare() {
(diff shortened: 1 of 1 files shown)


bin/install-scalafmt-native.sh
```diff
@@ -1,22 +1,86 @@
 #!/usr/bin/env bash
 set -eux
 
-VERSION="v${1:-2.3.0-RC2}"
+RAW_VERSION=${1:-2.3.0-RC2}
```
kitbellew, line 4: opt: time to update the default version?

bin/install-scalafmt-native.sh
```diff
@@ -1,22 +1,86 @@
...
+  do
+    cursor=$((cursor+1))
+  done
+  [ "${arr_version_a[$cursor]}" -gt "${arr_version_b[$cursor]}" ] && echo 1 || echo -1
+}
+
 NAME=scalafmt-linux
 if [ "$(uname)" == "Darwin" ]; then
   NAME=scalafmt-macos
+elif [ "$(semver_compare 2.7.0 "$RAW_VERSION")" -le 0 ]; then
```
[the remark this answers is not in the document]
poslegm: Good catch 🤔 
It works, but now I don't understand how exactly. Will check it twice.

bin/install-scalafmt-native.sh
poslegm: Yes, it makes sense

PR #2118, opened 2020-07-24
Fix a problem where parens are removed after a yield

### LINKED ISSUE #2117: RedundantParens removes parens surrounding in the context of a for yield
I've added a test after running scalafmt into some of our companys' code bases.

## Problem & expectation

The test shows the problem and also the expectation. The result of the formatting does not compile because of the missing space.

```bash
[info] FormatTests:
[info] - rewrite/RedundantParens.stat:274: yield without a space               | *** FAILED ***
[info]   ===========
[info]   => Obtained
[info]   ===========
[info]   object a {
[info]     val x = for {
[info]       a <- b
[info]     } yielda
[info]   }
[info]
[info]
[info]   =======
[info]   => Diff
[info]   =======
[info]        a <- b
[info]   -  } yield a
[info]   +  } yielda
[info]    } (RedundantParens.stat:274)
```

I'm working on a fix right now but I'm not sure what the fix should be. Should we perform the refactoring (remove the parens *but* add a space after the yield) or should we consider this an non-redundant parens set?

Personally, I'd like to see the refactoring happen so that's the solution I'm working on, but if you want something else, let me know.

Thanks, I'll keep the issue up to date.

daddykotex:
it's a problem when there is no space between the yield and the
parens because it leaves the code in a broken state (won't compile)

Fixes #2117

### DIFF
--- a/scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat
+++ b/scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat
@@ -271,3 +271,15 @@ object a {
 object a {
   print("") shouldBe (())
 }
+<<< yield without a space
+object a {
+  val x = for {
+    a <- b
+  } yield(a)
+}
+>>>
+object a {
+  val x = for {
+    a <- b
+  } yield a
+}
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantParens.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantParens.scala
@@ -99,6 +99,15 @@ class RedundantParens(implicit ctx: RewriteCtx) extends RewriteSession {
         }
         ctx.removeLFToAvoidEmptyLine(toks(beg + offBeg), toks(beg + offEnd))
         ctx.removeLFToAvoidEmptyLine(toks(end - offEnd), toks(end - offBeg))
+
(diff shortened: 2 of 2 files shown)


scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat
```diff
@@ -271,3 +271,15 @@ object a {
 object a {
   print("") shouldBe (())
 }
+<<< yield without a space
+object a {
+  val x = for {
+    a <- b
+  } yield(a)
```
kitbellew, line 278: is this specific to `yield`? what about:

```
  a match {
    case a if(a) => a
  }
```

daddykotex: I'm testing it as we speak, I'll open another MR if it also fails there

daddykotex: Ahh I just saw you PR, great. I was not sure if extra spaces would be removed later on

kitbellew: i already fixed it, see linked pr.

kitbellew: and thank you for not only reporting a problem but also fixing it. doesn't happen too often :)

PR #2119, opened 2020-07-25
Router: prefer case break after `=>` over `if`

kitbellew:
By changing the line associated with the newline split after `=>`, we give higher priority to the State ending in this split compared to the State ending in the space split when cost and depth are the same.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -415,7 +415,7 @@ class Router(formatOps: FormatOps) {
         val caseStat = leftOwner.asInstanceOf[Case]
         if (isCaseBodyABlock(tok, caseStat)) Seq(Split(Space, 0))
         else {
-          def newlineSplit(cost: Int) = {
+          def newlineSplit(cost: Int)(implicit line: sourcecode.Line) = {
             val noIndent = rhsIsCommentedOut(tok)
             val isDouble = tok.hasBlankLine && caseStat.body.tokens.isEmpty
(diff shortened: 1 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -415,7 +415,7 @@ class Router(formatOps: FormatOps) {
         val caseStat = leftOwner.asInstanceOf[Case]
         if (isCaseBodyABlock(tok, caseStat)) Seq(Split(Space, 0))
         else {
-          def newlineSplit(cost: Int) = {
+          def newlineSplit(cost: Int)(implicit line: sourcecode.Line) = {
```
poslegm, line 418: Does `sourcecode.Line` somehow affects formatting?

Or is this PR just for unit test addition?

kitbellew: it absolutely does. the ordering of State uses the line of the associated Split, so this line affects the position in the BestFirstSearch priority queue.

kitbellew: and let me mention once again that i normally add a test in a separate commit, before modifying the code in which the test also changes, so you could review each commit separately and see for yourself.

kitbellew: added explanation in the commit message and the PR description

poslegm: got it, thank you for explanation

PR #2139, opened 2020-08-08
Release moved from Travis to GitHub Actions

poslegm:


.github/workflows/release.yml
```diff
@@ -0,0 +1,20 @@
...
+      - uses: olafurpg/setup-scala@v2
+      - uses: olafurpg/setup-gpg@v2
+      - run: git fetch --tags --unshallow
+      - name: Publish ${{ github.ref }}
+        run: sbt ci-release docs/docusaurusPublishGhpages
+        env:
+          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
+          PGP_SECRET: ${{ secrets.PGP_SECRET }}
+          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
```
poslegm, line 20: @olafurpg Hello! I have not access to repo settings. Can you please check are this secrets specified?

.github/workflows/release.yml
```diff
@@ -0,0 +1,20 @@
+name: Release
+on:
+  push:
+    branches: [master]
+    tags: ["*"]
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
```
kitbellew, line 10: don't know if it's important; `ci.yml` uses:
* `actions/checkout@v2`, not `v1`
* `olafurpg/setup-scala@v7`, not `v2`

.github/workflows/release.yml
```diff
@@ -0,0 +1,20 @@
...
+      - uses: olafurpg/setup-scala@v2
+      - uses: olafurpg/setup-gpg@v2
+      - run: git fetch --tags --unshallow
+      - name: Publish ${{ github.ref }}
+        run: sbt ci-release docs/docusaurusPublishGhpages
+        env:
```
kitbellew, line 16: `scalameta` uses something very similar, a few additional variables: https://github.com/scalameta/scalameta/blob/master/.github/workflows/release.yml#L19

poslegm: Hm, I don't know what they can be used for. Website deploy may be?

kitbellew: idk

.github/workflows/release.yml
olafurpg: These four variables should already be configured for this repo but I think you are still missing some secret variables for the deploy key to publish the website.

poslegm: I added missing variables just like in scalameta repo

kitbellew: is there a way to test it, other than merging and trying to push a release?

poslegm: Release job pushes snapshot to the [sonatype repo](https://oss.sonatype.org/content/repositories/snapshots/org/scalameta/scalafmt-core_2.13/) on every merge. So we can test sonatype publishing process after merge.

But I don't know how to test website publishing

kitbellew: if there's no better way, let's try merging. if necessary, we create some test release, out of order with the proper sequence.

olafurpg: There’s no way to test this via a PR but you can merge into a non-master branch. I normally just merge and push fixes into master until it works. 

The google credentials are not needed for the website, only the deploy key.

.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
...
+          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
+          PGP_SECRET: ${{ secrets.PGP_SECRET }}
+          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+          GIT_USER: scalameta@scalameta.org
+          GITHUB_DEPLOY_KEY: ${{ secrets.GITHUB_DEPLOY_KEY }}
```
olafurpg, line 22: I recall some user saying it GitHub Actions secrets can't start with "GITHUB_". Make sure we are on the latest mdoc version (2.2.4)
```suggestion
          GIT_DEPLOY_KEY: ${{ secrets.GIT_DEPLOY_KEY }}
```

olafurpg: I just updated the deploy keys and secrets following instructions here https://scalameta.org/mdoc/docs/docusaurus.html#publish-to-github-pages-from-ci

olafurpg:
Looks like the release worked 🎉

PR #2153, opened 2020-08-18
Policy: refactor, change how exclusion is represented

kitbellew:
Follow-on to #2141.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -248,50 +248,30 @@ class FormatOps(
   final def startsStatement(token: Token): Option[Tree] =
     statementStarts.get(hash(token))
 
-  def parensRange(token: Token, exclusive: Boolean = true): Option[Range] =
-    matchingOpt(token).map(matchingParensRange(exclusive)(token, _))
-
-  def matchingParensRange(exclusive: Boolean)(a: Token, b: Token): Range = {
-    def eval(lt: Token, rt: Token): Range =
-      Range(lt.start, if (exclusive) rt.start else rt.end)
-    if (a.start < b.end) eval(a, b) else eval(b, a)
-  }
-
-  def matchingParensRangeTupled(exclusive: Boolean): ((Token, Token)) => Range =
-    (matchingParensRange(exclusive) _).tupled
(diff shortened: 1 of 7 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -248,50 +248,28 @@ class FormatOps(
...
-  def matchingParensRangeTupled(exclusive: Boolean): ((Token, Token)) => Range =
-    (matchingParensRange(exclusive) _).tupled
+  def parensTuple(token: Token): Option[(T, T)] =
+    matchingOpt(token).map(matchingParensTuple(token, _))
 
   def getExcludeIf(
       end: Token,
       cond: Token => Boolean = _.is[T.RightBrace]
-  ): Set[Range] = {
+  ): Map[T, T] =
```
poslegm, line 257: I thought for a while on this data structure. It is not immediately clear that `Map[T, T]` is a set of open-close pairs.

But I have no ideas how to make it simply. Type alias may be?

Another question: with `Set[(T, T)]` I can have `Set((1, 2), (1, 3))` - pairs with same start but different end. With `Map[T, T]` it will be a collision. Was it done on purpose here?

poslegm (approved):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
kitbellew: i don't know if there's an ideal data structure here. it needs to represent a list of non-overlapping ranges, i.e.
* k1 != k2
* k1 < v1
* if k1 < k2, then v1 <= k2 as well.

poslegm: If keys should be unique than `Set[(T, T)]` not suitable here

may be at least specify it in the type signature?
```scala
type NonOverlappingRanges[T] = Map[T, T]
```

kitbellew: added an explicit implementation, take a look, please.

poslegm: It's much clearer now, thank you!

PR #2182, opened 2020-09-06
Scalafmt: support `@`-reloaded Ammonite scripts

### LINKED ISSUE #1966: Format .sc Ammonite and Mill files
It would be cool if scalafmt was formatting .sc files which are used for Ammonite and Mill.
Generally it works except 2 corner cases:
- Top level definitions such as:
  ```scala
  def a = {
    "sdcfs"
    ""
  }
  ```
  are formatted:
  ```scala
  def a = {
      "sdcfs"
      ""
    }
  ```
- With Ammonite (thus Mill too) you can reload the interpreter with `@`. Useful for example to add repositories (see [an example here](http://www.lihaoyi.com/mill/page/thirdparty-modules.html#ensime)). That `@` is failing scalafmt to format.

Thanks

kitbellew:
Fixes #1966.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
@@ -16,6 +16,7 @@ import org.scalafmt.internal.BestFirstSearch
 import org.scalafmt.internal.FormatOps
 import org.scalafmt.internal.FormatWriter
 import org.scalafmt.rewrite.Rewrite
+import org.scalafmt.util.FileOps
 
 /**
   * WARNING. This API is discouraged when integrating with Scalafmt from a build tool
@@ -61,7 +62,7 @@ object Scalafmt {
       if (filename == defaultFilename) baseStyle
       else { // might throw for invalid conf
         val style = baseStyle.getConfigFor(filename)
-        val isSbt = filename.endsWith(".sc") || filename.endsWith(".sbt")
+        val isSbt = FileOps.isAmmonite(filename) || FileOps.isSbt(filename)
         if (isSbt) style.forSbt else style
       }
     val isWin = code.contains(WinLineEnding)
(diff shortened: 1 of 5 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
```diff
@@ -62,48 +63,70 @@ object Scalafmt {
...
+      @tailrec
+      def iter(input: Seq[String], output: Seq[String]): Try[Seq[String]] =
+        if (input.isEmpty) Success(output)
+        else
+          doFormatOne(input.head, style, range, file) match {
+            case Failure(x) => Failure(x)
+            case Success(x) => iter(input.tail, output :+ x)
+          }
+      // see: https://ammonite.io/#Save/LoadSession
+      iter(code.split("(?:\\s*\\n@)+"), Seq.empty).map(_.mkString("\n@\n"))
```
poslegm, line 99: Just propose to generalize it as reusable function:
```scala
def traverse[A, B](xs: Seq[A])(f: A => Try[B]): Try[Seq[B]] = {
    @tailrec
    def iter(rest: Seq[A], res: Seq[B]): Try[Seq[B]] =
      if (rest.isEmpty) Success(res)
      else
        f(rest.head) match {
          case Success(x) => iter(rest.tail, res :+ x)
          case Failure(e) => Failure(e)
        }
    iter(xs, Seq.empty)
  }
```

and here

```scala
traverse(code.split("(?:\\s*\\n@)+"), code => doFormatOne(code, style, range, file)).map(_.mkString("\n@\n"))
```

poslegm (approved):

scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
kitbellew: done. also, removed processing for ranges, because they wouldn't work after the split.

poslegm (approved):

PR #2223, opened 2020-10-15
Router: preserve breaks for source.keep in `case`

### LINKED ISSUE #2219: Unable to align on pattern alternative, even by setting align.tokens
- **Version**: 2.4.2
- **Integration**: SBT
- **Configuration**:
```
version = 2.7.4

style = default

maxColumn = 100

align.tokens = [{code = "|", owner = "Pat.Alternative"}]

danglingParentheses = true

trailingCommas = preserve

verticalAlignMultilineOperators = true

rewrite.rules = [
  AvoidInfix
  RedundantBraces
  RedundantParens
  AsciiSortImports
  PreferCurlyFors
]

project.excludeFilters = [
   "scalafix-inputs",
   "scalafix-outputs"
]
```

## Steps

Given code like this:
```scala
case ( Root / "foo"
     | Root / "bar"
     | Root / "baz" / _ / "qux"
     ) => unifiedResult
```

```bash
sbt scalafmtAll
```

## Problem

```scala
case (Root / "foo" | Root / "bar" | Root / "baz" / _ / "qux") => unifiedResult
```

## Expectation

I would like the formatted output to look like this (same as original):

```scala
case ( Root / "foo"
     | Root / "bar"
     | Root / "baz" / _ / "qux"
     ) => unifiedResult
```

## Workaround

My only option right now is disabling the region with `// format: off` and `// format: on`.

## Notes

I've been trying to solve this by setting `align.tokens` to various values. I think `[{code = "|", owner = "Pat.Alternative"}]` is the right value based on:

```scala
scala> p"(2 | 3)".productPrefix
val res10: String = Pat.Alternative
```

But I've also trying setting `owner` to `Case`, `Term.Match`, and `Term.PartialFunction`, all with no success.

kitbellew:
Also, fix closing paren when `spaces.inParentheses` is set. Currently, one case is excluded for the opening paren (`if`/`while`/etc.), but a completely different one (`isDefnOrCallSite`) applies to the closing paren. Let's fix that.

Fixes #2219.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1918,12 +1918,16 @@ class Router(formatOps: FormatOps) {
         Seq(
           Split(NoSplit, 0)
         )
-      case FormatToken(_, T.RightParen(), _) =>
-        val mod =
-          Space(style.spaces.inParentheses && isDefnOrCallSite(rightOwner))
-        Seq(
-          Split(mod, 0)
-        )
+      case FormatToken(_, close: T.RightParen, _) =>
+        def modNoNL = {
+          def allowSpace = rightOwner match {
+            case _: Term.If | _: Term.While | _: Term.For | _: Term.ForYield =>
+              isLastToken(close, rightOwner)
+            case _ => true
+          }
+          Space(style.spaces.inParentheses && allowSpace)
+        }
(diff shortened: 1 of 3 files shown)


poslegm (approved):

kitbellew:
@poslegm GitHub actions had some warnings, found explanation here: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

PR #2225, opened 2020-10-22
Bump deps coursier,sbt

kpbochenek:
Current coursier version fails to parse non SemVer version deps.
You cannot build e.g. scalameta SNAPSHOT and test it with scalafmt.
It is fixed in latest coursier.
But latest coursier conflicts with sbt 1.3.x
And latest sbt already contains sbt-coursier.

Fun stuff :)

Previously:
`Ignoring unparsed versions: Vector(4.3.24+30-63c1bd76-SNAPSHOT)`

Now works fine:
`val scalametaV  = "4.3.24+30-63c1bd76-SNAPSHOT"`

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -8,9 +8,6 @@ addSbtPlugin("org.scalameta" % "sbt-mdoc" % "2.2.9")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.2")
 addSbtPlugin("com.geirsson" % "sbt-ci-release" % "1.5.3")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.10.0")
-addSbtPlugin(
-  "io.get-coursier" % "sbt-coursier" % coursier.util.Properties.version
-)
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.15.0")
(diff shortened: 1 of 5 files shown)


project/plugins.sbt
```diff
@@ -8,9 +8,6 @@ addSbtPlugin("org.scalameta" % "sbt-mdoc" % "2.2.9")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.2")
 addSbtPlugin("com.geirsson" % "sbt-ci-release" % "1.5.3")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.10.0")
-addSbtPlugin(
-  "io.get-coursier" % "sbt-coursier" % coursier.util.Properties.version
```
kitbellew, line 12: is there a place where this `coursier.util.Properties.version` is defined -- and no longer needed?

kpbochenek: https://github.com/scalameta/scalafmt/commit/f644b79585a0fceff82102d8f5d266508b09786c

Was probably coming from meta sbt-coursier, that's why in project/project/plugins.sbt version was fixed.

kitbellew (approved):

poslegm (approved):

PR #2237, opened 2020-10-26
Revert "Bump deps coursier,sbt"

poslegm:
Reverts scalameta/scalafmt#2225

I need to revert it because snapshot release [fails](https://github.com/scalameta/scalafmt/runs/1308100322?check_suite_focus=true) on command `sbt ci-release docs/docusaurusPublishGhpages` 

cc @kpbochenek

PR #2238, opened 2020-10-26
Try fix docusaurusPublish with sbt 1.4.1

kpbochenek:
Try fix: #2237


kpbochenek:
It would fail anyway on CI

Issue #2228, opened 2020-10-23
Spaces inside ${ curly.braced.parameters } of interpolated strings


arbitrary-dev:
- **`sbt-scalafmt` version**: 2.4.2
- **Integration**: SBT
- **Configuration**:
```
version = "2.7.5"

preset=IntelliJ
maxColumn = 110
trailingCommas = always

align.preset = some
continuationIndent.extendSite = 2
docstrings.blankFirstLine = yes
indentOperator.exclude = "^(&&|\\|\\||~)$"
newlines.source = keep
newlines.implicitParamListModifierPrefer = after
spaces.beforeContextBoundColon = always

rewrite.rules = [
  SortImports,
  RedundantBraces,
  RedundantParens,
  SortModifiers,
]
```

## Problem

Scalafmt formats code like this:
```diff
-val msg = s"Request took ${ timeElapsed.toMillis }ms"
+val msg = s"Request took ${timeElapsed.toMillis}ms"
```

## Expectation

I would like the formatter to have it configurable:
```
spaces.inInterpolatedStringCurlyBraces = always
```

arbitrary-dev:
Also would be nice to have an option for simplification of parameters:
```diff
-val msg = s"Pretty ${ simple } parameter"
+val msg = s"Pretty $simple parameter"

val msg2 = s"Can't be ${ simpli }fied"
```

kitbellew:
didn't use the word `fixes`.

arbitrary-dev:
```
spaces.inInterpolatedStringCurlyBraces = true
```

PR #2249, opened 2020-10-28
Add spaces.inInterpolatedStringCurlyBraces

arbitrary-dev:
Implements #2228

Your comments were addressed, please review once again @kitbellew 

For the exception of this one:
>i would suggest moving this class/object inside object Spaces, just like we did with Newlines.

I've looked at `Spaces` & `Newlines`, and didn't quite get the point)


arbitrary-dev:
I've simplified the implementation, please take a look @olafurpg

kitbellew:
Code of conduct needs to be resolved with olafurpg before proceeding.

PR #2252, opened 2020-10-29
Fix wart that scalafmt will reformt `behavior of` in scalatest.

sequencer:
In scalatest, scala format will reformat:
```diff
- behavior of "something"
+ behavior.of("something")
```
This patch will fix this.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Pattern.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Pattern.scala
@@ -33,6 +33,7 @@ object Pattern {
       "in",
       "ignore",
       "be",
+      "of",
       "taggedAs",
       "thrownBy",
       "synchronized",


poslegm:
Hello! Not sure about default for this word. You can just specify some filters you want in your `.scalafmt.conf`:

```
  until
  to
  by
  eq
  ne
  "should.*"
  "contain.*"
  "must.*"
  in
  ignore
  be
  taggedAs
  thrownBy
  synchronized
  have
  when
  size
  only
  noneOf
  oneElementOf
  noElementsOf
  atLeastOneElementOf
  atMostOneElementOf
  allElementsOf
  inOrderElementsOf
  theSameElementsAs
  of
  include
]
```

Why is this method not suitable for you?

kitbellew:
here's an example of the config we use:

```
rewrite {
  rules = [
    AvoidInfix,
    ...
  ]
  neverInfix.excludeFilters = [
    until
    to
    by
    eq
    ne
    synchronized
  ]
}

fileOverride {
  "glob:**/{test,it,src/test/scala}/{*.scala,**/*.scala}" {
      until
      to
      by
      eq
      ne
      "should.*"
      "contain.*"
      "must.*"
      in
      ignore
      be
      of
      message
      taggedAs
      thrownBy
      synchronized
      have
      when
      size
      only
      noneOf
      "one.*Of"
      "atLeast.*Of"
      "atMost.*Of"
      "all.*Of"
      noElementsOf
      "inOrder.*"
      "theSame.*As"
      length
      regex
      key
      include
      and
      or
    ]
  }
}
```

sequencer:
Thanks! I see, I wonder maybe I can try to upstream the keywords in scalatest? Maybe that will be usefull?

kitbellew:
> Thanks! I see, I wonder maybe I can try to upstream the keywords in scalatest? Maybe that will be usefull?

What do you mean by "upstream" here?

kitbellew (changes requested):
Discussing as a config change

sequencer:
> What do you mean by "upstream" here?
Yes I think scalatest is widely used, and it's elements in `neverInfix` are already scalatest keywords. 
So maybe complement it would be better.

kitbellew:
> > What do you mean by "upstream" here?
> 
> Yes I think scalatest is widely used, and it's elements in `neverInfix` are already scalatest keywords.
> So maybe complement it would be better.

the problem is, some of scalatest's keywords are also popular methods (such as size, in, etc.; for instance, in spark functions), and i don't think they should be formatted as infix. scalatest exceptions should only be applied to test code.

sequencer:
I see, that's sound reasonable!
I'll use @poslegm's solution to override!

PR #2254, opened 2020-11-02
Docstrings: add the Preserve style explicitly

kitbellew:


scalafmt-core/shared/src/main/scala/org/scalafmt/config/Docstrings.scala
```diff
@@ -79,6 +73,9 @@ object Docstrings {
   sealed abstract class Style {
     def skipFirstLine: Boolean
   }
+  case object Preserve extends Style {
+    override def skipFirstLine: Boolean = ???
```
poslegm, line 77: not important, but it may help in the future

```suggestion
    override def skipFirstLine: Boolean = throw new NotImplementedError("skipFirstLine called on Docstrings#Style#Preserve, it's a bug in scalafmt")
```

poslegm (approved):

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Docstrings.scala
kitbellew: will do

PR #2260, opened 2020-11-04
upgrade to Scalameta 4.3.24

SethTisue:
context: I'm trying to get the Scala community build onto a newer version of Scalameta over at https://github.com/scala/community-build/pull/1278

not sure yet if this upgrade would help me, but it's one of the paths I'm exploring, so let's see what CI has to say


SethTisue:
okay, CI is green, good to know.

leaving this out there to merge or not, as you wish.

PR #2261, opened 2020-11-11
Case values alignment

### LINKED ISSUE #2258: continuationIndent.caseSite = 5
## Problem

```diff
 a match {
   case _: A |
-       _: B => ...
+      _: B => ...
 }
```

## Expectation

Would be nice to have an option to specify `continuationIndent.caseSite = 5`, so that case statements look evenly aligned.

arbitrary-dev:
Fixes #2258

Made it by default, to not introduce additional configuration, as it seems to me not much of a use.


kitbellew:
Can't change behavior without a fallback. Code of conduct needs to be resolved with olafurpg before proceeding.

Issue #2216, opened 2020-10-05
Support Scala 3


tgodzik:
Thanks to https://github.com/scalameta/scalameta/pull/2104 and some previous PRs we should be able now to start work on formatting Scala 3 code.  This might bit of work on top of the parser, but I was thinking of coordinating the efforts via this issue.

The list of things we need to support or check:
- [x] enums https://github.com/scalameta/scalafmt/pull/2262 https://github.com/scalameta/scalafmt/pull/2283
- [x] extension methods https://github.com/scalameta/scalafmt/pull/2298
- [x] given definitions https://github.com/scalameta/scalafmt/pull/2332 
- [x] given export https://github.com/scalameta/scalafmt/pull/2331
- [x] given patterns `case a @ given Context =>` https://github.com/scalameta/scalafmt/pull/2327
- [x] given imports `import a.{given Context[_]}` https://github.com/scalameta/scalafmt/pull/2327 
- [x] using -  https://github.com/scalameta/scalafmt/pull/2284
- [x] export clause (`export a.b.{ ... }`) https://github.com/scalameta/scalafmt/pull/2329
- [x] macros/quote/splice (https://github.com/scalameta/scalafmt/pull/2286)
- [x] macro patterns https://github.com/scalameta/scalafmt/pull/2316
- [x] inline and transparent modifier https://github.com/scalameta/scalafmt/pull/2324
- [x] inline if (`inline if (a > b) a else b`) https://github.com/scalameta/scalafmt/pull/2353 
- [x] inline match https://github.com/scalameta/scalafmt/pull/2353
- [x] match types (`type List[X] = X match { ... }`) https://github.com/scalameta/scalafmt/pull/2309
- [x] match chaining and `.match` syntax https://github.com/scalameta/scalafmt/pull/2386
- [x] extending multiple traits with coma (`trait C extends A, B`) https://github.com/scalameta/scalafmt/pull/2326
- [x] trait parameters (`trait X(val i: Int`) https://github.com/scalameta/scalafmt/pull/2295
- [x] ~~super trait~~ **(removed in 3.0.0-M2)**
- [x] open classes (PR https://github.com/scalameta/scalafmt/pull/2285)
- [x] infix modifier https://github.com/scalameta/scalafmt/pull/2325
- [x] derivation (`class X derives Eql`)  https://github.com/scalameta/scalafmt/pull/2371
[...]

PR #2262, opened 2020-11-17
Support enum in dotty

MichalBednarz:
The first change related to #2216. 

We've discussed with @tgodzik that it would be reasonable to introduce consecutive features in separate PRs so that it simplifies both the review process and introduction of further changes. Let me know if you prefer me to do this differently.

Here I attempted to support the enum in the exact same way that the class keyword is implemented right now.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -945,13 +945,12 @@ class FormatOps(
       }
 
   def binPackParentConstructorSplits(
-      chain: Either[Tree, Seq[Type.With]],
-      inits: Option[List[Init]],
+      owners: Set[Tree],
       lastToken: Token,
-      indentLen: Int
+      indentLen: Int,
+      extendsThenWith: Boolean = false
   )(implicit line: sourcecode.Line, style: ScalafmtConfig): Seq[Split] = {
     val nlMod = NewlineT(alt = Some(Space))
-    val owners = chain.fold[Set[Tree]](Set(_), x => x.toSet)
     val nlPolicy = ctorWithChain(owners, lastToken)
     val nlOnelineTag = style.binPack.parentConstructors match {
       case BinPack.ParentCtors.Oneline => Right(true)
@@ -963,7 +962,6 @@ class FormatOps(
         Right(style.newlines.source eq Newlines.fold)
     }
     val indent = Indent(Num(indentLen), lastToken, ExpiresOn.After)
-    val extendsThenWith = inits.exists(_.length > 1)
     Seq(
       Split(Space, 0).withSingleLine(lastToken, noSyntaxNL = extendsThenWith),
       Split(nlMod, 0)
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -945,7 +945,8 @@ class FormatOps(
       }
 
   def binPackParentConstructorSplits(
-      chain: Either[Template, Seq[Type.With]],
+      chain: Either[Tree, Seq[Type.With]],
+      inits: Option[List[Init]],
       lastToken: Token,
       indentLen: Int
   )(implicit line: sourcecode.Line, style: ScalafmtConfig): Seq[Split] = {
@@ -962,7 +963,7 @@ class FormatOps(
         Right(style.newlines.source eq Newlines.fold)
     }
     val indent = Indent(Num(indentLen), lastToken, ExpiresOn.After)
-    val extendsThenWith = chain.left.exists(_.inits.length > 1)
+    val extendsThenWith = inits.exists(_.length > 1)
     Seq(
       Split(Space, 0).withSingleLine(lastToken, noSyntaxNL = extendsThenWith),
       Split(nlMod, 0)
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -556,8 +556,12 @@ class Router(formatOps: FormatOps) {
         Seq(
           Split(modification, 0)
         )
-      // Defn.{Object, Class, Trait}
-      case FormatToken(_: T.KwObject | _: T.KwClass | _: T.KwTrait, r, _) =>
+      // Defn.{Object, Class, Trait, Enum}
+      case FormatToken(
+            _: T.KwObject | _: T.KwClass | _: T.KwTrait | _: T.KwEnum,
+            r,
+            _
+          ) =>
         val expire = defnTemplate(leftOwner)
           .flatMap(templateCurly)
           .getOrElse(leftOwner.tokens.last)
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1394,10 +1394,10 @@ class Router(formatOps: FormatOps) {
         val lastToken = rightOwner.tokens.last
         val enumCase = rightOwner.asInstanceOf[Defn.EnumCase]
         binPackParentConstructorSplits(
-          Left(rightOwner),
-          Some(enumCase.inits),
+          Set(rightOwner),
           lastToken,
-          style.continuationIndent.extendSite
+          style.continuationIndent.extendSite,
+          enumCase.inits.length > 1
         )
       // Template
       case FormatToken(_, right @ T.KwExtends(), _) =>
@@ -1407,10 +1407,10 @@ class Router(formatOps: FormatOps) {
           .orElse(template.map(_.tokens.last))
           .getOrElse(rightOwner.tokens.last)
         binPackParentConstructorSplits(
-          template.toLeft(Seq.empty),
-          template.map(_.inits),
+          template.toSet,
           lastToken,
-          style.continuationIndent.extendSite
+          style.continuationIndent.extendSite,
+          template.exists(_.inits.length > 1)
         )
       case FormatToken(_, T.KwWith(), _) =>
         def isFirstWith(t: Template) =
@@ -1426,8 +1426,9 @@ class Router(formatOps: FormatOps) {
               } =>
             splitWithChain(
               isFirstWith(template),
-              Left(template),
-              templateCurly(template).getOrElse(template.tokens.last)
+              Set(template),
+              templateCurly(template).getOrElse(template.tokens.last),
+              template.inits.length > 1
             )
 
           case template: Template =>
@@ -1460,7 +1461,7 @@ class Router(formatOps: FormatOps) {
           case t @ WithChain(top) =>
             splitWithChain(
               !t.lhs.is[Type.With],
-              Right(withChain(top)),
+              withChain(top).toSet,
               top.tokens.last
             )
 
@@ -2027,15 +2028,16 @@ class Router(formatOps: FormatOps) {
 
   private def splitWithChain(
       isFirstWith: Boolean,
-      chain: => Either[Template, Seq[Type.With]],
-      lastToken: => Token
+      owners: => Set[Tree],
+      lastToken: => Token,
+      extendsThenWith: => Boolean = false
   )(implicit line: sourcecode.Line, style: ScalafmtConfig): Seq[Split] =
     if (isFirstWith) {
       binPackParentConstructorSplits(
-        chain,
-        chain.left.toOption.map(_.inits),
+        owners,
         lastToken,
-        IndentForWithChains
+        IndentForWithChains,
+        extendsThenWith
       )
     } else {
       Seq(Split(Space, 0), Split(Newline, 1))
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1388,6 +1388,17 @@ class Router(formatOps: FormatOps) {
         // Add space if right starts with a symbol
         Seq(Split(identModification(right), 0))
 
+      // Enum Case
+      case FormatToken(_, T.KwExtends(), _)
+          if rightOwner.isInstanceOf[Defn.EnumCase] =>
+        val lastToken = rightOwner.tokens.last
+        val enumCase = rightOwner.asInstanceOf[Defn.EnumCase]
+        binPackParentConstructorSplits(
+          Left(rightOwner),
+          Some(enumCase.inits),
+          lastToken,
+          style.continuationIndent.extendSite
+        )
       // Template
       case FormatToken(_, right @ T.KwExtends(), _) =>
         val template = defnTemplate(rightOwner)
@@ -1397,6 +1408,7 @@ class Router(formatOps: FormatOps) {
           .getOrElse(rightOwner.tokens.last)
         binPackParentConstructorSplits(
           template.toLeft(Seq.empty),
+          template.map(_.inits),
           lastToken,
           style.continuationIndent.extendSite
         )
@@ -1452,6 +1464,14 @@ class Router(formatOps: FormatOps) {
               top.tokens.last
             )
 
+          case enumCase: Defn.EnumCase =>
+            val indent = style.continuationIndent.withSiteRelativeToExtends
+            val expire = enumCase.tokens.last
+            Seq(
+              Split(Space, 0).withIndent(indent, expire, ExpiresOn.After),
+              Split(Newline, 1)
+                .withIndent(indent, expire, ExpiresOn.After)
+            )
           case _ =>
             Seq(Split(Space, 0))
         }
@@ -2011,7 +2031,12 @@ class Router(formatOps: FormatOps) {
       lastToken: => Token
   )(implicit line: sourcecode.Line, style: ScalafmtConfig): Seq[Split] =
     if (isFirstWith) {
-      binPackParentConstructorSplits(chain, lastToken, IndentForWithChains)
+      binPackParentConstructorSplits(
+        chain,
+        chain.left.toOption.map(_.inits),
+        lastToken,
+        IndentForWithChains
+      )
     } else {
       Seq(Split(Space, 0), Split(Newline, 1))
     }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenClasses.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenClasses.scala
@@ -10,18 +10,18 @@ object Keyword {
   def unapply(token: Token): Boolean = {
     token.is[KwAbstract] || token.is[KwCase] || token.is[KwCatch] ||
     token.is[KwClass] || token.is[KwDef] || token.is[KwDo] ||
-    token.is[KwElse] || token.is[KwExtends] || token.is[KwFalse] ||
-    token.is[KwFinal] || token.is[KwFinally] || token.is[KwFor] ||
-    token.is[KwForsome] || token.is[KwIf] || token.is[KwImplicit] ||
-    token.is[KwImport] || token.is[KwLazy] || token.is[KwMatch] ||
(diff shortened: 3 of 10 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenClasses.scala
```diff
@@ -10,18 +10,18 @@ object Keyword {
...
-    token.is[KwImport] || token.is[KwLazy] || token.is[KwMatch] ||
-    token.is[KwMacro] || token.is[KwNew] || token.is[KwNull] ||
-    token.is[KwObject] || token.is[KwOverride] || token.is[KwPackage] ||
-    token.is[KwPrivate] || token.is[KwProtected] || token.is[KwReturn] ||
-    token.is[KwSealed] || token.is[KwSuper] || token.is[KwThis] ||
-    token.is[KwThrow] || token.is[KwTrait] || token.is[KwTrue] ||
-    token.is[KwTry] || token.is[KwType] || token.is[KwVal] ||
-    token.is[KwVar] || token.is[KwWhile] || token.is[KwWith] ||
-    token.is[KwYield]
+    token.is[KwElse] || token.is[KwEnum] || token.is[KwExtends] ||
```
kitbellew, line 13: if you are adding enum only, would it make sense to append it so that only one line changes?

scalafmt-tests/src/test/resources/test/Enum.stat
```diff
@@ -0,0 +1,51 @@
...
+
+  case Mercury extends Planet(3.303e+23, 2.4397e6)
+  case Venus extends Planet(4.869e+24, 6.0518e6)
+  case Earth extends Planet(5.976e+24, 6.37814e6)
+  case Mars extends Planet(6.421e+23, 3.3972e6)
+  case Jupiter extends Planet(1.9e+27, 7.1492e7)
+  case Saturn extends Planet(5.688e+26, 6.0268e7)
+  case Uranus extends Planet(8.686e+25, 2.5559e7)
+  case Neptune extends Planet(1.024e+26, 2.4746e7)
+}
```
kitbellew, line 51: add newline

kitbellew (changes requested):

kitbellew:
> ... introduce consecutive features in separate PRs ... if you prefer me to do this differently.

that's exactly how I'd prefer to have it done. unlike gerrit, github doesn't allow you to create a sequence of prs, one on top of the other, but if some are completely independent, you can submit them separately.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -556,8 +556,12 @@ class Router(formatOps: FormatOps) {
         Seq(
           Split(modification, 0)
         )
-      // Defn.{Object, Class, Trait}
-      case FormatToken(_: T.KwObject | _: T.KwClass | _: T.KwTrait, r, _) =>
+      // Defn.{Object, Class, Trait, Enum}
```
kitbellew, line 559: forgot to ask. should we assume that if the dialect doesn't support enum, KwEnun would not be parsed?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenClasses.scala
MichalBednarz: Of course, in similar cases in the future, I will simply prefer the number of lines changed over the alphabetical order if it will make sense.

kitbellew: didn't realize they are sorted. then keep as is.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Yes, it will not be parsed in that case. I will be a normal Ident.

kitbellew:
@poslegm there's some problem with environment, i think you looked into that before, any ideas how to fix?

@olafurpg some of the errors are coming from your module, are you the only one who can create an updated version?

gabro:
@kitbellew we identified and fix the same issue in other scalameta projects. I think @tgodzik is working on a fix for the CI (it's just a matter of updating the action versions)

tgodzik:
> @poslegm there's some problem with environment, i think you looked into that before, any ideas how to fix?
> 
> @olafurpg some of the errors are coming from your module, are you the only one who can create an updated version?

I am on it!

olafurpg:
Upgrading olafurpg/setup-scala to v10 should fix the CI errors.

Screenshot 2020-11-17 at 14 26 38

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
MichalBednarz: That's a good question. As I play with the dialect-specific features right know I can see that they parse even if they're not enabled in the dialect. I assume that it should be parsed then?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenClasses.scala
MichalBednarz: can be added as the fourth in the line as well

MichalBednarz: However, it is formatted in a strange way then so I will then keep it as it is

scalafmt-tests/src/test/resources/test/Enum.stat
```diff
@@ -0,0 +1,51 @@
...
+enum Color {
+case Red, Green, Blue
+}
+>>>
+enum Color {
+  case Red, Green, Blue
+}
+<<< parameterized enum
+enum Color (val rgb: Int) {
+case Red   extends Color(0xff0000)
```
tgodzik, line 12: Could you add some cases that a longer and might break the line? You can add an additional setting to make the max column shorter.

tgodzik (changes requested):
Could you add some tests that could try to break the line?

scalafmt-tests/src/test/resources/test/Enum.stat
kitbellew: Actually, this is interesting. So enums are represented as cases, which means the formatter might be looking at them as match cases, since there was nothing else before. Do we need to modify the processing of KwCase as well? At least add tests which will make enum cases look very different from match cases and show that it still works and uses the right amount of indentation.

scalafmt-tests/src/test/resources/test/Enum.stat
```diff
@@ -0,0 +1,51 @@
...
+case Green extends Color(0x00ff00)
+case Blue  extends Color(0x0000ff)
+}
+>>>
+enum Color (val rgb: Int) {
```
kitbellew, line 17: Shouldn't there be no space before opening paren?

scalafmt-tests/src/test/resources/test/Enum.stat
```diff
@@ -0,0 +1,51 @@
...
+case Venus extends Planet(4.869e+24, 6.0518e6)
+case Earth extends Planet(5.976e+24, 6.37814e6)
+case Mars extends Planet(6.421e+23, 3.3972e6)
+case Jupiter extends Planet(1.9e+27, 7.1492e7)
+case Saturn extends Planet(5.688e+26, 6.0268e7)
+case Uranus extends Planet(8.686e+25, 2.5559e7)
+case Neptune extends Planet(1.024e+26, 2.4746e7)
+}
+>>>
+enum Planet (mass: Double, radius: Double) {
```
kitbellew, line 38: Can you force the line to break and see what indentation you get, and whether it breaks on every comma? Add a few more parameters, to have more commas.

scalafmt-tests/src/test/resources/test/Enum.stat
```diff
@@ -0,0 +1,51 @@
...
+  case Green extends Color(0x00ff00)
+  case Blue extends Color(0x0000ff)
+}
+<<< user-defined members of enums
+enum Planet (mass: Double, radius: Double) {
+private final val G = 6.67300e-11
+def surfaceGravity = G * mass / (radius * radius)
+def surfaceWeight(otherMass: Double) = otherMass * surfaceGravity
+
+case Mercury extends Planet(3.303e+23, 2.4397e6)
```
kitbellew, line 28: Can you add some additional `with` statements and see where it breaks and what indentation is used?

kitbellew (changes requested):
Make your test cases way more complex, so that more rules would need to be applied and potentially break

scalafmt-tests/src/test/resources/test/Enum.stat
MichalBednarz: Good point, there is no space after the class identifier

scalafmt-tests/src/test/resources/test/Enum.stat
MichalBednarz: Implementation needs changes because it ends up being:
```
enum Planet (mass: Double,
radius: Double,
name: String,
id: Int) {
```

MichalBednarz: In addition, another error occurs (though it doesn't seem to be related to the enum).
```
- test/Enum.stat:24: user-defined members of enums                      | *** FAILED ***
org.scalafmt.Error$FormatterChangedAST: Formatter changed AST
=====================
6.67300e-11 != 6.67300e-11
(scala.meta.Lit$Double$LitDoubleImpl != scala.meta.Term$Block$TermBlockImpl)
```

scalafmt-tests/src/test/resources/test/Enum.stat
MichalBednarz: This is the result for now:
```
case Mercury
    extends Planet(
      3.303e+23,
      2.4397e6
    ) with Alpha with Beta with Gamma with Delta with Epsilon
```

scalafmt-tests/src/test/resources/test/Enum.stat
tgodzik: We check if the AST changed after reformatting, which should not happen. This looks suspicious, but might be related to parser. I will take a look.

scalafmt-tests/src/test/resources/test/Enum.stat
MichalBednarz: done

scalafmt-tests/src/test/resources/test/Enum.stat
MichalBednarz: The indentation works as expected for parameter lists now. 
Besides, I threw away the code causing a failing assertion with AST.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -315,7 +315,7 @@ object TreeOps {
     tree match {
       case _: Decl.Def | _: Defn.Def | _: Defn.Macro | _: Defn.Class |
           _: Defn.Trait | _: Ctor.Secondary | _: Decl.Type | _: Defn.Type |
-          _: Type.Apply | _: Type.Param | _: Type.Tuple =>
+          _: Type.Apply | _: Type.Param | _: Type.Tuple | _: Defn.Enum =>
```
kitbellew, line 318: I have found 5 invocations of `Defn.Class` in `TreeOps`, 2 more in `TokenOps`. Should we extend all those cases with `Defn.Enum`?

scalafmt-tests/src/test/resources/test/Enum.stat
```diff
@@ -35,17 +32,47 @@ case Uranus extends Planet(8.686e+25, 2.5559e7)
 case Neptune extends Planet(1.024e+26, 2.4746e7)
 }
 >>>
-enum Planet (mass: Double, radius: Double) {
-  private final val G = 6.67300e-11
-  def surfaceGravity = G * mass / (radius * radius)
-  def surfaceWeight(otherMass: Double) = otherMass * surfaceGravity
-
-  case Mercury extends Planet(3.303e+23, 2.4397e6)
```
kitbellew: is `case` within `enum` parsed using `KwCase` token? is the tree parsed as `Case` or something else? i only see handling of the parent `enum`, but not the `case` trees inside.

kitbellew (changes requested):

scalafmt-tests/src/test/resources/test/Enum.stat
MichalBednarz: yes, I'm going to add such a case. It doesn't work though yet - the changes that fixed the case with the opening parenthesis and parameter list don't apply to the above case.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
MichalBednarz: yeah, I've also noticed these. I'm gonna push them

scalafmt-tests/src/test/resources/test/Enum.stat
MichalBednarz: There is an `EnumCase` token that I was not aware of at the very beginning.

scalafmt-tests/src/test/resources/test/Enum.stat
MichalBednarz: > by the way, can enum itself extend anything, or only the cases?

they can. Following code parses:
```
  enum A extends Alpha with Beta with Gamma with Delta with Epsilon {
    case B
  }
```

MichalBednarz: Changes in `TokenOps` made the above case pass.

MichalBednarz: > additionally, it would be good to have tests to see how this works with `binPack.parentConstructors`, specifically the Oneline and OnelineIfPrimaryOneline. might need adjusted implementation since for enum primary is the extends, since the case doesn't take parameters. we have plenty of examples of how to adjust config for a single test case.

Here is the example passing test case related to this setting:
```
<<<
binPack.parentConstructors = Never
maxColumn = 30
===
enum Foo(val x: Int) {
  case Bar extends Foo(1) with X with Z
}
>>>
enum Foo(val x: Int) {
  case Bar
      extends Foo(
        1
      ) with X with Z
}
```
Here we use `class` kw instead of `enum` and the following passes:
```
<<<
binPack.parentConstructors = Never
maxColumn = 30
===
enum Foo(val x: Int) {
  class Bar extends Foo(1) with X with Z
}
>>>
enum Foo(val x: Int) {
  class Bar
      extends Foo(1)
      with X
      with Z
}
```
I'm not sure if we expect the exact same result with the enum as with class? 

BTW there seems to be a mistake in the docs:
```
Parent constructors are C and D in
"class A extends B with C and D" // probably should be "class A extends B with C with D"
```
In addition - is not B also the parent constructor of A?

kitbellew:
@MichalBednarz i might have lost track... Are you still working on it, or is it ready for review?

MichalBednarz:
> @MichalBednarz i might have lost track... Are you still working on it, or is it ready for review?
There is still some work left with parent constructors - never, and always work, however one-line and onelineIfPrimaryOneline need some more adjustment. I'll let you know as soon as I submit the changes.

MichalBednarz:
@kitbellew actually I need your help with `Oneline` and `OnelineIfPrimaryOnelin` settings. I attempted to see how does `Oneline` setting work with classes. The examples from the docs work as expected however, when slightly changed the following test passes:
```
<<< ONLY class test
binPack.parentConstructors = Oneline
maxColumn = 30
===
object a {
  class A(x: Int)
  extends Foo
  with X
  with Y
  with Z


  class B(
    x: Int
  )
  extends Foo
  with X
  with Y
  with Z
}
>>>
object a {
  class A(x: Int)
      extends Foo
      with X
      with Y
      with Z

  class B(
      x: Int
  ) extends Foo
      with X
      with Y
      with Z
}
```
Though, the result should be like this (am I right?):
```
object a {
  class A(x: Int)
      extends Foo with X 
      with Y with Z

  class B(
      x: Int
  ) extends Foo with X with Y
      with Z
}
```
Is this a bug or expected behavior?

MichalBednarz:
I pushed the changes for `Never` and `Always` parent constructor settings - these are review ready.

kitbellew:
> @kitbellew actually I need your help with `Oneline` and `OnelineIfPrimaryOnelin` settings. I attempted to see how does `Oneline` setting work with classes.

`Oneline` means it will be formatted on a single line if and only if the entire thing fits on one line. Otherwise, it will break on every `extends` and `with`.

`OnelineIfPrimaryOneline` will format `extends` and `with` on a single (possibly separate) line if they fit and if the primary ctor (what comes before `extends`) can be formatted on a single line:

```
// everything fits
class Foo(a: A) extends C

// primary fits, and others fit separately
class Foo(a: A, b: B)
  extends C with D
```

MichalBednarz:
The changes are ready for the review.

tgodzik (approved):
LGTM from me, but obviously we need another approve from @kitbellew 

Thanks for looking into this!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -945,7 +945,8 @@ class FormatOps(
       }
 
   def binPackParentConstructorSplits(
-      chain: Either[Template, Seq[Type.With]],
+      chain: Either[Tree, Seq[Type.With]],
+      inits: Option[List[Init]],
```
kitbellew, line 949: should we add
```
  require(chain.isRight == inits.isEmpty)
```

tgodzik: Maybe we could have an additional method :

```scala
  def binPackParentConstructorSplits(
      chain: Tree
      inits: Option[List[Init]],
      lastToken: Token,
      indentLen: Int) = {
  val extendsThenWith = inits.exists(_.length > 1)
  binPackParentConstructorSplits(Left(chain), extendsThenWith, lastToken, indentLen)
 }
```

and change the second chain parameter to `extendsThenWith : Boolean = false` (we might need to move it last to make sense).
I think it would work a bit nicer than `require`.

MichalBednarz: It would require to overload the `binPackParentConstructorSplits` three times. Moreover we have `splitWithChain` that uses it and thereby should be overloaded too in a similar manner. I would go with @kitbellew suggestion.

tgodzik: One overload would be enough I think. `require` will throw a runtime exception, which is why I am not keen on it.

kitbellew (approved):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -2003,11 +2028,17 @@ class Router(formatOps: FormatOps) {
 
   private def splitWithChain(
       isFirstWith: Boolean,
-      chain: => Either[Template, Seq[Type.With]],
-      lastToken: => Token
+      owners: Set[Tree],
```
tgodzik, line 2031:
```suggestion
      owners: => Set[Tree],
```
Let's use byName also, not sure if that wasn't an optimisation thing. We will not need to calculate it in `else`

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -2003,11 +2028,17 @@ class Router(formatOps: FormatOps) {
...
+      lastToken: => Token,
+      extendsThenWith: Boolean = false
```
tgodzik, line 2033:
```suggestion
      extendsThenWith: => Boolean = false
```

tgodzik (changes requested):
Just a tiny nitpick from my side.

tgodzik (approved):
Thanks a lot! Great work @MichalBednarz ! :tada:

tgodzik:
Also, thanks for the feedback @kitbellew ! We would be honestly lost without it.

kitbellew:
> Also, thanks for the feedback @kitbellew ! We would be honestly lost without it.

i wouldn't go as far as "lost", but happy to help in any way.

PR #2263, opened 2020-11-17
Update actions versions to fix CI

tgodzik:
I also added the dependabot file to make the process automatic.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,11 +14,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [openjdk@1.8, openjdk@1.11]
+        java: [adopt@1.8, adopt@1.11]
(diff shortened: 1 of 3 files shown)


gabro (approved):

.github/workflows/ci.yml
```diff
@@ -14,11 +14,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [openjdk@1.8, openjdk@1.11]
+        java: [adopt@1.8, adopt@1.11]
```
tgodzik, line 17: Seems that jabba no longer lists any 1.8 openjdk. `adopt` should be good to use

olafurpg (approved):
LGTM 👍 It's cool that dependabot can help upgrade actions. Feel like it should be enabled by default.

PR #2282, opened 2020-12-07
Upgrade dependencies

olafurpg:
* Latest Scala.
* Latest sbt, disable BSP server because it doesn't work well with Metals.
* Latest Scalameta.
* Use only ScalaTest FunSuite because matchers introduce compile errors
  in 2.13.3 due to "multiple infix" deprecation warnings.
* Remove unnecessary sbt-coursier dependency because it's included with sbt.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,8 +10,8 @@ def localSnapshotVersion: String = s"$parseTagVersion-SNAPSHOT"
 def isCI = System.getenv("CI") != null
 
 def scala211 = "2.11.12"
-def scala212 = "2.12.8"
-def scala213 = "2.13.1"
+def scala212 = "2.12.12"
+def scala213 = "2.13.3"
 
 inThisBuild(
   List(
--- a/build.sbt
+++ b/build.sbt
@@ -211,7 +211,8 @@ lazy val tests = project
       scalametaTestkit,
       scalatest.value
     ),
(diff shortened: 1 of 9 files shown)


build.sbt
```diff
@@ -10,8 +10,8 @@ def localSnapshotVersion: String = s"$parseTagVersion-SNAPSHOT"
 def isCI = System.getenv("CI") != null
 
 def scala211 = "2.11.12"
-def scala212 = "2.12.8"
-def scala213 = "2.13.1"
+def scala212 = "2.12.12"
+def scala213 = "2.13.3"
```
kitbellew, line 14: this says that the warning is a linter problem in this specific version and can be disabled:

https://contributors.scala-lang.org/t/multiarg-infix-application-considered-warty/4490

do you really believe that scalatest matchers will cease to exist completely?

olafurpg: Idiomatic usage of ScalaTest matchers heavily relies on nested infix operators, which are deprecated to my best understanding. It's maybe possible to tweak the build settings to suppress these deprecation warnings so we can continue to use matchers. However, matchers are only used in one test suite so I figured it was easier to remove them altogether.

Personally, I think matchers are difficult to read and write so I think it's nice to get rid of them.

kitbellew (approved):

PR #2284, opened 2020-12-07
Support using soft keyword in parameter and argument lists

MichalBednarz:
The next one based on #2216.

The `using` soft keyword in parameter lists seems to be one-to-one with the `implicit`. However, `using` can also be used within argument lists. 

We've discussed with @tgodzik that we can add aliases to `newlines.implicitParamListModifierPrefer` and `newlines.implicitParamListModifierForce`, namely `newlines.usingParamListModifierPrefer` and `newlines.usingParamListModifierForce` (and apply this existing setting to the using argument lists).

Please take a look at the changes in the docs, I put the information about the whole `implicit vs using` matter in one place. In addition, we can update every occurrence of `implicit` that applies to the `using` at the same time and replace it with `implicit or using` for the case of specificity and maybe add some additional examples. What do you think?

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1213,6 +1213,18 @@ something.map { x => f(x) }
 
 > Since v2.5.0.
 
+#### Newlines around `using` parameter and argument list modifier
+
+> Since v3.0.0
+
+`using` soft keyword was introduced in Scala 3 and is supported in scalafmt.
+The settings `newlines.implicitParamListModifierPrefer` and
+`newlines.implicitParamListModifierForce` are from now on aliases for
+`newlines.usingParamListModifierPrefer` and `newlines.usingParamListModifierForce`.
+By picking one of them formatting will be applied at the same time to both `implicit` and `using`.
+Besides parameter lists, `using` can also be used with argument lists hence
+provided rules will then also be applied to the argument lists.
+
 #### Prefer After (default)
 
 > Prefers newline after `implicit`. Newline will be added unless the entire
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -45,6 +45,8 @@ class FormatOps(
     baseStyle: ScalafmtConfig,
     val filename: String = ""
 ) {
+  import FormatOps._
+
   val initStyle = {
     val queue = new mutable.Queue[Tree]
     queue += tree
@@ -65,6 +67,7 @@ class FormatOps(
   implicit val dialect = initStyle.runner.dialect
   private val ownersMap = getOwners(tree)
   val tokens: FormatTokens = FormatTokens(tree.tokens, owners)
+  private[internal] val soft = new SoftKeywordClasses(dialect)
   private val statementStarts = getStatementStarts(tree)
   val dequeueSpots = getDequeueSpots(tree) ++ statementStarts.keys
   private val matchingParentheses: Map[TokenHash, Token] =
@@ -882,20 +885,20 @@ class FormatOps(
   )(implicit style: ScalafmtConfig): Boolean = {
     def opensImplicit =
       (style.newlines.forceAfterImplicitParamListModifier ||
-        next(ft).hasBreak) && opensConfigStyleImplicitOrUsingParamList(ft)
+        next(ft).hasBreak) && opensConfigStyleImplicitParamList(ft)
     (ft.hasBreak || opensImplicit) && {
       val close = matching(ft.left)
       tokens(close, -1).hasBreak
     }
   }
 
-  def opensConfigStyleImplicitOrUsingParamList(
+  /** Works for `using` as well */
+  def opensConfigStyleImplicitParamList(
       formatToken: FormatToken
   )(implicit style: ScalafmtConfig): Boolean =
-    (formatToken.right
-      .is[T.KwImplicit] || formatToken.right.text == "using") &&
+    formatToken.right.is[soft.ImplicitOrUsing] &&
       style.newlines.notBeforeImplicitParamListModifier &&
-      opensImplicitOrUsingParamList(formatToken).isDefined
+      opensImplicitParamList(formatToken).isDefined
 
   def styleAt(tree: Tree): ScalafmtConfig = {
     val style = styleMap.at(tree.tokens.head)
@@ -1136,7 +1139,7 @@ class FormatOps(
         def isDefinition = ownerCheck(owners(close2))
 
         val shouldAddNewline = {
-          if (right.is[T.KwImplicit] || right.text == "using")
+          if (right.is[soft.ImplicitOrUsing])
             style.newlines.forceBeforeImplicitParamListModifier ||
             style.verticalMultiline.newlineBeforeImplicitKW
           else
@@ -1147,7 +1150,7 @@ class FormatOps(
           Split(NoSplit.orNL(!shouldAddNewline), 0)
             .withIndent(indentParam, close2, ExpiresOn.Before)
         )
-      case Decision(t @ FormatToken(T.KwImplicit() | T.Ident("using"), _, _), _)
+      case Decision(t @ FormatToken(soft.ImplicitOrUsing(), _, _), _)
           if style.newlines.forceAfterImplicitParamListModifier ||
             style.verticalMultiline.newlineAfterImplicitKW =>
         Seq(Split(Newline, 0))
@@ -1291,22 +1294,22 @@ class FormatOps(
   def getApplyArgs(
       ft: FormatToken,
       isRight: Boolean
-  )(implicit style: ScalafmtConfig): (Tree, Seq[Tree]) = {
+  )(implicit style: ScalafmtConfig): TreeArgs = {
     val paren = if (isRight) ft.right else ft.left
     val owner = if (isRight) ft.meta.rightOwner else ft.meta.leftOwner
     def getArgs(argss: Seq[Seq[Tree]]): Seq[Tree] =
       findArgsFor(paren, argss).getOrElse(Seq.empty)
     owner match {
-      case InfixApp(ia) if style.newlines.formatInfix => (ia.op, ia.rhs)
+      case InfixApp(ia) if style.newlines.formatInfix => TreeArgs(ia.op, ia.rhs)
       case t @ SplitDefnIntoParts(_, name, tparams, paramss) =>
         if (if (isRight) paren.is[T.RightParen] else paren.is[T.LeftParen])
-          (name, getArgs(paramss))
+          TreeArgs(name, getArgs(paramss))
         else
-          (name, tparams)
+          TreeArgs(name, tparams)
       case SplitCallIntoParts(tree, either) =>
         either match {
-          case Left(args) => (tree, args)
-          case Right(argss) => (tree, getArgs(argss))
+          case Left(args) => TreeArgs(tree, args)
+          case Right(argss) => TreeArgs(tree, getArgs(argss))
         }
       case _ =>
         logger.debug(s"""Unknown tree
@@ -1316,9 +1319,7 @@ class FormatOps(
     }
   }
 
-  def opensImplicitOrUsingParamList(
-      ft: FormatToken
-  ): Option[Seq[Term.Param]] = {
+  def opensImplicitParamList(ft: FormatToken): Option[Seq[Term.Param]] = {
     val paramsOpt = splitDefnIntoParts.lift(ft.meta.leftOwner).flatMap {
       case (_, _, _, paramss) =>
         findArgsFor(ft.left, paramss)
@@ -1327,14 +1328,12 @@ class FormatOps(
     paramsOpt.filter(!_.exists(TreeOps.hasExplicitImplicit))
   }
 
-  def opensImplicitOrUsingParamList(
-      ft: FormatToken,
-      args: Seq[Tree]
-  ): Boolean =
+  /** Works for `using` as well */
+  def opensImplicitParamList(ft: FormatToken, args: Seq[Tree]): Boolean =
     ft.right.is[T.KwImplicit] && args.forall {
       case t: Term.Param => !hasExplicitImplicit(t)
       case _ => true
-    } || ft.right.text == "using"
+    } || ft.right.is[soft.KwUsing]
 
   def isEnclosedInMatching(tree: Tree, open: T, close: T): Boolean =
     tree.tokens.headOption.contains(open) && (tree.tokens.last eq close)
@@ -1785,4 +1784,17 @@ class FormatOps(
     }
   }
 
+  // For using to be the soft kw it also has to be preceded by paren
+  def isRightImplicitOrUsingSoftKw(
+      ft: FormatToken,
+      soft: SoftKeywordClasses
+  ): Boolean = ft.left match {
+    case _: T.KwImplicit => true
+    case soft.KwUsing() => prevNonComment(prev(ft)).left.is[T.LeftParen]
+    case _ => false
+  }
+}
+
+object FormatOps {
+  case class TreeArgs(tree: Tree, args: Seq[Tree])
 }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -882,19 +882,20 @@ class FormatOps(
   )(implicit style: ScalafmtConfig): Boolean = {
     def opensImplicit =
       (style.newlines.forceAfterImplicitParamListModifier ||
-        next(ft).hasBreak) && opensConfigStyleImplicitParamList(ft)
+        next(ft).hasBreak) && opensConfigStyleImplicitOrUsingParamList(ft)
     (ft.hasBreak || opensImplicit) && {
       val close = matching(ft.left)
       tokens(close, -1).hasBreak
     }
   }
 
-  def opensConfigStyleImplicitParamList(
+  def opensConfigStyleImplicitOrUsingParamList(
       formatToken: FormatToken
   )(implicit style: ScalafmtConfig): Boolean =
-    formatToken.right.is[T.KwImplicit] &&
+    (formatToken.right
+      .is[T.KwImplicit] || formatToken.right.text == "using") &&
       style.newlines.notBeforeImplicitParamListModifier &&
-      opensImplicitParamList(formatToken).isDefined
+      opensImplicitOrUsingParamList(formatToken).isDefined
 
   def styleAt(tree: Tree): ScalafmtConfig = {
     val style = styleMap.at(tree.tokens.head)
@@ -1135,7 +1136,7 @@ class FormatOps(
         def isDefinition = ownerCheck(owners(close2))
 
         val shouldAddNewline = {
-          if (right.is[T.KwImplicit])
+          if (right.is[T.KwImplicit] || right.text == "using")
             style.newlines.forceBeforeImplicitParamListModifier ||
             style.verticalMultiline.newlineBeforeImplicitKW
           else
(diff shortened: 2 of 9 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1760,10 +1762,21 @@ class Router(formatOps: FormatOps) {
...
-        opensImplicitParamList(prevNonComment(prev(formatToken))).fold {
+            !style.verticalMultiline.atDefnSite && (prev(formatToken) match {
+              // check if using is the soft kw
+              case FormatToken(_: T.LeftParen, T.Ident("using"), _) => true
+              case FormatToken(_, _: T.KwImplicit, _) => true
+              case _ => false
+            }) =>
+        val argsOrParamsOpt = formatToken.meta.leftOwner match {
+          // for the using argument list
+          case _: ApplyUsing => Some(getApplyArgs(formatToken, false)._2)
```
tgodzik, line 1775:
```suggestion
          case _: ApplyUsing => Some{
            val (_, args) = getApplyArgs(formatToken, false)
            args
          }
```
maybe? But this is just a nitpick for sure,

scalafmt-tests/src/test/resources/dotty/Using.stat
```diff
@@ -0,0 +1,235 @@
...
+>>>
+def max[T](x: T, y: T)(using
+    a: A[T],
+    b: B[T],
+    c: C[T]
+): T = {}
+<<< apply method with context argument
+maxColumn = 40
+===
+max(2, 3)(using intOrd)
```
tgodzik, line 22:
```suggestion
max(2, 3)(using          intOrd)
```
just to show that something is happening maybe?

scalafmt-tests/src/test/resources/dotty/Using.stat
```diff
@@ -0,0 +1,235 @@
...
+>>>
+maximum(xs)(using
+  descending(using
+    listOrd(using intOrd)
+  )
+)
+<<< multiple using clauses
+maxColumn = 40
+===
+def f(u: Universe)(using ctx: u.Context)(using s: ctx.Symbol, k: ctx.Kind) = {}
```
tgodzik, line 56: Does it split both if the second clause is really long?

scalafmt-tests/src/test/resources/dotty/Using.stat
```diff
@@ -0,0 +1,235 @@
...
+>>>
+def method2(using
+    a: Int,
+    b: String,
+    c: String
+): Boolean
+<<< using as identifier
+maxColumn = 20
+===
+def using(using using: using, using: using, using: using)
```
tgodzik, line 229: Could you maybe add a case where it's just in a val? Outside of parenthesis?

tgodzik (changes requested):
Looks really good! I mostly have some minor comments.

docs/configuration.md
```diff
@@ -1208,6 +1208,18 @@ something.map { x => f(x) }
 
 > Since v2.5.0.
 
+#### Newlines around `using` parameter and argument list modifier
+
+> Since v2.7.6
```
kitbellew, line 1213: i'd symbolically switch to 3.0.0 when scalafmt fully supports scala 3 :)

tgodzik (approved):
LGTM from my side :tada:

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -876,19 +876,20 @@ class FormatOps(
...
-        next(ft).hasBreak) && opensConfigStyleImplicitParamList(ft)
+        next(ft).hasBreak) && opensConfigStyleImplicitOrUsingParamList(ft)
     (ft.hasBreak || opensImplicit) && {
       val close = matching(ft.left)
       tokens(close, -1).hasBreak
     }
   }
 
-  def opensConfigStyleImplicitParamList(
+  def opensConfigStyleImplicitOrUsingParamList(
```
kitbellew, line 886: perhaps i'd suggest not renaming these methods but in scaladoc adding a comment to specify it covers "using" as well. same for other cases.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1760,10 +1762,22 @@ class Router(formatOps: FormatOps) {
...
-            !style.verticalMultiline.atDefnSite =>
-        opensImplicitParamList(prevNonComment(prev(formatToken))).fold {
+            !style.verticalMultiline.atDefnSite && isRightImplicitOrUsingSoftKw(
+              prev(formatToken)
+            ) =>
+        val argsOrParamsOpt = formatToken.meta.leftOwner match {
+          // for the using argument list
+          case _: ApplyUsing =>
+            Some {
+              val (_, args) = getApplyArgs(formatToken, false)
```
kitbellew, line 1774: why not just `getApplyArgs(...)._2`?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1318,11 +1321,14 @@ class FormatOps(
...
-  def opensImplicitParamList(ft: FormatToken, args: Seq[Tree]): Boolean =
+  def opensImplicitOrUsingParamList(
+      ft: FormatToken,
+      args: Seq[Tree]
+  ): Boolean =
     ft.right.is[T.KwImplicit] && args.forall {
       case t: Term.Param => !hasExplicitImplicit(t)
       case _ => true
-    }
+    } || ft.right.text == "using"
```
kitbellew, line 1331: i'd use `ft.meta.right.text` (a `val`).

`def Token.text` calls `def Token.pos.text` which in turn calls `def Position.Range.text` which calls `new String(input.chars, start, end - start)`.

also, should we define a match for `using` in `util.TokenClasses` (and reuse it in the proposed match for both `implicit` and `using`)?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1140,7 +1141,7 @@ class FormatOps(
           Split(NoSplit.orNL(!shouldAddNewline), 0)
             .withIndent(indentParam, close2, ExpiresOn.Before)
         )
-      case Decision(t @ FormatToken(T.KwImplicit(), _, _), _)
+      case Decision(t @ FormatToken(T.KwImplicit() | T.Ident("using"), _, _), _)
```
kitbellew, line 1144: should we define a match like this in `util.TokenClasses` and reuse everywhere?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
```diff
@@ -222,4 +222,10 @@ object TokenOps {
     }
   }
 
+  def isRightImplicitOrUsingSoftKw(ft: FormatToken): Boolean = ft match {
+    case FormatToken(_: LeftParen, Ident("using"), _) => true
```
kitbellew, line 226: can use the proposed `TokenClasses.KwUsing()` match.

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: That was my suggestion. I really dislike `_2` etc. but I think maybe it's way too complex now.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: We could have something similar to [SkUsing](https://github.com/scalameta/scalameta/blob/9a8781772b32a92fba6abd709d01ff328a20f34b/scalameta/parsers/shared/src/main/scala/scala/meta/internal/parsers/ScalametaParser.scala#L848) in scalameta

kitbellew: can we expose those `SkUsing` and others in scalameta, and just use them directly? they do multiple important things: check the dialect support and remove backticks, so that
```
Ident("using")
Ident("`using`")
```
would both work. it would be ideal to avoid duplication here, no?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: we can also create a new case class and have `getApplyArgs` return it, instead of the tuple.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: What do you think about https://github.com/scalameta/scalameta/pull/2193 ?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1315,7 +1320,9 @@ class FormatOps(
     }
   }
 
-  def opensImplicitParamList(ft: FormatToken): Option[Seq[Term.Param]] = {
+  def opensImplicitParamList(
```
kitbellew, line 1323: can you restore the original formatting?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -65,6 +65,7 @@ class FormatOps(
   implicit val dialect = initStyle.runner.dialect
   private val ownersMap = getOwners(tree)
   val tokens: FormatTokens = FormatTokens(tree.tokens, owners)
+  val soft = new SoftKeywordClasses(dialect)
```
kitbellew, line 68: can it be package-private instead?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1324,11 +1331,15 @@ class FormatOps(
     paramsOpt.filter(!_.exists(TreeOps.hasExplicitImplicit))
   }
 
-  def opensImplicitParamList(ft: FormatToken, args: Seq[Tree]): Boolean =
+  /** Works for `using` as well */
+  def opensImplicitParamList(
```
kitbellew, line 1335: can you restore the original formatting?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -1017,7 +1017,7 @@ class FormatWriter(formatOps: FormatOps) {
     ft.meta.rightOwner match {
       case x: Importer => x.importees.length > 1
       case _ =>
-        val args = getApplyArgs(ft, true)._2
+        val TreeArgs(_, args) = getApplyArgs(ft, true)
```
kitbellew, line 1020: why not
```
getApplyArgs(ft, true).args
```
otherwise, there was no point in creating a case class with _named_ fields.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -62,6 +63,7 @@ class Router(formatOps: FormatOps) {
 
   private def getSplitsImpl(formatToken: FormatToken): Seq[Split] = {
     implicit val style = styleMap.at(formatToken)
+    val soft = new SoftKeywordClasses(dialect)
```
kitbellew, line 66: you don't need it, see line 62, where `formatOps._` is imported.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -759,14 +761,15 @@ class Router(formatOps: FormatOps) {
             Policy.NoPolicy
           }
 
-        val handleImplicit =
```
kitbellew, line 762: can we avoid renaming and restore the original code? can add comment that `using` is also covered, if you'd like

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1762,10 +1765,19 @@ class Router(formatOps: FormatOps) {
       case FormatToken(c: T.Comment, _, _) =>
         Seq(Split(getMod(formatToken), 0))
 
-      case FormatToken(_: T.KwImplicit, _, _)
+      case FormatToken(soft.ImplicitOrUsing(), _, _)
           if !style.binPack.unsafeDefnSite &&
-            !style.verticalMultiline.atDefnSite =>
-        opensImplicitParamList(prevNonComment(prev(formatToken))).fold {
+            !style.verticalMultiline.atDefnSite && isRightImplicitOrUsingSoftKw(
```
kitbellew, line 1770: can you insert a newline after `&&`? that way, you are likely to get a more compact, single-line formatting for
```
isRightImplicitOrUsingSoftKw(prev(formatToken), soft) =>
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1762,10 +1765,19 @@ class Router(formatOps: FormatOps) {
...
+              prev(formatToken),
+              soft
+            ) =>
+        val argsOrParamsOpt = formatToken.meta.leftOwner match {
+          // for the using argument list
+          case _: ApplyUsing => Some(getApplyArgs(formatToken, false).args)
+          case _ =>
```
kitbellew, line 1777: is the newline needed?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
```diff
@@ -222,4 +222,14 @@ object TokenOps {
     }
   }
 
+  // For using to be the soft kw it also has to be preceded by paren
+  def isRightImplicitOrUsingSoftKw(
+      ft: FormatToken,
+      soft: SoftKeywordClasses
+  ): Boolean = ft match {
+    case FormatToken(_: LeftParen, soft.KwUsing(), _) => true
```
kitbellew, line 230: there is only one problem with this approach: this will not be recognized (and we need to add a test, too)
```
def foo( /* go scala3 */ using ... )
```
you might need to move this method into `FormatOps` instead and implement something like this:
```
  def isLeftImplicitOrUsingSoftKw(
      ft: FormatToken,
      soft: SoftKeywordClasses
  ): Boolean = ft.left match {
    case _: KwImplicit => true
    case soft.KwUsing() => prevNonComment(prev(ft)).is[LeftParen]
    case _ => false
  }
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -889,10 +890,12 @@ class FormatOps(
     }
   }
 
+  /** Works for `using` as well */
   def opensConfigStyleImplicitParamList(
       formatToken: FormatToken
   )(implicit style: ScalafmtConfig): Boolean =
-    formatToken.right.is[T.KwImplicit] &&
+    formatToken.right
```
kitbellew, line 897: do we need the line break before `.is` here?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1287,25 +1290,27 @@ class FormatOps(
       else tokens(nextNonComment(maybeArrow), 1)
     }(x => prevNonComment(tokens(x, -1)))
 
+  case class TreeArgs(tree: Tree, args: Seq[Tree])
```
kitbellew, line 1293: i'd probably define this class outside of `FormatOps` class, as it doesn't need to refer to the `FormatOps` instance; can put it inside companion `object FormatOps` if you'd like

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
MichalBednarz: Speaking of `ApplyUsing` there is no similar counterpart for the definition so I would stay with the paren check.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1779,4 +1784,19 @@ class FormatOps(
...
 
+  // For using to be the soft kw it also has to be preceded by paren
+  def isRightImplicitOrUsingSoftKw(
+      ft: FormatToken,
+      soft: SoftKeywordClasses
+  ): Boolean = ft.left match {
+    case _: T.KwImplicit => true
+    case soft.KwUsing() =>
+      prevNonComment(prev(ft)).left
+        .is[T.LeftParen] //revNonComment(prev(ft)).isInstanceOf[LeftParen]
```
kitbellew, line 1795: leftovers?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -759,8 +761,10 @@ class Router(formatOps: FormatOps) {
             Policy.NoPolicy
           }
 
+        // covers using as well
         val handleImplicit =
-          if (onlyConfigStyle) opensConfigStyleImplicitParamList(formatToken)
+          if (onlyConfigStyle)
```
kitbellew, line 766: restore original formatting, please

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1762,10 +1766,16 @@ class Router(formatOps: FormatOps) {
...
-      case FormatToken(_: T.KwImplicit, _, _)
+      case FormatToken(soft.ImplicitOrUsing(), _, _)
           if !style.binPack.unsafeDefnSite &&
-            !style.verticalMultiline.atDefnSite =>
-        opensImplicitParamList(prevNonComment(prev(formatToken))).fold {
+            !style.verticalMultiline.atDefnSite &&
+            isRightImplicitOrUsingSoftKw(formatToken, soft) =>
+        val argsOrParamsOpt = formatToken.meta.leftOwner match {
+          // for the using argument list
+          case _: ApplyUsing => Some(getApplyArgs(formatToken, false).args)
```
kitbellew, line 1775: here, you're checking for the owner of `using`. why can't the same be done in the newly added method you call just above?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenOps.scala
```diff
@@ -221,5 +221,4 @@ object TokenOps {
       Some((if (nonWs < 0) part.length else nonWs) - afterLastNL)
     }
   }
-
```
kitbellew, line 224: ?

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
MichalBednarz: the reason is that `isRightImplicitOrUsingSoftKw` returns true both for `using` at the definition site and at the call site. this match/case separates these because `using` def site doesn't have its `ApplyUsing` counterpart.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
MichalBednarz: yes, sorry for that

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: thank you, understood!

kitbellew (approved):

PR #2285, opened 2020-12-07
Scala 3: Add support to format open classes.

olafurpg:
This PR implements support to format open classes, which is a new soft
modifier in Scala 3. Details
https://dotty.epfl.ch/docs/reference/other-new-features/open-classes.html

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
@@ -87,10 +87,13 @@ object ScalafmtRunner {
     val scala213 = Scala213
       .withAllowTrailingCommas(true)
     val default = scala213
-      .withAllowOrTypes(true) // New feature in Dotty
-      .withAllowAndTypes(true) // New feature in Dotty
-      .withAllowTraitParameters(true) // New feature in Dotty
-      .withAllowColonForExtractorVarargs(true) // New feature in Dotty
+      // Supported Dotty features.
+      .withAllowOrTypes(true)
+      .withAllowAndTypes(true)
+      .withAllowTraitParameters(true)
+      .withAllowColonForExtractorVarargs(true)
+      .withAllowExportClause(true)
+      .withAllowOpenClass(true)
 
     implicit lazy val decoder: ConfDecoder[Dialect] = {
       ReaderUtil.oneOf[Dialect](
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
@@ -1,24 +1,34 @@
 package org.scalafmt.config
 
 import metaconfig._
+import scala.meta.classifiers.Classifier
+import scala.meta.Tree
+import scala.meta.Mod
+import sourcecode.Name
+import sourcecode.Text
 
 case class SortSettings(
     order: List[SortSettings.ModKey]
 )
 
 object SortSettings {
-
-  implicit val SortSettingsModKeyReader: ConfCodec[ModKey] =
-    ReaderUtil.oneOf[ModKey](
-      `implicit`,
-      `final`,
-      `sealed`,
-      `abstract`,
-      `override`,
-      `private`,
-      `protected`,
-      `lazy`
-    )
+  final case class ModKey(
+      name: String,
+      matches: Mod => Boolean
+  ) extends Product
+  private def createMod[T <: Mod](implicit
+      name: Name,
+      c: Classifier[Tree, T]
+  ): ModKey = ModKey(name.value, m => c(m))
+  val `private` = createMod[Mod.Private]
+  val `protected` = createMod[Mod.Protected]
+  val `final` = createMod[Mod.Final]
+  val `sealed` = createMod[Mod.Sealed]
+  val `abstract` = createMod[Mod.Abstract]
+  val `implicit` = createMod[Mod.Implicit]
+  val `override` = createMod[Mod.Override]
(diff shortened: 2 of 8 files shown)


tgodzik (approved):
LGTM! And we thought it would be a super quick feature :sweat_smile:

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
```diff
@@ -87,10 +87,13 @@ object ScalafmtRunner {
     val scala213 = Scala213
       .withAllowTrailingCommas(true)
     val default = scala213
-      .withAllowOrTypes(true) // New feature in Dotty
-      .withAllowAndTypes(true) // New feature in Dotty
-      .withAllowTraitParameters(true) // New feature in Dotty
-      .withAllowColonForExtractorVarargs(true) // New feature in Dotty
+      // Supported Dotty features.
```
kitbellew, line 90: unrelated question: for the other changes that @MichalBednarz and @tgodzik have been working on (like `using` and `enum`), do we need to add these properties?

I can see one for `.withAllowEnums`, and something about `.withAllowGivenUsing` (although i am not sure if that's the same as `using`).

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
```diff
@@ -1,24 +1,34 @@
...
-    )
+  final case class ModKey(
+      name: String,
+      matches: Mod => Boolean
+  ) extends Product
+  private def createMod[T <: Mod](implicit
+      name: Name,
+      c: Classifier[Tree, T]
+  ): ModKey = ModKey(name.value, m => c(m))
+  val `private` = createMod[Mod.Private]
```
kitbellew, line 23: i'd add an empty line before this

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
```diff
@@ -1,24 +1,34 @@
...
-      `private`,
-      `protected`,
-      `lazy`
-    )
+  final case class ModKey(
+      name: String,
+      matches: Mod => Boolean
+  ) extends Product
+  private def createMod[T <: Mod](implicit
+      name: Name,
```
kitbellew, line 20: also, question (on scala, probably): how does this work? why is there a unique implicit `Name` and `Classifier`?

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
```diff
@@ -1,24 +1,34 @@
...
-      `abstract`,
-      `override`,
-      `private`,
-      `protected`,
-      `lazy`
-    )
+  final case class ModKey(
+      name: String,
+      matches: Mod => Boolean
+  ) extends Product
```
kitbellew, line 18: a case class always extends `Product`, no?

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
```diff
@@ -1,24 +1,34 @@
...
+  private def createMod[T <: Mod](implicit
```
kitbellew, line 19: i'd add an empty line before this

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
```diff
@@ -1,24 +1,34 @@
...
+      name: Name,
+      c: Classifier[Tree, T]
+  ): ModKey = ModKey(name.value, m => c(m))
```
kitbellew, line 22: why not `c.apply`?

kitbellew (changes requested):
there's some scala magic happening here :) but a few questions still...

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
poslegm: there is [macro machinery](https://github.com/lihaoyi/sourcecode/blob/master/sourcecode/src-2/sourcecode/Macros.scala#L72-L77) around it

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
tgodzik: I think we will use the separate Dotty dialect for now (until parser is fully stable) and require users to add `runner.dialect`. I think @olafurpg already included it in another PR for macros.

olafurpg:
Superseded by #2286, where this commit https://github.com/scalameta/scalafmt/pull/2286/commits/6e0878fb8654e3cb64a97655e91bed3c032cfb0c addresses the comments from this PR

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
olafurpg: Correct, removed.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
olafurpg: Removed this method. Inserted a blank line after the class definition.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
olafurpg: No specific reason. I admit this method was unnecessarily complicated so I removed the method.

PR #2286, opened 2020-12-08
Scala 3: Implement support for quote and splice

olafurpg:
Builds on top of #2285. See commit history for details.

Scala 3 macros use a new syntax called "quote" `'expr` and "splice" `$expr`.
This PR implements support to format quotes and splices.

This PR includes a commit that requires users to specify `runner.dialect = Scala3` in order to format Scala 3 syntax. This change is related to quote and splice because enabling quote/splice parsing support introduces a regression for Scala 2 users https://github.com/scalameta/scalameta/issues/2191.

Towards #2216.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
@@ -94,6 +94,7 @@ object ScalafmtRunner {
       .withAllowColonForExtractorVarargs(true)
       .withAllowExportClause(true)
       .withAllowOpenClass(true)
+      .withAllowWhiteboxMacro(true)
 
     implicit lazy val decoder: ConfDecoder[Dialect] = {
       ReaderUtil.oneOf[Dialect](
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
@@ -87,14 +87,7 @@ object ScalafmtRunner {
     val scala213 = Scala213
       .withAllowTrailingCommas(true)
     val default = scala213
-      // Supported Dotty features.
-      .withAllowOrTypes(true)
-      .withAllowAndTypes(true)
-      .withAllowTraitParameters(true)
-      .withAllowColonForExtractorVarargs(true)
-      .withAllowExportClause(true)
-      .withAllowOpenClass(true)
-      .withAllowWhiteboxMacro(true)
+    val Scala3 = Dotty
 
     implicit lazy val decoder: ConfDecoder[Dialect] = {
       ReaderUtil.oneOf[Dialect](
@@ -105,6 +98,7 @@ object ScalafmtRunner {
         Sbt0137,
         Sbt1,
         Dotty,
+        Scala3,
         Paradise211,
         Paradise212
       )
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
@@ -87,10 +87,13 @@ object ScalafmtRunner {
     val scala213 = Scala213
       .withAllowTrailingCommas(true)
     val default = scala213
-      .withAllowOrTypes(true) // New feature in Dotty
-      .withAllowAndTypes(true) // New feature in Dotty
-      .withAllowTraitParameters(true) // New feature in Dotty
-      .withAllowColonForExtractorVarargs(true) // New feature in Dotty
+      // Supported Dotty features.
+      .withAllowOrTypes(true)
+      .withAllowAndTypes(true)
+      .withAllowTraitParameters(true)
+      .withAllowColonForExtractorVarargs(true)
+      .withAllowExportClause(true)
+      .withAllowOpenClass(true)
 
     implicit lazy val decoder: ConfDecoder[Dialect] = {
       ReaderUtil.oneOf[Dialect](
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
@@ -1,10 +1,7 @@
 package org.scalafmt.config
 
 import metaconfig._
-import scala.meta.classifiers.Classifier
-import scala.meta.Tree
 import scala.meta.Mod
-import sourcecode.Name
 import sourcecode.Text
 
 case class SortSettings(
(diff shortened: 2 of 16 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
```diff
@@ -1,24 +1,34 @@
...
-      `implicit`,
-      `final`,
-      `sealed`,
-      `abstract`,
-      `override`,
-      `private`,
-      `protected`,
-      `lazy`
-    )
+  final case class ModKey(
```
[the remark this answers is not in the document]
olafurpg: I based this PR on top of #2285 because they have conflicting changes in `build.sbt` and `ScalafmtRunner.scala`. I will try to separate Scala 3 PRs in the future. I suspect it will be easier to separate PRs once once the `Scala3` dialect changes in this PR are merged.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1075,7 +1076,9 @@ class Router(formatOps: FormatOps) {
             .withSingleLine(close, killOnFail = true)
         ) ++ oneArgPerLineSplits
 
-      case FormatToken(_, _: T.LeftBrace, _) =>
+      case FormatToken(_: T.MacroSplice | _: T.MacroQuote, _: T.LeftBrace, _) =>
+        Seq(Split(NoSplit, 0))
+      case FormatToken(left, _: T.LeftBrace, _) =>
```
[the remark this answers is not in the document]
olafurpg: Good catch, removed. This was leftovers from debugging.

olafurpg (commented):
Thank you for the review!

scalafmt-tests/src/test/resources/dotty/Enum.stat
```diff
@@ -1,3 +1,4 @@
+runner.dialect = Scala3
```
tgodzik, line 1: Dialect is inferred based on the fact that it's in the dotty directory. Similar to how it was done for ScalaJS. No need to set it here.

scalafmt-tests/src/test/resources/unit/Dialect.stat
```diff
@@ -17,11 +17,7 @@ val x: 42 = 42
 val x : A   &  B = 42
 >>>
 val x: A & B = 42
-<<< extractor
-val List(a : _ *) = 2
->>>
-val List(a: _*) = 2
-<<< trait parameters
+<<< trait parameters - supported in default dialect to produce consistent trees in Scala 2 and Scala 3.
```
kitbellew, line 20: is this related?

olafurpg: This is related to the dialect changes. This PR adds a requirement to add `runner.dialect = Scala3` in order to format Scala 3 syntax. Trait parameters are a new syntax in Scala 3 but they're enabled by default in Scalafmt. The purpose of the comment here was make this decision explicit, it's not accidental

scalafmt-tests/src/test/resources/dotty/Enum.stat
olafurpg: Nice, removed.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
```diff
@@ -87,10 +87,7 @@ object ScalafmtRunner {
     val scala213 = Scala213
       .withAllowTrailingCommas(true)
     val default = scala213
-      .withAllowOrTypes(true) // New feature in Dotty
-      .withAllowAndTypes(true) // New feature in Dotty
-      .withAllowTraitParameters(true) // New feature in Dotty
-      .withAllowColonForExtractorVarargs(true) // New feature in Dotty
+    val Scala3 = Dotty
```
tgodzik, line 90: Just a side note - we might never be able to cover all dialects with scalafmt. Scala 3 has the new keywords such as `given`, which would break the Scala 2 code.

kitbellew: @tgodzik can you explain this, please? i thought that `given` wouldn't be parsed unless the dialect has the flag confirming support for this keyword. how would that break scala2?

kitbellew (approved):

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
tgodzik: Sorry! I just meant that if we would enable everything in the default dialect this would be an issue. Having two separate dialect is the way to go. Dialect for Scala 2 has givens disabled and tokenizes it as a normal ident, while Scala 3 dialect will think it's a keyword. And because of that we cannot have a common dialect for both.

kitbellew: ah, naturally! and in cases of a mixed codebase, we have `fileOverride` configuration parameter.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -527,7 +527,8 @@ class Router(formatOps: FormatOps) {
       // Opening ( with no leading space.
       case FormatToken(
             RightParenOrBracket() | T.KwSuper() | T.KwThis() | T.Ident(_) |
-            T.RightBrace() | T.Underscore(),
+            T.RightBrace() | T.Underscore() | T.MacroQuotedIdent(_) |
+            T.MacroSplicedIdent(_),
```
kitbellew, line 531: so this was rolled back, right?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1075,6 +1076,8 @@ class Router(formatOps: FormatOps) {
             .withSingleLine(close, killOnFail = true)
         ) ++ oneArgPerLineSplits
 
+      case FormatToken(_: T.MacroSplice | _: T.MacroQuote, _: T.LeftBrace, _) =>
```
kitbellew, line 1079: so is this the only splice/quote change? the rest of the code in this pr seems to be about modifiers.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: That's right. We remove the additional Idents, because `$a` is actually a valid name.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: I think this might be it, yeah :sweat_smile:

PR #2292, opened 2020-12-14
Fix Macros after fixes in scalameta

tgodzik:
We recently merged the support for Macros together with an update to scalameta, which caused the master to stop compiling due to removal of two token types in scalameta. This should fix the issue. 

I also needed to fix some tests, since `'identifier` is only allowed within splices or quotes.


olafurpg (approved):

kitbellew:
> I also needed to fix some tests, since `'identifier` is only allowed within splices or quotes.

that's in dotty only, correct? symbol literals are still supported for scala2?

tgodzik:
Yes, they will work correctly.

Issue #2288, opened 2020-12-08
Failing build on Windows computer


olafurpg:
Running `sbt tests/test` fails on Windows with the following error
```
[info] ScalaTest
[info] Run completed in 1 minute, 23 seconds.
[info] Total number of tests run: 78
[info] Suites: completed 10, aborted 0
[info] Tests: succeeded 77, failed 1, canceled 0, ignored 0, pending 0
[info] *** 1 TEST FAILED ***
[error] Failed: Total 2864, Failed 112, Errors 0, Passed 2752, Ignored 7
[error] Failed tests:
[error]         org.scalafmt.EmptyFileTest
[error]         org.scalafmt.cli.CliTest
[error]         org.scalafmt.util.GitOpsTest
[error]         org.scalafmt.FormatTests
[error] (tests / Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 108 s (01:48), completed 8 d├®c. 2020 13:51:55
```

Reported by @sjrd. The Scalafmt CI ran on Windows via Appveyor at some point but it got removed. It should be fairly easy to setup the CI on Windows again with GitHub Actions.

PR #2293, opened 2020-12-14
Add tests for windows

tgodzik:
Fixes https://github.com/scalameta/scalafmt/issues/2288

Additionally, to help with some [cryptic errors](https://github.com/scalatest/scalatest/issues/556) I got back to munit (reverted https://github.com/scalameta/scalafmt/pull/1697). The support in Intellij is much better now, so I think it's worth keeeping. There was a recent release with some nice changes in Intellij - https://blog.jetbrains.com/scala/2020/12/01/intellij-scala-plugin-2020-3/#munit-support

What do you think?

I commented on parts that are related to windows failures.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+* eol=lf
+*.png eol=autocrlf
--- a/build.sbt
+++ b/build.sbt
@@ -213,6 +213,7 @@ lazy val tests = project
       scalatest.value
     ),
     scalacOptions ++= scalacJvmOptions.value,
+    javaOptions += "-Dfile.encoding=UTF8",
     // Fork in CI to avoid memory limitation issues. Disable forking locally
     // because ScalaTest error reporting fails with cryptic serialization errors
     // when forking is enabled.
--- a/scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
+++ b/scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
@@ -34,6 +34,13 @@ abstract class AbstractCliTest extends AnyFunSuite with DiffAssertions {
     Cli.getConfig(args, baseCliOptions).get
   }
 
+  def assertContains(out: String, expected: String) = {
+    assert(
+      out.replace("\r\n", "\n").contains(expected.replace("\r\n", "\n")),
+      out + "\n should have contained \n" + expected
+    )
+  }
+
   val unformatted = """
     |object a    extends   App {
     |pr("h")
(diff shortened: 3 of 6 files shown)


scalafmt-tests/src/test/scala/org/scalafmt/util/DiffTest.scala
```diff
@@ -1,17 +1,27 @@
...
-    loc: Position,
+    loc: Location,
     original: String,
     expected: String,
     skip: Boolean,
     only: Boolean,
     style: ScalafmtConfig
 ) {
-  val fullName = s"${loc.fileName}:${loc.lineNumber}: $name"
+  val file = DiffTest.testDir.relativize(Paths.get(loc.path)).toString()
```
tgodzik, line 19: This is needed to have test names like `vertical-multiline/beforeImplicitKW.stat`

.gitattributes
```diff
@@ -0,0 +1,2 @@
+* eol=lf
+*.png eol=autocrlf
```
kitbellew, line 2: LF missing?

build.sbt
```diff
@@ -36,10 +36,11 @@ inThisBuild(
     crossScalaVersions := List(scala213, scala212, scala211),
     resolvers += Resolver.sonatypeRepo("releases"),
     libraryDependencies ++= List(
-      scalatest.value % Test,
```
kitbellew, line 39: wait a minute... are you switching to munit, from scalatest? can we support windows without that?

we tried it before, intellij doesn't run tests anymore.

kitbellew (changes requested):
if you really need to add munit, let's have that as a completely separate pr, and not hide behind adding tests for windows.

but we tried it before, and it didn't work:

https://github.com/scalameta/scalafmt/pull/1681
https://github.com/scalameta/scalafmt/pull/1697

build.sbt
tgodzik: I switched to munit, because it was not possible to read the logs in the CI :/ 

Intellij should work now: https://blog.jetbrains.com/scala/2020/12/01/intellij-scala-plugin-2020-3/#munit-support

I can remove it and keep the changes for the Windows job only.

tgodzik:
> if you really need to add munit, let's have that as a completely separate pr, and not hide behind adding tests for windows.
> 
> but we tried it before, and it didn't work:
> 
> #1681
> #1697

Sure, I will put it into a separate PR. It just helped me fix the other issues.

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
```diff
@@ -34,6 +34,13 @@ abstract class AbstractCliTest extends AnyFunSuite with DiffAssertions {
     Cli.getConfig(args, baseCliOptions).get
   }
 
+  def assertContains(out: String, expected: String) = {
+    assert(
+      out.replace("\r\n", "\n").contains(expected.replace("\r\n", "\n")),
+      out + "\n shoudld have contained \n" + expected
```
kitbellew, line 40: typo

kitbellew (changes requested):
thank you for simplifying?

kitbellew (approved):

.gitattributes
tgodzik: I just copied over https://github.com/scalameta/munit/blob/master/.gitattributes

PR #2294, opened 2020-12-15
Add munit test framework to the codebase

tgodzik:
This reverts commit 07403e1e5fd1e76ca3002be92d3b67619f889f2f.

Previously, in https://github.com/scalameta/scalafmt/pull/1697 munit was disabled because of the lacking support in Intellij. Currently, both issues mentioned in the PR are resolved and the support is much improved:

https://blog.jetbrains.com/scala/2020/12/01/intellij-scala-plugin-2020-3/#munit-support

I think it would be useful to add it back as it will for example simplify running a single tests in FormatTests:

```
sbt:scalafmtRoot> ~tests/testOnly org.scalafmt.FormatTests -- *TypeWith*
[warn] javaOptions will be ignored, fork is set to false
org.scalafmt.FormatTests:
  + default/TypeWith.stat:2: with                                         | 0.347s
  + default/TypeWith.stat:12: #1133                                       | 0.069s
  + default/TypeWith.stat:22: with-chain types #1125                      | 0.029s
  + default/TypeWith.stat:33: abstract def #1125                          | 0.089s
  + default/TypeWith.stat:48: class param #1125                           | 0.033s
  + default/TypeWith.stat:63: def param #1125                             | 0.031s
  + default/TypeWith.stat:74: def param negative #1125                    | 0.024s
FormatTests:86 Total explored: 827
[info] Passed: Total 7, Failed 0, Errors 0, Passed 7
[success] Total time: 2 s, completed Dec 15, 2020 10:02:52 AM
[info] 1. Monitoring source files for scalafmt/tests/testOnly org.scalafmt.FormatTests -- *TypeWith*...
[info]    Press <enter> to interrupt or '?' for more options.
```


scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
```diff
@@ -2,72 +2,72 @@ package org.scalafmt.util
...
-  override type FixtureParam = GitOpsImpl
-
   // DESNOTE(2017-08-16, pjrt): Create a temporary git directory for each
   // test.
-  override def withFixture(test: OneArgTest): Outcome = {
-    val f = Files.createTempDirectory(dirName)
-    val absFile = AbsoluteFile.fromPath(f.toString).get
-    val ops = new GitOpsImpl(absFile)
-    init(ops)
+  private implicit var gitOps: GitOpsImpl = _
```
kitbellew, line 22: why not call this `ops`, just like before, and vastly reduce the size of this patch?

scalafmt-tests/src/main/scala/org/scalafmt/util/DiffAssertions.scala
```diff
@@ -1,122 +0,0 @@
...
-      obtained: String,
-      diff: String
-  )(implicit pos: Position)
-      extends TestFailedException(
-        (_: StackDepthException) =>
-          Some(title + "\n" + error2message(obtained, diff)),
-        None: Option[Throwable],
-        pos
-      ) {
-    /* fool Intellij, prepend a stack trace element pointing to the test case
```
kitbellew, line 29: this thing was very important. we need to achieve the same with `munit` somehow.

kitbellew (changes requested):
unfortunately, the biggest problem i have with this implementation is that when one of the `xxx.stat` tests fails (in Intellij), the error message no longer clicks through to the location of the `.stat` file (the stack trace simply contains the useless location of the `FormatTests` file). if you can fix that problem, we can move forward.

scalafmt-tests/src/test/scala/org/scalafmt/util/DiffTest.scala
```diff
@@ -1,17 +1,27 @@
...
-    loc: Position,
+    loc: Location,
     original: String,
     expected: String,
     skip: Boolean,
     only: Boolean,
     style: ScalafmtConfig
 ) {
-  val fullName = s"${loc.fileName}:${loc.lineNumber}: $name"
+  val file = DiffTest.testDir.relativize(Paths.get(loc.path)).toString()
```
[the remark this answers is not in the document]
tgodzik: I added the parent of the stat file so that we have the same name as previously. `binPack/LiteralList.stat` instead of just `LiteralList.stat`

scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
tgodzik: Sure, it really came from the revert.

scalafmt-tests/src/main/scala/org/scalafmt/util/DiffAssertions.scala
tgodzik: I will check it out, I was using VS Code and it has a nice functionality where it tries to guess the location of the file, based on the name. Works pretty well, wonder why Intellij does not allow for anything like this.

olafurpg (approved):
@tgodzik Thank you for taking this initiative! LGTM 👍 

I took the liberty to push a commit that points the location of the assertion errors to the original `*.stat` file so that you can cmd+click on the filename in the IntelliJ terminal.

A test failure looks like this when run from IntelliJ

Screenshot 2020-12-15 at 17 14 46

The "Click to see differences" button opens a GUI diff viewer that makes it easier to scrutinize the diff

Screenshot 2020-12-15 at 17 14 54

@kitbellew Do these changes address your concerns about migrating to MUnit from ScalaTest? Please let me know if there's anything you discover is missing from the original ScalaTest setup.

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
```diff
@@ -78,7 +75,7 @@ trait HasTests extends FormatAssertions {
     val sep =
       if (content.contains(System.lineSeparator)) System.lineSeparator
       else "\n"
-    val spec = filename.stripPrefix(testDir + File.separator)
+    val spec = filename.stripPrefix(BuildInfo.resourceDirectory.toURI.toString)
```
kitbellew, line 78: it will only work if `toURI` ends with `/` (or `\` on Windows), right? to have a relative name.

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
```diff
@@ -43,9 +43,6 @@ trait HasTests extends FormatAssertions {
     )
 
   lazy val debugResults = mutable.ArrayBuilder.make[Result]
-  val testDir = new File(
```
kitbellew, line 46: `ManualTests` and `UnitTests` is failing for me now.

kitbellew (changes requested):

kitbellew:
> I took the liberty to push a commit that points the location of the assertion errors to the original `*.stat` file so that you can cmd+click on the filename in the IntelliJ terminal.

@olafurpg this looks really good (better than the stacktrace hack i had before). can you explain how it works? :)

olafurpg:
As soon as I pushed the last commit, I realized the tests were failing 😓 I pushed a new commit that fixes the issue for me when I test locally.

>  can you explain how it works? :)

MUnit assertions accept an `munit.Location` argument that is used to pretty-print a helpful error message with the full filename + line number of the call site. By default, the `Location` argument is generated by macros at compile-time but in this PR I tweaked the `Location` to point to the `*.stat` file

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
olafurpg: The logic here was buggy. I refactored it to use only `java.nio.file.Path`

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
olafurpg: This should be fixed now.

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
```diff
@@ -90,15 +96,10 @@ trait HasTests extends FormatAssertions {
               |$c""".stripMargin
           )
       }
-    val style: ScalafmtConfig = {
-      val pathPatternToReplace =
-        if (OsSpecific.isWindows) s"""\\\\.*""" else "/.*"
```
kitbellew, line 98: is this because `BuildInfo` uses `/` regardless of OS?

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
```diff
@@ -72,10 +74,14 @@ trait HasTests extends FormatAssertions {
   def extension(filename: String): String = filename.replaceAll(".*\\.", "")
 
   def parseDiffTests(content: String, filename: String): Seq[DiffTest] = {
-    val sep =
+    val sep = {
```
kitbellew, line 77: optional: i think the `{}` is not needed here

kitbellew (commented):
on my mac, in 2020.2.3, the link is not clickable. let me upgrade to 2020.2.4, to try.

kitbellew:
> on my mac, in 2020.2.3, the link is not clickable. let me upgrade to 2020.2.4, to try.

2020.2.4 also doesn't seem to work. upgrading to 2020.3.

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
olafurpg: Whoops, removed. Remnants from some debugging session. I don't mind enabling more aggressive Scalafmt options if you have preferences

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
olafurpg: I refactored the code to use `nio.file.Path.relativize(..).getName(0)` to avoid dealing with regular expressions and OS-specific file separators.

olafurpg (commented):
The IntelliJ MUnit support is only available in 2020.3 (not 2020.2). When you run a test you now have two options, one for vanilla JUnit and one for MUnit (Scala logo background), see screenshot

Screenshot 2020-12-15 at 17 52 52

The run configuration should be of type "MUnit"

scalafmt-tests/src/test/scala/org/scalafmt/util/HasTests.scala
kitbellew: i do not, really. it's just our general approach to code changes is to reduce the amount of red and green, to make reviewing easier :)

kitbellew:
> The IntelliJ MUnit support is only available in 2020.3 (not 2020.2). When you run a test you now have two options, one for vanilla JUnit and one for MUnit (Scala logo background), see screenshot
...
> The run configuration should be of type "MUnit"

Upgraded to 2020.3, used run configuration of `MUnit`. I must be doing something wrong, because it's not clickable still.

Screen Shot 2020-12-15 at 09 12 37
Screen Shot 2020-12-15 at 09 11 59
Screen Shot 2020-12-15 at 09 12 09

olafurpg:
Hmm, I'm not sure why the filename is not clickable for you. I suspect there's a setting to enable it but I couldn't find anything. I tried to disable "Highlight hyperlinks" under "Tools > Terminal" but it didn't have an effect. Do you have any plugins that affect the terminal? I have a mostly vanilla installation, only install the IdeaVim and Scala plugins.

Screenshot 2020-12-16 at 11 01 47

Screenshot 2020-12-16 at 10 57 07

I'm using the community edition but it shouldn't make a difference.
Screenshot 2020-12-16 at 10 56 29

unkarjedy:
> because it's not clickable still

@kitbellew @olafurpg 
Could you please tell whether you have Android plugin enabled or disabled? 

I found out that this "file link" behaviour comes from this plugin (which is strange itself).
Probably I should add a dedicated logic to test console in our test framework support not to rely on Android plugin.

olafurpg:
I can confirm that disabling the Android plugin removes the "file link" behavior.

kitbellew:
> I've created an issue in YouTrack to clean/unify up this behaviour: https://youtrack.jetbrains.com/issue/IDEA-257969.
> As a workaround, you can enable Android plugin, if it doesn't hurt you much :)
> As another workaround, I could probably add a dedicated logic to test console in our test framework support not to rely on the Android plugin, but it will require some time.

@unkarjedy thank you for helping us with this mystery, and thank you for offering to add it to main application.

kitbellew (approved):
in light of @unkarjedy 's discovery and proposed workarounds, I'm ok moving forward.

@olafurpg @tgodzik perhaps we can document the workarounds somewhere, such as where we explain how to contribute to scalafmt and format the `x.stat` tests.

project/Dependencies.scala
```diff
@@ -21,10 +21,7 @@ object Dependencies {
   val scalametaTestkit = "org.scalameta" %% "testkit" % scalametaV
 
   val scalacheck = "org.scalacheck" %% "scalacheck" % scalacheckV
-  // NOTE(olafur): this repo uses only ScalaTest FunSuite with asserts. Matchers
-  // are intentionally not included in the build because they rely on "multiple
-  // infix" syntax, which is deprecated in the latest Scala version.
-  val scalatest = Def.setting("org.scalatest" %%% "scalatest-funsuite" % scalatestV)
+  val munit      = Def.setting("org.scalameta" %%% "munit" % munitV)
```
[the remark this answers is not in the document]
tgodzik: Weirdly enough, that's actually formatted by scalafmt like, so it will fail the format check if I remove them.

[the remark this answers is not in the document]
kitbellew: Perhaps we do have some form of alignment after all.

tgodzik:
> in light of @unkarjedy 's discovery and proposed workarounds, I'm ok moving forward.
> 
> @olafurpg @tgodzik perhaps we can document the workarounds somewhere, such as where we explain how to contribute to scalafmt and format the `x.stat` tests.

Do you mean adding information about how to get the links (via android plugin) ? I guess we could add it to https://scalameta.org/scalafmt/docs/contributing-scalafmt.html , what do you think?  I also commented in the code on how this is actually achieved.

scalafmt-tests/src/test/scala/org/scalafmt/util/GitOpsTest.scala
tgodzik: Done!

kitbellew:
> > in light of @unkarjedy 's discovery and proposed workarounds, I'm ok moving forward.
> > @olafurpg @tgodzik perhaps we can document the workarounds somewhere, such as where we explain how to contribute to scalafmt and format the `x.stat` tests.
> 
> Do you mean adding information about how to get the links (via android plugin) ? I guess we could add it to https://scalameta.org/scalafmt/docs/contributing-scalafmt.html , what do you think? I also commented in the code on how this is actually achieved.

yes, something like that.

kitbellew:
> > > in light of @unkarjedy 's discovery and proposed workarounds, I'm ok moving forward.
> > > @olafurpg @tgodzik perhaps we can document the workarounds somewhere, such as where we explain how to contribute to scalafmt and format the `x.stat` tests.
> > 
> > 
> > Do you mean adding information about how to get the links (via android plugin) ? I guess we could add it to https://scalameta.org/scalafmt/docs/contributing-scalafmt.html , what do you think? I also commented in the code on how this is actually achieved.
> 
> yes, something like that.

actually, maybe we should add it to https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/resources/readme.md and mention it in `contributing-scalafmt.html`.

tgodzik:
> https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/resources/readme.md

Added some additional comments in https://github.com/scalameta/scalafmt/pull/2294/commits/90c8c939718dce83fcf893b9085de90d0c2ce8c1 What do you think @kitbellew  ?

kitbellew (approved):

PR #2295, opened 2020-12-16
[Scala 3] Add tests for formatting trait parameters, local type declaration and new type wildcard symbol

tgodzik:
~~Scala 3 added support for specifying trait constructor parameters, which seems to work out of the box. I added some tests just in case.~~

~~EDIT: Seems that it's also working fine in case of local type declarations, that can now be defined inside a local block. Added tests for that also.~~ 

It seems that a number of new features work out of the box in scalafmt so I added tests that check:
- local type declaration
- new wildcard symbol `?`
- trait parameters

I decided to put everything into single PR instead of each in a separate one, because it seems all those are small features/changes.


scalafmt-tests/src/test/resources/dotty/Type.stat
```diff
@@ -0,0 +1,44 @@
...
+}
+>>>
+def hello() = {
+  type T[
+      AAAAAAAAAAAAA <: BBBBBBBBBBBBBB[
+        ?
+      ]
+  ]
+  val a: T
+}
```
kitbellew, line 44: missing LF

olafurpg (approved):
LGTM 👍 This is safe to merge IMO given it has no code changes

kitbellew (changes requested):
I looked over the codebase looking to make sure that (`Token.KwTrait`, `Token.KwClass`) and (`Defn.Class`, `Defn.Trait`) are always used together.

There is at least example when this is not true: `FormatOps.verticalMultiline`.

scalafmt-tests/src/test/resources/dotty/Type.stat
tgodzik: Fixed!

tgodzik:
> I looked over the codebase looking to make sure that (`Token.KwTrait`, `Token.KwClass`) and (`Defn.Class`, `Defn.Trait`) are always used together.
> 
> There is at least example when this is not true: `FormatOps.verticalMultiline`.

I added it there and added some more tests for verticalMultiline. I haven't really been able to find a test case that would fail without the change though.

kitbellew (approved):

PR #2298, opened 2020-12-21
Scala 3: add support for extension methods

olafurpg:
Previously, Scalafmt didn't support formatting extension methods.
This commit fixes that issue for code using braces. There's still no
support for indentation syntax (which is still being designed and is
likely to change soon).

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/DanglingParentheses.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/DanglingParentheses.scala
@@ -36,10 +36,11 @@ object DanglingParentheses {
     case object `class` extends Exclude
     case object `trait` extends Exclude
     case object `enum` extends Exclude
+    case object `extension` extends Exclude
     case object `def` extends Exclude
 
     implicit val reader: ConfCodec[Exclude] =
-      ReaderUtil.oneOf[Exclude](`class`, `trait`, `enum`, `def`)
+      ReaderUtil.oneOf[Exclude](`class`, `trait`, `enum`, `extension`, `def`)
   }
 
 }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -466,6 +466,14 @@ class Router(formatOps: FormatOps) {
           Split(NewlineT(isDouble = tok.hasBlankLine), 0)
         )
 
+      case tok @ FormatToken(_: T.RightParen, _: T.KwDef, _)
+          if leftOwner.is[Defn.ExtensionGroup] =>
+        // Force a newline with the body of a Defn.ExtensionGroup if the
+        // method doesn't fit on a single line.
+        Seq(
+          Split(Space, 0).withSingleLine(leftOwner.tokens.last),
+          Split(Newline, 1).withIndent(2, leftOwner.tokens.last, After)
+        )
       case tok @ FormatToken(left, right, _)
           if startsStatement(right).isDefined =>
         val expire = rightOwner.tokens
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -466,14 +466,16 @@ class Router(formatOps: FormatOps) {
           Split(NewlineT(isDouble = tok.hasBlankLine), 0)
         )
 
-      case tok @ FormatToken(_: T.RightParen, _: T.KwDef, _)
+      case tok @ FormatToken(_: T.RightParen, _: T.KwDef | _: T.Comment, _)
           if leftOwner.is[Defn.ExtensionGroup] =>
+        val expireToken = leftOwner.tokens.last
         // Force a newline with the body of a Defn.ExtensionGroup if the
         // method doesn't fit on a single line.
         Seq(
-          Split(Space, 0).withSingleLine(leftOwner.tokens.last),
-          Split(Newline, 1).withIndent(2, leftOwner.tokens.last, After)
+          Split(Space, 0).withSingleLine(expireToken),
+          Split(Newline, 1).withIndent(2, expireToken, After)
         )
+
       case tok @ FormatToken(left, right, _)
           if startsStatement(right).isDefined =>
         val expire = rightOwner.tokens
@@ -560,6 +562,9 @@ class Router(formatOps: FormatOps) {
           case name: Term.Name
               if style.spaces.afterSymbolicDefs &&
                 isSymbolicName(name.value) && name.parent.exists(isDefDef) =>
+            Space
+          case _: Defn.ExtensionGroup
+              if formatToken.left.is[soft.KwExtension] =>
             Space
           case _ => NoSplit
         }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenClasses.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenClasses.scala
@@ -7,16 +7,6 @@ import scala.meta.tokens.Token
 import scala.meta.tokens.Token._
 
 @classifier
-trait ExtensionKeyword
-object ExtensionKeyword {
-  def unapply(token: Token): Boolean = token match {
-    case _: KwExtension => true
-    case Ident("extension") => true
-    case _ => false
-  }
-}
-
-@classifier
 trait Keyword
 object Keyword {
   def unapply(token: Token): Boolean = {
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenClasses.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenClasses.scala
@@ -7,6 +7,16 @@ import scala.meta.tokens.Token
 import scala.meta.tokens.Token._
 
 @classifier
+trait ExtensionKeyword
+object ExtensionKeyword {
+  def unapply(token: Token): Boolean = token match {
+    case _: KwExtension => true
+    case Ident("extension") => true
+    case _ => false
+  }
+}
+
+@classifier
 trait Keyword
 object Keyword {
   def unapply(token: Token): Boolean = {
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
@@ -114,7 +114,12 @@ object TreeOps {
       t.tokens.find(!_.is[Trivia]).foreach(addTok(_, tree))
     def addAll(trees: Seq[Tree]) = trees.foreach(x => addTree(x, x))
 
-    def addDefn[T: ClassTag](mods: Seq[Mod], tree: Tree): Unit = {
+    def addDefnTokens[T](
+        mods: Seq[Mod],
+        tree: Tree,
+        what: String,
+        isMatch: Token => Boolean
+    ): Unit = {
       // Each @annotation gets a separate line
       val annotations = mods.filter(_.is[Mod.Annot])
       addAll(annotations)
@@ -123,11 +128,20 @@ object TreeOps {
         case Some(x) => addTree(x, tree)
         case _ =>
           // No non-annotation modifier exists, fallback to keyword like `object`
-          tree.tokens.find(classTag[T].runtimeClass.isInstance) match {
+          tree.tokens.find(isMatch) match {
             case Some(x) => addTok(x, tree)
-            case None => throw Error.CantFindDefnToken[T](tree)
+            case None =>
+              throw Error.CantFindDefnToken(what, tree)
           }
       }
+    }
+    def addDefn[T: ClassTag](mods: Seq[Mod], tree: Tree): Unit = {
+      addDefnTokens[T](
+        mods,
+        tree,
+        classTag[T].runtimeClass.getSimpleName(),
+        t => classTag[T].runtimeClass.isInstance(t)
+      )
     }
 
     def loop(x: Tree): Unit = {
@@ -137,6 +151,8 @@ object TreeOps {
         case t: Defn.Macro => addDefn[KwDef](t.mods, t)
         case t: Decl.Def => addDefn[KwDef](t.mods, t)
         case t: Defn.Enum => addDefn[KwEnum](t.mods, t)
+        case t: Defn.ExtensionGroup =>
+          addDefnTokens(Nil, t, "extension", t => ExtensionKeyword.unapply(t))
         case t: Defn.Object => addDefn[KwObject](t.mods, t)
         case t: Defn.Trait => addDefn[KwTrait](t.mods, t)
         case t: Defn.Type => addDefn[KwType](t.mods, t)
@@ -299,7 +315,8 @@ object TreeOps {
   def isDefnSiteWithParams(tree: Tree): Boolean =
     tree match {
       case _: Decl.Def | _: Defn.Def | _: Defn.Macro | _: Defn.Class |
-          _: Defn.Trait | _: Defn.Enum | _: Defn.EnumCase | _: Ctor.Secondary =>
+          _: Defn.Trait | _: Defn.Enum | _: Defn.EnumCase |
+          _: Defn.ExtensionGroup | _: Ctor.Secondary =>
         true
       case x: Ctor.Primary =>
         x.parent.exists(isDefnSiteWithParams)
@@ -316,7 +333,7 @@ object TreeOps {
       case _: Decl.Def | _: Defn.Def | _: Defn.Macro | _: Defn.Class |
           _: Defn.Trait | _: Ctor.Secondary | _: Decl.Type | _: Defn.Type |
           _: Type.Apply | _: Type.Param | _: Type.Tuple | _: Defn.Enum |
-          _: Defn.EnumCase =>
+          _: Defn.EnumCase | _: Defn.ExtensionGroup =>
         true
       case _: Term.Function | _: Type.Function => true
       case x: Ctor.Primary => x.parent.exists(isDefnSite)
@@ -419,6 +436,8 @@ object TreeOps {
     case t: Defn.Trait => (t.mods, t.name, t.tparams, t.ctor.paramss)
     case t: Defn.Enum => (t.mods, t.name, t.tparams, t.ctor.paramss)
     case t: Defn.EnumCase => (t.mods, t.name, t.tparams, t.ctor.paramss)
+    case t: Defn.ExtensionGroup =>
+      (Nil, Name.Anonymous(), t.tparams, List(t.eparam) :: t.uparams)
     case t: Ctor.Primary => (t.mods, t.name, Seq.empty, t.paramss)
     case t: Ctor.Secondary => (t.mods, t.name, Seq.empty, t.paramss)
   }
@@ -581,6 +600,7 @@ object TreeOps {
             DanglingParentheses.Exclude.`class`
           case _: Defn.Trait => DanglingParentheses.Exclude.`trait`
           case _: Defn.Enum => DanglingParentheses.Exclude.`enum`
+          case _: Defn.ExtensionGroup => DanglingParentheses.Exclude.`extension`
           case _: Defn.Def => DanglingParentheses.Exclude.`def`
           case _ => null
         }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
@@ -24,7 +24,6 @@ import scala.meta.tokens.Token
 import scala.meta.tokens.Token._
 import scala.meta.tokens.Tokens
 import scala.reflect.ClassTag
-import scala.reflect.classTag
 
 import org.scalafmt.CompatCollections.JavaConverters._
 import org.scalafmt.Error
@@ -105,7 +104,10 @@ object TreeOps {
     ret.result()
   }
 
-  def getStatementStarts(tree: Tree): Map[TokenHash, Tree] = {
+  def getStatementStarts(
+      tree: Tree,
+      soft: SoftKeywordClasses
+  ): Map[TokenHash, Tree] = {
     val ret = Map.newBuilder[TokenHash, Tree]
     ret.sizeHint(tree.tokens.length)
 
@@ -114,7 +116,7 @@ object TreeOps {
       t.tokens.find(!_.is[Trivia]).foreach(addTok(_, tree))
     def addAll(trees: Seq[Tree]) = trees.foreach(x => addTree(x, x))
 
-    def addDefnTokens[T](
+    def addDefnTokens(
         mods: Seq[Mod],
         tree: Tree,
         what: String,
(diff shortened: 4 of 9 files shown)


scalafmt-tests/src/test/scala/org/scalafmt/FormatTests.scala
```diff
@@ -54,6 +54,9 @@ class FormatTests extends FunSuite with CanRunTests with FormatAssertions {
         e.partialOutput
       case x => x.get
     }
+    if (!onlyManual) {
+      assertEquals(obtained, t.expected, "Obtained formatting output does not match the expected formatting output" )
```
olafurpg, line 58: `assertEquals` produces copy-pasteable output instead of `.stripMargin` formatted multiline strings from `assertNoDiff`

project/Dependencies.scala
```diff
@@ -6,7 +6,7 @@ import org.portablescala.sbtplatformdeps.PlatformDepsPlugin.autoImport._
 
 object Dependencies {
   val metaconfigV = "0.9.10"
-  val scalametaV  = "4.4.2"
+  val scalametaV  = "4.4.3"
```
olafurpg, line 9: This release includes an important fix for positions

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -464,6 +464,14 @@ class Router(formatOps: FormatOps) {
           Split(NewlineT(isDouble = tok.hasBlankLine), 0)
         )
 
+      case tok @ FormatToken(_: T.RightParen, _: T.KwDef, _)
+          if leftOwner.is[Defn.ExtensionGroup] =>
+        // Force a newline with the body of a Defn.ExtensionGroup if the
+        // method doesn't fit on a single line.
+        Seq(
+          Split(Space, 0).withSingleLine(leftOwner.tokens.last),
+          Split(Newline, 1).withIndent(2, leftOwner.tokens.last, After)
```
olafurpg, line 473: The 2 space indentation is hardcoded in several places so I hardcoded the `2` literal here as well. We may need to make this indentation configuration down the road with significant indentation syntax.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenClasses.scala
```diff
@@ -4,6 +4,16 @@ import scala.meta.internal.classifiers.classifier
 import scala.meta.tokens.Token
 import scala.meta.tokens.Token._
 
+@classifier
+trait ExtensionKeyword
+object ExtensionKeyword {
+  def unapply(token: Token): Boolean = token match {
+    case _: KwExtension => true
+    case Ident("extension") => true
```
tgodzik, line 12: So `extension` is still a soft keyword, but it was confirmed that is should be changed as it's impossible distinguish. I reopened the issue in Dotty, since it was closed and plainly ignored https://github.com/lampepfl/dotty/issues/10076

In case it becomes a keyword this case should not be needed and currently scalameta always parses extension as a keyword.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -137,6 +151,8 @@ object TreeOps {
         case t: Defn.Macro => addDefn[KwDef](t.mods, t)
         case t: Decl.Def => addDefn[KwDef](t.mods, t)
         case t: Defn.Enum => addDefn[KwEnum](t.mods, t)
+        case t: Defn.ExtensionGroup =>
+          addDefnTokens(Nil, t, "extension", t => ExtensionKeyword.unapply(t))
```
tgodzik, line 155: Why not just `addDefn[KwExtension](Nil, t)` ?

tgodzik (changes requested):
I think we should currently assume that extension is a keyword. We can change it back in the future, but that's how it will be currently tokenized for the Dotty dialect.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -464,6 +464,14 @@ class Router(formatOps: FormatOps) {
           Split(NewlineT(isDouble = tok.hasBlankLine), 0)
         )
 
+      case tok @ FormatToken(_: T.RightParen, _: T.KwDef, _)
```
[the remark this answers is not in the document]
kitbellew: one other thing: what about
```
extension [A](...) /* new feature */ // or this
  def ...
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: i think significant is only important for parsing. for formatting, you can fix it to 2, no?

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -464,6 +464,14 @@ class Router(formatOps: FormatOps) {
...
+          if leftOwner.is[Defn.ExtensionGroup] =>
```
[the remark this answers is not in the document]
tgodzik: It will not be parsed as an extension if the dialect does not support it.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenClasses.scala
tgodzik: I cannot promise that it will be enough for the future, since there is still changes being made, but I am 90 % sure. There is a number of cases that extension as a soft keyword will not just work like:
```
extension(a : Int){
  def b = a
}
```

This can be an extension or invocation of `def extension(a : Int)( c : => Unit)`

kitbellew: if we end up keeping this logic, i think that
- this commit needs to be rebased on the `master`
- this trait moved inside @MichalBednarz's new `SoftKeywordClasses` type
- `true` on this line replaced with `dialect.allowExtensionMethods`.

tgodzik: So actually, it will be a soft keyword :/ I was told otherwise and now again they changed their minds. Sorry about that, I will fix it in the parser.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/DanglingParentheses.scala
```diff
@@ -36,10 +36,11 @@ object DanglingParentheses {
     case object `class` extends Exclude
     case object `trait` extends Exclude
     case object `enum` extends Exclude
+    case object `extension` extends Exclude
     case object `def` extends Exclude
 
     implicit val reader: ConfCodec[Exclude] =
-      ReaderUtil.oneOf[Exclude](`class`, `trait`, `enum`, `def`)
+      ReaderUtil.oneOf[Exclude](`class`, `trait`, `enum`, `extension`, `def`)
```
MichalBednarz, line 43: A corresponding part in the documentation can be updated. In `danglingParentheses.exclude` it's stated that `currently, class, trait and def are supported`. I haven't included there the enum by myself. You can add both extension and enum if you want.

tgodzik: Added it!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -464,6 +464,14 @@ class Router(formatOps: FormatOps) {
...
+        // Force a newline with the body of a Defn.ExtensionGroup if the
+        // method doesn't fit on a single line.
+        Seq(
+          Split(Space, 0).withSingleLine(leftOwner.tokens.last),
+          Split(Newline, 1).withIndent(2, leftOwner.tokens.last, After)
+        )
```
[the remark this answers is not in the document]
tgodzik: I added it. I also updated indent `expire` to take into account comments

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
tgodzik: Actually, this was needed, but I changed it to reuse SoftKEywords class.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -123,12 +128,21 @@ object TreeOps {
         case Some(x) => addTree(x, tree)
         case _ =>
           // No non-annotation modifier exists, fallback to keyword like `object`
-          tree.tokens.find(classTag[T].runtimeClass.isInstance) match {
+          tree.tokens.find(isMatch) match {
             case Some(x) => addTok(x, tree)
-            case None => throw Error.CantFindDefnToken[T](tree)
+            case None =>
```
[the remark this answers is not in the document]
tgodzik: Fixed!

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -114,7 +114,12 @@ object TreeOps {
       t.tokens.find(!_.is[Trivia]).foreach(addTok(_, tree))
     def addAll(trees: Seq[Tree]) = trees.foreach(x => addTree(x, x))
 
-    def addDefn[T: ClassTag](mods: Seq[Mod], tree: Tree): Unit = {
+    def addDefnTokens[T](
```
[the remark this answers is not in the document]
tgodzik: Removed!

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -123,12 +128,21 @@ object TreeOps {
...
+              throw Error.CantFindDefnToken(what, tree)
           }
       }
     }
+    def addDefn[T: ClassTag](mods: Seq[Mod], tree: Tree): Unit = {
+      addDefnTokens[T](
+        mods,
+        tree,
+        classTag[T].runtimeClass.getSimpleName(),
+        t => classTag[T].runtimeClass.isInstance(t)
```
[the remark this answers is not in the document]
tgodzik: Changed it a bit taking into account your comment, thanks!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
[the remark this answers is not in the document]
tgodzik: Good catch, I added support for the edge case with comments and added some tests, let me know what you think!

tgodzik:
@olafurpg @kitbellew I updated the PR after fixing `extension` to be a soft keyword again in scalameta. I also added some tests to make sure it all works, let me know what you think!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -561,6 +566,8 @@ class Router(formatOps: FormatOps) {
               if style.spaces.afterSymbolicDefs &&
                 isSymbolicName(name.value) && name.parent.exists(isDefDef) =>
             Space
+          case ext: Defn.ExtensionGroup if tok.left.is[soft.KwExtension] =>
```
MichalBednarz, line 569: nitpick: `_: Defn.ExtensionGroup`

scalafmt-tests/src/test/resources/dotty/Extension.stat
```diff
@@ -94,3 +94,50 @@ extension [AAaaaaaaaaaa](
...
+===
+extension [A](cat: Cat){ /* new feature */ // or this
+  def a = 1
+}
+>>>
+extension [A](
+    cat: Cat
+) { /* new feature */ // or this
+  def a = 1
+}
```
MichalBednarz, line 143: add newline

MichalBednarz (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -466,6 +466,19 @@ class Router(formatOps: FormatOps) {
           Split(NewlineT(isDouble = tok.hasBlankLine), 0)
         )
 
+      case tok @ FormatToken(_: T.RightParen, _: T.KwDef | _: T.Comment, _)
```
kitbellew, line 469: what happens if right is comment but after that we don't get a KwDef?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -466,6 +466,19 @@ class Router(formatOps: FormatOps) {
...
+          if leftOwner.is[Defn.ExtensionGroup] =>
+        val expireToken = tok.right match {
+          case _: T.KwDef => leftOwner.tokens.last
+          case _ => nextNonComment(next(tok)).right
```
kitbellew, line 473: why right? looks inconsistent with the KwDef handling.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -527,7 +540,7 @@ class Router(formatOps: FormatOps) {
         )
       // Opening [ with no leading space.
       // Opening ( with no leading space.
-      case FormatToken(
+      case tok @ FormatToken(
```
kitbellew, line 543: i wouldn't add tok. you can always refer to the actual variable, formatToken.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: yep

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -137,6 +155,8 @@ object TreeOps {
         case t: Defn.Macro => addDefn[KwDef](t.mods, t)
         case t: Decl.Def => addDefn[KwDef](t.mods, t)
         case t: Defn.Enum => addDefn[KwEnum](t.mods, t)
+        case t: Defn.ExtensionGroup =>
+          addDefnTokens(Nil, t, "extension", t => soft.KwExtension.unapply(t))
```
kitbellew, line 159: i may have asked. why do we need a lambda here? can't we just use KwExtension.unapply directly?

scalafmt-tests/src/test/scala/org/scalafmt/FormatTests.scala
```diff
@@ -54,6 +54,13 @@ class FormatTests extends FunSuite with CanRunTests with FormatAssertions {
         e.partialOutput
       case x => x.get
     }
+    if (!onlyManual) {
```
kitbellew, line 57: why did you reorder? i think the original idea was to put higher preference on idempotence violations, although I can't remember why. generally, if you believe this way is better, I'd strongly recommend doing that as part of a separate pull request, so the intent is clearly recorded in history.

scalafmt-tests/src/test/scala/org/scalafmt/FormatTests.scala
```diff
@@ -74,10 +81,7 @@ class FormatTests extends FunSuite with CanRunTests with FormatAssertions {
       )
       .get
     debug2.printTest()
-    assertNoDiff(formattedAgain, obtained, "Idempotency violated")
-    if (!onlyManual) {
-      assertNoDiff(obtained, t.expected)
-    }
+    assertEquals(formattedAgain, obtained, "Idempotency violated")
```
kitbellew, line 84: looks like we're also changing one type of assert to another. can we do that in a separate pr? i think it deserves its own commit, with explicit explanation.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
```diff
@@ -88,6 +88,10 @@ object ScalafmtRunner {
       .withAllowTrailingCommas(true)
     val default = scala213
     val Scala3 = Dotty
+      // Disable significant indentation because the exact syntax is still under
+      // development and is likely to change soon. See discussions at
+      // https://contributors.scala-lang.org/t/feedback-sought-optional-braces/4702
+      .withAllowSignificantIndentation(false)
```
kitbellew, line 95: not a big fan of latent changes.

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
tgodzik: The problem is that `extension` is a soft keyword so a tokenizer treats it as an Ident and `ident` will not match `extension`

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
tgodzik: Unfortunately, I think until it all is stable we shouldn't enable it in scalafmt.

scalafmt-tests/src/test/scala/org/scalafmt/FormatTests.scala
tgodzik: Not sure what @olafurpg meant here, but I guess I can remove the changes.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
tgodzik: I created a separate PR for some of the unrelated changes in this PR: https://github.com/scalameta/scalafmt/pull/2312

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: That cannot happen, the code will not parse then.  Nothing significant can be added between `)` and `def`

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: `right` will be the `def` token. It doesn't work with `left`, which will be still a comment.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
tgodzik: I will rebase once the change is made.

scalafmt-tests/src/test/scala/org/scalafmt/FormatTests.scala
tgodzik: Moved the change to https://github.com/scalameta/scalafmt/pull/2312

scalafmt-tests/src/test/scala/org/scalafmt/FormatTests.scala
tgodzik: Moved to https://github.com/scalameta/scalafmt/pull/2312

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Fixed!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -466,6 +466,19 @@ class Router(formatOps: FormatOps) {
           Split(NewlineT(isDouble = tok.hasBlankLine), 0)
         )
 
+      case tok @ FormatToken(_: T.RightParen, _: T.KwDef | _: T.Comment, _)
+          if leftOwner.is[Defn.ExtensionGroup] =>
+        val expireToken = tok.right match{
```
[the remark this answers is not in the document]
tgodzik: Fixed

scalafmt-tests/src/test/resources/dotty/Extension.stat
tgodzik: Fixed

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Fixed

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: in that case, does it still make sense to match `KwDef` or `Comment` here? shouldn't `_` be equivalent, given the `leftOwner` check below?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: so for comment, you will return `KwDef`. but in case of `KwDef`, you are not returning the `KwDef` token, you are returning the last token of the entire `ExtensionGroup`.

wouldn't `leftOwner.tokens.last` be sufficient for both cases? or `lastToken(leftOwner)` (which stops before trailing comments/whitespace)? i am not clear why there's asymmetry here.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
kitbellew: ok, i am not sure i understood. what i was asking was whether this code would work equally:
```
  addDefnTokens(Nil, t, "extension", soft.KwExtension.unapply)
```
that is, instead of `t => soft.KwExtension.unapply(t)`, can we use the simpler `soft.KwExtension.unapply`. and if we can't, can we at least use a different variable (because the case also uses `t`, so it might be a bit confusing)?

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Ach, actually you can have a using clause `(using A)` and an opening bracket `{`  We want to format it differently in that case.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: I was overthinking it, `leftOwner.tokens.last` works for both cases, thanks!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: If we leave it as just `_` it will format it a bit weirdly and also break the parser, since there is an issue when the formatting is a bit off. I will fix it separately, but the way it's currently formatted looks good to me.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
tgodzik: Ach, sorry! I totally misread it - or in reality I somehow never really read it :O  Changed!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: > Ach, actually you can have a using clause `(using A)` and an opening bracket `{` We want to format it differently in that case.

we can merge this as-is, but it would definitely be useful to have a follow-up pr to handle this:
```
extension [A, B](
  a: A
) /* foo */ (
  using b: B /* bar */
) /* baz */ {
  def a = 1
}
```

kitbellew (approved):

PR #2308, opened 2020-12-29
Scala 3: context functions

MichalBednarz:
#2216 
One of the tests requires `withAllowSignificantIndentation` to be set to false (also introduced in #2298). 
As you can see, the main problem with context functions is to find the optimal way to reuse the existing methods that have been used for regular functions so far. I wait for your suggestions on this matter.


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -597,8 +606,13 @@ class Router(formatOps: FormatOps) {
 
       // Term.Apply and friends
       case FormatToken(T.LeftParen(), _, _)
-          if getLambdaAtSingleArgCallSite(formatToken).isDefined =>
-        val lambda = getLambdaAtSingleArgCallSite(formatToken).get
+          if getLambdaAtSingleArgCallSite(formatToken).productIterator
+            .map(_.asInstanceOf[Option[Any]])
+            .exists(_.isDefined) =>
+        val (funLambda, contextFunLambda) = getLambdaAtSingleArgCallSite(
```
tgodzik, line 612: I think we should try to use `Tree` in this case.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -63,14 +63,26 @@ object TreeOps {
     ret.result()
   }
 
-  @tailrec
-  def isBlockFunction(fun: Term.Function): Boolean =
-    fun.parent match {
-      case Some(b: Term.Block) => isSingleElement(b.stats, fun)
+  def isBlockFunction(
+      fun: Either[Term.Function, Term.ContextFunction]
+  ): Boolean =
```
tgodzik, line 68: I think we will need to use Tree here also.

tgodzik (changes requested):
Looks good overall, thanks for the great work! However, I think we will need to replace `Either[Term.Function, Term.ContextFunction]` as we will be getting more (Term.PolyFunction for example).

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1246,42 +1250,58 @@ class FormatOps(
 
   def getLambdaAtSingleArgCallSite(
       ft: FormatToken
-  )(implicit style: ScalafmtConfig): Option[Term.Function] =
+  )(implicit style: ScalafmtConfig): Option[Tree] = {
+    def infixArg(fun: Tree) = fun.parent.exists({
+      case Term.ApplyInfix(_, _, _, List(`fun`)) => true
```
kitbellew, line 1255: i don't think you need parentheses around curlies.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1246,42 +1250,58 @@ class FormatOps(
...
-            case _ => false
-          }) =>
-        Some(fun)
+      case Term.Apply(_, List(fun: Term.ContextFunction)) => Some(fun)
+      case fun: Term.Function if infixArg(fun) => Some(fun)
+      case fun: Term.ContextFunction if infixArg(fun) => Some(fun)
       case t: Init =>
-        findArgsFor(ft.left, t.argss).collect { case List(f: Term.Function) =>
-          f
+        val args = findArgsFor(ft.left, t.argss)
```
kitbellew, line 1264: isn't this unnecessarily complex? why not just add another case to the original collect?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -529,6 +530,7 @@ class FormatOps(
         child.parent.collect {
           case _: Term.Block | _: Term.If | _: Term.While | _: Source => true
           case fun: Term.Function if isBlockFunction(fun) => true
+          case fun: Term.ContextFunction if isBlockFunction(fun) => true
```
kitbellew, line 533: this is a general comment which will allow to simplify this code massively. can we instead modify scalameta to derive both Function and ContextFunction from the same trait or abstract class, with common methods and fields declared there?

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
```diff
@@ -87,7 +87,7 @@ object ScalafmtRunner {
     val scala213 = Scala213
       .withAllowTrailingCommas(true)
     val default = scala213
-    val Scala3 = Dotty
+    val Scala3 = Dotty.withAllowSignificantIndentation(false)
```
kitbellew, line 91: this actually duplicates the change in extension group. probably should have been submitted as a separate pr.

MichalBednarz: @tgodzik we might merge #2298 with the above change first or you can extract it into separate pr as suggested by @kitbellew

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
MichalBednarz: this is a common issue with scalameta api for me so far, @tgodzik is this even possible to introduce such change?

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
tgodzik: We can just rebase one of the PRs, should be easy enough.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: So this is all related to macros, which are bit complex. Besides, that would be just `ContextFunction` and `Function` that could share some methods, nothing else really.

tgodzik: Trying it out in https://github.com/scalameta/scalameta/pull/2207

kitbellew (approved):

PR #2309, opened 2020-12-30
[Scala 3] Add support for match types

tgodzik:
Scala 3 allows to specify types using the match clause such as:
```scala
type T = R match{
  case Char => String
  case Array[t] => t
}
```
Implementing this required mostly copying whatever was done for normal matches.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -33,7 +33,8 @@ import scala.meta.{
   Template,
   Term,
   Tree,
-  Type
+  Type,
+  TypeCase
 }
 import scala.meta.tokens.Token
 import scala.meta.tokens.{Token => T}
@@ -1293,6 +1294,10 @@ class FormatOps(
       else tokens(nextNonComment(maybeArrow), 1)
     }(x => prevNonComment(tokens(x, -1)))
 
+  // look for arrow before body, if any, else after cond/pat
+  def getCaseArrow(term: TypeCase): FormatToken =
+    tokens(nextNonComment(tokens(term.pat.tokens.last)), 1)
+
   def getApplyArgs(
       ft: FormatToken,
       isRight: Boolean
@@ -1628,7 +1633,8 @@ class FormatOps(
         case _: Term.Try | _: Term.TryWithHandler =>
           if (hasStateColumn) getSplits(getSpaceSplit(1), false)
           else getSlbSplits()
-        case _: Term.Block | _: Term.Match | _: Term.NewAnonymous =>
+        case _: Term.Block | _: Term.Match | _: Type.Match |
+            _: Term.NewAnonymous =>
           if (!hasMatching(blast)) getSlbSplits()
           else getSplits(getSpaceSplit(1), true)
         case Term.ForYield(_, b) =>
@@ -1767,21 +1773,27 @@ class FormatOps(
   }
 
   // Redundant () delims around case statements
-  def isCaseBodyEnclosedAsBlock(ft: FormatToken, caseStat: Case)(implicit
+  def isCaseBodyEnclosedAsBlock(ft: FormatToken, caseStat: Tree)(implicit
       style: ScalafmtConfig
   ): Boolean = {
-    val body = caseStat.body
-    (ft.noBreak || style.newlines.getBeforeMultiline.ignoreSourceSplit) &&
-    body.eq(ft.meta.rightOwner) && (body match {
-      case _: Lit.Unit | _: Term.Tuple => false
-      case t: Term.ApplyInfix =>
-        val op = t.op.value
-        op != "->" && op != "→"
-      case _ => true
-    }) && {
-      val btoks = body.tokens
-      btoks.headOption.exists { head =>
-        head.is[Token.LeftParen] && matchingOpt(head).contains(btoks.last)
+    val bodyOpt = caseStat match {
+      case tc: TypeCase => Some(tc.body)
+      case c: Case => Some(c.body)
+      case _ => None
+    }
+    bodyOpt.exists { body =>
+      (ft.noBreak || style.newlines.getBeforeMultiline.ignoreSourceSplit) &&
+      body.eq(ft.meta.rightOwner) && (body match {
+        case _: Lit.Unit | _: Term.Tuple => false
+        case t: Term.ApplyInfix =>
+          val op = t.op.value
+          op != "->" && op != "→"
+        case _ => true
+      }) && {
+        val btoks = body.tokens
+        btoks.headOption.exists { head =>
+          head.is[Token.LeftParen] && matchingOpt(head).contains(btoks.last)
+        }
       }
     }
   }
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
@@ -19,6 +19,7 @@ import scala.meta.Template
 import scala.meta.Term
 import scala.meta.Tree
 import scala.meta.Type
+import scala.meta.TypeCase
 import scala.meta.classifiers.Classifier
 import scala.meta.tokens.Token
 import scala.meta.tokens.Token._
@@ -86,12 +87,14 @@ object TreeOps {
       case t: Term.For => getEnumStatements(t.enums)
       case t: Term.ForYield => getEnumStatements(t.enums)
       case t: Term.Match => t.cases
+      case t: Type.Match => t.cases
       case t: Term.PartialFunction => t.cases
       case t: Term.Try => t.catchp
       case t: Type.Refine => t.stats
       case t: scala.meta.Source => t.stats
       case t: Template => t.stats
       case t: Case if t.body.tokens.nonEmpty => Seq(t.body)
+      case t: TypeCase if t.body.tokens.nonEmpty => Seq(t.body)
       case _ => Seq.empty[Tree]
     }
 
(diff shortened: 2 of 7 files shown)


MichalBednarz (approved):
Good job, LGTM!

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -676,7 +679,13 @@ object TreeOps {
     }
 
   // Redundant {} block around case statements
-  def isCaseBodyABlock(ft: FormatToken, caseStat: Case): Boolean =
-    ft.right.is[Token.LeftBrace] && (caseStat.body eq ft.meta.rightOwner)
+  def isCaseBodyABlock(ft: FormatToken, caseStat: Tree): Boolean = {
+    val body = caseStat match {
+      case tc: TypeCase => tc.body
+      case c: Case => c.body
```
kitbellew, line 685: this will throw if parameter is neither

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -676,7 +679,13 @@ object TreeOps {
...
+    }
+    ft.right.is[Token.LeftBrace] && (body eq ft.meta.rightOwner) &&
+    (caseStat.is[Case] || caseStat.is[TypeCase])
```
kitbellew, line 688: it's too late to check that parameter is one of them. perhaps you should set body to null in the match above instead.

scalafmt-tests/src/test/resources/dotty/MatchType.stat
```diff
@@ -0,0 +1,83 @@
...
+    x1
+}
+<<< recursive match type
+type Len[X] <: Int = X match {
+           case Unit => 0
+           case x *: xs => S[Len[xs]]}
+>>>
+type Len[X] <: Int = X match {
+  case Unit    => 0
+  case x *: xs => S[Len[xs]]
```
kitbellew, line 29: side question: do we have a rule which formats this type of expression (*)?

scalafmt-tests/src/test/resources/dotty/MatchType.stat
```diff
@@ -0,0 +1,83 @@
...
+type Concat[
+    X,
+    Y
+] <: TupleTupleTupleTupleTupleTupleTuple =
+  X match {
+    case Unit =>
+      Y
+    case x1 *: xs1 =>
+      x1 *: Concat[xs1, Y]
+  }
```
kitbellew, line 83: missing eol

scalafmt-tests/src/test/resources/dotty/MatchType.stat
```diff
@@ -0,0 +1,83 @@
...
+           case Unit => 
+          Y
+           case x1 *: xs1 => 
+          x1 *: Concat[xs1, Y]
+         }
+>>>
+type Concat[
+    X <: Tuple,
+    Y <: Tuple
+] <: Tuple = X match {
```
kitbellew, line 58: can we add cases with different values of newlines.source (or beforeMultiline)? would like to see a forced newline for unfold, and single line for next case with fold, for instance.

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1767,23 +1773,32 @@ class FormatOps(
...
+      }) && {
+        val btoks = caseBody.tokens
+        btoks.headOption.exists { head =>
+          head.is[Token.LeftParen] && matchingOpt(head).contains(btoks.last)
+        }
       }
     }
+    val body = caseStatement match {
+      case tc: TypeCase => tc.body
+      case c: Case => c.body
```
kitbellew, line 1799: should we add an explicit default case here and throw?

scalafmt-tests/src/test/resources/dotty/MatchType.stat
tgodzik: I think this gets translated as an Type.ApplyInfix so basically it should format anything according to existing settings. It seems to be checked in a couple of places.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1767,23 +1773,32 @@ class FormatOps(
   }
 
   // Redundant () delims around case statements
-  def isCaseBodyEnclosedAsBlock(ft: FormatToken, caseStat: Case)(implicit
+  def isCaseBodyEnclosedAsBlock(ft: FormatToken, caseStatement: Tree)(implicit
```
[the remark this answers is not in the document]
tgodzik: I haven't noticed I changed it! Thanks!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1767,23 +1773,32 @@ class FormatOps(
...
-      case _: Lit.Unit | _: Term.Tuple => false
-      case t: Term.ApplyInfix =>
-        val op = t.op.value
-        op != "->" && op != "→"
-      case _ => true
-    }) && {
-      val btoks = body.tokens
-      btoks.headOption.exists { head =>
-        head.is[Token.LeftParen] && matchingOpt(head).contains(btoks.last)
+
```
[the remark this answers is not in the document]
tgodzik: I reworked the function.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1767,23 +1773,32 @@ class FormatOps(
...
+    def isEnclosed(ft: FormatToken, caseBody: Tree)(implicit
```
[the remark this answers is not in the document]
tgodzik: Yeah, it was actually not needed. I reworked it and took into account if the tree is not a case.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
tgodzik: Added an option and just return false in that case.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
tgodzik: Added an option and I check it here.

scalafmt-tests/src/test/resources/dotty/MatchType.stat
tgodzik: Added

scalafmt-tests/src/test/resources/dotty/MatchType.stat
tgodzik: I added some more cases, let me know what you think!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: I added an option here.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -281,7 +281,8 @@ object ScalafmtConfig {
   )
 
   val dotty: ScalafmtConfig = default.copy(
-    runner = ScalafmtRunner.default.copy(dialect = scala.meta.dialects.Dotty)
+    runner =
+      ScalafmtRunner.default.copy(dialect = ScalafmtRunner.Dialect.Scala3)
```
tgodzik, line 285: I found another issue with indentation :/ I think we can always rebase all other PR depending on which gets merged first.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1767,22 +1773,31 @@ class FormatOps(
...
-    }) && {
-      val btoks = body.tokens
-      btoks.headOption.exists { head =>
-        head.is[Token.LeftParen] && matchingOpt(head).contains(btoks.last)
+    val bodyOpt = caseStat match {
+      case tc: TypeCase => Some(tc.body)
+      case c: Case => Some(c.body)
+      case _ => None
+    }
+    if (bodyOpt.isDefined) {
```
kitbellew, line 1784: may i suggest replacing with a functional
```
bodyOpt.exists { body =>
  ...
}
```

another option is to preserve the original code virtually unchanged, use the fact that `body.eq(rightOwner)` is part of the logical expression and simply use `null` for body (or anything else that's not `rightOwner`) for non-case trees.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -676,7 +679,14 @@ object TreeOps {
...
   // Redundant {} block around case statements
-  def isCaseBodyABlock(ft: FormatToken, caseStat: Case): Boolean =
-    ft.right.is[Token.LeftBrace] && (caseStat.body eq ft.meta.rightOwner)
+  def isCaseBodyABlock(ft: FormatToken, caseStat: Tree): Boolean = {
+    val body = caseStat match {
+      case tc: TypeCase => Some(tc.body)
+      case c: Case => Some(c.body)
+      case _ => None
+    }
+    body.isDefined && ft.right
```
kitbellew, line 688: same comment as above, with using exists or null.

yet another option, check left brace first, then body:
```
ft.right.is[...] && ft.meta.rightOwner.eq(
  caseStat match {
    case x: Case => x.body
    ...
    case _ => null
 }
)

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
kitbellew: i merged your other pr, time to rebase?

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
tgodzik: Added exists here and above, good idea!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Fixed! Added exists.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
tgodzik: Rebased!

kitbellew (approved):

PR #2314, opened 2021-01-08
Use recently added CaseTree instead of Tree for cases

tgodzik:
Follow up from https://github.com/scalameta/scalameta/pull/2207

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -27,7 +27,8 @@ import scala.meta.{
   Term,
   Tree,
   Type,
-  TypeCase
+  TypeCase,
+  CaseTree
 }
 
 object Constants {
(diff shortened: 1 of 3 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -703,7 +704,7 @@ object TreeOps {
     }
 
   // Redundant {} block around case statements
-  def isCaseBodyABlock(ft: FormatToken, caseStat: Tree): Boolean = {
+  def isCaseBodyABlock(ft: FormatToken, caseStat: CaseTree): Boolean = {
```
kitbellew, line 707: should we have added the body def to CaseTree, using the common type, since they are different? would have simplified this code, too.

tgodzik: Ach, yeah I can fix that.

tgodzik: https://github.com/scalameta/scalameta/pull/2212

kitbellew: 4.4.6 ready to go?

tgodzik: Yep! Took me a while today to get it working for 2.12.13, but it's finally up.

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -419,11 +420,12 @@ class Router(formatOps: FormatOps) {
...
-          if leftOwner.is[Case] || leftOwner.is[TypeCase] =>
-        val (body, condIsDefined) = leftOwner match {
-          case c: Case => (c.body, c.cond.isDefined)
-          case tc: TypeCase => (tc.body, false)
+      case tok @ FormatToken(_: T.RightArrow, _, _) if leftOwner.is[CaseTree] =>
+        val owner = leftOwner.asInstanceOf[CaseTree]
+        val body = owner.body
+        val condIsDefined = leftOwner match {
+          case c: Case => c.cond.isDefined
+          case tc: TypeCase => false
```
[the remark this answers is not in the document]
tgodzik: Missed that! Thanks!

kitbellew (approved):

PR #2323, opened 2021-01-10
Config: allow "keep" where "preserve" is used

### LINKED ISSUE #2319: Terminology: `keep` vs `preserve`
Some options use the `keep` keyword while others use the `preserve` keyword.  These words seem similar to me.  Is there a convention as to which options use which of these keywords?  Is there a reason to use two different words?

If not, I'd propose allowing either keyword and making them aliases of each other.  (Either that or standardize on one and deprecate the other.)

kitbellew:
Fixes #2319.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1179,7 +1179,7 @@ something.map { x => f(x) }
 ```
 
 ```scala mdoc:scalafmt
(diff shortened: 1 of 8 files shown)


tgodzik (approved):
LGTM!

PR #2324, opened 2021-01-11
[Scala3] Add inline and transparent modifiers

tgodzik:
Scala 3 introduced inline and transparent modifiers that can be add to variables or definitions. More info at https://dotty.epfl.ch/docs/reference/metaprogramming/inline.html

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
@@ -27,7 +27,9 @@ object SortSettings {
     ModKey("protected", _.is[Mod.Protected]),
     //
     ModKey("lazy", _.is[Mod.Lazy]),
-    ModKey("open", _.is[Mod.Open])
+    ModKey("open", _.is[Mod.Open]),
+    ModKey("transparent", _.is[Mod.Transparent]),
+    ModKey("inline", _.is[Mod.Inline])
   )
 
   implicit val SortSettingsModKeyReader: ConfCodec[ModKey] =
--- a/scalafmt-tests/src/test/resources/rewrite/SortModifiersInlineTransparent.stat
+++ b/scalafmt-tests/src/test/resources/rewrite/SortModifiersInlineTransparent.stat
@@ -1,0 +1,18 @@
+maxColumn = 200 #to not disturb the output too much
+rewrite {
+  rules = [SortModifiers]
+  sortModifiers {
+    order = ["inline", "transparent"]
+  }
+}
+runner.dialect = Scala3
+<<< transparent inline method
+transparent inline def choose(b: Boolean): A = {
+  if (b) { new A }
+  else { new B }
+}
+>>>
+inline transparent def choose(b: Boolean): A = {
+  if (b) { new A }
+  else { new B }
+}
--- a/scalafmt-tests/src/test/resources/scala3/InlineTransparent.stat
+++ b/scalafmt-tests/src/test/resources/scala3/InlineTransparent.stat
@@ -1,0 +1,32 @@
+<<< simple inline
+inline val logging =     false
+>>>
+inline val logging = false
+<<< complex inline
+inline val logging = {
+    if (a) { true}
+    else false
+}
+>>>
+inline val logging = {
+  if (a) { true }
+  else false
+}
+<<< inline transparent
+transparent inline def choose(b: Boolean): A = {if(b) {new A} else {new B}}
+>>>
+transparent inline def choose(b: Boolean): A = {
+  if (b) { new A }
+  else { new B }
+}
+<<< inline transparent break
+maxColumn = 20
+===
+transparent inline override protected def choose(b: Boolean): A = {if(b) {new A} else {new B}}
+>>>
+transparent inline override protected def choose(
+    b: Boolean
+): A = {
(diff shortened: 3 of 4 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
```diff
@@ -27,7 +27,9 @@ object SortSettings {
     ModKey("protected", _.is[Mod.Protected]),
     //
     ModKey("lazy", _.is[Mod.Lazy]),
-    ModKey("open", _.is[Mod.Open])
+    ModKey("open", _.is[Mod.Open]),
+    ModKey("transparent", _.is[Mod.Transparent]),
+    ModKey("inline", _.is[Mod.Inline])
```
kitbellew, line 32: I'd really put inline first. might it not replace `@inline` which is usually there?

scalafmt-tests/src/test/resources/scala3/InlineTransparent.stat
```diff
@@ -0,0 +1,22 @@
...
+  if (a) { true }
+  else false
+}
+<<< inline transparent
+transparent inline def choose(b: Boolean): A = {if(b) {new A} else {new B}}
+>>>
+transparent inline def choose(b: Boolean): A = {
+  if (b) { new A }
+  else { new B }
+}
```
kitbellew, line 22: eol

scalafmt-tests/src/test/resources/scala3/InlineTransparent.stat
```diff
@@ -0,0 +1,22 @@
+maxColumn = 20
```
kitbellew, line 1: can you add tests so that lines would need to bed broken around these new modifiers, and see see how the code is formatted? indentation is interesting.

scalafmt-tests/src/test/resources/rewrite/SortModifiersInlineTransparent.stat
```diff
@@ -0,0 +1,18 @@
+maxColumn = 200 #to not disturb the output too much
+rewrite {
+  rules = [SortModifiers]
+  sortModifiers {
+    order = ["inline", "transparent"]
+  }
+}
+runner.dialect = Scala3
+<<< transparent inline method
+transparent inline def choose(b: Boolean): A = {
```
kitbellew, line 10: do you need a test with all possible modifiers in the wrong order to show how it gets reordered?

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
tgodzik: Hmm... do you mean we should move inline to the top of the list? Then we should have for example:
`inline override def foo = ???`
instead of:
`override inline def foo = ???`
?
I feel like the second one reads better though.

scalafmt-tests/src/test/resources/rewrite/SortModifiersInlineTransparent.stat
tgodzik: There are some tests, which don't include `transparent` or `inline`, but I think that's really not needed. Ordering is being tested anyway.

scalafmt-tests/src/test/resources/scala3/InlineTransparent.stat
tgodzik: Added!

scalafmt-tests/src/test/resources/scala3/InlineTransparent.stat
tgodzik: I added an additional test, but the modifiers don't seem to be ever broken into lines.

scalafmt-tests/src/test/resources/scala3/InlineTransparent.stat
kitbellew: i see still a "do not enter" sign :)

scalafmt-tests/src/test/resources/scala3/InlineTransparent.stat
```diff
@@ -0,0 +1,32 @@
...
+transparent inline def choose(b: Boolean): A = {
+  if (b) { new A }
+  else { new B }
+}
+<<< inline transparent break
+maxColumn = 20
+===
+transparent inline override protected def choose(b: Boolean): A = {if(b) {new A} else {new B}}
+>>>
+transparent inline override protected def choose(
```
kitbellew, line 27: for another PR: should they be broken if they overflow?

kitbellew (approved):

scalafmt-tests/src/test/resources/scala3/InlineTransparent.stat
tgodzik: Damn, I really thought I changed it! Hopefully now it's fine.

scalafmt-tests/src/test/resources/scala3/InlineTransparent.stat
tgodzik: I don't think so, I doubt that a larger number of modifiers shows up often :thinking:  Besides, I think they should in most cases fit in the usually used 80 maxColumn limit

PR #2325, opened 2021-01-12
[Scala 3] Add infix modifier

tgodzik:
This PR adds the infix modifier as well as makes sure it's properly sorted when using SortModifiers and that it works within extensions.

Infix modifier is mentioned here: https://dotty.epfl.ch/docs/reference/changed-features/operators.html

This also handles https://github.com/scalameta/scalafmt/pull/2298#pullrequestreview-562910763


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -470,8 +470,10 @@ class Router(formatOps: FormatOps) {
           Split(NewlineT(isDouble = tok.hasBlankLine), 0)
         )
 
-      case tok @ FormatToken(_: T.RightParen, _: T.KwDef | _: T.Comment, _)
-          if leftOwner.is[Defn.ExtensionGroup] =>
+      case tok @ FormatToken(_: T.RightParen, _, _)
+          if leftOwner.is[Defn.ExtensionGroup] && nextNonComment(
```
kitbellew, line 474: can you please manually break after && and format rest on one line?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -470,8 +470,10 @@ class Router(formatOps: FormatOps) {
...
+            formatToken
+          ).right.isNot[LeftParenOrBrace] =>
```
kitbellew, line 476: should this be .is[KwDef]?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -470,8 +470,10 @@ class Router(formatOps: FormatOps) {
           Split(NewlineT(isDouble = tok.hasBlankLine), 0)
         )
 
-      case tok @ FormatToken(_: T.RightParen, _: T.KwDef | _: T.Comment, _)
-          if leftOwner.is[Defn.ExtensionGroup] =>
+      case tok @ FormatToken(_: T.RightParen, _, _)
```
kitbellew, line 473: this change, and its tests, seem unrelated to infix, no? should they be part of a different pr?

scalafmt-tests/src/test/resources/rewrite/SortModifiersInfix.stat
```diff
@@ -0,0 +1,12 @@
+maxColumn = 200 #to not disturb the output too much
```
kitbellew, line 1: do we need a separate .stat file for this test?

kitbellew (changes requested):
i think two unrelated changes are mixed here

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: So the change is actually related to `infix` as we can get `infix def` etc. So having it as `.isNot[LeftParenOrBrace]` is safer, because we want to match only if we don't have a `{` block or another `(` param list.

kitbellew: is this cleaner than, say, .is[KwDef] or .is[Mod.xxx]?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: At least `.isNot[LeftParenOrBrace]` is needed for this change and since the list of accepted tokens started growing it turned out to be easier to just use it together with `nextNonComment`, which simplifies this logic a lot.

Because, this also fixes the issue with the comments I added the tests here. I can add it in a different PR, but it would just be a PR with those tests.

scalafmt-tests/src/test/resources/rewrite/SortModifiersInfix.stat
tgodzik: Not really, there were separate files for modifiers already that's why I added another one. I do agree that we can just put it into a single file. I will put it together with `inline`, `transparent` and `open`. What do you think?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: yes, i understand now. what type do you get for rightOwner of next non comment when it is def, infix, (, {, etc?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Yes, because we would need at least 3 with `Ident` as `infix` is a soft keyword. Also, it's more logical `LeftParenOrBrace` means we are not starting a block and not doing another param list, so the only other alternative is that it is the definition.

scalafmt-tests/src/test/resources/rewrite/SortModifiersInfix.stat
kitbellew: i like this approach 🙂

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: thank you, understood. unless there's a more logical way with rightOwner tree, vs right token, I'm ok with this. thanks again for looking into this.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: it should be `Defn.ExtensionGroup`  for `(`, `Term.Block` for `{` for the others `Defn.Def` - so nothing I think to help us here.

scalafmt-tests/src/test/resources/rewrite/SortModifiersInfix.stat
tgodzik: I put them all together into `SortModifiersScala3.stat`

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Fixed!

kitbellew (approved):

PR #2327, opened 2021-01-14
Scala 3: given imports, pattern-bound givens

MichalBednarz:
#2216

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -129,7 +129,7 @@ class Router(formatOps: FormatOps) {
           Split(NoSplit, 0)
         )
       // Import
-      case FormatToken(_: T.Dot, _: T.LeftBrace | _: T.Underscore, _)
+      case FormatToken(_: T.Dot, _, _)
           if existsParentOfType[Import](rightOwner) =>
         Seq(Split(NoSplit, 0))
       // Import left brace
--- a/scalafmt-tests/src/test/resources/scala3/GivenPatterns.stat
+++ b/scalafmt-tests/src/test/resources/scala3/GivenPatterns.stat
@@ -1,0 +1,43 @@
+<<< for comprehension
+for given Context <- applicationContexts yield ()
+>>>
+for given Context <- applicationContexts yield ()
+<<< for comprehension with parens
+for (given Context <- applicationContexts)
+yield ()
+>>>
+for (given Context <- applicationContexts)
+  yield ()
+<<< match case
+context match {
+  case (ctx @ given Context) => ctx
+}
+>>>
+context match {
+  case (ctx @ given Context) => ctx
+}
+<<< match case tuple
+pair match {
+  case (ctx @ given Context, y) => ctx
+}
+>>>
+pair match {
+  case (ctx @ given Context, y) => ctx
+}
+<<< don't break before nor after given
+maxColumn = 10
+===
+context match {
+  case (ctx @ given Context) => ctx
+  case (ctx @ given Context) => ctx
+  case (ctx @ given Context) => ctx
+}
+>>>
+context match {
+  case (ctx @ given Context) =>
+    ctx
+  case (ctx @ given Context) =>
+    ctx
+  case (ctx @ given Context) =>
+    ctx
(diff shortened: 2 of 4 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -129,8 +129,11 @@ class Router(formatOps: FormatOps) {
           Split(NoSplit, 0)
         )
       // Import
-      case FormatToken(_: T.Dot, _: T.LeftBrace | _: T.Underscore, _)
-          if existsParentOfType[Import](rightOwner) =>
+      case FormatToken(
```
kitbellew, line 132: should this just be `case FormatToken(_: T.Dot, _, _)`?

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
MichalBednarz: good point, it works as well!

tgodzik (approved):
LGTM!

scalafmt-tests/src/test/resources/scala3/ImportingGivens.stat
```diff
@@ -0,0 +1,46 @@
...
+<<< merged import clauses
+object B {
+  import A.{given, _}
+}
+>>>
+object B {
+  import A.{given, _}
+}
+<<< import given of type
+maxColumn = 20
```
kitbellew, line 20: can you add the same but with 10?

scalafmt-tests/src/test/resources/scala3/GivenPatterns.stat
```diff
@@ -0,0 +1,27 @@
+#
```
kitbellew, line 1: this one is not related to a code change, right?

scalafmt-tests/src/test/resources/scala3/GivenPatterns.stat
```diff
@@ -0,0 +1,27 @@
...
+for given Context <- applicationContexts yield ()
+<<< for comprehension with parens
+for (given Context <- applicationContexts)
+yield ()
+>>>
+for (given Context <- applicationContexts)
+  yield ()
+<<< match case
+context match {
+  case (ctx @ given Context) => ctx
```
kitbellew, line 14: can you add two more cases, with shorter max column, to force break before and after given? indent is of interest.

kitbellew (changes requested):
my apologies, previous review was incomplete, i forgot something

scalafmt-tests/src/test/resources/scala3/GivenPatterns.stat
MichalBednarz: do we want the given to break in such a situation? in the case of `case (ctx @ Context) => ctx` the line breaks only after the arrow.

scalafmt-tests/src/test/resources/scala3/GivenPatterns.stat
MichalBednarz: right, for test purposes only

scalafmt-tests/src/test/resources/scala3/ImportingGivens.stat
MichalBednarz: looks same, corresponds with the situation without given

scalafmt-tests/src/test/resources/scala3/GivenPatterns.stat
kitbellew: I thought it would break in the middle of the `x @ y` expression...

scalafmt-tests/src/test/resources/scala3/ImportingGivens.stat
```diff
@@ -0,0 +1,55 @@
...
+  given T2,
+  given T3
+}
+<<< import given of type doesn't break
+maxColumn = 10
+===
+import A.{given T1, given T2, given T3}
+>>>
+import A.{
+  given T1,
```
kitbellew, line 34: Looks like it doesn't break in the middle... Well, that's fine.

kitbellew (approved):

PR #2329, opened 2021-01-15
[Scala 3] Format export clauses

tgodzik:
Scala 3 has a complimentary structure to imports, which is the export cluase: https://dotty.epfl.ch/docs/reference/other-new-features/export.html

All of the existing options for import make sense in case of export.

Not sure if it makes sense to add aliases to all of the import options or whether to separate them. I left it as is for now.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/ExpandImportSelectors.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/ExpandImportSelectors.scala
@@ -21,7 +21,9 @@ class ExpandImportSelectors(implicit ctx: RewriteCtx) extends RewriteSession {
 
   override def rewrite(tree: Tree): Unit =
     tree match {
-      case q"import ..$imports" =>
+      case stat: ImportExportStat =>
+        val imports = stat.importers
+        val keyword = stat.tokens.head.text
         val groupedPatches = mutable.Map.empty[Token, Group]
         imports.foreach { `import` =>
           val parentTokens = `import`.parent.get.tokens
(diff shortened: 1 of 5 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/ExpandImportSelectors.scala
```diff
@@ -21,7 +21,12 @@ class ExpandImportSelectors(implicit ctx: RewriteCtx) extends RewriteSession {
 
   override def rewrite(tree: Tree): Unit =
     tree match {
-      case q"import ..$imports" =>
+      case stat: ImportExportStat =>
+        val imports = stat.importers
+        val keyword = stat match {
+          case _: Export => "export"
```
kitbellew, line 27: is there a KwImport that can be used instead?

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/ExpandImportSelectors.scala
tgodzik: Good idea, changed to `val keyword = stat.tokens.head.text`

tgodzik: I mean, I went a bit further here, since the first token is either `export` or `import`, so this works nicely.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/ExpandImportSelectors.scala
```diff
@@ -21,7 +21,9 @@ class ExpandImportSelectors(implicit ctx: RewriteCtx) extends RewriteSession {
 
   override def rewrite(tree: Tree): Unit =
     tree match {
-      case q"import ..$imports" =>
+      case stat: ImportExportStat =>
+        val imports = stat.importers
+        val keyword = stat.tokens.head.text
```
kitbellew, line 26: this is clever! :)

kitbellew (approved):

PR #2330, opened 2021-01-16
[Scala 3] Add formatting for opaque types

tgodzik:
More information about opaque types here: http://dotty.epfl.ch/docs/reference/other-new-features/opaques.html

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
@@ -30,7 +30,8 @@ object SortSettings {
     ModKey("open", _.is[Mod.Open]),
     ModKey("transparent", _.is[Mod.Transparent]),
     ModKey("inline", _.is[Mod.Inline]),
-    ModKey("infix", _.is[Mod.Infix])
+    ModKey("infix", _.is[Mod.Infix]),
+    ModKey("opaque", _.is[Mod.Opaque])
   )
 
   implicit val SortSettingsModKeyReader: ConfCodec[ModKey] =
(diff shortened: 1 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/SortSettings.scala
```diff
@@ -30,7 +30,8 @@ object SortSettings {
     ModKey("open", _.is[Mod.Open]),
     ModKey("transparent", _.is[Mod.Transparent]),
     ModKey("inline", _.is[Mod.Inline]),
-    ModKey("infix", _.is[Mod.Infix])
+    ModKey("infix", _.is[Mod.Infix]),
```
kitbellew, line 33: should we simply modify the rule so that all unknown `Mod` tokens are sorted alphabetically at the end?

tgodzik: We could do it to be on the safe side.

tgodzik: Although, the current behaviour is that unknown ones go to the start, for example in a case that a non default order was defined and it does not contain all mods. We could make it differently if the mod does not exist even in the default order, but I feel that maybe we should leave as is?

If it's not found in the list it goes to the start.

tgodzik: Also, in reality all modifiers should be handled once we finish

kitbellew: sounds good

kitbellew (approved):

PR #2337, opened 2021-01-27
[Scala 3] Add formatting for dependent function types

tgodzik:
Dependent functions allow us to use named parameters in function types and base their return types on the type of that parameter.

They are defined here: http://dotty.epfl.ch/docs/reference/new-types/dependent-function-types.html

### DIFF
--- a/scalafmt-tests/src/test/resources/scala3/DependentFunction.stat
+++ b/scalafmt-tests/src/test/resources/scala3/DependentFunction.stat
@@ -1,0 +1,61 @@
+
+<<< simple dependent function
+val extractor:   (e: Entry) => e.Key =    extractKey
+>>>
+val extractor: (e: Entry) => e.Key = extractKey
+<<< simple dependent function term
+type T = 
+ (e: Entry) 
+=> e.Key
+>>>
+type T =
+  (e: Entry) => e.Key
+<<< long dependent function
+maxColumn = 30
+===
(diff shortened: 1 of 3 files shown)


scalafmt-tests/src/test/resources/scala3/DependentFunction.stat
```diff
@@ -0,0 +1,61 @@
...
+  (e: Entry) ?=> e.Key
+<<< long context dependent function
+maxColumn = 30
+===
+val extractor: (entryWithALongName: EntryWithALongName) ?=> e.KeyOfALongName = extractKey
+>>>
+val extractor: (
+    entryWithALongName: EntryWithALongName
+) ?=> e.KeyOfALongName =
+  extractKey
```
kitbellew, line 61: my usual nitpick :)

kitbellew (approved):

kitbellew:
@tgodzik a small request. since you always create your pull requests as a single commit, including subsequent amends, can i suggest to merge them using "Squash and merge" (rather than the usual merge), which creates a nice single commit without a side branch.

scalafmt-tests/src/test/resources/scala3/DependentFunction.stat
tgodzik: fixed!

PR #2338, opened 2021-01-27
[Scala 3] Add formatting for polymorphic functions

tgodzik:
Polymorphic types are a way to define a function parametrized by a type, complementary to `def method[T]`

More information here: http://dotty.epfl.ch/docs/reference/new-types/polymorphic-function-types.html

I was also looking at fold/unfold, but polymorphic function types cannot have curly braces around their body, so this will not make much sense.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -356,6 +356,7 @@ class Router(formatOps: FormatOps) {
 
       case FormatToken(T.RightArrow() | T.ContextArrow(), right, _)
           if leftOwner.is[Term.FunctionTerm] ||
+            leftOwner.is[Term.PolyFunction] ||
             (leftOwner.is[Template] &&
               leftOwner.parent.exists(_.is[Term.NewAnonymous])) =>
         val (endOfFunction, expiresOn) = leftOwner match {
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
@@ -342,6 +342,7 @@ object TreeOps {
           _: Defn.EnumCase | _: Defn.ExtensionGroup =>
         true
       case _: Term.FunctionTerm | _: Type.FunctionType => true
+      case _: Term.PolyFunction | _: Type.PolyFunction => true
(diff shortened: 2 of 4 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -356,7 +356,7 @@ class Router(formatOps: FormatOps) {
 
       case FormatToken(T.RightArrow() | T.ContextArrow(), right, _)
           if leftOwner.is[Term.FunctionTerm] ||
-            (leftOwner.is[Template] &&
+            leftOwner.is[Term.PolyFunction] || (leftOwner.is[Template] &&
```
kitbellew, line 359: completely optional: the new term looks to go better, logically, with the previous line:
```
if leftOwner.is[Term.FunctionTerm] || leftOwner.is[Term.PolyFunction] ||
  (leftOwner.is[Template] && // unchanged
```
if it fits, of course.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -341,7 +341,9 @@ object TreeOps {
           _: Type.Apply | _: Type.Param | _: Type.Tuple | _: Defn.Enum |
           _: Defn.EnumCase | _: Defn.ExtensionGroup =>
         true
-      case _: Term.FunctionTerm | _: Type.FunctionType => true
+      case _: Term.FunctionTerm | _: Term.PolyFunction | _: Type.FunctionType |
```
kitbellew, line 344: also completely optional:
```
case _: Term.FunctionTerm | _: Type.FunctionType => true // unchanged
case _: Term.PolyFunction | _: Type.PolyFunction => true
```

kitbellew (approved):
ocd

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
tgodzik: Changed!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: The if line is still to long to changed it to:
```scala
 if leftOwner.is[Term.FunctionTerm] ||
   leftOwner.is[Term.PolyFunction] ||      
  (leftOwner.is[Template] &&
```

PR #2340, opened 2021-01-28
[Scala 3] Add support for union and intersection types

tgodzik:
Intersection and union types are defined respectively by `&` and `|`, you might read more about it here http://dotty.epfl.ch/docs/reference/new-types/union-types.html and http://dotty.epfl.ch/docs/reference/new-types/intersection-types.html

Since, they both look pretty similar to alternatives, that's what I based my implementation on.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1821,7 +1821,19 @@ class Router(formatOps: FormatOps) {
             Seq(Split(Space.orNL(noNL), 0))
         }
 
-      // Pat
+      // Union/Intersection types
+      case FormatToken(T.Ident(value), _, _)
+          if (value == "|" && leftOwner.is[Type.Or]) ||
+            (value == "&" && leftOwner.is[Type.And]) =>
+        if (style.newlines.source eq Newlines.keep)
+          Seq(Split(Space.orNL(newlines == 0), 0))
+        else
+          Seq(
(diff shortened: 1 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1821,7 +1821,19 @@ class Router(formatOps: FormatOps) {
             Seq(Split(Space.orNL(noNL), 0))
         }
 
-      // Pat
+      // Union/Intersection types
+      case FormatToken(T.Ident(value), _, _)
```
kitbellew, line 1825: optional: should `Type.And` and `Type.Or` be derived from
```
object Type {
  trait BinaryType {
    def lhs: Type
    def rhs: Type
    val op: Type.Name
  }
}
```
and match
```
case FormatToken(ident: T.Ident, _, _) if leftOwner match {
  case x: Type.Binary => x.op.tokens.head eq ident // unless there's a better check
  case _ => false
} =>
```

kitbellew: was just a thought. i frequently have them 😉

kitbellew (approved):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: I really appreciate the comments and the reviews, thank you!

PR #2341, opened 2021-01-28
Add spaces.inInterpolatedStringCurlyBraces

arbitrary-dev:
Implements #2228

Previous misunderstandings with @olafurpg over CoC were resolved in Discord chat.

Please review again @kitbellew

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2406,6 +2406,8 @@ import a.b.{c, d}
 
 ### `spaces.inInterpolatedStringCurlyBraces`
 
+> Since v3.0.0.
+
 ```scala mdoc:defaults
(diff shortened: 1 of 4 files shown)


kitbellew (commented):
implementation looks good. @olafurpg please approve if prior concerns have indeed been resolved.

olafurpg (approved):
LGTM. Many Boolean options often end up being enums so it might be worth considering using an enum from the start if it makes sense in your opinion

arbitrary-dev:
I used enum in the original review, but it was deemed as overcomplicated, so let's leave it as boolean.

kitbellew (approved):

PR #2345, opened 2021-01-30
Align case values

### LINKED ISSUE #2258: continuationIndent.caseSite = 5

arbitrary-dev:
Fixes #2258


docs/configuration.md
```diff
@@ -182,6 +182,27 @@ class A(
 }
 ```
 
+### `continuationIndent.caseSite`
+
+> Since v2.7.6.
```
kitbellew, line 187: the next version will be v3.0.0.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1734,7 +1734,11 @@ class Router(formatOps: FormatOps) {
           Split(Space, 0).withSingleLine(expire, killOnFail = true),
           Split(Space, 0, policy = policy)
             .withIndent(if (bodyBlock) 0 else 2, expire, After)
-            .withIndent(if (bodyBlock) 4 else 2, arrow, After)
+            .withIndent(
+              if (bodyBlock) 4 else style.continuationIndent.caseSite,
```
kitbellew, line 1738: this won't work. indent before arrow differs for `bodyBody = true/false` cases.

scalafmt-tests/src/test/resources/unit/Case.case
```diff
@@ -81,3 +81,15 @@ case q"""...${stats: Seq[Stat]} """ =>
...
+newlines.source = keep
+includeCurlyBraceInSelectChains = true
+continuationIndent.caseSite = 3
+===
+case _: Aaaaaa |
+    _: Bbbbbb |
+    _: Cccccc =>
+>>>
+case _: Aaaaaa |
+     _: Bbbbbb |
```
kitbellew, line 94: you specified indent 3. why is it 5 here?

scalafmt-tests/src/test/resources/unit/Case.case
```diff
@@ -81,3 +81,15 @@ case q"""...${stats: Seq[Stat]} """ =>
...
 >>>
 case List(xs @ _*) => a
+<<< #2258 Evenly align case values
+newlines.source = keep
+includeCurlyBraceInSelectChains = true
+continuationIndent.caseSite = 3
+===
+case _: Aaaaaa |
+    _: Bbbbbb |
+    _: Cccccc =>
```
kitbellew, line 91: you need to add plenty of tests, with various level of complexity (including `if` guards, case body in a block or without, etc.).

kitbellew (changes requested):

scalafmt-tests/src/test/resources/unit/Case.case
```diff
@@ -81,3 +81,15 @@ case q"""...${stats: Seq[Stat]} """ =>
 case  List(xs @ _*) => a
 >>>
 case List(xs @ _*) => a
+<<< #2258 Evenly align case values
+newlines.source = keep
```
kitbellew, line 85: btw, other `source` options should also be tested.

arbitrary-dev: Refactored it without `source`

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
arbitrary-dev: Updated, please check.

scalafmt-tests/src/test/resources/unit/Case.case
arbitrary-dev: Added, please check.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1735,7 +1735,7 @@ class Router(formatOps: FormatOps) {
           Split(Space, 0, policy = policy)
             .withIndent(if (bodyBlock) 0 else 2, expire, After)
             .withIndent(
-              if (bodyBlock) 4 else style.continuationIndent.caseSite,
+              style.continuationIndent.caseSite + (if (bodyBlock) 0 else -2),
```
kitbellew, line 1738: please add extensive tests with values 0 and 1.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1735,7 +1735,7 @@ class Router(formatOps: FormatOps) {
           Split(Space, 0, policy = policy)
             .withIndent(if (bodyBlock) 0 else 2, expire, After)
             .withIndent(
-              if (bodyBlock) 4 else style.continuationIndent.caseSite,
```
kitbellew: please add a variable `val foo = if (bodyBlock) 0 else 2` and use it both here and two lines above.

kitbellew (changes requested):

scalafmt-tests/src/test/resources/unit/Case.case
kitbellew: i don't think you added any tests where bodyBlock is true

kitbellew:
your pr comment uses the word "implements". i don't think this word has any significance in GitHub.

please change it to "fixes", as that creates a link with your issue and will close it automatically when this is merged.

scalafmt-tests/src/test/resources/unit/Case.case
arbitrary-dev: Can you help me on this?

I tried…
```
case {
  _: SomeClass
} => ???
```
…but it fails with syntactic error.

kitbellew: i think `bodyBlock` refers to `{}` after the arrow

arbitrary-dev: Added

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
arbitrary-dev: Added

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1730,15 +1730,12 @@ class Router(formatOps: FormatOps) {
           ) NoPolicy
           else defaultPolicy
 
+        val foo = if (bodyBlock) 0 else 2
```
kitbellew, line 1733: my apologies, i didn't mean that the variable should be named `foo`, i just left it to you to provide a meaningful name.

scalafmt-tests/src/test/resources/test/ContinuationIndentCaseSite0.case
```diff
@@ -0,0 +1,160 @@
+continuationIndent.caseSite = 0
+<<< Long if
+case tok if tok.left.name.startsWith("xml") && tok.right.name.startsWith("xml") => List( NoSplit0 )
```
kitbellew, line 3: i would prefer if you put inside a `match` or a partial function, because `case` by itself is meaningless and there's likely code which checks the parent of a case statement.

scalafmt-tests/src/test/resources/test/ContinuationIndentCaseSite5.case
```diff
@@ -0,0 +1,164 @@
+continuationIndent.caseSite = 5
+<<< Long if
+case tok if tok.left.name.startsWith("xml") && tok.right.name.startsWith("xml") => List( NoSplit0 )
```
kitbellew, line 3: i just noticed that these are all bare `case` statements (commented for another file), it's better to use a valid language construct.

kitbellew (changes requested):
almost there

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
arbitrary-dev: I was unable to come up with one 😀

Could be also simple `a` or `i`, what do you think?

kitbellew: variables may not have cryptic, meaningless names. how about `entireClauseIndent`?

scalafmt-tests/src/test/resources/test/ContinuationIndentCaseSite0.case
arbitrary-dev: Updated, please check again.

scalafmt-tests/src/test/resources/test/ContinuationIndentCaseSite5.case
arbitrary-dev: Updated, please check again.

kitbellew (approved):

PR #2351, opened 2021-02-09
Update Scalameta to 4.4.8 and add formatting for type bounds

tgodzik:
Previously, we would not split on type bounds which might sometimes be useful especially with the new Scala 3 types. Now, I added a separate case of type bounds in type declarations and definitions.

This also popped up due to the fixes done recently in scalameta. Not super sure about the changes, it is in a way similar to extends now, but none of it looks perfect.

We could remove the case and go back to having type bounds only ever split by space.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -14,6 +14,7 @@ import scala.meta.tokens.{Token, Tokens}
 import scala.meta.tokens.{Token => T}
 import scala.meta.{
   Case,
+  Decl,
   Defn,
   Enumerator,
   Importer,
@@ -1908,6 +1909,18 @@ class Router(formatOps: FormatOps) {
             Split(Newline, 1).withIndent(2, lastToken, After)
           )
         }
+      /* Type bounds in type definitions and declarations such as:
+       * type `Tuple <: Alpha & Beta = Another` or `Tuple <: Alpha & Beta`
+       */
+      case FormatToken(_, _: T.Subtype, _)
+          if isScala3Dialect &&
+            rightOwner.is[Type.Bounds] &&
+            rightOwner.parent.exists(t => t.is[Defn.Type] || t.is[Decl.Type]) =>
+        val typeBoundEnd = rightOwner.tokens.last
+        Seq(
+          Split(Space, 0).withSingleLineNoOptimal(typeBoundEnd),
+          Split(Newline, 1).withIndent(2, typeBoundEnd, After)
+        )
       // Interpolation
       case FormatToken(_, _: T.Interpolation.Id, _) =>
         Seq(
--- a/scalafmt-tests/src/test/resources/scala3/MatchType.stat
+++ b/scalafmt-tests/src/test/resources/scala3/MatchType.stat
@@ -52,10 +52,8 @@ type Concat[X <: Tuple, Y <: Tuple] <: Tuple = X match {
           x1 *: Concat[xs1, Y]
          }
 >>>
-type Concat[
-    X <: Tuple,
-    Y <: Tuple
-] <: Tuple = X match {
+type Concat[X <: Tuple, Y <: Tuple]
+  <: Tuple = X match {
   case Unit =>
     Y
   case x1 *: xs1 =>
@@ -71,10 +69,8 @@ type Concat[X, Y] <: TupleTupleTupleTupleTupleTupleTuple = X match {
           x1 *: Concat[xs1, Y]
          }
 >>>
-type Concat[
-    X,
-    Y
-] <: TupleTupleTupleTupleTupleTupleTuple =
+type Concat[X, Y]
+  <: TupleTupleTupleTupleTupleTupleTuple =
   X match {
     case Unit =>
       Y
--- a/scalafmt-tests/src/test/resources/scala3/OpaqueType.stat
+++ b/scalafmt-tests/src/test/resources/scala3/OpaqueType.stat
@@ -13,10 +13,10 @@ opaque type HelloIAmAnOpaqueTypeTerry =
 <<< opaque type bounds
 maxColumn = 15
 ===
-opaque type VeryLongClassName <: A & B = AB
+opaque type VeryLongClassName <: AType = AB
 >>>
-opaque type VeryLongClassName <: A & B =
-  AB
+opaque type VeryLongClassName
+  <: AType = AB
 <<< mixed modifiers
 private opaque   type T =   List[Int] 
 >>>
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -6,7 +6,7 @@ import org.portablescala.sbtplatformdeps.PlatformDepsPlugin.autoImport._
 
 object Dependencies {
   val metaconfigV = "0.9.10"
-  val scalametaV  = "4.4.7"
+  val scalametaV  = "4.4.8"
   val scalacheckV = "1.15.2"
   val coursier    = "1.0.3"
   val munitV      = "0.7.21"
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -67,6 +67,7 @@ class FormatOps(
   import TokenOps._
   import TreeOps._
   implicit val dialect = initStyle.runner.dialect
+  def isScala3Dialect = dialect == ScalafmtRunner.Dialect.scala3
   private val ownersMap = getOwners(tree)
   val tokens: FormatTokens = FormatTokens(tree.tokens, owners)
   private[internal] val soft = new SoftKeywordClasses(dialect)
--- a/scalafmt-tests/src/test/resources/scala3/UnionIntersectionType.stat
+++ b/scalafmt-tests/src/test/resources/scala3/UnionIntersectionType.stat
@@ -129,3 +129,26 @@ def hello(
 ) = {
   val a: A & B
 }
+<<< opaque type bounds
+maxColumn = 20
+===
+opaque type VeryLongClassName <: Alpha & Beta & Gamma = AB
+>>>
+opaque type VeryLongClassName
+  <: Alpha & Beta &
+    Gamma = AB
+<<< single split type bounds
+maxColumn = 20
+===
+opaque type VeryLongClassName <: Alpha & Beta = AB
+>>>
+opaque type VeryLongClassName
+  <: Alpha & Beta =
+  AB
+<<< single split type decl bounds
+maxColumn = 20
+===
+type VeryLongClassName <: Alpha & Beta
+>>>
+type VeryLongClassName
+  <: Alpha & Beta


scalafmt-tests/src/test/resources/scala3/OpaqueType.stat
```diff
@@ -15,8 +15,15 @@ maxColumn = 15
 ===
 opaque type VeryLongClassName <: A & B = AB
 >>>
-opaque type VeryLongClassName <: A & B =
```
tgodzik, line 18: So this started do have A and B on separate lines due to the recent fixes and I thought it might be better to split on `<:`

tgodzik:
I think we have 3 options, but none of them seems perfect:

- split on type bound and `&`/`|`
- split only on `&`/`|`
- split on none of the above and assume in most cases we would not have a very long chain here.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1908,6 +1909,15 @@ class Router(formatOps: FormatOps) {
             Split(Newline, 1).withIndent(2, lastToken, After)
           )
         }
+      // type bounds in type definitions and declarations
```
kitbellew, line 1912: how do we guarantee that this rule only applies for scala3? (otherwise, it would change formatting for existing code)

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1908,6 +1909,15 @@ class Router(formatOps: FormatOps) {
...
+      case FormatToken(_: T.Subtype, _, _)
```
kitbellew, line 1913: not sure if the language allows... can the split be before the `<:` symbol? (it works like an `extends`, and we break before `extends`.) would it look nice?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1908,6 +1909,15 @@ class Router(formatOps: FormatOps) {
...
+          if leftOwner.is[Type.Bounds] &&
+            leftOwner.parent.exists(t => t.is[Defn.Type] || t.is[Decl.Type]) =>
+        val lastToken = leftOwner.parent.get.tokens.last
+        Seq(
+          Split(Space, 0),
```
kitbellew, line 1918: should there be a single-line policy here? not sure, just asking, really.

scalafmt-tests/src/test/resources/scala3/MatchType.stat
```diff
@@ -52,15 +52,13 @@ type Concat[X <: Tuple, Y <: Tuple] <: Tuple = X match {
...
-    X <: Tuple,
-    Y <: Tuple
-] <: Tuple = X match {
-  case Unit =>
-    Y
-  case x1 *: xs1 =>
-    x1 *: Concat[xs1, Y]
-}
+type Concat[X <: Tuple, Y <: Tuple] <:
+  Tuple = X match {
```
kitbellew, line 56: would it break after `=` instead if we increase `maxColumn`?

scalafmt-tests/src/test/resources/scala3/OpaqueType.stat
kitbellew: definitely better than on `&`. doesn't look like you did anything to penalize breaking on `&`, though, right?

kitbellew (changes requested):

kitbellew:
> I think we have 3 options, but none of them seems perfect:
> 
> * split on type bound and `&`/`|`

this sounds like the best idea, although i'd test for cases where only one split is needed -- and that type bound is chosen _instead of_ the logical.

> * split only on `&`/`|`

not my choice.

> * split on none of the above and assume in most cases we would not have a very long chain here.

when we don't have a long chain, this would be equivalent to your first option. but that one has more to offer.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Currently, we might change the formatting for existing code, though Scala 2 only allows `<:` for declarations. We could check if the dialect is Scala3, which would make sure that we don't change this behaviour for existing users.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: This would be something like:
```scala
type Concat[X <: Tuple, Y <: Tuple]
  <: Tuple = X match {
    case Unit =>
      Y
    case x1 *: xs1 =>
      x1 *: Concat[xs1, Y]
  }
```
Looks also ok I think. Should we change it?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Not sure how that works. If we specify the single line policy, how is it different from the current behaviour. Or is that policy for everything after `<:`?

scalafmt-tests/src/test/resources/scala3/MatchType.stat
tgodzik: Checked and yes.

scalafmt-tests/src/test/resources/scala3/OpaqueType.stat
tgodzik: No, would it work when we wrap or/and with singleLine policy then? Did you mean this in the other comment?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: @olafurpg thoughts on backwards compatibility here?
@tgodzik should we use `newlines.beforeMultiline` to make a decision here? (and for default, aka `classic`, case, i would leave formatting as was before.)

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: 1. does it compile?

2. if the definition is long and need to have another break (say, after `&`), will the next line be indented further? (examples below _do not_ show the additional indent but could.)

3. do you like it yourself? would you prefer A, B or C? (i don't like B, i think.)

```
// A
type Concat[X, Y] <:
  Tuple & Shmuple &
  B = X match {
    ...
  }
// B
type Concat[X, Y]
  <: Tuple & Shmuple &
  B = X match {
    ...
  }
// C
type Concat[X, Y]
  <: Tuple & Shmuple
  & B = X match {
    ...
  }
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: single-line policy would mean no breaks after space. a penalized-break policy means breaks are allowed but a weight is added to each one. there are plenty of examples of both.

scalafmt-tests/src/test/resources/scala3/OpaqueType.stat
kitbellew: see comment above.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
olafurpg: I think it would be good to limit Scala 3 changes to Scala 3 users where possible. It’s too early IMO to introduce changes in Scala 2 formatting for hypothetical Scala 3 use-cases.

tgodzik: I will check the dialect in this case then. As for `newlines.beforeMultiline` I don't think it really applies here. Usually it's after `=` or `=>`

kitbellew: then `newlines.source`?

tgodzik: Ok, so when we add that it means we would split on `&` and `|` instead of `<:` always besides `unfold`, I added instead the check for the Scala 3 dialect, which makes sure that we don't modify the behaviour for existing users.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Ok, so I think a bit modified C makes sense here (that's what I changed to):
```scala
type Concat[X, Y]
  <: Tuple & Shmuple
    & B = X match {
      ...
    }
```
what do you think? It does compile from what I checked.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: That makes sense, I added `withSingleLineNoOptimal`

scalafmt-tests/src/test/resources/scala3/OpaqueType.stat
tgodzik: Changed the test here and added some into `UnionIntersectionType.stat`

scalafmt-tests/src/test/resources/scala3/MatchType.stat
```diff
@@ -52,15 +52,13 @@ type Concat[X <: Tuple, Y <: Tuple] <: Tuple = X match {
...
-type Concat[
-    X <: Tuple,
-    Y <: Tuple
-] <: Tuple = X match {
-  case Unit =>
-    Y
-  case x1 *: xs1 =>
-    x1 *: Concat[xs1, Y]
-}
+type Concat[X <: Tuple, Y <: Tuple]
```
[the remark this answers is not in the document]
tgodzik: Yeah, the indentation was a bit excessive before. Changed!

scalafmt-tests/src/test/resources/scala3/MatchType.stat
```diff
@@ -71,16 +69,14 @@ type Concat[X, Y] <: TupleTupleTupleTupleTupleTupleTuple = X match {
...
-    X,
-    Y
-] <: TupleTupleTupleTupleTupleTupleTuple =
-  X match {
-    case Unit =>
-      Y
-    case x1 *: xs1 =>
-      x1 *: Concat[xs1, Y]
-  }
+type Concat[X, Y]
```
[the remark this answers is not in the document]
tgodzik: Fixed!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1908,6 +1909,19 @@ class Router(formatOps: FormatOps) {
...
           )
         }
+      /* Type bounds in type definitions and declarations such as:
+       * type `Tuple <: Alpha & Beta = Another` or `Tuple <: Alpha & Beta`
+       */
+      case FormatToken(_, _: T.Subtype, _)
+          if isScala3Dialect &&
+            rightOwner.is[Type.Bounds] &&
+            rightOwner.parent.exists(t => t.is[Defn.Type] || t.is[Decl.Type]) =>
+        val lastIndentToken = rightOwner.parent.get.tokens.last
```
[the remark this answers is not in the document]
tgodzik: Changed to only indenting the bounds themselves.

kitbellew (approved):
@tgodzik hopefully you don't hate me yet... i turned your minor scalameta version update into a massive modification :)

tgodzik:
I thought it might not be that simple 😅

PR #2352, opened 2021-02-10
[Scala 3] Allow enums to be aligned with classes, objects and traits

tgodzik:
More info about enums here: https://dotty.epfl.ch/docs/reference/enums/enums.html


kitbellew (approved):
alignments are weird :) są gusty i guściki...

tgodzik:
> alignments are weird :) są gusty i guściki...

Yeah, I wouldn't format my code like that, but this seemed like a small enough omission :D

PR #2362, opened 2021-02-23
Update scalameta to 4.4.10

tgodzik:
The release added support for Scala 2.13.5.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
@@ -455,8 +455,7 @@ object TreeOps {
     case t: Defn.Trait => (t.mods, t.name, t.tparams, t.ctor.paramss)
     case t: Defn.Enum => (t.mods, t.name, t.tparams, t.ctor.paramss)
     case t: Defn.EnumCase => (t.mods, t.name, t.tparams, t.ctor.paramss)
-    case t: Defn.ExtensionGroup =>
-      (Nil, Name.Anonymous(), t.tparams, List(t.eparam) :: t.uparams)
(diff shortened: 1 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -455,8 +455,7 @@ object TreeOps {
     case t: Defn.Trait => (t.mods, t.name, t.tparams, t.ctor.paramss)
     case t: Defn.Enum => (t.mods, t.name, t.tparams, t.ctor.paramss)
     case t: Defn.EnumCase => (t.mods, t.name, t.tparams, t.ctor.paramss)
-    case t: Defn.ExtensionGroup =>
-      (Nil, Name.Anonymous(), t.tparams, List(t.eparam) :: t.uparams)
```
kitbellew, line 459: did the new version replace `eparam/uparams` with more typical `paramss`?

kitbellew (approved):

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
tgodzik: Yes, with `3.0.0-RC1` we can now have `using` parameters on both sides of the main parameter, so it didn't make any sense to keep it separate. It is very similar to the `Defn.Def` now with the only difference that in `paramss` there will be at most one parameter list that doesn't have `using` modifier.

Good news is that we should support Scala 3.0.0-RC1 fully now and we should be able to finish the work here on the integration for Scala 3. Though, it might take me same time still. Trying to work on derives next, almsot have something ready there.

PR #2366, opened 2021-03-11
[Scala 3] Format import/export when using `as` soft keyword

tgodzik:
Previously, renames could only be done using `=>`, now it's also possible to use `as`. This also makes it possible to not use braces when doing a single rename/unimport.

More information can be found [here](https://dotty.epfl.ch/docs/reference/changed-features/imports.html)

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -9,6 +9,8 @@ import org.scalafmt.internal.TreeSyntacticGroup
 import scala.meta._
 import scala.meta.tokens.Token.LeftBrace
 import scala.meta.tokens.Token.RightBrace
+import scala.meta.tokens.Token.Dot
+import scala.meta.tokens.Token.RightArrow
 
 import org.scalafmt.util.InfixApp
 import org.scalafmt.util.TreeOps._
@@ -104,8 +106,33 @@ class RedundantBraces(implicit ctx: RewriteCtx) extends RewriteSession {
       case t: Term.Interpolate if settings.stringInterpolation =>
         processInterpolation(t)
 
+      case Importer(_, List(importee)) if ctx.dialect.allowAsForImportRename =>
+        processSingleImport(importee)
+
       case _ =>
     }
+
+  private def processSingleImport(tree: Importee) = {
+    val tokens = tree.tokens
+    tree match {
+      case _: Importee.Rename | _: Importee.Unimport
+          if !tokens.exists(_.is[RightArrow]) =>
+        ctx.tokenTraverser
+          .findBefore(tokens.head) {
+            case _: LeftBrace => Some(true)
+            case _: Dot => Some(false)
+            case _ => None
+          }
+          .foreach { leftBrace =>
+            val rightBrace = ctx.getMatching(leftBrace)
+            implicit val builder = Seq.newBuilder[TokenPatch]
+            removeBraces(leftBrace, rightBrace)
+            ctx.addPatchSet(builder.result(): _*)
+          }
+      case _ =>
+    }
+
+  }
 
   private def processInit(tree: Init): Unit =
     tree.argss.foreach(processMultiArgApply)
--- a/scalafmt-tests/src/test/resources/scala3/RenameImportExport.stat
+++ b/scalafmt-tests/src/test/resources/scala3/RenameImportExport.stat
@@ -1,0 +1,81 @@
+
+<<< single import
+import scala   as  java
+>>>
+import scala as java
+<<< select import
+import java.util.List   as  `J-List`
+>>>
+import java.util.List as `J-List`
+<<< multiple imports
+import java.util.{List   as  `J-List`, LinkedList as Linked}
+>>>
+import java.util.{List as `J-List`, LinkedList as Linked}
+<<< multiple imports multiline
+maxColumn = 20
+===
+import java.util.{List   as  `J-List`, LinkedList as Linked}
+>>>
+import java.util.{
+  List as `J-List`,
+  LinkedList as Linked
+}
+<<< redundant braces
+import java.util.{List   as  `J-List`}
+>>>
+import java.util.{List as `J-List`}
+<<< remove redundant braces in import 
+rewrite.rules = [RedundantBraces]
+===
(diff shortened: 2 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -104,9 +106,44 @@ class RedundantBraces(implicit ctx: RewriteCtx) extends RewriteSession {
       case t: Term.Interpolate if settings.stringInterpolation =>
         processInterpolation(t)
 
+      case i: Importer =>
+        processImport(i)
```
kitbellew, line 110: would this be more readable:
```
  case Importer(_, List(importee)) => processSingleImport(importee)
```

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -104,9 +106,44 @@ class RedundantBraces(implicit ctx: RewriteCtx) extends RewriteSession {
...
+
+    if (tree.importees.size == 1) {
+      tree.importees.head match {
+        case _: Importee.Rename | _: Importee.Unimport =>
+          val tokens = tree.importees.head.tokens
+
+          val start = ctx.tokenTraverser.findBefore(tokens.head) { t =>
+            if (t.is[LeftBrace]) Some(true)
+            else if (t.is[Dot]) Some(false)
+            else None
```
kitbellew, line 125: would this be more readable:
```
  ... findBefore(...) {
    case _: LeftBrace => Some(true)
    case _: Dot => Some(false)
    case _ => None
  }
```

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -104,9 +106,44 @@ class RedundantBraces(implicit ctx: RewriteCtx) extends RewriteSession {
...
+          }
+
+          val end = ctx.tokenTraverser.findAfter(tokens.last) { t =>
+            if (t.is[RightBrace]) Some(true)
+            else if (t.is[LF]) Some(false)
+            else None
+          }
+
+          (start, end) match {
```
kitbellew, line 134: to avoid doing both when one fails... would this be more readable:
```
for {
  lbrace <- ....findBefore {...}
  rbrace <- ...findAfter {...}
} yield {
  implicit val builder = Seq.newBuilder[TokenPatch]
  removeBraces(lbrace, rbrace)
  ctx.addPatchSet(builder.result(): _*)
}
```

alternatively:
```
  ...findBefore {
  }.foreach { lbrace =>
    ...findAfter {
    }.foreach { rbrace =>
       // remove braces
    }
  }
```

finally, should we simply look for preceding `LeftBrace` and then use `ctx.getMatching`?

scalafmt-tests/src/test/resources/scala3/RenameImportExport.stat
```diff
@@ -0,0 +1,75 @@
...
+maxColumn = 20
+===
+import java.util.{List   as  `J-List`, LinkedList as Linked}
+>>>
+import java.util.{
+  List as `J-List`,
+  LinkedList as Linked
+}
+<<< redundant braces
+import java.util.{List   as  `J-List`}
```
kitbellew, line 24: what if it had been `.{List => JList}`? do you need to check for `as` vs `=>`? or even, if the language is known to be scala3, rewrite `=>` to `as`?

kitbellew (changes requested):

scalafmt-tests/src/test/resources/scala3/RenameImportExport.stat
tgodzik: We need to make sure it's `as`. I forgot about that one. Rewriting `=>` to `as` would probably be a nice feature, but maybe we can follow up with it if there's interest?

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
tgodzik: Good idea, changed and added a condition for the dialect.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
tgodzik: Looks much better, thanks!

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
tgodzik: Decided to do a single foreach and then `ctx.getMatching`, looks better.

scalafmt-tests/src/test/resources/scala3/RenameImportExport.stat
tgodzik: Fixed and add an additional point to the issue.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -104,9 +106,34 @@ class RedundantBraces(implicit ctx: RewriteCtx) extends RewriteSession {
...
+        processSingleImport(importee)
+
       case _ =>
     }
 
+  private def processSingleImport(tree: Importee) = {
+    val tokens = tree.tokens
+    tree match {
+      case _: Importee.Rename | _: Importee.Unimport
+          if !tokens.exists(_.is[RightArrow]) =>
```
kitbellew, line 119: should this be a positive check rather than negative? i.e., look for `Ident("as")` explicitly?

kitbellew (approved):
minor suggestion

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
tgodzik: That is the magic of soft keywords, you can actually do:
`import as as as` or `import a.{ as => as }`

and it compiles :D

No `=>` is a safer choice.

PR #2367, opened 2021-03-11
[Scala 3] Add tests for new wildcard imports

tgodzik:
It's now possible to use `*` instead of `_` for wildcard imports. More info [here](https://dotty.epfl.ch/docs/reference/changed-features/imports.html)

### DIFF
--- a/scalafmt-tests/src/test/resources/scala3/StarImport.stat
+++ b/scalafmt-tests/src/test/resources/scala3/StarImport.stat
@@ -1,0 +1,17 @@
+
+<<< simple wildcard import
+import a . b . c . *
+>>>
+import a.b.c.*
+<<< star named import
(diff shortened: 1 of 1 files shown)


scalafmt-tests/src/test/resources/scala3/StarImport.stat
```diff
@@ -0,0 +1,17 @@
+
+<<< simple wildcard import
+import a . b . c . *
```
kitbellew, line 3: btw, should we modify our import rewrite rules to change `_` to `*` for scala3?

tgodzik: We could, we should probably have something like `rewrite.Scala3` setting that would change the older syntax. I will add a task to the issue.

PR #2368, opened 2021-03-11
[Scala 3] Fix formatting for the new vararg splices

tgodzik:
It's now possible specify arguments for vararg parameters by adding `*`:
```scala
def method(a: String, b: String*)
method(arg1, args*)
```
this also works for pattern matches.

More info about the new syntax is available [here](https://dotty.epfl.ch/docs/reference/changed-features/vararg-splices.html)

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1851,7 +1851,11 @@ class Router(formatOps: FormatOps) {
       case FormatToken(_, T.Ident("|"), _) if rightOwner.is[Pat.Alternative] =>
         val noNL = style.newlines.source.ne(Newlines.keep) || newlines == 0
         Seq(Split(Space.orNL(noNL), 0))
-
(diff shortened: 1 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1851,7 +1851,11 @@ class Router(formatOps: FormatOps) {
       case FormatToken(_, T.Ident("|"), _) if rightOwner.is[Pat.Alternative] =>
         val noNL = style.newlines.source.ne(Newlines.keep) || newlines == 0
         Seq(Split(Space.orNL(noNL), 0))
-
+      case FormatToken(_, T.Ident("*"), _)
```
kitbellew, line 1854: should we check that the left is also Ident?

tgodzik: I don't think it's needed, the `Term.Repeated` condition is much more important here, which must be after Ident.

kitbellew (approved):

PR #2369, opened 2021-03-11
[Scala 3] Add tests for modified extension methods

tgodzik:
Previously, extension methods could not have `using` clauses before the main parameter or type params in definitions, which is now possible.

### DIFF
--- a/scalafmt-tests/src/test/resources/scala3/Extension.stat
+++ b/scalafmt-tests/src/test/resources/scala3/Extension.stat
@@ -167,3 +167,94 @@ extension [A, B](a: A) /* foo */ (using
 )
   /* baz */
   def a = 1
+<<< multiple using clauses
+maxColumn = 40
+===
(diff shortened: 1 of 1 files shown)


scalafmt-tests/src/test/resources/scala3/Extension.stat
```diff
@@ -167,3 +167,19 @@ extension [A, B](a: A) /* foo */ (using
 )
   /* baz */
   def a = 1
+<<< multiple using clauses
```
kitbellew, line 170: do you want to add variants of this test with different values of `implicitParamListModifierForce` and `implicitParamListModifierPrefer`?

tgodzik: I think we should be fine, looks like everything is working irrelevant which param list it is.

tgodzik: I added some more tests, it seems to work ok as far as I can tell.

kitbellew: lovely, it really does work! :)

kitbellew (approved):

PR #2371, opened 2021-03-18
[Scala 3] Add formatting for derives clause

tgodzik:
Derives is a new concept from Scala 3 that can appear after `extends` clauses and may take a form of a soft keyword `derives` followed up by a list of types separated by comas.

More information can be found here: https://dotty.epfl.ch/docs/reference/contextual/derivation.html


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -411,6 +411,16 @@ class FormatOps(
     template.tokens.find(x => x.is[T.LeftBrace] && owners(x) == template)
   }
 
+  def templateDerivesOrCurly(template: Template): Option[Token] = {
+    val tokens = template.tokens
+    val index = tokens.indexWhere(x =>
+      (x.is[soft.KwDerives] || x.is[T.LeftBrace]) && owners(x) == template
+    )
+    if (index < 0) None
+    else if (tokens(index).is[soft.KwDerives]) Some(tokens(index - 1))
```
kitbellew, line 420: and if `index == 0`? :)

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -942,14 +952,21 @@ class FormatOps(
...
       case Some(_: Defn.Given) =>
         findLast(template.tokens)(x => x.is[T.KwWith] && owners(x) == template)
-      case _ => templateCurly(template)
-    }).getOrElse(template.tokens.last)
+      case _ if isDerives => templateCurly(template)
+      case _ => templateDerivesOrCurly(template)
+    }
+
+    val (expire, when) = curlyOrDerivesOrWith match {
+      case Some(value) => (value, ExpiresOn.Before)
```
kitbellew, line 967: why do we expire before the token which comes before `derives`? and not before `derives`, or after that preceding token?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1432,28 +1433,52 @@ class Router(formatOps: FormatOps) {
...
+          when
         )
+
       // trait A extends B, C, D, E
       case FormatToken(T.Comma(), right, _) if leftOwner.is[Template] =>
-        val template = leftOwner.asInstanceOf[Template]
-        val prevOwner = prevNonComment(prev(formatToken)).meta.leftOwner
+        val template: Template = leftOwner.asInstanceOf[Template]
+        val previousToken = prevNonComment(prev(formatToken))
+        val prevOwner = previousToken.meta.leftOwner
```
kitbellew, line 1471: looks like we didn't need to change these lines. `previousToken` is not used, and others are equivalent.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -733,4 +733,9 @@ object TreeOps {
       owner.is[Init] && init == owner || init.tpe == owner
     }
 
+  def isFirstDerives(t: Template, owner: Tree) =
+    t.derives.headOption.exists { tp =>
+      owner.is[Type] && tp == owner
+    }
```
kitbellew, line 739: or

```
owner.is[Type] && t.derives.headOption.contains(owner)
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1432,28 +1433,52 @@ class Router(formatOps: FormatOps) {
...
+        val firstDerives = isFirstDerives(
+          template,
+          prevOwner
+        )
+        def firstInit = isFirstInit(template, prevOwner)
         val ident =
-          if (isFirstInit(template, prevOwner))
+          if (firstDerives || firstInit)
```
kitbellew, line 1478: opt: could keep it inlined
```
if (firstDerives || isFirstInit(template, prevOwner))
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1432,28 +1433,52 @@ class Router(formatOps: FormatOps) {
...
+        // we should expire before derives or else we will get additional indentation
+        val (lastToken, when) = derivesOrCurly match {
+          case Some(tk) => (tk, ExpiresOn.Before)
+          case None =>
+            val last = template
+              .map(_.tokens.last)
+              .getOrElse(rightOwner.tokens.last)
+
+            (last, ExpiresOn.After)
+        }
```
kitbellew, line 1458: optional:
```
        // we should expire before derives or else we will get additional indentation
        val (when, lastToken) = template
          .flatMap {
            if (right.is[soft.KwDerives]) templateCurly else templateDerivesOrCurly
          }.map {
            (ExpiresOn.Before, _)
          }.getOrElse {
            (ExpiresOn.After, template.getOrElse(rightOwner).tokens.last)
          }
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -590,8 +590,9 @@ class Router(formatOps: FormatOps) {
           .flatMap(templateCurly)
           .getOrElse(leftOwner.tokens.last)
         val forceNewlineBeforeExtends = Policy.before(expire) {
-          case Decision(t @ FormatToken(_, _: T.KwExtends, _), s)
-              if t.meta.rightOwner == leftOwner =>
+          case Decision(t @ FormatToken(_, right, _), s)
+              if (right.is[T.KwExtends] || right.is[soft.KwDerives]) &&
```
kitbellew, line 594: optional: should this check be added as a trait to `SoftKeywordClasses` within `util/TokenClasses.scala`? it's used several times.

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Hmm... honestly, I am not sure. It didn't work any other way. I will dig in and see what's going on.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Changed to `index <= 0` - in reality this should never happen (hopefully)

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Ok, I found the issue, when moving to the previous token we would stop at a Space token, we need to get back to a non whitespace one, which is now changed.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Good catch, I had a lot of different approaches, so that probably stems from that.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
tgodzik: Changed

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Changed

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Reworked it to not need `when`.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Added a new trait there, good idea.

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ContinuationIndent.scala
```diff
@@ -15,6 +15,7 @@ case class ContinuationIndent(
     defnSite: Int = 4,
     caseSite: Int = 4,
     ctorSite: Option[Int] = None,
+    @annotation.ExtraName("deriveSite")
```
kitbellew, line 18: by the way, do we need to mention this in the documentation?

tgodzik: We don't, I was thinking to go through all the PRs and update it all just before the release. Does it make sense or should I update it now?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1432,28 +1432,45 @@ class Router(formatOps: FormatOps) {
...
+              templateCurly(t)
+            else
+              templateDerivesOrCurly(t)
+          }
+
+        val lastToken = derivesOrCurly
+          .orElse(
+            template
+              .map(_.tokens.last)
+          )
```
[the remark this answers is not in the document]
tgodzik: I changed the code a lot and figured that the comment did not make sense within the new context, but I haven't read it properly. Sorry about that! I changed it now, looks much clearer.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1432,28 +1432,45 @@ class Router(formatOps: FormatOps) {
           style.continuationIndent.extendSite,
           enumCase.inits.length > 1
         )
+
       // Template
-      case FormatToken(_, right @ T.KwExtends(), _) =>
+      case FormatToken(_, right, _) if right.is[soft.ExtendsOrDerives] =>
```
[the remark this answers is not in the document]
kitbellew: Did this change?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -942,14 +949,18 @@ class FormatOps(
 
   def typeTemplateSplits(
       template: Template,
-      indent: Int
-  ): Seq[Split] = {
+      indent: Int,
+      isDerives: Boolean = false
+  )(implicit line: sourcecode.Line): Seq[Split] = {
     val hasSelfAnnotation = template.self.tokens.nonEmpty
-    val expire = (template.parent match {
```
kitbellew, line 948: I think this line didn't need to change, sorry didn't notice before

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -942,14 +949,18 @@ class FormatOps(
...
+    val curlyOrDerivesOrWith = template.parent match {
       case Some(_: Defn.Given) =>
         findLast(template.tokens)(x => x.is[T.KwWith] && owners(x) == template)
-      case _ => templateCurly(template)
-    }).getOrElse(template.tokens.last)
```
kitbellew, line 952: And this line also didn't need to change

kitbellew (changes requested):
This should be the last round

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Fixed

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Good catch, fixed

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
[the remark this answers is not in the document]
tgodzik: Forgot about that one! Thanks! It's fixed now.

kitbellew (approved):

PR #2372, opened 2021-03-22
TokenTraverser: fix bug with traversing back

kitbellew:
For consistency, also slightly rewrite the forward check.


tgodzik (approved):
LGTM

PR #2374, opened 2021-03-23
RedundantBraces: keep around some new anonymous expressions with select

### LINKED ISSUE #2296: [BUG] rewrite.rules.RedundantBraces removes braces when it should not.
- **Version**: 2.7.5
- **Integration**: VsCode metals extension
- **Metals server version**: 0.9.7
- **Configuration**:
```
version = "2.7.5"
align.preset = more

align {
  preset = more
}

binPack {
  literalArgumentLists = true
  parentConstructors = Oneline
}

continuationIndent {
  extendSite = 4
  withSiteRelativeToExtends = 2
}

includeNoParensInSelectChains = true

newlines {
  alwaysBeforeElseAfterCurlyIf = true
  avoidInResultType = true
  beforeCurlyLambdaParams = multilineWithCaseOnly
}

project {
  excludeFilters = [
    ".metals"
  ]
}

rewrite {
  rules = [
    PreferCurlyFors,
    RedundantBraces,
    RedundantParens,
    SortImports,
    SortModifiers,
  ]

  sortModifiers {
    order = [
      "final",
      "sealed",
      "abstract",
      "override",
      "implicit",
      "private",
      "protected",
      "lazy"
    ]
  }
}

spaces {
  inImportCurlyBraces = true
}

verticalMultiline {
  arityThreshold = 4
  atDefnSite = true
  excludeDanglingParens = []
  newlineAfterImplicitKW = true
  newlineBeforeImplicitKW = false
  newlineAfterOpenParen = true
}
```

## Steps

Given code like this:
```scala
class A
trait B {
  def foo: Int = 5
}

val res = { new A with B }.foo
```

When I run scalafmt by running "format document" command in VsCode

## Problem

```scala
class A
trait B {
  def foo: Int = 5
}

val res = new A with B.foo
```

## Explanation of the problem

I would like the formatted output to not change. This has to do with `rewrite.rules.RedundantBraces`, when I remove it from my `.scalafmt` config files, the code does not change. The problem is, as it is probably obvious, that formatted code does not compile due to syntax error.

kitbellew:
Fixes #2296.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -251,7 +251,7 @@ class RedundantBraces(implicit ctx: RewriteCtx) extends RewriteSession {
         def disqualifiedByUnit =
           !settings.includeUnitMethods && d.decltpe.exists(_.syntax == "Unit")
         settings.methodBodies &&
-        getSingleStatIfLineSpanOk(b).exists(innerOk) &&
+        getSingleStatIfLineSpanOk(b).exists(innerOk(b)) &&
         !isProcedureSyntax(d) &&
         !disqualifiedByUnit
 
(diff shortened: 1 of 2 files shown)


kitbellew:
@shotexa can you check the added tests and let me know if any other cases need to be handled?

poslegm (approved):

tgodzik (approved):
LGTM!

PR #2375, opened 2021-03-23
Documentation: reduce nesting, reword, caution on all defaults

### LINKED ISSUE #2344: generation of example with all-configuration settings produces wrong config
Just copy the huge config from
https://scalameta.org/scalafmt/docs/configuration.html#other 
(2.7.5 version at this moment)

Try format any scala file e.g. with IntelliJ.
Observe that it's not parsed due to many different issues with the config, some examples:
1. regexp are not escaped:
```
rewrite.neverInfix.includeFilters = [
  "[\w\d_]+"
]
```
it should be
```
rewrite.neverInfix.includeFilters = [
  "[\\w\\d_]+"
]
```
2. null values are present:
```
newlines.source = null
```

Looks like default values generation logic is broken.

Here is the list of the lines I had to change to make the config parsarable (I will omit all the error messages that those lines caused):
```
#continuationIndent.ctorSite = null
#continuationIndent.commaSiteRelativeToExtends = 2
#rewrite.redundantBraces.parensForOneLineApply = null
  implicit
  final
  sealed
  abstract
  override
  private
  protected
  lazy
  #open
  #transparent
  #inline
  #inline
  #infix
  #opaque
]
rewrite.neverInfix.includeFilters = [
  "[\\w\\d_]+"
]
indentOperator.exclude = "^(&&|\\|\\|)$"
#newlines.source = null
#newlines.implicitParamListModifierPrefer = null
#newlines.beforeMultiline = null
#newlines.beforeMultilineDef = null
#newlines.afterInfix = null
#runner.parser = "<Parse[Tree]>"
#runner.dialect = "<Dialect>"
project.includeFilters = [
  ".*\\.scala$"
  ".*\\.sbt$"
  ".*\\.sc$"
]
```

kitbellew:
Fixes #2344.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1084,21 +1084,23 @@ something.map { x =>
 something.map { x => f(x) }
 ```
 
-### Newlines around `implicit` parameter list modifier
+### `newlines.implicitParamListModifierXXX`
+
+These parameters control newlines around `implicit` parameter list modifier.
 
 > Since v2.5.0.
 
 #### Newlines around `using` parameter and argument list modifier
 
-> Since v3.0.0
-
-`using` soft keyword was introduced in Scala 3 and is supported in scalafmt.
-The settings `newlines.implicitParamListModifierPrefer` and
-`newlines.implicitParamListModifierForce` are from now on aliases for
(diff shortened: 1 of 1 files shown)


poslegm (approved):

tgodzik (approved):
LGTM!

PR #2376, opened 2021-03-24
[Scala 3] Add formatting for match used as an operator

tgodzik:
In Scala 3 match can be used as any other operator, which means it can chained and can be invoked with an additional `.`

More details here: https://dotty.epfl.ch/docs/reference/changed-features/match-syntax.html

I tried to keep most of the changes behind the dialect, so this will only be noticed by Scala 3 dialect users

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -1458,6 +1458,40 @@ class FormatOps(
     })
   }
 
+  /* Find last tree in a chain if the chain contains a match*/
+  def lastTreeInChainContaingMatch(
+      term: Tree,
+      matchFound: Boolean = false
+  ): Option[Tree] = {
+    term.parent match {
+      case Some(p: Term.Match) =>
+        lastTreeInChainContaingMatch(p, matchFound = true)
+      case Some(p) if p.is[Term.Select] || p.is[Term.Apply] =>
+        lastTreeInChainContaingMatch(p, matchFound)
+      case _ if term.is[Term.Match] || matchFound => Some(term)
+      case _ => None
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


kitbellew (commented):
@tgodzik i have written a bunch of random cosmetic comments below; however, please don't implement them for now, as i need a little more time to understand whether the general functional approach is the right one.

my thinking is around the fact that `.match` now looks exactly like any other `.foo`, hence perhaps the solution should be around existing rules handling `Term.Select` and adding `Term.Match` to them...

tgodzik:
> @tgodzik i have written a bunch of random cosmetic comments below; however, please don't implement them for now, as i need a little more time to understand whether the general functional approach is the right one.
> 
> my thinking is around the fact that `.match` now looks exactly like any other `.foo`, hence perhaps the solution should be around existing rules handling `Term.Select` and adding `Term.Match` to them...

I tried that and failed miserably :/ I had to change much more code and I never managed to get it to work. I could try that again, but it took me around a day and led nowhere.

kitbellew:
> I tried that and failed miserably :/ I had to change much more code and I never managed to get it to work. I could try that again, but it took me around a day and led nowhere.

I think I got it to work. How could I share it with you? I don't know if I can push it to this PR... but I can open another one. Please let me know.

tgodzik:
Maybe another PR then? Thanks for taking a look at it! I was really frustrated by trying to implement it 😰

kitbellew:
> I tried that and failed miserably :/ I had to change much more code and I never managed to get it to work. I could try that again, but it took me around a day and led nowhere.

i just read it again 😀 it took me several attempts, probably longer than a month altogether, to figure out how to format select chains properly, including comment handling, indents etc. perhaps that's why i thought that reusing could be simpler.

tgodzik:
Succeeded by https://github.com/scalameta/scalafmt/pull/2386

PR #2386, opened 2021-03-25
[Scala 3] Add formatting for match used as an operator

kitbellew:
Based on #2376 from @tgodzik:

*In Scala 3 match can be used as any other operator, which means it can chained and can be invoked with an additional `"."`*

*More details here: https://dotty.epfl.ch/docs/reference/changed-features/match-syntax.html*

*I tried to keep most of the changes behind the dialect, so this will only be noticed by Scala 3 dialect users*

This approach applies the formatting rules used for `Term.Select` to appropriate `Term.Match` expressions.


tgodzik (approved):
Looks awesome! Thanks for taking care of this one, I was starting to get nightmares about it :sweat_smile:

PR #2387, opened 2021-03-26
Update scalatags to 0.9.4

tgodzik:
Scaltags are now published again for 2.11 :tada:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -207,10 +207,7 @@ lazy val tests = project
     skip in publish := true,
     libraryDependencies ++= Seq(
       // Test dependencies
-      CrossVersion.partialVersion(scalaVersion.value) match {
-        case Some((2, 13)) => "com.lihaoyi" %% "scalatags" % "0.9.3"
-        case _ => "com.lihaoyi" %% "scalatags" % "0.9.4"
-      },
+      "com.lihaoyi" %% "scalatags" % "0.9.4",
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -207,10 +207,7 @@ lazy val tests = project
     skip in publish := true,
     libraryDependencies ++= Seq(
       // Test dependencies
-      CrossVersion.partialVersion(scalaVersion.value) match {
-        case Some((2, 13)) => "com.lihaoyi" %% "scalatags" % "0.9.3"
```
kitbellew, line 211: do you know why we had a split? what does the commit message say, if you use git blame?

tgodzik: Looks like it was introduced while adding support for 2.13 https://github.com/scalameta/scalafmt/pull/1522

Doesn't seem there is any particular reason other than maybe it was only released for 2.13?

kitbellew: last question. should we keep this check for 2.13 and just update its version? just in case the next version of scalatags will once again not support an older version of scala, and we won't need to remember how to write this construct?

tgodzik: I don't think they should soon start to stop supporting it, but we have this construct all over the place in Metals and I think this is also done for parallel-collections https://github.com/scalameta/scalafmt/blob/master/build.sbt#L119

kitbellew (approved):

PR #2396, opened 2021-04-02
BFS: keep policies and indents for provided splits

kitbellew:
Otherwise, we were using an incorrect split which doesn't work outside of the `format:off` area, by losing any policies and indents the original split had.

We now also no longer need the special handling of `{...}`.

In `scala-repos`, only one file is modified, one that `scalafmt` couldn't format before.


tgodzik (approved):
LGTM!

PR #2397, opened 2021-04-02
State: move format:off detection to FormatToken

kitbellew:
This logic does not depend on the splits (hence, no state dependency) and requires only access to the tokens.

Also, make sure that `format:off` applies to the split just before it while `format:on` only applies to the split just after it.


tgodzik (approved):
Looks good!

PR #2401, opened 2021-04-03
ScalafmtConfig: define a new `indent.main` parameter

### LINKED ISSUE #1493: Definition indentation tab width doesn't work
Version: 2.0.0-RC4
Integration: VS Code with Scala (Metals) extension
Configuration:
```
  version = "2.0.0-RC4"
  maxColumn = 80
  continuationIndent.callSite = 8
  continuationIndent.defnSite = 8
  continuationIndent.extendSite = 8
```

## Steps

When I run VS code format document (that calls scalafmt), scalafmt formats code like this:
```scala
final class Goods(
        _food: Int = 0, 
        _materials: Int = 0, 
        _products: Int = 0, 
        _fuel: Int = 0 
) {
  var raw = Array(_food, _materials, _products, _fuel)
  def food: Int = raw(0)
  def materials: Int = raw(1)
  def products: Int = raw(2)
  def fuel: Int = raw(3)
  def weight = raw.sum
}
```
## Expectation

```scala
final class Goods(
        _food: Int = 0, 
        _materials: Int = 0, 
        _products: Int = 0, 
        _fuel: Int = 0 
) {
        var raw = Array(_food, _materials, _products, _fuel)
        def food: Int = raw(0)
        def materials: Int = raw(1)
        def products: Int = raw(2)
        def fuel: Int = raw(3)
}

### LINKED ISSUE #2385: Please support a 4-space indent.
## Problem

Why does this formatter not support 4-space indentation?
I know that 2-space is widely known in the scala community, as talked about in a past issue ( #1493 ), but I think there are many engineers who want to use 4-space.
Are there any technical difficulties?

## Expectation

support a 4-space indent.

### PR DESCRIPTION
**Pull request by kitbellew**
Fixes #2385. Fixes #1493.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
@@ -381,6 +381,7 @@ object ScalafmtConfig {
         "binPack.unsafeXXX && newlines.implicitParamListModifierXXX (not implemented)"
       )
       checkPositive(
+        indent.main,
         indent.callSite,
         indent.defnSite,
         indent.commaSiteRelativeToExtends
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
@@ -105,7 +105,8 @@ case class ScalafmtConfig(
     comments: Comments = Comments(),
     optIn: OptIn = OptIn(),
     binPack: BinPack = BinPack(),
-    continuationIndent: ContinuationIndent = ContinuationIndent(),
+    @annotation.ExtraName("continuationIndent")
+    indent: Indents = Indents(),
     align: Align = Align(),
     spaces: Spaces = Spaces(),
     literals: Literals = Literals(),
@@ -151,7 +152,7 @@ case class ScalafmtConfig(
   private implicit def spacesReader = spaces.reader
   private implicit def literalsReader = literals.reader
   private implicit def xmlLiteralsDecoder = xmlLiterals.decoder
-  private implicit def continuationIndentReader = continuationIndent.reader
+  private implicit def indentReader = indent.reader
   private implicit def binpackReader = binPack.decoder
   private implicit def newlinesReader = newlines.reader
   private implicit def optInReader = optIn.reader
@@ -236,7 +237,7 @@ object ScalafmtConfig {
   val default = ScalafmtConfig()
 
   val intellij: ScalafmtConfig = default.copy(
-    continuationIndent = ContinuationIndent(2, 2),
+    indent = Indents(callSite = 2, defnSite = 2),
     align = default.align.copy(openParenCallSite = false),
     optIn = default.optIn.copy(
       configStyleArguments = false
@@ -261,7 +262,7 @@ object ScalafmtConfig {
       unsafeCallSite = true,
       parentConstructors = BinPack.ParentCtors.Always
     ),
-    continuationIndent = ContinuationIndent(4, 4),
+    indent = Indents(callSite = 4, defnSite = 4),
     importSelectors = ImportSelectors.binPack,
     newlines = default.newlines.copy(
       avoidInResultType = true,
@@ -379,7 +380,7 @@ object ScalafmtConfig {
         },
         "binPack.unsafeXXX && newlines.implicitParamListModifierXXX (not implemented)"
       )
-      addIfNegative(continuationIndent.callSite, continuationIndent.defnSite)
+      addIfNegative(indent.callSite, indent.defnSite)
     }
     if (allErrors.isEmpty) Configured.ok(cfg)
     else {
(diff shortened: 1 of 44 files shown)

### CODE REVIEW

File: scalafmt-tests/src/test/resources/unit/Import.source [hunk 1]
```diff
@@ -181,17 +185,17 @@ object blah {
 >>>
 object blah {
   import a.{
-    // format: off
+      // format: off
```
kitbellew, line 188: NB: this is a bug that #2397 fixes.

olafurpg (commented):
I’m on the fence about this change. Are there feature requests from users asking about these particular settings?

The type signature indentation could maybe share the setting with “defnSite” 🤔 

With Scala 3 and significant indentation around the corner I think we should be careful not to introduce too many new indentation settings.

kitbellew:
@olafurpg not really. i was mainly playing around with the primary indent, partly because of the impending significant indentation and partly because of the general howl around availability of this setting, and wanted to debug some changes one type at a time (and found a couple of bugs along the way, fixing in other changes).

regarding the primary indent: as long as code is not more complex, i don't have any problem with giving people the rope if that's what they want. the perennial tabs-vs-spaces debacle is one example of wisdom of reserving one's opinions.

olafurpg:
My personal opinion remains that Scalafmt shouldn’t support configurable indentation. @kitbellew if your opinion is that this setting should exist then that’s a good enough argument for me. This is not a hill I will die defending! I trust you to make a decision.

kitbellew:
> My personal opinion remains that Scalafmt shouldn’t support configurable indentation. @kitbellew if your opinion is that this setting should exist then that’s a good enough argument for me. This is not a hill I will die defending! I trust you to make a decision.

i think that most people will stick with 2 but usually the most vocal ones are the detractors. as i said, i don't want to tell people how much salt to use in their food. @tgodzik, what do you think?

tgodzik:
> > My personal opinion remains that Scalafmt shouldn’t support configurable indentation. @kitbellew if your opinion is that this setting should exist then that’s a good enough argument for me. This is not a hill I will die defending! I trust you to make a decision.
> 
> i think that most people will stick with 2 but usually the most vocal ones are the detractors. as i said, i don't want to tell people how much salt to use in their food. @tgodzik, what do you think?

I think configurable indentation is fine, especially when it comes to Scala 3, however, I would not split it into too many options. I would prefer to have single option for all of those if possible. Too many options result in verbose configuration and more complexity for maintainers. We've been discussing this back and forth and the overall consensus is that people might think they need all the options, but that just results in useless arguments about the perfect style. Setting one option for indentation though should not cause that mich problem I think. 

That's at least my opinion, I am by far not an expert. I usually am happy with whatever the formatter gives me and each day I am thankful that I don't get formatting comments on most PRs expect in some edge cases.

kitbellew:
> > > My personal opinion remains that Scalafmt shouldn’t support configurable indentation. @kitbellew if your opinion is that this setting should exist then that’s a good enough argument for me. This is not a hill I will die defending! I trust you to make a decision.
> > 
> > 
> > i think that most people will stick with 2 but usually the most vocal ones are the detractors. as i said, i don't want to tell people how much salt to use in their food. @tgodzik, what do you think?
> 
> I think configurable indentation is fine, ... I would prefer to have single option for all of those if possible.

@olafurpg @tgodzik single option it is. the updated commits have been uploaded.

olafurpg (approved):
I think it would be nice to group all of the indentation-related settings under the same object. For example, we could rename `continuationIndent` into `indent` and introduce `indent.default` (exact name up for debate). 

Either way, LGTM.

tgodzik (approved):
LGTM! Thanks for taking care of this. Hopefully, we don't regret adding the option :sweat_smile:

kitbellew:
> I think it would be nice to group all of the indentation-related settings under the same object. For example, we could rename `continuationIndent` into `indent` and introduce `indent.default` (exact name up for debate).

Done.

@olafurpg @tgodzik @poslegm please help me confirm that the first commit i added (removal of indent settings from the dynamic config) is not going to affect anything. (it's package-private, can't imagine why it would be visible outside the repo.)

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -105,7 +105,8 @@ case class ScalafmtConfig(
     comments: Comments = Comments(),
     optIn: OptIn = OptIn(),
     binPack: BinPack = BinPack(),
-    continuationIndent: ContinuationIndent = ContinuationIndent(),
+    @annotation.ExtraName("continuationIndent")
```
kitbellew, line 108: old name is preserved

olafurpg:
> please help me confirm that the first commit i added (removal of indent settings from the dynamic config) is not going to affect anything. 

The only public library API for Scalafmt is the Java interfaces in the `scalafmt-interfaces` module. Everything else can be considered private so feel free to remove/change methods/classes.

olafurpg:
FWIW, `ScalafmtConfig` is a case class, which means that every addition of settings is a binary breaking change on the auto-generated `copy()` method. There shouldn't be any client that's calling those APIs directly outside of the Scalafmt repo.

PR #2403, opened 2021-04-03
Router: simplify and refactor `getSplitsValEquals`

kitbellew:
This method is only applied when `newlines.beforeMultiline` is classic, which in reality means both this parameter and `newlines.source` are not explicitly specified.

Hence, append `Classic` to the name and remove non-classic logic. Use the original method name to put the overall logic.

Also, merge the two similar rules, for `def a = ...` and `val a = ...`.


olafurpg (approved):
LGTM. I’m a bit out of context to comment on the changes. Is this a pure refactoring or are there formatting changes?

kitbellew:
> LGTM. I’m a bit out of context to comment on the changes. Is this a pure refactoring or are there formatting changes?

No formatting changes (at least, not yet). This is to simplify the code and remove unnecessary paths.

PR #2404, opened 2021-04-05
Router: force break before multiline assignments

### LINKED ISSUE #2400: newlines.beforeMultiline = unfold doesn't seem to work correctly
- **Version**: 2.7.5
- **Integration**: cli
- **Configuration**:
```
newlines.beforeMultiline = unfold
```

## Steps

Given code like this (test.scala):
```scala
class Test {
  val a = functionCall(
    1,
    2
  )
}
```

when I run it
```bash
scalafmt tests.scala --stdout
```

Output looks like
```scala
class Test {
  val a = functionCall(
    1,
    2
  )
}
```

while I expected it to be

```scala
class Test {
  val a = 
    functionCall(
      1,
      2
    )
}
```

## Workaround

I couldn't find workaround.

## Notes

See also..

kitbellew:
Expand the deprecated `alwaysBeforeMultilineDef` to have more granular control. Also, if the deprecated `beforeMultilineDef` is not explicitly specified, apply the behaviour of `beforeMultiline` parameter instead.

As discussed in #2400, the option to have consistent formatting is desirable. Fixes #2400.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -798,9 +798,9 @@ package core {
 > Since v2.7.0
 
 This parameter controls whether to force a new line before a multi-line body of
-`case/if/val` and how to format it if the space is allowed. (For multi-line
-bodies of method definitions, please see
-[`newlines.beforeMultilineDef`](#newlinesbeforemultilinedef) below.)
+`case/if/def/val` and how to format it if the space is allowed. (For
+additional control with assignment expressions, please also see
+[`newlines.forceBeforeMultilineAssign`](#newlinesforcebeforemultilineassign) below.)
 
 It accepts the same values as [`newlines.source`](#newlinessource) (and defaults
 to that parameter's setting).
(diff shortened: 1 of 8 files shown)


kitbellew:
@olafurpg @tgodzik @poslegm I'm on the fence about whether this is the right approach.

the desire is to continue supporting the old functionality which forced a break for `def/macro` definitions only, and honor the request to have it optionally supported for all definitions (including `val` etc.). which leads to at least two levels of inclusion.

so now the question is where to do it, and i personally can see it being useful at the class field level but not inside a method, while others want it everywhere...

tgodzik (approved):
I am not totally sure myself what should be the right approach here. My opinion would be to just split everywhere if the user wants that and keep it to a single option while also deprecating the old option. 

It seems a bit complex to add `where` and `what` clauses. Maybe would could just expose a couple of combinations for now?

`anywhere` and `anyassign` would be  `beforeMultiline` setting with `keep` `unfold` etc. And we would use the proper combination for `alwaysBeforeMultilineDef` while also deprecating it.

If the users ever really need that we would expose more combinations, but for now we would not add too much.

Honestly, I don't have that experience with formatting options (usually use defaults) and I am hearing some war stories about too many config options just making it again difficult to decide on the style. That's why I would maybe always keep it to the minimum? I don't feel too strongly here really and the PR looks really solid.

kitbellew:
@tgodzik what about getting rid of `where` (so it's anywhere) and supporting just three values for the parameter: none, any, def?

PR #2405, opened 2021-04-06
[Scala 3] Add support for optional braces

tgodzik:
Optional braces are discussed in depth here https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html

There 4 different ways indentation can start:
- after a keyword token such as:
```
=  =>  ?=>  <-  catch  do  else  finally  for
if  match  return  then  throw  try  while  yield
```
- after `:` at end of line in templates
- after `)` in if or an extension group
- after `with` for given definitions

I created a PR just in case someone wanted to take a look at it. Otherwise, I should be able to finish it by the next week. 

I still need to do:
- [ ] use indent from https://github.com/scalameta/scalafmt/pull/2401
- [ ] more tests!
- [ ] try to figure out if some settings can break optional braces

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
@@ -88,10 +88,6 @@ object ScalafmtRunner {
       .withAllowTrailingCommas(true)
     val default = scala213
     val scala3 = Scala3
-      // Disable significant indentation because the exact syntax is still under
-      // development and is likely to change soon. See discussions at
-      // https://contributors.scala-lang.org/t/feedback-sought-optional-braces/4702
-      .withAllowSignificantIndentation(false)
 
     implicit lazy val decoder: ConfDecoder[Dialect] = {
       ReaderUtil.oneOf[Dialect](
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -177,6 +177,37 @@ class Router(formatOps: FormatOps) {
           if rightOwner.is[SomeInterpolate] =>
         Seq(Split(Space(style.spaces.inInterpolatedStringCurlyBraces), 0))
 
+      // Optional braces after any token that can start indentation:
+      // =  =>  ?=>  <-  catch  do  else  finally  for
+      // if  match  return  then  throw  try  while  yield
+      case FormatToken(left, right, _)
+          if dialect.allowSignificantIndentation &&
+            left.is[CanStartIndent] &&
+            startsStatement(right).isDefined &&
+            rightOwner.parent.exists(_.is[Term.Block]) =>
+        val block = rightOwner.parent.get.asInstanceOf[Term.Block]
+        val last = lastToken(block)
+        Seq(Split(Newline, 0).withIndent(2, last, After))
+
+      // Optional braces in templates after `:` or `with`
+      case FormatToken(_: T.Colon | _: T.KwWith, right, _)
+          if dialect.allowSignificantIndentation &&
+            leftOwner.is[Template] &&
+            startsStatement(right).isDefined =>
+        Seq(Split(Newline, 0).withIndent(2, lastToken(leftOwner), After))
+
+      // Optional braces in templates after if (cond) or extension groups
+      case FormatToken(T.RightParen(), right, _)
+          if dialect.allowSignificantIndentation &&
+            (leftOwner.is[Defn.ExtensionGroup] || leftOwner.is[Term.If]) &&
+            startsStatement(right).isDefined &&
+            rightOwner.parent.exists(_.is[Term.Block]) =>
+        val expire = leftOwner match {
+          case ifClause: Term.If => lastToken(ifClause.thenp)
+          case ext: Defn.ExtensionGroup => lastToken(ext)
+        }
+        Seq(Split(Newline, 0).withIndent(2, expire, After))
+
       // { ... } Blocks
       case tok @ FormatToken(open @ T.LeftBrace(), right, between) =>
         val close = matching(open)
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenClasses.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TokenClasses.scala
@@ -21,6 +21,7 @@ object Keyword {
     token.is[KwObject] || token.is[KwOverride] || token.is[KwPackage] ||
     token.is[KwPrivate] || token.is[KwProtected] || token.is[KwReturn] ||
     token.is[KwSealed] || token.is[KwSuper] || token.is[KwThis] ||
+    token.is[KwThen] ||
     token.is[KwThrow] || token.is[KwTrait] || token.is[KwTrue] ||
     token.is[KwTry] || token.is[KwType] || token.is[KwVal] ||
     token.is[KwVar] || token.is[KwWhile] || token.is[KwWith] ||
@@ -117,4 +118,18 @@ class SoftKeywordClasses(dialect: Dialect) extends SoftKeywords(dialect) {
       tok.is[KwExtends] || tok.is[KwDerives]
     }
   }
+}
+
+/*
+=  =>  ?=>  <-  catch  do  else  finally  for
+if  match  return  then  throw  try  while  yield
+ */
+@classifier
+trait CanStartIndent {
+  def unapply(tok: Token): Boolean =
+    tok.is[Equals] || tok.is[RightArrow] || tok.is[ContextArrow] ||
+      tok.is[LeftArrow] || tok.is[KwCatch] || tok.is[KwDo] || tok.is[KwElse] ||
+      tok.is[KwElse] || tok.is[KwFinally] || tok.is[KwFor] || tok.is[KwMatch] ||
+      tok.is[KwReturn] || tok.is[KwThen] || tok.is[KwThrow] || tok.is[KwTry] ||
+      tok.is[KwWhile] || tok.is[KwYield]
 }
--- a/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
+++ b/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
@@ -1,0 +1,112 @@
+
+<<< simple value equals
+val test = 
+    val a = ""
+     a  +    ""
+>>>
+val test =
+  val a = ""
+  a + ""
+<<< if else
+trait A:
+  val cond = if true then
+    stat1
+    stat2
+  else
+      stat3
+      stat4
+>>>
+trait A:
+  val cond =
+    if true then
+      stat1
+      stat2
+    else
+      stat3
+      stat4
+<<< object
+object Obj:
+  def hello = 
+      1
+       2
+>>>
+object Obj:
+  def hello =
+    1
+    2
+<<< object with braces
+object Obj{
+  def hello = 
+    1
+    2
+}
+>>>
+object Obj {
+  def hello =
+    1
+    2
+}
+<<< extension method
+maxColumn = 40
+===
+extension [A](a: Map[A, Foooooooooooooooo[B]]) 
+    def add(b: A) = a + b
+     def add2(b: A) = a + b
+  
+    def add3(b: A) = a + b
+>>>
+extension [A](
+    a: Map[A, Foooooooooooooooo[B]]
+)
+  def add(b: A) = a + b
+  def add2(b: A) = a + b
+
+  def add3(b: A) = a + b
+<<< extension multi
+maxColumn = 40
+===
+extension [A](a: Map[A, Foooooooooooooooo[B]]) (using b: Map[A, Foooooooooooooooo[B]])
+    def add(b: A) = a + b
+     def add2(b: A) = a + b
+  
+    def add3(b: A) = a + b
+>>>
+extension [A](
+    a: Map[A, Foooooooooooooooo[B]]
+)(using b: Map[A, Foooooooooooooooo[B]])
+  def add(b: A) = a + b
+  def add2(b: A) = a + b
+
+  def add3(b: A) = a + b
+<<< if(cond) indentation 
+trait A:
+  val cond =
+    if (true)
+        stat1
+         stat2
+    else
+       stat3
+       stat4
+>>>
+trait A:
+  val cond =
+    if (true)
+      stat1
+      stat2
+    else
+      stat3
+      stat4
+<<< given with
+given intOrd: Ord[Int] with Eq[Int] with
+    def compare(x: Int, y: Int) =
+      if x < y then -1 else if x > y then +1 else 0
+     def compare2(x: Int, y: Int) =
+      if x < y then -1 else if x > y then +1 else 0
+>>>
+given intOrd: Ord[Int]
+  with Eq[Int]
+  with
+    def compare(x: Int, y: Int) =
+      if x < y then -1 else if x > y then +1 else 0
+    def compare2(x: Int, y: Int) =
+      if x < y then -1 else if x > y then +1 else 0


kitbellew:
@tgodzik once again, i have a slightly simplified version for this commit. let me know how to show you.

i added more tests (comments, end markers) and generalized the block logic (instead of checking for token classes, i assumed the parser would have done it, so just looking for a `Term.Block` without a `{`).

one possible bug i found in the parser: `end if` marker is not included as part of the `Term.If` tree and leads to incorrect indentation.

tgodzik:
> @tgodzik once again, i have a slightly simplified version for this commit. let me know how to show you.
> 
> i added more tests (comments, end markers) and generalized the block logic (instead of checking for token classes, i assumed the parser would have done it, so just looking for a `Term.Block` without a `{`).

I think we can again do a follow up PR. Thanks for taking a look! I wasn't sure when I would be able to look at it further and this was mostly done (`ish`).

> 
> one possible bug i found in the parser: `end if` marker is not included as part of the `Term.If` tree and leads to incorrect indentation.

Hmm... that is actually on purpose since we made the end statement a separate statement and were hoping that would work out alright. The indentation level should be the same as `if`'s - do you have a test case this is not working in? I'd rather think it would not work if it were to be included in `Term.If`

kitbellew:
> > one possible bug i found in the parser: `end if` marker is not included as part of the `Term.If` tree and leads to incorrect indentation.
> 
> Hmm... that is actually on purpose since we made the end statement a separate statement and were hoping that would work out alright. The indentation level should be the same as `if`'s - do you have a test case this is not working in? I'd rather think it would not work if it were to be included in `Term.If`

example of incorrect formatting:
```
trait A /* comm */:
   val cond =
     if true then
        stat1
        stat2
     else
        stat3
        stat4
   end if
```

others do include the marker.

kitbellew:
actually, the examples where end marker _seemed_ to work are for cases where indentation covers a template and hence includes any end markers.

let me see if the example in the previous comment can be handled in the formatter differently, without changing the parser.

kitbellew:
> let me see if the example in the previous comment can be handled in the formatter differently, without changing the parser.

wait, the body (`rhs`) of the `val` needs to include the marker, right? so it needs to be some tree which includes both `if` end `end if`. should there be a `Term.WithEndMarker` which contains two fields?

tgodzik:
> > let me see if the example in the previous comment can be handled in the formatter differently, without changing the parser.
> 
> wait, the body (`rhs`) of the `val` needs to include the marker, right? so it needs to be some tree which includes both `if` end `end if`. should there be a `Term.WithEndMarker` which contains two fields?

It should be a Block that contains `Term.If` and `Term.End`

tgodzik:
Indetation that we accept should be:
```scala
trait A /* comm */:
   val cond =
     if true then
        stat1
        stat2
     else
        stat3
        stat4
     end if
```

Does Scala 3 accept indentation on the level of `val`?

kitbellew:
> Indetation that we accept should be:
> 
> ```scala
> trait A /* comm */:
>    val cond =
>      if true then
>         stat1
>         stat2
>      else
>         stat3
>         stat4
>      end if
> ```
> 
> Does Scala 3 accept indentation on the level of `val`?

the parser will accept this expression but the formatter dedents end marker because it's not part of the `val cond` body.

generally, i keep getting parse errors if there are several end markers:
```
// this works
trait A:
  val cond =
    if true then
      stat1
      stat2
    else
      stat3
      stat4
  val cond =
    if true then
      stat1
      stat2
    else
      stat3
      stat4

// this fails in the parser
trait A:
  val cond =
    if true then
      stat1
      stat2
    else
      stat3
      stat4
    end if
  val cond =
    if true then
      stat1
      stat2
    else
      stat3
      stat4
    end if
end trait
```

kitbellew:
> > > let me see if the example in the previous comment can be handled in the formatter differently, without changing the parser.
> > 
> > 
> > wait, the body (`rhs`) of the `val` needs to include the marker, right? so it needs to be some tree which includes both `if` end `end if`. should there be a `Term.WithEndMarker` which contains two fields?
> 
> It should be a Block that contains `Term.If` and `Term.End`

block is too generic, no? this needs to be structured. either a special wrapper type, or stick `end` as part of `if` (add an optional `end` marker field as the end).

i'd go with the second approach if possible, and add a method to figure out the right end marker label as well:
```
class If(...., end: Option[...]) {
  def getEndLabel: String = "if"
}
class Class(...., end: Option[...]) {
  def getEndLabel: String = name.value
}
```

tgodzik:
> block is too generic, no? this needs to be structured. either a special wrapper type,

`end` just acts like a statement, I don't think we need to include it in different trees. Especially that I don't think we have currently a use for it.

> stick end as part of if (add an optional end marker field as the end).

That would be adding the field to a lot of different trees :O It might be more correct, but we wanted to avoid it since it's really more of a visual marker than something that really affects the code.

I need to think about this one a bit :thinking:

kitbellew:
> > block is too generic, no? this needs to be structured. either a special wrapper type,
> 
> `end` just acts like a statement, I don't think we need to include it in different trees. Especially that I don't think we have currently a use for it.
> 
> > stick end as part of if (add an optional end marker field as the end).
> 
> That would be adding the field to a lot of different trees :O It might be more correct, but we wanted to avoid it since it's really more of a visual marker than something that really affects the code.
> 
> I need to think about this one a bit 🤔

why do you need a separate tree for this? why not treat is as a pair of tokens at the end of the tree? you don't extract the braces which surround a block (and some blocks don't have braces).

tgodzik:
> 
> why do you need a separate tree for this? why not treat is as a pair of tokens at the end of the tree? you don't extract the braces which surround a block (and some blocks don't have braces).

We thought of different approaches and this one seemed to cause least amount of trouble, might have been totally wrong about that one.

kitbellew:
> > why do you need a separate tree for this? why not treat is as a pair of tokens at the end of the tree? you don't extract the braces which surround a block (and some blocks don't have braces).
> 
> We thought of different approaches and this one seemed to cause least amount of trouble, might have been totally wrong about that one.

let me see if i missed something in the formatter.

kitbellew:
> > > why do you need a separate tree for this? why not treat is as a pair of tokens at the end of the tree? you don't extract the braces which surround a block (and some blocks don't have braces).
> > 
> > 
> > We thought of different approaches and this one seemed to cause least amount of trouble, might have been totally wrong about that one.
> 
> let me see if i missed something in the formatter.

nope: here's the tree
```
 Template(
  Nil,
  Nil,
  Self(Name(""), None),
  List(
   Defn.Val(
    Nil, 
    List(
     Pat.Var(
      Term.Name("cond")
     )
    ),
    None, 
    Term.If(
     Lit.Boolean(true), 
     Term.Block(List(Term.Name("stat1"), Term.Name("stat2"))), 
     Term.Block(List(Term.Name("stat3"), Term.Name("stat4"))), 
     Nil
    )
   ),
   Term.EndMarker(Term.Name("if"))
  ),
  Nil
)
```
and the end marker is not part of val. i'd say that's a bug.

tgodzik:
> and the end marker is not part of val. i'd say that's a bug.

Yeah, that's a bug. I wonder though if we should treat the `end` like `}` - include it, but not additional specific structure maybe?
I think we don't really need it from the parser perspective, it's kind of like a comment. I think you mentioned this solution somewhere. What do you think? I can work on it in a bit in the parser

kitbellew:
> > and the end marker is not part of val. i'd say that's a bug.
> 
> Yeah, that's a bug. I wonder though if we should treat the `end` like `}` - include it, but not additional specific structure maybe?
> I think we don't really need it from the parser perspective, it's kind of like a comment. I think you mentioned this solution somewhere. What do you think? I can work on it in a bit in the parser

do you remember the reasons you were discussing which led to having a separate tree for the end?

from the formatter perspective, i don't see a reason and would include it just as a pair of tokens, provided that these tokens are part of class/object/trait/enum/def/macro/val/if etc but **not** part of its template or block which would normally contain braces.

a lot of the work in the formatter went into looking for a closing brace so that we can expire indent before it. if we don't need to look for it, code is a lot cleaner.

tgodzik:
> 
> do you remember the reasons you were discussing which led to having a separate tree for the end?
> 

I think when we discussed it with Olaf, he mentioned it would be a more permissive parser and that's why we decided to have it as a separate statement. We didn't decide on having `BlockWithEnd` or `TemplateWithEnd` since that would create a larger number of trees and having a separate Boolean flag is something we that we never do.

> from the formatter perspective, i don't see a reason and would include it just as a pair of tokens, provided that these tokens are part of class/object/trait/enum/def/macro/val/if etc but **not** part of its template or block which would normally contain braces.
> 

I think we wanted to keep it for scalafix.

kitbellew:
> > do you remember the reasons you were discussing which led to having a separate tree for the end?
> 
> We didn't decide on having `BlockWithEnd` or `TemplateWithEnd` since that would create a larger number of trees and having a separate Boolean flag is something we that we never do.

we do have optional fields in many trees, like `decltpe`, so adding optional end to block and template wouldn't be unreasonable.

tgodzik:
Btw. I just realised that the parser is failing correctly:
```
trait A:
  val cond =
    if true then
      stat1
      stat2
    else
      stat3
      stat4
    end if
  val cond =
    if true then
      stat1
      stat2
    else
      stat3
      stat4
    end if
end trait
```

`end trait` should be `end A` - anything that is named should have the name after the end marker.

kitbellew:
> > > do you remember the reasons you were discussing which led to having a separate tree for the end?
> > 
> > 
> > We didn't decide on having `BlockWithEnd` or `TemplateWithEnd` since that would create a larger number of trees and having a separate Boolean flag is something we that we never do.
> 
> we do have optional fields in many trees, like `decltpe`, so adding optional end to block and template wouldn't be unreasonable.

oh :( i am contradicting myself... i wanted the marker to be out of template/block and in the body of `val` :( ok, if it makes the parser easier to modify and satisfies other constraints, we can add them into block and template, and the formatter will have to handle these two cases.

and if we do that, can we put the closing `}` in the end marker when braces are used (and only keep it `None` if neither braces nor the `end` marker are present)?

tgodzik:
> 
> oh :( i am contradicting myself... i wanted the marker to be out of template/block and in the body of `val` :( ok, if it makes the parser easier to modify and satisfies other constraints, we can add them into block and template, and the formatter will have to handle these two cases.
> 
> and if we do that, can we put the closing `}` in the end marker when braces are used (and only keep it `None` if neither braces nor the `end` marker are present)?

Not sure I understand. What are you thinking that the tree should look like? Keep the current behaviour and add `}` as a possible value for the end marker?

kitbellew:
> oh :( i am contradicting myself... i wanted the marker to be out of template/block and in the body of `val` :( ok, if it makes the parser easier to modify and satisfies other constraints, we can add them into block and template, and the formatter will have to handle these two cases.
> 
> and if we do that, can we put the closing `}` in the end marker when braces are used (and only keep it `None` if neither braces nor the `end` marker are present)?

well, it would also be ideal to have a begin marker as well :) one that refers to opening `{` if used, or `:`, `then`, `)`, `with` or any other token after which the indented contents of template/block start.

kitbellew:
> > and if we do that, can we put the closing `}` in the end marker when braces are used (and only keep it `None` if neither braces nor the `end` marker are present)?
> 
> Not sure I understand. What are you thinking that the tree should look like? Keep the current behaviour and add `}` as a possible value for the end marker?

```
trait BeginMarker extends Tree
// points to opening brace of template/block
class BraceBeginMarker extends BeginMarker
// points to colon before missing-braces template/block
class ColonEolBeginMarker extends BeginMarker
// points to final "with" before missing-braces block (or we can push into next one)
class GivenWithBeginMarker extends BeginMarker

trait EndMarker extends Tree
// points to closing brace of template/block
class BraceEndMarker extends EndMarker
// points to "end label" after missing-braces template/block
class LabelEndMarker extends EndMarker

class Block {
  val begin: Option[BeginMarker]
  val stats: Seq[Stat]
  val end: Option[EndMarker]
}

// simple option
class Template {
  val inits...
  val derives...
  val early...
  val beginContents: Option[BeginMarker]
  val self: Option[Self]
  val stats: Seq[Stat]
  val endContents: Option[EndMarker]
}

// complex option
class Template {
  val inits...
  val derives...
  val early...
  val contents: TemplateContents
} {
  def self = contents.self
  def stats = contents.stats
}
class TemplateContents {
  val begin: Option[BeginMarker]
  val self: Option[Self]
  val stats: Seq[Stat]
  val end: Option[EndMarker]
}
```

by the way, what is `early` in template definition?

tgodzik:
> > oh :( i am contradicting myself... i wanted the marker to be out of template/block and in the body of `val` :( ok, if it makes the parser easier to modify and satisfies other constraints, we can add them into block and template, and the formatter will have to handle these two cases.
> > and if we do that, can we put the closing `}` in the end marker when braces are used (and only keep it `None` if neither braces nor the `end` marker are present)?
> 
> well, it would also be ideal to have a begin marker as well :) one that refers to opening `{` if used, or `:`, `then`, `)`, `with` or any other token after which the indented contents of template/block start.

I think I'd rather just remove `Term.End` altogether. This seems quite a big change and I don't really see the benefit here. 

After thinking about it and rethinking it, my opinion is that a separate statement fits nicely with the Scala 3 logic. It needs to be indented the same the tree it's closing and also it needs to be in an Indented Block. The current behavior seems to work ok, so maybe let's leave it as is?

The example in the tests https://github.com/scalameta/scalafmt/pull/2406/files#r608136671 is not actually correct, so we are good there.

tgodzik:
Also things like this are actually correct:
```scala
    val cond = {
      if true then
        stat1
        stat2
      else { // c1
        stat3
        stat4
      } 
      end if
    }
```
we would need to have two closing tokens here :O

I think statement solves the problem really.

kitbellew:
> The current behavior seems to work ok, so maybe let's leave it as is?
> 
> The example in the tests https://github.com/scalameta/scalafmt/pull/2406/files#r608136671 is not actually correct, so we are good there.

@tgodzik sorry, i must have made a typo. but, the parser is failing with `end A`:
```
.../scala3/OptionalBraces.stat:18: error: end of file expected but identifier found
end A
^

trait A:
  val cond =
    if true then
      stat1
      stat2
    else
      stat3
      stat4
    end if
  val cond =
    if true then
      stat1
      stat2
    else
      stat3
      stat4
    end if
end A
```

tgodzik:
> OptionalBraces.stat

It's because it's a `stat` file and `end` is a separate statement, we would need to do `OptionalBraces.source` file.

kitbellew:
the example that is incorrectly indented in the tests is this:
```
trait A extends B:
  val cond = if true then
    stat1
    stat2
  else
     stat3
     stat4
  end if
=== becomes
trait A extends B:
   val cond =
     if true then
        stat1
        stat2
     else
        stat3
        stat4
   end if
```
and this fails parsing:
```
.../scala3/OptionalBraces.stat:9: error: ; expected but identifier found
  end val
  ^

trait A extends B:
  val cond = if true then
    stat1
    stat2
  else
     stat3
     stat4
  end if
  end val
```
and this one, too:
```
/Users/ameltze/dev/code/scalafmt/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat:2: error: illegal start of simple expression
  val cond =
            ^
    /* c1*/ if true then
     stat1
trait A extends B:
  val cond =
    /* c1*/ if true then
     stat1
     stat2
    else
     stat3
     stat4
    end if
  end val
```
the question is: how should it work? is the end marker allowed at all if the beginning didn't start on its own line?

this page doesn't explain how end markers need to be indented: https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html#the-end-marker

tgodzik:
The first two seem to fail correctly (Dotty does not accept that syntax), the last one might be a bug on our side.

> the question is: how should it work? is the end marker allowed at all if the beginning didn't start on its own line?
> 

I think that is the case from what I tested. It needs to be in an indented block if we want to use `end` marker.

> this page doesn't explain how end markers need to be indented: https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html#the-end-marker

Yeah, optional braces are not explained perfectly and I am getting into a lot of edge cases :/

tgodzik:
Closing this one in favour of the other PR

kitbellew:
@tgodzik optional braces is not done. `match` and `catch` are not followed by block or template.

kitbellew:
@tgodzik potential bug in the parser: in this expression
```
   this match
      case a =>
         that match
            case b => bb
            case c => cc
         end match
      case b =>
         that match
            case c => cc
            case _ => dd
```
the tree for `case a` includes `case b`. if i remove the end marker for nested match within `case a`, it works.

PR #2406, opened 2021-04-06
[Scala 3] Add support for optional braces

kitbellew:
Optional braces are discussed in depth in
https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html

There 4 different ways indentation can start:
- after a keyword token such as:
```
=  =>  ?=>  <-  catch  do  else  finally  for
if  match  return  then  throw  try  while  yield
```
- after `:` at end of line in templates
- after `)` in if or an extension group
- after `with` for given definitions

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Indents.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Indents.scala
@@ -3,6 +3,7 @@ package org.scalafmt.config
 import metaconfig._
 
 /** @param main the primary indentation used in the code
+  * @param significant the indentation used when optional braces are omitted
   * @param defnSite indentation around class/def
   * @param ctorSite indentation around class constructor parameters
   * @param caseSite indentation for case values before arrow
@@ -13,6 +14,7 @@ import metaconfig._
   */
 case class Indents(
     main: Int = 2,
+    private[config] val significant: Option[Int] = None,
     callSite: Int = 2,
     defnSite: Int = 4,
     caseSite: Int = 4,
@@ -24,6 +26,8 @@ case class Indents(
 ) {
   implicit val reader: ConfDecoder[Indents] =
     generic.deriveDecoder(this).noTypos
+
+  lazy val getSignificant = significant.getOrElse(main)
 
   def getDefnSite(tree: meta.Tree): Int =
     (tree match {
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -1907,6 +1907,30 @@ class FormatOps(
     }
   }
 
+  // Optional braces in templates after `:|with`
+  object OptionalBracesTemplate {
+    def unapply(meta: FormatToken.Meta): Option[Template] =
+      meta.leftOwner match {
+        case t: Template if templateCurly(t).contains(tokens(meta.idx).left) =>
+          Some(t)
+        case _ => None
+      }
+  }
+
+  // Optional braces after any token that can start indentation:
+  // )  =  =>  ?=>  <-  catch  do  else  finally  for
+  // if  match  return  then  throw  try  while  yield
+  object OptionalBracesBlock {
+    def unapply(meta: FormatToken.Meta): Option[Term.Block] = {
+      val nft = nextNonComment(tokens(meta.idx))
+      nft.meta.rightOwner.parent match {
+        case Some(t: Term.Block) if t.tokens.headOption.contains(nft.right) =>
+          Some(t)
+        case _ => None
+      }
+    }
+  }
+
 }
 
 object FormatOps {
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -173,6 +173,28 @@ class Router(formatOps: FormatOps) {
           if rightOwner.is[SomeInterpolate] =>
         Seq(Split(Space(style.spaces.inInterpolatedStringCurlyBraces), 0))
 
+      // optional braces: template follows
+      case FormatToken(
+            _: T.Colon | _: T.KwWith,
+            _,
+            OptionalBracesTemplate(owner)
+          ) if dialect.allowSignificantIndentation =>
+        val indent = style.indent.getSignificant
+        Seq(Split(Newline, 0).withIndent(indent, lastToken(owner), After))
+
+      // optional braces: block follows
+      case FormatToken(
+            _: T.Equals | _: T.RightArrow | _: T.ContextArrow | _: T.LeftArrow |
+            _: T.KwCatch | _: T.KwDo | _: T.KwElse | _: T.KwFinally |
+            _: T.KwFor | _: T.KwMatch | _: T.KwReturn | _: T.KwThen |
+            _: T.KwThrow | _: T.KwTry | _: T.KwWhile | _: T.KwYield |
+            _: T.RightParen,
+            _,
+            OptionalBracesBlock(owner)
+          ) if dialect.allowSignificantIndentation =>
+        val indent = style.indent.getSignificant
+        Seq(Split(Newline, 0).withIndent(indent, lastToken(owner), After))
+
       // { ... } Blocks
       case tok @ FormatToken(open @ T.LeftBrace(), right, between) =>
         val close = matching(open)
--- a/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
+++ b/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
@@ -1,0 +1,277 @@
+indent.significant = 3
+<<< simple value equals
+val test = // comm
+    val a = ""
+     a  +    ""
+>>>
+val test = // comm
+   val a = ""
+   a + ""
+<<< if else with comment before colon
+trait A /* comm */ :
+  val cond =
+   if true then
+    stat1
+    stat2
+   else { // c1
+     stat3
+     stat4
+     }
+   end if
+>>>
+trait A /* comm */:
+   val cond =
+      if true then
+         stat1
+         stat2
+      else { // c1
+        stat3
+        stat4
+      }
+      end if
+<<< nested class with end marker
+trait A /* comm */ :
+  class B:
+    val a = ""
+  end B
+>>>
+trait A /* comm */:
+   class B:
+      val a = ""
+   end B
+<<< object
+object Obj:
+  def hello = 
+      1
+       2
+  end hello
+>>>
+object Obj:
+   def hello =
+      1
+      2
+   end hello
+<<< object with braces
+object Obj{
+  def hello = 
+    1
+    2
+}
+>>>
+object Obj {
+  def hello =
+     1
+     2
+}
+<<< extension method
+maxColumn = 40
+===
+extension [A](a: Map[A, Foooooooooooooooo[B]]) 
+    def add(b: A) = a + b
+     def add2(b: A) = a + b
+  
+    def add3(b: A) = a + b
+>>>
+extension [A](
+    a: Map[A, Foooooooooooooooo[B]]
+)
+   def add(b: A) = a + b
+   def add2(b: A) = a + b
+
+   def add3(b: A) = a + b
+<<< extension multi
+maxColumn = 40
+===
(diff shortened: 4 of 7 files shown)


kitbellew:
@tgodzik this is based on your #2405, ptal.

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -0,0 +1,130 @@
...
+trait A /* comm */:
+   val cond =
+     if true then
+        stat1
+        stat2
+     else { // c1
+       stat3
+       stat4
+     }
+   end if
```
kitbellew, line 30: wrong indentation here

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -0,0 +1,130 @@
...
+<<< nested class with end marker
+trait A /* comm */ :
+  class B:
+    val a = ""
+  end B
+>>>
+trait A /* comm */:
+   class B:
+      val a = ""
+   end B
```
kitbellew, line 40: indentation is correct but because it's all part of a template

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1909,6 +1909,39 @@ class FormatOps(
...
   }
 
+  object OptionalBraces {
+    def unapply(ft: FormatToken): Option[(FormatToken, Tree)] = ft.left match {
+      case _ if !dialect.allowSignificantIndentation => None
+      case _: T.Comment | _: T.LeftBrace => None
+      case _: T.Colon => // Optional braces in templates after `:`
+        ft.meta.leftOwner match {
+          case t: Template =>
+            val ok = t.tokens.headOption.contains(ft.left)
```
kitbellew, line 1919: this uses the fact that template starts with `:`, although i would have expected colon to serve as a separator.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1909,6 +1909,39 @@ class FormatOps(
...
+            if (ok) Some((ft, t)) else None
+          case _ => None
+        }
+      case _: T.KwWith => // Optional braces in templates after `with`
+        ft.meta.leftOwner match {
+          case t: Template =>
+            val nextFt = nextNonComment(ft)
+            val ok = nextFt.meta.rightOwner.parent.contains(t.self)
```
kitbellew, line 1927: this is slightly hacky, mostly because it's unintuitive to me why the first `def` in this template is owned by the `self` field of the template, and not the template itself. i wonder if there's some other token possible which would start the template but wouldn't be owned by `self`...

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1909,6 +1909,39 @@ class FormatOps(
...
+          case _ => None
+        }
+      case _ =>
+        // Optional braces after any token that can start indentation:
+        // =  =>  ?=>  <-  catch  do  else  finally  for
+        // if  match  return  then  throw  try  while  yield
+        val nextFt = nextNonComment(ft)
+        nextFt.meta.rightOwner.parent match {
+          case Some(t: Term.Block) =>
+            val ok = t.tokens.headOption.contains(nextFt.right)
```
kitbellew, line 1938: this is simple enough. it wouldn't match regular `{}` blocks because block is the parent of `{`, but with optional braces, block is the grandparent of the next non-comment token, and that that non-comment token is the first in the block.

kitbellew (commented):
listing some assumptions from the parser. not sure if they are reliable.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Indents.scala
```diff
@@ -13,6 +14,7 @@ import metaconfig._
   */
 case class Indents(
     main: Int = 2,
+    significant: Int = 3,
```
tgodzik, line 17: I would just use  `main` this way the diffs will be minimal if people switch to 3 and start using it.

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
tgodzik: I haven't dug into that too much. I will try to figure exactly what is the problem here.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Indents.scala
kitbellew: you mean:
```
significant: Int = 2
```
or
```
significant: Option[Int] = None
```
followed by
```
def getSignificant = significant.getOrElse(main)
```

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
kitbellew: i will look, too, if you said these two (`if` and `end if`) should be packaged as a block.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Indents.scala
tgodzik: I would just use main and not add a new setting.

kitbellew: oh... i didn't expect that :) i thought you said in another discussion that martin recommended 3 for significant indentation...

tgodzik: That's the recommendation, but we were far from consensus there. If we do default of 2 then at least when people drop optional braces it will not cause additional changes. They can change to 3 or 4 if they want to.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
kitbellew: i got rid of it, using `templateCurly` instead. `self` is inconsistent; sometimes its tokens include the first body token, sometimes they don't (braces and no self annotation).

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
kitbellew: this doesn't work either. if there are extends, colon is not the first token. also switched to `templateCurly`.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -173,6 +173,28 @@ class Router(formatOps: FormatOps) {
...
+            _: T.Colon | _: T.KwWith,
+            _,
+            OptionalBracesTmpl(owner)
+          ) if dialect.allowSignificantIndentation =>
+        val indent = style.indent.getSignificant
+        Seq(Split(Newline, 0).withIndent(indent, lastToken(owner), After))
+
+      // optional braces: block follows
+      case ft @ FormatToken(
+            _: T.Equals | _: T.RightArrow | _: T.ContextArrow | _: T.LeftArrow |
```
kitbellew, line 187: tried avoiding these, since the parser realistically already knows which tokens can be followed by a block without braces. but java kept blowing up, probably because matching works better on case classes and not on `unapply` extractors. so using an extractor only on the `meta` field (which is really sufficient to match on the entire token because it contains the token index).

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1909,6 +1910,30 @@ class FormatOps(
     }
   }
 
+  // Optional braces in templates after `:|with`
+  object OptionalBracesTmpl {
+    def unapply(meta: FormatToken.Meta): Option[Template] =
+      meta.leftOwner match {
+        case t: Template if templateCurly(t).contains(tokens(meta.idx).left) =>
```
kitbellew, line 1917: no more `self`, just using `templateCurly` (which also correctly handles comments).

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1909,6 +1910,30 @@ class FormatOps(
...
+        case _ => None
+      }
+  }
+
+  // Optional braces after any token that can start indentation:
+  // ) =  =>  ?=>  <-  catch  do  else  finally  for
+  // if  match  return  then  throw  try  while  yield
+  object OptionalBracesBlck {
+    def unapply(meta: FormatToken.Meta): Option[Term.Block] = {
+      val nft = nextNonComment(tokens(meta.idx))
```
kitbellew, line 1928: allows comments between keyword and the start of the braceless block. also, handles `)` the same way, no need to clarify extension or if.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -406,16 +406,17 @@ class FormatOps(
...
-
-  def templateDerivesOrCurly(template: Template): Option[Token] = {
-    template.tokens.collectFirst {
-      case t: T.LeftBrace if owners(t) eq template => t
-      case t @ soft.KwDerives() if owners(t) eq template => tokens(t, -1).left
-    }
-  }
+  def templateCurly(template: Template): Option[Token] =
+    getStartOfTemplateBody(template)
+      .map(x => prevNonComment(tokens(x, -1)).left)
```
kitbellew, line 411: avoids iteration over all tokens, finds it exactly where it should be. handles `colon at end of line` as well.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -406,16 +406,17 @@ class FormatOps(
...
+
+  def templateDerivesOrCurly(template: Template): Option[Token] =
```
kitbellew, line 413: if `inits` are non-empty, finds next token after the end. else, falls back to `templateCurly`.

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
tgodzik: Actually, this doesn't parse in Dotty:
```
  trait A /* comm */ :
    val cond = if true then
      stat1
      stat2
    else  // c1
      stat3
      stat4
    end if
```

It complains about misaligned end statement. You need an actual indentation to have an end statement (I think?)

kitbellew: thank you, modified this!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1907,6 +1907,30 @@ class FormatOps(
...
+        case t: Template if templateCurly(t).contains(tokens(meta.idx).left) =>
+          Some(t)
+        case _ => None
+      }
+  }
+
+  // Optional braces after any token that can start indentation:
+  // )  =  =>  ?=>  <-  catch  do  else  finally  for
+  // if  match  return  then  throw  try  while  yield
+  object OptionalBracesBlck {
```
tgodzik, line 1923:
```suggestion
  object OptionalBracesBlock {
```
 it's just a letter more?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1907,6 +1907,30 @@ class FormatOps(
     }
   }
 
+  // Optional braces in templates after `:|with`
+  object OptionalBracesTmpl {
```
tgodzik, line 1911:
```suggestion
  object OptionalBracesTemplate {
```
?

tgodzik (approved):
LGTM! Thanks for working on this! Pretty excited about it, we should be able to publish an alpha? I just added two small comments, but feel free to ignore them.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
kitbellew: :) will do. i think it was overflowing the line somewhere.

kitbellew: done, uploaded

PR #2411, opened 2021-04-08
Router: handle optional braces in this/catch/match/for-do/for-yield

kitbellew:
These two keywords are not followed by a block, hence the existing logic doesn't work for them.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -1943,8 +1943,8 @@ class FormatOps(
   }
 
   // Optional braces after any token that can start indentation:
-  // )  =  =>  ?=>  <-  catch  do  else  finally  for
-  // if  match  return  then  throw  try  while  yield
+  // )  =>  ?=>  <-  do  else  finally  for
+  // if  return  then  throw  try  while  yield
   object OptionalBracesBlock {
     def unapply(meta: FormatToken.Meta): Option[OptionalBraces] = {
       val nft = nextNonComment(meta)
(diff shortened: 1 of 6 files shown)


scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -275,3 +275,57 @@ object a {
...
+>>>
+enum IndentWidth:
+   def foo =
+     this match
+        case A =>
+          that match
+             case b => bb
+             case c => cc
+          end match
+          case B => // should be at same level as case A
```
tgodzik, line 300: Looks like an issue with the parser, taking a look into it!

tgodzik: https://github.com/scalameta/scalameta/pull/2276

tgodzik: That's merged, should we release a new version, bump it here with the fix and tag the release candidate? Or maybe did you find some more issues to fix?

kitbellew: i am adding as many examples i can find online, to see how this works.

i also tried to implement the rewrite dropping braces, and that's also where some of the issues were detected.

tgodzik: ok cool, I will hold off on releasing then

kitbellew: i think we could release the parser, hopefully there are no more issues with it. that way, we can test the formatter with the updated version.

kitbellew:
> These two keywords are not followed by a block, hence the existing logic doesn't work for them.

@tgodzik general question: should the parser emit a block even when braces are missing and only one statement is there?

tgodzik:
We don't emit a block then. It seemed to work similar in Dotty. Single statements don't require indentation, so I did not want the AST change if we just formatted the code.

tgodzik (approved):
Looks good! I will take a look at the other issue tomorrow, but I think we could release the RC with this ready to get people find more issues 😅

kitbellew:
@tgodzik found another case (secondary ctor). its body looks like a block but in reality it splits into init/stats, kind of like match and catch.

i need to test a few more things; the `Trees.scala` file contains a few more with stats (`Pkg`, `Source`, `PartialFunction`, `For`, `ForYield`, `EnumCase`, `Importer`).

There are also `Match`, `Refine` and `Existential` types, I don't know what they look like.

kitbellew:
> i need to test a few more things; the `Trees.scala` file contains a few more with stats (`Pkg`, `Source`, `PartialFunction`, `For`, `ForYield`, `EnumCase`, `Importer`).

Added `for-do`, `for-yield`, they required a forced newline before `do` and `yield`. Also added single-line option for single-stat "blocks". Finally, added keep/fold/unfold versions.

Any thoughts on what else could be missing? Other test cases to add?

kitbellew:
> There are also `Match`, `Refine` and `Existential` types, I don't know what they look like.

added `Type.Match`. don't think others use optional braces...

tgodzik:
I'll take a look and maybe add some tests later on. Most of the testing on parser was done here: https://github.com/scalameta/scalameta/blob/master/tests/shared/src/test/scala/scala/meta/tests/parsers/dotty/SignificantIndentationSuite.scala - there might be some useful test cases there.

tgodzik (approved):
Looks good! Amazing amount of tests. I will merge this one a follow up with some edge cases, that I had to deal with while parsing :tada:

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1907,30 +1910,121 @@ class FormatOps(
...
   // Optional braces after any token that can start indentation:
-  // )  =  =>  ?=>  <-  catch  do  else  finally  for
-  // if  match  return  then  throw  try  while  yield
+  // )  =>  ?=>  <-  do  else  finally  for
+  // if  return  then  throw  try  while  yield
   object OptionalBracesBlock {
-    def unapply(meta: FormatToken.Meta): Option[Term.Block] = {
-      val nft = nextNonComment(tokens(meta.idx))
+    def unapply(meta: FormatToken.Meta): Option[OptionalBraces] = {
+      val nft = nextNonComment(meta)
```
tgodzik, line 1950: Seems we can't use `rightOwner` :/ for example:
```scala
def printTextDocument(doc: TextDocument): String =
      if cond then
        sb.append(doc.text.substring(offset, end))
        sb.append(doc.text.substring(offset, end))
```

rightOwner will actually not be Block.

kitbellew: let me see

tgodzik: Something like:
```
      val nft = nextNonComment(meta)
      val foundBlock = nft.meta.leftOwner match {
        case t: Term.If if meta.left.text == "then" => Some(t.thenp)
        case t: Term.If if meta.left.text == "else" => Some(t.elsep)
        case _ => Nil
      }
      foundBlock match {
        case Some(t: Term.Block) if t.tokens.headOption.contains(nft.right) =>
          Some(new OptionalBraces(t, mainIndent = t.stats.length == 1))
        case _ => None
      }
```
seems to work, but we need to extract block from each leftOwner.

tgodzik: I started doing some tests on the Dotty codebase and this popped up.

tgodzik: Getting to:
```
val foundBlock: Option[Term] = nft.meta.leftOwner match {
        case t: Term.If if meta.left.text == "then" || meta.left.text == ")" => Some(t.thenp)
        case t: Term.If if meta.left.text == "else" => Some(t.elsep)
        case t: Term.Try if meta.left.text == "try" => Some(t.expr)
        case t: Term.Try if meta.left.text == "finally" => t.finallyp
        case t: Term.Return => Some(t.expr)
        case t: Term.While if meta.left.text == "while"  =>  Some(t.expr)
        case t: Term.While if meta.left.text == "do"  =>  Some(t.body)
        case t: Term.If if meta.left.text == "then" || meta.left.text == ")" => Some(t.thenp)
        case _ => None
      }
```

Is there any easier way? :sweat_smile:

kitbellew: #2413

somehow i feel that a lot of the complexity with scalafmt is because scalameta trees are a bit top-down and randomly structured. most expressions look to me like an infix; they have an op, a left and a right.

kitbellew: > ```
> val foundBlock: Option[Term] = nft.meta.leftOwner match {
>         case t: Term.If if meta.left.text == "then" || meta.left.text == ")" => Some(t.thenp)
>         case t: Term.If if meta.left.text == "else" => Some(t.elsep)
> ...
> ```

Sorry, didn't see that earlier. I'd probably use `.is[KwThen]` or `RightParen` and avoid `text` comparisons, but i like that it doesn't traverse any trees. the only question, is this list exhaustive? lambdas, partial functions, `do` coming from `for`... let me think.

tgodzik: It's not yet exhaustive, it started getting too too long, so I actually started looking at a similar approach to what you did 😅

But maybe it's just a couple of cases more.

PR #2413, opened 2021-04-09
FormatOps: find any ancestor of right as a block

kitbellew:


tgodzik:
It seems that indentation is double in some places:
```scala
object a:
   def foo =
     try foo
     catch
        case A =>
             that match
                case b =>
                  bb
                case c =>
                  cc
             end match
        case B =>
          that match
             case c =>
               cc
             case _ =>
               dd
```

Not exactly sure about why.

tgodzik:
Maybe this an issue if we are in a case and look for a Block.

kitbellew:
let me see. also, some tests are failing...

kitbellew:
@tgodzik do we even need to handle `=>` since case already handles that in scala2? are there any other possibilities for `=>` which are not `Case`?

is this allowed in scala3:
```
val func: A => B = (a: A) =>
  a + a
  a * a
func(a)
```
```

**Comment by tgodzik** [comment]
> @tgodzik do we even need to handle `=>` since case already handles that in scala2? are there any other possibilities for `=>` which are not `Case`?
> 
> is this allowed in scala3:
> 
> ```
> val func: A => B = (a: A) =>
>   a + a
>   a * a
> func(a)
> ```

This one works, lambdas work with indentation.

**Comment by kitbellew** [comment]
done with `=>`. do you have an example when `<-` is used to start a block?

**Comment by tgodzik** [comment]
Surprisingly enough I don't think we ever tested that. I can cook up something later. Didn't see anything in existing codebases.

**Comment by kitbellew** [comment]
@tgodzik nested if is not working, need to fix. optional braces is hard 😀

**Comment by tgodzik** [comment]
After the first PR I thought it might not be that hard as anticipated after all, but turns out it is 😂 I did test it out on Dotty and it seemed to work mostly aside from some exceptions.

**Comment by kitbellew** [comment]
@tgodzik question: in
```
  val a = {
    if (a)
      aaa // c1
    else bbb
  }
```
when I print `if` tree and also its `then` child, i get
```
TYPE: Term$If$TermIfImpl
SOURCE: if (a)
      aaa // c1
    else bbb
STRUCTURE: Term.If(Term.Name("a"), Term.Name("aaa"), Term.Name("bbb"), Nil)
TOKENS: if,∙,(,a,),¶,∙,∙,∙,∙,∙,∙,aaa,∙,//∙c1,¶,∙,∙,∙,∙,else,∙,bbb

TYPE: Term$Name$TermNameImpl
SOURCE: aaa
STRUCTURE: Term.Name("aaa")
TOKENS: aaa
```
why is `// c1` not part of `thenp`? makes it harder to set indentation correctly.

**Comment by kitbellew** [comment]
continuing. in the context of optional braces, comments should be handled differently in these:
```
    // comment should belong to thenp
    if a then
      aaa // c1
    else bbb

    // comment should belong to thenp
    if a then
      aaa
      // c1
    else bbb

    // comment should NOT belong to thenp nor else
    if a then
      aaa
    // c1
    else bbb
```

**Comment by kitbellew** [comment]
continuing. even without optional braces, same idea:
```
    // comment should belong to thenp
    if (a)
      aaa // c1
    else bbb

    // comment should belong to thenp
    if (a)
      aaa
      // c1
    else bbb

    // comment should NOT belong to thenp nor else
    if (a)
      aaa
    // c1
    else bbb
```

tgodzik:
I haven't looked into the comments much, so there might so weird things going on there. They are kind of magically skipped.

tgodzik:
I think we might need to open an issue for improvements when it comes to enclosing comments in trees. I think currently it's mostly done randomly, because we haven't looked into comments when doing optional braces.

PR #2415, opened 2021-04-11
Router: handle `then/else/do/try/finally` with optional braces

kitbellew:
Also, abide by `newlines.source` for optional braces.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -1902,6 +1902,8 @@ class FormatOps(
           case _: T.Equals => unapplyEquals(ft, nft)
           case _: T.KwCatch => unapplyCatch(ft, nft)
           case _: T.KwMatch => unapplyMatch(ft, nft)
+          case _: T.KwThen => unapplyThen(ft, nft)
+          case _: T.KwElse => unapplyElse(ft, nft)
           case _ => unapplyBlock(ft, nft)
         }
       }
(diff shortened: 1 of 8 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1982,32 +2051,135 @@ class FormatOps(
...
+      }
+  }
+
+  // Optional braces after any token that can start indentation:
+  // else
+  object OptionalBracesElse {
+    def unapply(
+        ftMeta: FormatToken.Meta
+    )(implicit style: ScalafmtConfig): Option[OptionalBraces] =
+      getMissingBrace(ftMeta).flatMap { nft =>
```
tgodzik, line 2117: It seems that we need to check `getMissingBrace(ftMeta)` on each of the new unapplies, which makes us do it over and over again. I wonder whether we should do it all in a single `OptionalBraces` object? Looks like we should be able to merge that. Or am I too worried about performance impact that does not exist there?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1982,32 +2051,135 @@ class FormatOps(
...
+        ftMeta.leftOwner match {
+          case t: Term.If =>
+            def nestedIf(x: Term.If) = {
+              val forceNL = shouldBreakInOptionalBraces(nft) ||
+                !ifWithoutElse(t) && existsIfWithoutElse(x)
+              Some(new OptionalBraces(x, forceNL))
+            }
+            t.thenp match {
+              case x: Term.If => nestedIf(x)
+              case Term.Block(List(x: Term.If)) => nestedIf(x)
```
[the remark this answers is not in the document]
tgodzik: Also, what was the issue with nested ifs? It seems it should work normally :thinking:

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -472,3 +506,70 @@ object a:
...
+    foo
+    bar
+  while (foo) do
+    foo
+  while (foo) do
+    foo
+    bar
+>>>
+object a:
+   if (a) then foo
```
tgodzik, line 523: Should we also test nested `if (a)`?

tgodzik (approved):
Wow, this is getting more complex than I imagined. Great work! I left a couple of comments, but nothing blocking.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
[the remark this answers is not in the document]
kitbellew: this covers a block which contains only an if, not one which simply starts with an if.

if the block has multiple statements, a break is a must.

if the block only has an if, you must break, too, if it doesn't have an else but the outer one does:
```
if a then if b then c
else d
```
this `else` could belong to either of them since their indent is the same.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
kitbellew: good comment. i think it's time to go back to a single matcher. i will send an update in the morning.

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
kitbellew: will do

kitbellew (commented):
still more work to do

PR #2419, opened 2021-04-12
FormatTokens: remove optional braces, rewrite to new syntax

kitbellew:


kitbellew:
@tgodzik @olafurpg @poslegm  before i write too much complicated code, should we have
- one setting to rewrite everything to scala3 (drop optional braces, replace parens with `then/do`, wildcards etc.)
- one setting for each type of rewrite

(rewriting in the opposite direction is not part of this, it looks easier to remove things, just like we did with various other rewrite rules)

kitbellew:
> (rewriting in the opposite direction is not part of this, it looks easier to remove things, just like we did with various other rewrite rules)

apparently, the compiler can do this, too (including rewriting to old style):
- https://dotty.epfl.ch/docs/reference/other-new-features/control-syntax.html#rewrites
- https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html#settings-and-rewrites

the only difference is our code can rewrite and indent in one pass, instead of two that the compiler seems to require.

tgodzik:
> @tgodzik @olafurpg @poslegm before i write too much complicated code, should we have
> 
> * one setting to rewrite everything to scala3 (drop optional braces, replace parens with `then/do`, wildcards etc.)
> * one setting for each type of rewrite
> 
> (rewriting in the opposite direction is not part of this, it looks easier to remove things, just like we did with various other rewrite rules)

I wonder whether we should actually put in the rest of the proposed rewrites here. I am thinking that this should actually be done Scalafix, where people can just migration rules to make sure that the old syntax is replaced. These rules will only be necessary until 3.1, where we will most likely drop the old syntax altogether. It's a different case with optional braces, which is going to stay and people might decide to follow that style in their codebase. They could use the compiler rewrites, but that is much less interactive and besides removing `{` is already in line with what scalafmt does.

So overall I am thinking:

- scalafix rule RewriteToScala3Syntax for (and anything else that comes along):
```
- wildcard imports from _ to *
- => to as for renamed imports
- old varargs splices syntax to new one
```
these rules are kind of one time.

- scalafmt rewrite rule `braces to no braces`
- not support `no braces to braces` anywhere, because people can already just switch that syntax off and I imagine they can use the compiler rewrite if needed. So if any codebase decides on not having those, they can just decide it via compiler option. We can come back to that later on if needed.

What do you all think? @mlachkar was already thinking of putting those rules in scalafix once scalafix support for Scala 3 is done. This rewrite will also be much more needed before 3.1

olafurpg:
I think it makes sense to have this rewrite in Scalafmt. Users will continue to write braces by habit so I think it's nice to have Scalafmt fix it on file save. It's not a one-time thing.

It also makes sense to have one-time Scalafix rewrites for users who prefer to run it like that (or don't use Scalafmt).

Does it make sense to keep both?

tgodzik:
> I think it makes sense to have this rewrite in Scalafmt. Users will continue to write braces by habit so I think it's nice to have Scalafmt fix it on file save. It's not a one-time thing.

@olafurpg what do you think about the syntax migration rules though? Those are kind of one time.

olafurpg:
@tgodzik I suspect people will continue to write common Scala 2 syntax by habit for a while. There's no such thing as a "one-time" migration when new code gets written all the time with old habits :)

Scalafix and Scalafmt have similar rewrite APIs so I feel like it shouldn't be too difficult to support the same rewrites in both tool if there's demand for it. Maybe this will get confusing for users in the end, I'm not sure 🤷 There is a gray area between Scalafmt and Scalafix syntactic rewrites and we have never really ironed out how small/large the intersection should be between them.

kitbellew:
ok, let's hypothetically say we have the rewrites. should i just add one boolean flag and rewrite everything that can be rewritten into the given dialect? braces, parens, wildcards, etc?

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -629,3 +629,269 @@ object a:
...
+      } // c1
+    } else {
+      b
+    }
+  }
+}
+>>>
+object a:
+   val a =
+     if (a) {
```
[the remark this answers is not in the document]
kitbellew: probably, but it's not a scala3 rewrite with optional braces and significant indentation, because that requires a `then`, if I'm reading the doc correctly.

this is what our regular redundant braces rule would do.

tgodzik:
> ok, let's hypothetically say we have the rewrites. should i just add one boolean flag and rewrite everything that can be rewritten into the given dialect? braces, parens, wildcards, etc?

I would say at least two separate settings:
- rewrite to no braces
- rewrite to new syntax

I can imagine people not wanting to use optional braces, but they would still need to rewrite to the new syntax for anything that might become deprecated.

kitbellew:
one more question. if syntax is off and braces is on, should we keep braces for `if (a) {` or rewrite to `if a then`?

basically, is quiet syntax part of syntax or part of braces? in the parser, we use the same dialect flag for both, it looks like.

tgodzik:
> one more question. if syntax is off and braces is on, should we keep braces for `if (a) {` or rewrite to `if a then`?
> 
> basically, is quiet syntax part of syntax or part of braces? in the parser, we use the same dialect flag for both, it looks like.

So funily enough, you can use optional braces with `if(cond)`, this is valid:
```
if (cond)
  stat1
  stat2
else
  stat3
  stat4
```
and honestly it's another thing I argued about with Martin, but he said he wanted to make it easier for people to migrate.

:shrug: 

In that case I would just remove `{` and `}` for optional braces rewrite and only migrate to `then` under the new syntax rewrite.

kitbellew:
> So funily enough, you can use optional braces with `if(cond)`, this is valid:

undocumented. tested, works. however, inconsistently, `while (...)` and `for (...)` do not. at least, not in the parser.

need to update the formatter to support this case correctly.

tgodzik:
> > So funily enough, you can use optional braces with `if(cond)`, this is valid:
> 
> undocumented. tested, works. however, inconsistently, `while (...)` and `for (...)` do not. at least, not in the parser.
> 
> need to update the formatter to support this case correctly.

I haven't thought while would work with optional braces, but it seems to work. We need to fix that :/ 

Better yet,
```
  for{a <- 0 to 11}
      println(a)
      println(a)
```
also works :fearful:

tgodzik:
Opened up an issue here https://github.com/lampepfl/dotty/issues/12090

And we need to fix it in the parser, yay!

smarter:
> In that case I would just remove { and } for optional braces rewrite and only migrate to then under the new syntax rewrite.

IMO, this doesn't seem like the best thing to do. Right now Scala 2 code should look like:
```scala
if (foo) {
  a
  b
}
```
and Scala 3 code should look like:
```scala
if foo then
  a
  b
```
While it's true that technically the parser technically also supports:
```scala
if (foo)
  a
  b
```
this isn't a style that we want people to use. To me the main value of having it is that if you accidentally write `if (foo)` instead of `if foo then`, the compiler doesn't misinterpret your multi-line if as if it was a one-line if followed by another expression (and in fact I think we should emit a warning if we find code that looks like that).

An automatic rewrite that would output the third style might accidentally promote this style in the Scala community which isn't what we'd like to achieve.

kitbellew:
> this isn't a style that we want people to use. To me the main value of having it is that if you accidentally write `if (foo)` instead of `if foo then`, the compiler doesn't misinterpret your multi-line if as if it was a one-line if followed by another expression (and in fact I think we should emit a warning if we find code that looks like that).

@tgodzik @smarter I have implemented it via a third option for the rewrite parameter. However, if you feel strongly that nobody should be allowed to drop braces unless `then` is used, that parameter option can be removed.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RemoveScala3OptionalBraces.scala
```diff
@@ -0,0 +1,137 @@
...
+          case _: Term.For if allowOldSyntax || {
+                val rbFt = ftoks(ftoks.matching(ft.right))
+                ftoks.nextNonComment(rbFt).right.is[Token.KwDo]
+              } =>
+            removeToken
+          case _: Term.ForYield => removeToken
+          case _: Term.Match => removeToken
+          case _: Term.Try => removeToken
+          case _: meta.Ctor.Secondary =>
+            if (!ftoks.prevNonComment(ft).left.is[Token.Equals])
```
tgodzik, line 61: Procedure syntax is no longer allowed an will be disabled in the [Scala 3 dialect](https://github.com/scalameta/scalameta/pull/2313). This also doesn't compile in Scala 3.

The alternative approach would be to have the rewrite for Scala 2 dialect, but we do have it already in scalafix, which is used for example by the migration tool.

docs/configuration.md
```diff
@@ -1969,6 +1987,35 @@ method3(
...
+
+If this flag is enabled,
+[optional braces](https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html)
+will be removed and significant indentation applied.
+
+The flag takes the following values:
+
+- `never`: disabled
+- `always`: enabled
+- `newSyntaxOnly`: applies only to expressions using new syntax (which includes the case when
```
tgodzik, line 2016: I would change the conditions around to make it harder for people to use optional braces with the old syntax.

```suggestion
- `appliesToOldSyntax`: applies also expressions using deprecated syntax 
```

docs/configuration.md
```diff
@@ -1969,6 +1987,35 @@ method3(
...
+### `rewrite.scala3.removeOptionalBraces`
+
+If this flag is enabled,
+[optional braces](https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html)
+will be removed and significant indentation applied.
+
+The flag takes the following values:
+
+- `never`: disabled
+- `always`: enabled
```
tgodzik, line 2015:
```suggestion
- `always`: applies to only new control syntax expressions
```

tgodzik (changes requested):
Thanks a lot for the great work! It took me a while to think it all over, but I think it's a really solid. I do however have some more general comments.

docs/configuration.md
```diff
@@ -1969,6 +1987,35 @@ method3(
...
+
+- [control syntax](https://dotty.epfl.ch/docs/reference/other-new-features/control-syntax.html)
+- [vararg splices](https://dotty.epfl.ch/docs/reference/changed-features/vararg-splices.html)
+- [imports](https://dotty.epfl.ch/docs/reference/changed-features/imports.html)
+
+### `rewrite.scala3.removeOptionalBraces`
+
+If this flag is enabled,
+[optional braces](https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html)
+will be removed and significant indentation applied.
```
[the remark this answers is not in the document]
kitbellew: kept as-is for now, please let me know if you have another suggestion

docs/configuration.md
```diff
@@ -1954,6 +1972,35 @@ method3(
...
+
+The flag takes the following values:
+
+- `no`: disabled
+- `yes`: applies to expressions using the new control syntax (or
+  `rewrite.scala3.convertToNewSyntax` is set)
+- `oldSyntaxToo`: applies also to expressions using deprecated syntax
```
kitbellew, line 2002: renamed to no/yes/oldSyntaxToo

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RemoveScala3OptionalBraces.scala
```diff
@@ -0,0 +1,133 @@
...
+          case _: Term.For if allowOldSyntax || {
+                val rbFt = ftoks(ftoks.matching(ft.right))
+                ftoks.nextNonComment(rbFt).right.is[Token.KwDo]
+              } =>
+            removeToken
+          case _: Term.ForYield => removeToken
+          case _: Term.Match => removeToken
+          case _: Term.Try => removeToken
+          case _: meta.Ctor.Secondary
+              if ftoks.prevNonComment(ft).left.is[Token.Equals] =>
```
kitbellew, line 61: for now, simply making sure it's not procedure syntax, since it's still occasionally supported. if merge scalameta/scalameta#2313 and release a new version, will add the new flag here.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RemoveScala3OptionalBraces.scala
```diff
@@ -0,0 +1,133 @@
...
+          case _ => false
+        }
+        if (ok) removeToken else null
+      case _: Term.ForYield => removeToken
+      case _: Term.Try => removeToken
+      case _: Term.Throw => removeToken
+      case _: Term.Return => removeToken
+      case t: Defn.Def =>
+        if (tree ne t.body) null
+        else if (ftoks.prevNonComment(ft).left.is[Token.Equals]) removeToken
```
kitbellew, line 126: same comment as above

kitbellew:
> Thanks a lot for the great work! It took me a while to think it all over, but I think it's a really solid. I do however have some more general comments.

thank you for the review, i understand that this one was quite large. removed redundant braces logic, as requested, made all other changes except wording in documentation.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RemoveScala3OptionalBraces.scala
tgodzik: Sure, however I think we will never actually have a use for this. We need the Scala 3 dialect to use this rewrite and any codebase that uses Scala 3 will not be compiling with the procedure syntax. So once the new version of scalameta is released we will just remove this (or at least the tests).

tgodzik (approved):
Thanks! LGTM

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RemoveScala3OptionalBraces.scala
kitbellew: of course!

PR #2420, opened 2021-04-13
FormatOps: support optional braces for `if (a)`

kitbellew:


tgodzik (approved):
LGTM! We will also follow up with fixes to the parser for `for` and `while`

PR #2427, opened 2021-04-17
FormatOps: bugfix with enclosed `(apply(x => y))`

kitbellew:
The code incorrectly treats the first opening paren as precursor to the lambda parameter.


tgodzik (approved):
LGTM!

PR #2428, opened 2021-04-17
Prepare for scala3 rewrites (new syntax, optional braces)

kitbellew:
- FormatTokens: allow to look up an ealier token
- FormatOps: get stmt differently in rewritten block
- FormatTokens: move matchingParens from FormatOps
- FormatOps: use tokens from FormatTokens, not tree
- FormatWriter: recognize blk rewrites in `a{} -> a()`

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
@@ -196,7 +196,7 @@ object TreeOps {
     *
     * Contains lookup keys in both directions, opening [({ and closing })].
     */
-  def getMatchingParentheses(tokens: Tokens): Map[TokenHash, Token] = {
+  def getMatchingParentheses(tokens: Iterable[Token]): Map[TokenHash, Token] = {
     val ret = Map.newBuilder[TokenHash, Token]
     var stack = List.empty[Token]
     tokens.foreach {
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
@@ -96,9 +96,9 @@ object TreeOps {
       case _ => Seq.empty[Tree]
     }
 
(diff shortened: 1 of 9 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -173,6 +174,15 @@ object TreeOps {
         case t: Ctor.Secondary =>
           addDefn[KwDef](t.mods, t)
           addAll(t.stats)
+        // special handling for rewritten blocks
+        case t @ Term.Block(List(_)) // single-stat block
```
tgodzik, line 178: Why are these two cases needed? I am having difficulty following the logic here for `replacedWith` and what situation this is influenced by.

tgodzik (approved):
LGTM! I just have a question (just to understand it better) about one particular part of the code, but nothing blocking from my side.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
kitbellew: @tgodzik my apologies, i should have explained.

the logic for optional braces rewrite is different from the one we employed with older rewrites. the old rewrites traversed the tree, then collected and applied some token patches, converted to a string and re-parsed the code again.

the new approach is different. we parse the tree, create a `FormatTokens` object (a sequence of `FormatToken` pairs, with intervening whitespace) and then modify this sequence ONLY (no changes in trees or parsing).

therefore, this particular logic, which determines the start of a statement (but in reality is used for two primary purposes: to find where one statement ENDS and another BEGINS, and to find the beginning of a lambda enclosed in braces), breaks down a little bit.

for instance, if we remove an optional brace from from the start of a block, the `Router` rule which would have formatted `{?` before no longer applies; unless we ignore this rewritten block here, the wrong rule (which looks at starts of statements and was never intended to look block starts) would apply.

this is the minimum change to make this work. the method `extractStatementsIfAny` (line 80 above) is also used for a completely different purpose, or else we could have changed it to return `block.stats.tail` or `match.cases.tail` (although that still doesn't solve the problem with brace-enclosed lambdas).

PR #2429, opened 2021-04-17
Rewrite: fix the placeholder check

### LINKED ISSUE #2426: Option to remove redundant braces should not result in Scala 3 code that does not compile
- **Version**: 3.0.0-RC1
- **Integration**: Sbt (via sbt-scalafmt 2.4.2)
- **Configuration**:
```
version = "3.0.0-RC1"
maxColumn = 80

runner.dialect = scala3
newlines.source = unfold
```

## Steps

Given code like this (which compiles with Scala 3.0.0-RC2):
```scala
def main(args: Array[String]): Unit =
  args foreach {
    println(_)
  }
```

```bash
scalafmt
```

## Problem

Scalafmt formats the code like this:
```scala
def main(args: Array[String]): Unit = args foreach println(_)
```

This code does not compile with Scala 3.0.0-RC2:

```
sbt:test> run
[info] compiling 1 Scala source to C:\Git\Overload\target\scala-3.0.0-RC2\classes ...
[error] -- [E081] Type Error: C:\Git\Overload\src\main\scala\Main.scala:1:59 ---------
[error] 1 |def main(args: Array[String]): Unit = args foreach println(_)
[error]   |                                                           ^
[error]   |   Missing parameter type
[error]   |
[error]   |   I could not infer the type of the parameter _$1 of expanded function:
[error]   |   _$1 => args foreach println(_$1).
[error] one error found
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
[error] Total time: 0 s, completed 16 Apr 2021, 22:05:36
```

## Expectation

Scalafmt should not rewrite code such that it no longer compiles.

## Notes

It is only the combination of both `rewrite.rules = [RedundantBraces]` and `newlines.source = unfold` that causes this problem.

kitbellew:
Check the arguments of a Term.Apply (just like Term.ApplyInfix above) looking for a placeholder. Fixes #2426.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Rewrite.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Rewrite.scala
@@ -222,7 +222,7 @@ object RewriteCtx {
             queue ++= (t.lhs +: t.args)
             iter
           case t: Term.Apply =>
-            queue += t.fun
+            queue ++= (t.fun +: t.args)
             iter
           case t: Term.Select =>
             queue += t.qual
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM!

PR #2435, opened 2021-04-19
StyleMap: use binary search on change positions

kitbellew:
The purpose of this change is twofold:
- when tokens are rewritten, the original logic of looking up by a token would stop working, as new tokens wouldn't be registered in the `StyleMap`; instead, we need to use the token position rather than the actual token
- since format overrides are fairly rare, using binary search on a handful of change positions is cheaper than using a hashmap containing hundreds or even thousands of tokens following the first override

### DIFF
--- a/scalafmt-tests/src/test/scala/org/scalafmt/util/StyleMapTest.scala
+++ b/scalafmt-tests/src/test/scala/org/scalafmt/util/StyleMapTest.scala
@@ -110,10 +110,19 @@ class StyleMapTest extends FunSuite {
       """.stripMargin.parse[Source].get
     val fops1 = new FormatOps(code, ScalafmtConfig.default)
     val fops2 = new FormatOps(code, ScalafmtConfig(binPack = BinPack.enabled))
-    // all tokens starting with first "("
-    assertEquals(fops1.styleMap.numEntries, 43)
-    // all tokens starting with "// scalafmt:" override
-    assertEquals(fops2.styleMap.numEntries, 16)
+    /*
+     * - 1: initial style
+     * - 6: all 3 pairs of "()"
+     * - "// scalafmt:" override does nothing, since the value doesn't change
+     */
+    assertEquals(fops1.styleMap.numEntries, 7)
+    /*
+     * - 1: initial style
+     * - 1: "// scalafmt:" override
+     * - 2: last pair of "()"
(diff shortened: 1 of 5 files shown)


scalafmt-tests/src/test/scala/org/scalafmt/util/StyleMapTest.scala
```diff
@@ -89,4 +90,19 @@ class StyleMapTest extends FunSuite {
...
+      """object a {
+        |  println(1, 2, 3, 4, 5, 6)
+        |  println(1, 2, 3, 4, 5, 6)
+        |  // scalafmt: { binPack.unsafeCallSite = false }
+        |  println(1, 2, 3, 4, 5, 6)
+        |}
+      """.stripMargin.parse[Source].get
+    val fops1 = new FormatOps(code, ScalafmtConfig.default)
+    val fops2 = new FormatOps(code, ScalafmtConfig(binPack = BinPack.enabled))
+    assertEquals(fops1.styleMap.numEntries, 7)
```
tgodzik, line 104: What would be the different styles here? I thought it would be at max 2 - the default and the changed one. Could you maybe just add a comment why those numbers?

tgodzik (approved):
Looks good, although the test seems a bit unclear. Could you add a bit more explanation for people seeing this for the first time? (that would be me :D )

scalafmt-tests/src/test/scala/org/scalafmt/util/StyleMapTest.scala
kitbellew: will do. btw, not sure if you review one commit at a time (i make them specially), perhaps you would have seen how these numbers change with every commit.

tgodzik: I haven't looked at each commit, but it helps, thanks! Though I still wonder why we end up with those numbers.

kitbellew: added detailed breakdown for each commit, ptal

tgodzik: Thanks! That's super useful!

PR #2436, opened 2021-04-19
Router: use available style to check binpack

kitbellew:
The style for an intermediate token between two parentheses would have the same value for `binPack.unsafeCallSite`, so there's no need for a complex logic.


tgodzik (approved):
LGTM

PR #2438, opened 2021-04-21
BinPack: support indenting call sites only once

### LINKED ISSUE #2079: Double continuation indentation in nested fun call with bin packing
- **Version**: 2.6.2
- **Integration**: CLI
- **Configuration**:
```
version = 2.6.2
binPack.preset = true
continuationIndent.callSite = 4
```

## Steps

Given code like this:
```scala
object Enclosing {
  def apply(name: UTF8String): SimpleMethodName = {
    val len = name.length
    if (len == 0)
      throwInvalidEncodedName(name)

    if (false) {
      ???
    } else {
      // Normal method name
      new SimpleMethodName(
          validateSimpleEncodedName(name, 0, len, openAngleBracketOK = false))
    }
  }
}
```

```bash
scalafmt -c .scalafmt.conf --stdout Test.scala
```

## Problem

```scala
object Enclosing {
  def apply(name: UTF8String): SimpleMethodName = {
    val len = name.length
    if (len == 0)
      throwInvalidEncodedName(name)

    if (false) {
      ???
    } else {
      // Normal method name
      new SimpleMethodName(validateSimpleEncodedName(name, 0, len,
              openAngleBracketOK = false))
    }
  }
}
```

Note that the line `openAngleBracketOK = false)` is indented *8* spaces (i.e., 2x 4, the continuation indent).

## Expectation

```scala
object Enclosing {
  def apply(name: UTF8String): SimpleMethodName = {
    val len = name.length
    if (len == 0)
      throwInvalidEncodedName(name)

    if (false) {
      ???
    } else {
      // Normal method name
      new SimpleMethodName(validateSimpleEncodedName(name, 0, len,
          openAngleBracketOK = false))
    }
  }
}
```

with only *4* spaces of indentation.

## Workaround

No idea

kitbellew:
Instead of indenting for each level of "()" nesting, add a flag to allow indenting only once. Fixes #2079.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/BinPack.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/BinPack.scala
@@ -29,6 +29,7 @@ import metaconfig._
 case class BinPack(
     unsafeCallSite: Boolean = false,
     unsafeDefnSite: Boolean = false,
+    indentCallSiteOnce: Boolean = false,
     parentConstructors: BinPack.ParentCtors = BinPack.ParentCtors.MaybeNever,
     literalArgumentLists: Boolean = true,
     literalsIncludeSimpleExpr: Boolean = false,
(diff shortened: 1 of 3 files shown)


tgodzik (approved):
LGTM!

PR #2440, opened 2021-04-21
Don't pass CMSClassUnloading to sbt

raboof:
This has apparently been the default since 2013
(https://bugs.openjdk.java.net/browse/JDK-8000325)
and is no longer available on jdk15


olafurpg (approved):
The jvm options were copy pasted from some other project a long time ago. Thanks!

PR #2451, opened 2021-04-22
FormatOps: mimic `{}` pairs with optional braces

### LINKED ISSUE #2448: `SearchStateExploded` exception when significant indentation is used in pattern matching
- **Version**: 3.0.0-RC1
- **Integration**: CLI
- **Configuration**:
```
version = "3.0.0-RC1"
runner.dialect = Scala3
```

## Steps

Given code like this:
```scala
object Foo:
  def bar(): Option[RetType] = process(arg match
    case a: A if a.b =>
      None

    case a: A if a.c() =>
      None

    case a: A if a.c() =>
      None

    case a: A if a.b =>
      Some(someMethod().anotherMethod().fooBarMethod(ObjectWithLongName.method()))

    case a: A if a.c() =>
      val varName = a.method(a.someField.anotherField.method().map(ObjectWithLongName.A))
      val otherVarName = varName.method(a.someField.method(ObjectWithLongName.B))
      otherVarName

    case a: A if a.someField.otherField.function().exists(SomeObjectLongName.isTrue) =>
      None
  )
```

```bash
scalafmt
```

## Problem

Scalafmt throws `FailedToFormat` exception with a cause: `SearchStateExploded`.
```scala
org.scalafmt.cli.FailedToFormat: /path/to/Foo.scala
Caused by: org.scalafmt.Error$SearchStateExploded: Search state exploded on ')∙.[550:551]', line 19
	at org.scalafmt.internal.BestFirstSearch.shortestPath(BestFirstSearch.scala:174)
	at org.scalafmt.internal.BestFirstSearch.run$1(BestFirstSearch.scala:294)
	at org.scalafmt.internal.BestFirstSearch.getBestPath(BestFirstSearch.scala:295)
	at org.scalafmt.internal.BestFirstSearch$.apply(BestFirstSearch.scala:337)
	at org.scalafmt.Scalafmt$.$anonfun$doFormatOne$5(Scalafmt.scala:128)
```

## Expectation

No exception should be thrown. The file should be formatted.

### Workaround 1

We can make it work by adding to the configuration:
```
runner.maxStateVisits = 100000000
runner.optimizer.maxVisitsPerToken = 1000000
```
It works but formatting takes ~4 min.

### Workaround 2

The issue doesn't occur if we use curly brackets instead of significant indentation in pattern matching.

## Notes

The issue was found while running scalafmt on scaladoc project in Dotty. The code above is ~minimal code that is required to cause that issue. The original code is here: https://github.com/tdudzik/dotty/blob/master/scaladoc/src/dotty/tools/scaladoc/tasty/ClassLikeSupport.scala#L107 (`parseMember` method).

kitbellew:
With optional braces, we are missing the physical indication that the formatter has come to expect with some of its optimizations.

The BestFirstSearch algorithm has an optimization which recurses into a block and doesn't backtrack. Some split policies skip matching pairs of braces when applying a single-line constraint.

In the absence of any other indicators (like Indent/Outdent tokens), we need to identify cases where braces are optional.

Fixes #2448.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
@@ -20,8 +20,6 @@ private class BestFirstSearch private (
     range: Set[Range],
     formatWriter: FormatWriter
 )(implicit val formatOps: FormatOps) {
-  import Token._
-
   import LoggerOps._
   import TokenOps._
   import TreeOps._
@@ -69,15 +67,13 @@ private class BestFirstSearch private (
     if (!recurseOnBlocks || !isInsideNoOptZone(ft)) None
     else {
       val left = tokens(ft, -1)
-      if (!left.left.is[LeftBrace]) None
-      else {
-        val close = tokens.matching(left.left)
+      val closeOpt = formatOps.getEndOfBlock(left, false)(style)
+      closeOpt.filter(close =>
         // Block must span at least 3 lines to be worth recursing.
-        val ok = close != stop &&
+        close != stop &&
           distance(left.left, close) > style.maxColumn * 3 &&
           extractStatementsIfAny(left.meta.leftOwner).nonEmpty
-        if (ok) Some(close) else None
-      }
+      )
     }
 
   def stateColumnKey(state: State): StateHash = {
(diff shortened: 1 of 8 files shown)


tgodzik (approved):
Looks good! Thanks for fixing this! I think it was the one failing most in the Dotty codebase. It would be awesome to expose indentation, but I am not sure how to do it, so I opened up an issue here: https://github.com/scalameta/scalameta/issues/2303

Issue #2019, opened 2020-06-13
Adding Braces when maxline is below a certain limit


vito-c:
scalafmt.conf:
```conf
rewrite.rules = [AvoidInfix, PreferCurlyFors, SortImports, SortModifiers, RedundantBraces]
rewrite.redundantBraces.methodBodies = true
rewrite.redundantBraces.includeUnitMethods = true
rewrite.redundantBraces.maxLines = 100
danglingParentheses.defnSite = true
danglingParentheses.callSite = true
```

Input code:
```scala
  /** Redundant braces removes shouldn't remove braces */
  def bracesStay(): Option[Unit] = {
    Some("thing").map { u1 =>
      Unit
    }
  }
```

output code:
```scala
  /** Redundant braces please add some braces */
  def bracesAdd(): Option[Unit] =
    Some("thing").map { u1 =>
      Unit
    }
```

That checks out and works as expected but it  is impossible to undo the change by changing the maxLine

scalafmt.conf:
```conf
rewrite.rules = [AvoidInfix, PreferCurlyFors, SortImports, SortModifiers, RedundantBraces]
rewrite.redundantBraces.methodBodies = true
rewrite.redundantBraces.includeUnitMethods = true
rewrite.redundantBraces.maxLines = 1
danglingParentheses.defnSite = true
danglingParentheses.callSite = true
```

input code:
```scala
  /** Redundant braces please add some braces */
  def bracesAdd(): Option[Unit] =
    Some("thing").map { u1 =>
      Unit
    }
```
output code:
```scala
  /** Redundant braces please add some braces */
  def bracesAdd(): Option[Unit] =
    Some("thing").map { u1 =>
      Unit
    }
```

expected code:
```scala
  /** Redundant braces please add some braces */
  def bracesAdd(): Option[Unit] = {
    Some("thing").map { u1 =>
      Unit
    }
  }
```

kitbellew:
@vito-c as i mentioned, the tool doesn't add braces. the rule is called *redundant* braces. maxLines simply specifies that larger code blocks should not be touched. in both cases, you are looking for the opposite of what the tool does.

kitbellew:
other examples of why adding braces is going to be hard: #1463 #1222

vito-c:
@kitbellew thanks! I will check this out when I have time :)

vito-c:
I have something working on my local that adds braces when the body of a `def` function is over a certain maxlines.

It would result in changing some of the output of the tests for example:
[...]

PR #2452, opened 2021-04-22
FormatOps: use `keep` in binPackParentCtorSplits

### LINKED ISSUE #2226: How to prevent adding a newline before `extends`?
- **`sbt-scalafmt` version**: 2.4.2
- **Integration**: SBT
- **Configuration**:
```
version = "2.7.5"

preset=IntelliJ
maxColumn = 110
trailingCommas = always

align.preset = some
docstrings.blankFirstLine = yes
newlines.source = keep
newlines.implicitParamListModifierForce = [after]

rewrite.rules = [
  SortImports,
  RedundantBraces,
  RedundantParens,
  SortModifiers,
]
```

## Problem

```diff
-object Class extends Base
+object Class
+  extends Base
   with Trait1
   with Trait2 {
```

## Expectation

No newline added before `extends`

kitbellew:
Refactor `splitWithChain` into `binPackParentConstructorSplits`, automatically determine the intermediate or final expiration for `inits` or `derives` in a template.

That allows modifying a single rule to preserve existing breaks (in `newlines.source=keep`).

Fixes #2226.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -960,10 +960,11 @@ class FormatOps(
     nonWhitespaceOffset(right) - nonWhitespaceOffset(left)
   }
 
-  def typeTemplateSplits(
-      template: Template,
-      indentIfSecond: Int
-  )(implicit line: sourcecode.Line, ft: FormatToken): Seq[Split] = {
+  def typeTemplateSplits(template: Template, indentIfSecond: Int)(implicit
+      line: sourcecode.Line,
+      ft: FormatToken,
+      style: ScalafmtConfig
+  ): Seq[Split] = {
     def getPolicy(expire: T) = expire match {
       case lb: T.LeftBrace if template.self.tokens.isEmpty =>
         Policy.after(lb) {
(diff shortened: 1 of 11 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -403,12 +403,37 @@ class FormatOps(
...
+  private def findTreeInGroup[A](
+      trees: Seq[Tree],
+      func: (Tree, Seq[Tree], () => T) => A
+  )(expireFunc: Seq[Tree] => T)(implicit ft: FormatToken): Option[A] =
+    trees.find(_.pos.end >= ft.right.end).map { x =>
+      func(x, trees, () => expireFunc(trees))
     }
+
+  def findTemplateGroupOnRight[A](
+      func: (Tree, Seq[Tree], () => T) => A
```
tgodzik, line 421:
```suggestion
      func: (Tree, Seq[Tree], ExpireFunction) => A
```

Could we add maybe a type alias here? Or even:

```suggestion
      func: (Tree, Group, ExpireFunction) => A
```

It was a bit hard to understand this function. `T` looked to be a type parameter and it not clear what this function would be used for.

tgodzik (approved):
LGTM! Just a small comment, feel free to ignore, but might be useful for future readability.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
kitbellew: good idea, will do. thanks for the review.

btw, this is the last review, i don't have anything else pending.

tgodzik: Ok, cool! Let's release RC2 to gather some more feedback and I can work on the rewrite to braces.

kitbellew: on rc2, let's.

on "to braces": do we really need to? do you think it would be a common use case? https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html#settings-and-rewrites

kitbellew: adding back braces with multiple statements shouldn't be too hard. however, when it's one statement, we run into the same issue (mostly around infix) which has plagued #2019 and the issues it's superseded.

tgodzik: Hmm... we could hold off for the time being and just add it if people indeed want it. But I guess in most cases people should just add `-no-indent`

PR #2453, opened 2021-04-23
ProjectFiles: add `PathMatcher` patterns

### LINKED ISSUE #1648: Use PathMatcher for file filters
See #1645 #1646 

1. Implement `FilterGlobMatcher` using `java.nio.file.PathMatcher` as an alternative for [FilterMatcher](https://github.com/scalameta/scalafmt/blob/c92153e777984db6d69ec359dd1b1115bd2199d6/scalafmt-core/shared/src/main/scala/org/scalafmt/config/FilterMatcher.scala#L7)
2. Replace `exclude`/`include` in the documentation with `excludeGlobs`/`includeGlobs`
3. Keep old `exclude` and `include` in the config for compatibility purposes.

kitbellew:
Also, deprecate old regex-only patterns as the new option allows both. Fixes #1648.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2377,19 +2377,37 @@ val identity = Array(1, 0, 0,
 
 Configure which source files should be formatted in this project.
 
+#### `project.git`
+
+If this boolean flag is set, only format files tracked by git.
+
+#### `project.include/exclude`
+
+> Since v3.0.0.
+
+```scala mdoc:defaults
+project.include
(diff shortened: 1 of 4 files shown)


olafurpg (approved):
LGTM 👍 It's nice to consolidate towards globs. I didn't know about the Java glob API until a few years after releasing Scalafmt, it would have saved me so much time to use it from the beginning.

PR #2454, opened 2021-04-24
Documentation: point to ScalaFiddle, AST Explorer

### LINKED ISSUE #1364: Enhance align.tokens documentation example with scalafiddle embed
Augment the `align.tokens` example: https://scalameta.org/scalafmt/docs/configuration.html#aligntokens with an embedded scalafiddle, similar to: https://scalameta.org/docs/trees/scalafiddle.html

I would love to use this to simply copy/paste code into the documentation and get the correct `parent` rather than spinning anything up locally to figure it out.

kitbellew:
Fixes #1364.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -436,6 +436,10 @@ An align token is a pair of `code`, which is the string literal of an operator
 of token, and `owner`, which is the kind of the closest tree node that owns that
 token. If no `owner` is provided, then all tree kinds will be matched.
 
+> To find the `owner` part for a custom tree, look for its type prefix using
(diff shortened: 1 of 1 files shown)


olafurpg (approved):

kitbellew:
@keynmol @tgodzik this too, please? 😌 and don't need two.

kitbellew:
> > @keynmol @tgodzik this too, please? 😌 and don't need two.
> 
> I am not sure what you mean? The link to scalameta still works ok, so we should be good on that front.

Never mind, my apologies, I didn't realized the new tool is literally a drop-in replacement, and no URLs need to be modified.

PR #2463, opened 2021-04-26
FormatOps: optional braces in `= partial function`

### LINKED ISSUE #2425: [Scala 3] `error: outdent expected but case found` in braceless partial function
- **Version**: 3.0.0-RC1
- **Integration**: no
- **Configuration**:
```
version = 3.0.0-RC1
runner.dialect = scala3
```

## Steps

Given code like this:
```scala
val f: String => String =
  case "horses" => "are neat"
  case _ => "cows are sweet"
```

```bash
scalafmt Horses.scala
```

## Problem

Scalafmt prints:
```
scalafmt Horses.scala
/tmp/Horses.scala:2: error: outdent expected but case found
  case "horses" => "are neat"
  ^
```

## Expectation

I would expect `scalafmt` to format the partial function.

kitbellew:
Fixes #2425.


tgodzik (approved):
LGTM!

PR #2465, opened 2021-04-26
FormatTokensRewrite: migrate RewriteTrailingCommas

### LINKED ISSUE #1861: local comment configuration does not work

kitbellew:
Also, since the new implementation uses StyleMap, remove the check in StyleMap prohibiting overrides of this setting. Fixes #1861.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/FormatTokensRewrite.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/FormatTokensRewrite.scala
@@ -191,7 +191,8 @@ object FormatTokensRewrite {
 
   private val factories = Seq(
     RemoveScala3OptionalBraces,
-    ConvertToNewScala3Syntax
+    ConvertToNewScala3Syntax,
+    RewriteTrailingCommas
   )
 
   private[rewrite] trait Rule {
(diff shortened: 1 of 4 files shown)


tgodzik (approved):
LGTM!

Issue #1644, opened 2020-01-24
CLI: ignore most common typical generated files consistent to SBT integration


mroth:
This template is a guideline, not a strict requirement.
- **Version**: 2.3.2
- **Integration**: CLI
- **Configuration**: N/A (any)

## Steps

1. Run `scalafmt` binary from the CLI in a typical Scala project workspace.
2. Bombs/errors abound from generated files in `target` directories.

## Problem

Way back in #429 this was fixed by default, but that appears to have only been addressed for running via the SBT integration.

As direct CLI usage is becoming more common (especially with the native binary releases), this presents a confusing first usage scenario for a typical CLI user.

## Expectation

Should the CLI behavior be changed to essentially include a `project.excludeFilters` of `["target/"]` in a default invocation? My assumption as a user/developer is that invoking the CLI binary in a typical Scala project structure would ignore the most common build/generated destinations without requiring additional user configuration.

poslegm:
`^target/.*` regex may be reasonable in [default filters](https://github.com/scalameta/scalafmt/blob/fadab3012f9cb01d119270e6459f2a65ca885e55/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala#L11).

poslegm:
@olafurpg what do you think?

olafurpg:
Fine with me, we could also include `.metals/readonly`

olafurpg:
My reservation with this change is that users overwrite the defaults if they start tweaking the excludes. I think it might be better to change the default value to `project.git=true` if there exists a `.git/` directory 🤔

poslegm:
related issue #1526

poslegm:
You are right. `project.git = true` by default should solve problem for most use cases.

[Config reads here](https://github.com/scalameta/scalafmt/blob/84376adb3aa10410932ccba9bac39f2365f3d01b/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliOptions.scala#L228)

Also we have `git` [field](https://github.com/scalameta/scalafmt/blob/a90ca60f778c48fd1fd13c312500c5a4b2f16a3c/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala#L7) in the `ProjectFiles` class. It's never used and project builds successfully without it. What is the purpose of it?

kitbellew:
[...]

PR #2466, opened 2021-04-26
CliOptions: use config defaults from source class (and other refactors)

kitbellew:
This is to avoid duplicating information -- and to avoid deleting a setting accidentally since it's "not used" anywhere. See https://github.com/scalameta/scalafmt/issues/1644#issuecomment-578411802.


tgodzik (approved):
Looks good as far as I can tell, though I am not super familiar with this part of code. Seems to simplify it a lot of things, so big thumbs up from me!

PR #2467, opened 2021-04-26
FormatTokensRewrite: migrate PreferCurlyFors

kitbellew:


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/PreferCurlyFors.scala
```diff
@@ -24,70 +33,49 @@ object PreferCurlyFors extends Rewrite {
...
-      forToken <- forTokens.find(_.is[Token.KwFor])
-      leftParen <- findAfter(forToken) {
-        case _: Token.LeftParen => Some(true)
-        case Whitespace() => None
-        case _ => Some(false)
-      }
-      rightParen <- ctx.getMatchingOpt(leftParen)
-    } yield (leftParen, rightParen)
+  override def enabled(implicit style: ScalafmtConfig): Boolean =
+    PreferCurlyFors.enabled
```
tgodzik, line 42: Isn't this just always `true` ?

tgodzik (approved):
LGTM with just a minor comment

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/PreferCurlyFors.scala
kitbellew: for this specific rule, yes. but depending on `ScalafmtConfig` (which can potentially be overridden locally), it could actually be generally enabled (otherwise wouldn't be created) but not apply in some cases.

PR #2468, opened 2021-04-26
Scala 2.12: use Single Abstract Method lambdas

kitbellew:
Since we've deprecated 2.11, let's start using 2.12 features.


tgodzik (approved):
LGTM Looks much cleaner!

PR #2469, opened 2021-04-27
FormatTokensRewrite: migrate RedundantParen/Braces

kitbellew:
These two rules depend on each other and hence go hand in hand.


tgodzik (approved):
LGTM!

Issue #2479, opened 2021-04-29
Scala 3: Single-line definitions inappropriately wrapped and comments incorrectly indented


djneades:
- **Version**: 3.0.0-RC2
- **Integration**: Sbt via sbt-scalafmt 2.4.2
- **Configuration**:
```
version = 3.0.0-RC2
maxColumn = 79
runner.dialect = scala3
```

## Steps

Given code like this:
```scala
trait A

object Foo extends A
object Bar extends A

/** A doc comment */
object D extends A
```

When I run scalafmt like this from Sbt:

```bash
scalafmt
```

## Problem

Scalafmt formats code like this:
```scala
trait A

object Foo extends A
object Bar
    extends A

    /** A doc comment */
object D extends A
```

Notice:

1. The doc comment is incorrectly indented relative to the definition of object `D`.
2.  The definition of object `Bar` has been wrapped, even though it easily fits on a single line. (The definition of object `Foo`, which is exactly the same length, has not been wrapped.)

## Expectation

I should like the formatted output to look like this:
```scala
trait A

object Foo extends A
object Bar extends A

/** A doc comment */
object D extends A
```

djneades:
Non-doc comments are also processed infelicitously.

Before:

```scala
trait A

class B:
  trait Bar extends A
  trait Baz extends A
  // here is a comment

  trait From extends A
  lazy val x = 1
```

After scalafmt:

```scala
trait A

class B:
  trait Bar extends A
  trait Baz
      extends A
      // here is a comment

  trait From extends A
  lazy val x = 1
```

PR #2481, opened 2021-04-29
Use only the last non-comment token for binpack expire

tgodzik:
Fixes https://github.com/scalameta/scalafmt/issues/2479

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -426,7 +426,7 @@ class FormatOps(
       if (groups.lengthCompare(1) == 0)
         // for the last group, return '{' or ':'
         findTreeInGroup(groups.head, func) { x =>
-          templateCurly(template).getOrElse(tokenAfter(x).left)
+          templateCurly(template).getOrElse(getLastNonTrivialToken(x.last))
         }
       else {
         // for intermediate groups, return its last token
--- a/scalafmt-tests/src/test/resources/scala3/Issues.stat
+++ b/scalafmt-tests/src/test/resources/scala3/Issues.stat
@@ -19,3 +19,21 @@ def f(s: String): Double =
   s match
     case re(v): String => v.toDouble
     case _             => 0.0
+<<< extends with scalafmt/issues/2479
+object Outer:
+  trait A
+
+  object Foo extends A
+  object Bar extends A
+
+  /** A doc comment */
+  object D extends A
+>>>
+object Outer:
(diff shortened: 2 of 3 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -426,7 +426,7 @@ class FormatOps(
       if (groups.lengthCompare(1) == 0)
         // for the last group, return '{' or ':'
         findTreeInGroup(groups.head, func) { x =>
-          templateCurly(template).getOrElse(tokenAfter(x).left)
+          templateCurly(template).getOrElse(getLastNonTrivialToken(x.last))
```
kitbellew, line 429: @tgodzik when is this triggered? what does the test result look like without it?

scalafmt-tests/src/test/resources/scala3/Issues.stat
```diff
@@ -19,3 +19,21 @@ def f(s: String): Double =
...
     case re(v): String => v.toDouble
     case _             => 0.0
+<<< extends with scalafmt/issues/2479
+object Outer:
+  trait A
+
+  object Foo extends A
+  object Bar extends A
+
+  /** A doc comment */
```
kitbellew, line 29: is this comment parsed as part of `object Bar` before?

tgodzik: Yeah, that was the problem.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: It would set the expire at the comment:
```
trait A

object Foo extends A
object Bar
    extends A

    /** A doc comment */
object D extends A
```

This makes the extend go over the next line and the comment indented.

scalafmt-tests/src/test/resources/scala3/Issues.stat
kitbellew: hmm... should it be fixed in the parser and excluded from either tree (before or after)?

kitbellew (approved):
thanks, regardless of parser changes, safe to make it here.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Maybe it's fine to leave it here? Since it's an overall utility method? But I can also just inline it if you'd prefer it.

PR #2483, opened 2021-04-29
Add RedundantBraces import statements rewrite

hugo-vrijswijk:
I noticed when refactoring I would often remove unused imports and change code to:

```diff
- import x.{Y, Z}
+ import x.{Y}
```

This PR adds a new `RedundantBraces` rewrite that rewrites imports with a single 'importee' to:

```diff
- import x.{Y}
+ import x.Y
```


scalafmt-tests/src/test/resources/rewrite/RedundantBraces-import.stat
```diff
@@ -0,0 +1,32 @@
+rewrite.rules = [RedundantBraces]
```
kitbellew, line 1: please just append these tests to `RedundantBraces.stat`, we no longer need to create multiple test files since every case can have custom parameters.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/RedundantBracesSettings.scala
```diff
@@ -9,7 +9,8 @@ case class RedundantBracesSettings(
     stringInterpolation: Boolean = false,
     parensForOneLineApply: Option[Boolean] = None,
     generalExpressions: Boolean = true,
-    ifElseExpressions: Boolean = false
+    ifElseExpressions: Boolean = false,
+    importStatements: Boolean = true
```
kitbellew, line 13: i don't think we need a new parameter.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -124,6 +124,9 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
             (ConvertToNewScala3Syntax.enabled ||
               !x.tokens.exists(_.is[Token.RightArrow])) =>
         removeToken
+      case p: Importer
```
kitbellew, line 127: you need to put it in the case above, it also matches single-importee importers but mistakenly thinks that only rename and unimport are possible. something like this:
```
case meta.Importer(_, List(x))
  if !(x.is[Importee.Rename] || x.is[Importee.Unimport) ||
    ...
```

also, as you can see, we are not using any parameters to rewrite this, so my opinion is that we don't need a new parameter.

kitbellew (changes requested):
thank you for this self-service, very refreshing!

scalafmt-core/shared/src/main/scala/org/scalafmt/config/RedundantBracesSettings.scala
hugo-vrijswijk: Alright, I've removed it

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
hugo-vrijswijk: Correct, I've added it to the one above and added (and fixed) some extra tests for the name scenario's

PR #2484, opened 2021-04-29
Newlines: define templateBody parameters

### LINKED ISSUE #1542: Configuration for keep empty lines at the start of code blocks

### LINKED ISSUE #1975: Enforce newline before first def within class with extensions
- **Version**: 2.5.2
- **Integration**: IntelliJ
- **Configuration**:
```
version = 2.5.2
preset = IntelliJ
align.preset = some
continuationIndent.withSiteRelativeToExtends = 0
danglingParentheses.exclude = []
docstrings = ScalaDoc
maxColumn = 100
newlines.topLevelStatements = [before]
rewrite.rules = [AvoidInfix, SortModifiers, PreferCurlyFors]
  "`override`"
  "`private`"
  "`protected`"
  "`implicit`"
  "`final`"
  "`sealed`"
  "`abstract`"
  "`lazy`"
]
trailingCommas = multiple
verticalMultiline.newlineAfterOpenParen = true
verticalMultiline.excludeDanglingParens = []
```

## Steps

Given code like this:
```scala
trait Bar {}
trait Baz {}
case class Foo(
  a: String,
  b: String,
  c: String,
) extends Bar with Baz {
  def foo(): Int = 1
}
```

```bash
sbt scalafmtAll
```

## Problem

```scala
trait Bar {}
trait Baz {}

case class Foo(
  a: String,
  b: String,
  c: String,
) extends Bar
  with Baz {
  def foo(): Int = 1
}
```

## Expectation

```scala
trait Bar {}
trait Baz {}

case class Foo(
  a: String,
  b: String,
  c: String,
) extends Bar
  with Baz {

  def foo(): Int = 1
}
```

## Notes

The main issue is I'm trying to achieve the formatting recommended by the [official scala style guide for classes](https://docs.scala-lang.org/style/declarations.html#classes).

The docs provide two examples. The first without an extends statement:
> ```scala
> class Person(name: String, age: Int) {
>   …
> }
> 
> class Person(
>   name: String,
>   age: Int,
>   birthdate: Date,
>   astrologicalSign: String,
>   shoeSize: Int,
>   favoriteColor: java.awt.Color,
> ) {
>   def firstMethod: Foo = …
> }
> ```

The second, with extends:
> ```scala
> class Person(
>   name: String,
>   age: Int,
>   birthdate: Date,
>   astrologicalSign: String,
>   shoeSize: Int,
>   favoriteColor: java.awt.Color,
> ) extends Entity
>   with Logging
>   with Identifiable
>   with Serializable {
> 
>   def firstMethod: Foo = …
> }
> ```

The exact wording is:
> If a class/object/trait extends anything, the same general rule applies, put it on one line unless it goes over about 100 characters, and then put each item on its own line with trailing commas; closing parenthesis provides visual separation between constructor arguments and extensions; empty line should be added to further separate extensions from class implementation:

I believe this is related to issues brought up in #1653. In that issue, it was proposed to add a `newlines.alwaysInsideTopLevelStatements` configuration option.

The style guide seems to suggest adding the empty line for visual separation when extensions are added to the class.

Unfortunately, `newlines.topLevelStatements` only seems to focus on spacing between the top level elements and only when `newlines.topLevelStatementsMinBreaks` is meant. While I could partially achieve the desired result by setting `newlines.topLevelStatementsMinBreaks` to `0`, it's not ideal because:
1. Everything ends up being separated by a newline
2. It violates the first example where a newline isn't needed when there is no extensions

I _think_ what's needed is something along the lines of `newlines.alwaysAfterClassExtension`.

kitbellew:
Fixes #1975. Fixes #1542.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -796,6 +796,145 @@ package core {
 }
 ```
 
+### Newlines around template body
+
+> Since v3.0.0.
+
+This group of parameters controls whether to enforce a blank line before the first
+or after the last statement of a template body (i.e., body of a class, object, trait, enum).
+
+> These parameters will not cause any blank lines to be removed.
+
+#### `newlines.templateBodyIfMinStatements`
+
+`templateBodyIfMinStatements` can be `before` and/or `after`, while
+`templateBodyMinStatements` limits when the rule is applied.
+
+```scala mdoc:defaults
+newlines.templateBodyIfMinStatements
+newlines.templateBodyMinStatements
+```
+
+```scala mdoc:scalafmt
+newlines.templateBodyIfMinStatements = []
+---
+import org.scalafmt
+package core {
+  class C1 {
+    def one = 1
+  }
+  object O1 {
+    val one = 1
+    def two = 2
+  }
+  class C2 {}
+}
+```
+
+```scala mdoc:scalafmt
+newlines.templateBodyIfMinStatements = [before]
+---
+import org.scalafmt
+package core {
+  class C1 {
+    def one = 1
+  }
+  object O1 {
+    val one = 1
+    def two = 2
+  }
+  class C2 {}
+}
+```
+
+```scala mdoc:scalafmt
+newlines.templateBodyIfMinStatements = [after]
+---
+package core {
+  class C1 {
+    def one = 1
+  }
+  object O1 {
+    val one = 1
+    def two = 2
+  }
+  class C2 {}
+}
+```
+
+```scala mdoc:scalafmt
+newlines.templateBodyIfMinStatements = [before,after]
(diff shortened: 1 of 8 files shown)


tgodzik (approved):
LGTM, just a minor comment.

PR #2485, opened 2021-04-30
ScalafmtRunner: remove dialect parsing hack

kitbellew:


tgodzik:
I think we need to first release new Scalameta? I would wait until we get more info about possible bugs and @dos65 PR is merged.

kitbellew:
> I think we need to first release new Scalameta? I would wait until we get more info about possible bugs and @dos65 PR is merged.

of course! this is for visibility, so we don't forget it, after upgrading scalameta.

tgodzik:
@kitbellew are you looking into the scalameta upgrade? Or should I ? I've seen some tests start to fail for 4..4.16.

kitbellew:
> @kitbellew are you looking into the scalameta upgrade? Or should I ? I've seen some tests start to fail for 4.4.16.

i have it ready, some of the failing tests led to scalameta/scalameta#2334 and scalameta/scalameta#2335. i tagged 4.4.17, should send the pr soon.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
```diff
@@ -87,6 +81,7 @@ object ScalafmtRunner {
     val scala213 = Scala213
       .withAllowTrailingCommas(true)
     val default = scala213
+      .withAllowTraitParameters(true) // not true but backwards compatible
```
tgodzik, line 84: Why do we need it? Aren't trait params only available from Scala 3 ?

kitbellew: we have two options:
- keep this
- remove this test: https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/resources/unit/Dialect.stat#L20
thoughts?

tgodzik: I would remove the tests, although I am not sure why it was needed in the first place.

kitbellew: will do.

tgodzik (approved):

PR #2497, opened 2021-05-04
RedundantParens: support match, partial functions

kitbellew:


scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat
```diff
@@ -382,12 +382,168 @@ object a {
...
+      case false => false
+  )
+}
+>>>
+object a {
+  val b = c + (d.match
+    case true  => true
+    case false => false
+  )
+  val b = c + (d.match
```
tgodzik, line 538: And also here.

tgodzik (approved):
LGTM - left just comments about possible improvements.

scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat
kitbellew: if you remove parens, what would the result be? it tried to do
```
val b = c + d.match
   case true => true
   case false => false
   + e
```
is that valid? if it's not, parens are needed. if it is, at what level should the indentation be for `+ e`?

scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat
```diff
@@ -382,12 +382,168 @@ object a {
...
+      case true  => true
+      case false => false) + e
+  val b = c + (d match
+      case true  => true
+      case false => false
+  )
+}
+>>>
+object a {
+  val b = c + (d.match
```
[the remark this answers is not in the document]
kitbellew: if we can figure out the right conditions when they can be removed...

scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat
tgodzik: I think it would need to be:
```
val b = c + d.match
   case true => true
   case false => false
+ e
```

tgodzik: That's working for me in Dotty

tgodzik: Wait no, scratch that. It will wound up in the case. We need parens here.

scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat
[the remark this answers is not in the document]
tgodzik: Maybe let's leave it, automatically removing them might lead to some weird issues.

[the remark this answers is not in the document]
kitbellew: that's what i thought, too.

PR #2503, opened 2021-05-12
Scalameta: upgrade to 4.4.17

kitbellew:


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -199,7 +199,7 @@ class Router(formatOps: FormatOps) {
         val newlineBeforeClosingCurly = decideNewlinesOnlyBeforeClose(close)
         val selfAnnotation: Option[Tokens] = leftOwner match {
           // Self type: trait foo { self => ... }
-          case t: Template => Some(t.self.name.tokens).filter(_.nonEmpty)
+          case t: Template => Some(t.self.tokens).filter(_.nonEmpty)
```
kitbellew, line 202: this is because scalameta/scalameta#2334 is still not fixed.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -612,13 +612,14 @@ class Router(formatOps: FormatOps) {
           .getOrElse(getLastToken(leftOwner))
         val forceNewlineBeforeExtends = Policy.before(expire) {
           case Decision(t @ FormatToken(_, soft.ExtendsOrDerives(), _), s)
-              if t.meta.rightOwner == leftOwner =>
+              if t.meta.rightOwner.parent.contains(leftOwner) =>
```
kitbellew, line 615: this change is because `extends` is now part of template, not `object/class`.

this also means that this never matched `derives`.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -612,13 +612,14 @@ class Router(formatOps: FormatOps) {
...
             s.filter(x => x.isNL && !x.isActiveFor(SplitTag.OnelineWithChain))
         }
-        val policyExpire = defnBeforeTemplate(leftOwner).fold(r)(_.tokens.last)
-        val policyEnd = Policy.End.After(policyExpire)
-        val policy = delayedBreakPolicy(policyEnd)(forceNewlineBeforeExtends)
-        Seq(Split(Space, 0).withPolicy(policy))
+        val policy = defnBeforeTemplate(leftOwner).map { x =>
```
kitbellew, line 618: removing the policy for the immediate right token, doesn't look to be useful.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -136,7 +136,8 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
       case Left(_) => (left, removeToken)
       case Right(lft @ FormatToken(_, _: Token.LeftParen, _)) =>
         val rt = ft.right
-        val right = replaceToken("}") { // shifted right
+        val right = replaceToken("}", ft.meta.rightOwner.parent) {
```
kitbellew, line 139: ownership of braces moved to blocks, not enclosing tree.

scalafmt-tests/src/test/resources/scala3/Using.stat
```diff
@@ -230,13 +230,13 @@ def method2(using
 <<< using as identifier
 maxColumn = 20
 ===
-def using(using using: using, using: using, using: using)
+def using(using using: using, using: using, using: using): Unit
```
kitbellew, line 233: this is just because procedure syntax was deprecated.

scalafmt-tests/src/test/resources/unit/Lambda.stat
```diff
@@ -110,11 +110,10 @@ map(() => {
   }
 })
 >>>
-map(() => {
-  x =>
-    x match {
-      case e =>
-    }
+map(() => { x =>
```
kitbellew, line 113: this change is minor (and positive), but i don't know why it happened :)

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: btw. you should be able to test out with a snapshot version of Scalameta, they are created after each merge, so we would avoid doing release that has some issues still.

scalafmt-tests/src/test/resources/test/JavaDoc.stat
```diff
@@ -1815,3 +1815,126 @@ val a = 1
...
+===
+/** Nice scala code:
+  *   ```scala    sc:compile
+  * class Formatter(style: ScalaStyle, tree: Tree, toks: Array[FormatToken
+  * ```
+  */
+val a = 1
+>>>
+/** Nice scala code:
+  * ```scala sc:compile
```
tgodzik, line 1936: Hmm... this means if we could format it twice and each one would change the comment in a different way. Should we maybe diviate from the spec and just allow ``` after any number of spaces?

tgodzik (approved):
LGTM!

scalafmt-tests/src/test/resources/test/JavaDoc.stat
kitbellew: this part i didn't understand 🙂 can you explain, please?

tgodzik: I commented on a wrong part I think. I meant the situation in which fences are in a wrong spot (too many spaces) 
First ``` are not in a correct place:
```
/** Nice scala code:
  *    ```scala    sc:compile
  * class Formatter(style: ScalaStyle, tree: Tree, toks: Array[FormatToken])
  * ```
  */
```

after first formatting you get them into the correct place:
/** Nice scala code:
  * ```scala    sc:compile
  * class Formatter(style: ScalaStyle, tree: Tree, toks: Array[FormatToken])
  * ```
  */
```

and after second formatting you will format the code inside the fences since they are correct now:
```
/** Nice scala code:
 *  ```scala
 *  class Formatter(
 *    style: ScalaStyle,
 *    tree: Tree,
 *    toks: Array[FormatToken]
 *  )
 *  ```
 *  No more nice scala code.
 */
```

kitbellew: hmm... it could happen but only under very lucky circumstances. opening quotes would have to start a line, and closing quotes would have to start a line and be the last tokens in a paragraph.

how should we fix?

tgodzik: We should probably not adjust a line with spaces and ``` , because that actually changes the meaning of the comments. But again, it's not a big deal, so maybe it's fine to leave as is.

kitbellew: then we need to change the parser (in scalameta) to specify how many spaces were before the fence.

PR #2505, opened 2021-05-13
Remove optional braces in extension groups

tgodzik:


scalafmt-tests/src/test/resources/scala3/Extension.stat
```diff
@@ -258,3 +258,26 @@ extension (using
       TypeParameter2,
       TypeParameter3
   ](b: A) = a + b
+<<< drop optional braces
+rewrite.scala3.convertToNewSyntax = true
```
kitbellew, line 262: why do you need this and the next one? since we are testing just the removal here, and there isn't any interplay for that rule with these two others, plus these two don't really do anything in your test...

kitbellew (changes requested):

scalafmt-tests/src/test/resources/scala3/Extension.stat
tgodzik: Ach sorry, I just copied those over. I will remove them

tgodzik: Fixed

kitbellew (approved):

Issue #2480, opened 2021-04-29
Scala 3: error: outdent expected but case found


djneades:
- **Version**: 3.0.0-RC2
- **Integration**: Sbt via sbt-scalafmt 2.4.2
- **Configuration**:
```
version = 3.0.0-RC2
maxColumn = 79
runner.dialect = scala3
```

## Steps

Given code like this:
```scala
val f : String => PartialFunction[String, Int] = s =>
    case "Hello" =>
        5
    case "Goodbye" =>
        0
```

When I run scalafmt from Sbt like this:

```bash
scalafmt
```

## Problem

Scalafmt produces an error:
```bash
sbt:test> scalafmt
[info] Formatting 1 Scala sources...
[error] C:\Git\Overload\src\main\scala\Main.scala:2: error: outdent expected but case found
[error]     case "Hello" =>
[error]     ^: C:\Git\Overload\src\main\scala\Main.scala
[error] (Compile / scalafmt) C:\Git\Overload\src\main\scala\Main.scala:2: error: outdent expected but case found
[error]     case "Hello" =>
[error]     ^: C:\Git\Overload\src\main\scala\Main.scala
[error] Total time: 0 s, completed 29 Apr 2021, 15:18:36
```

## Expectation

Scalafmt should be able to format the code.

## Notes

This may be related to https://github.com/scalameta/scalameta/issues/2289 (fixed in 3.0.0-RC2), which I reported against Scalafmt 3.0.0-RC1. The code I gave there reproduced the error I was seeing, but was obviously overly simplified from my actual code and the sample above, both of which still break Scalafmt 3.0.0-RC2.

tgodzik:
Thanks for reporting! This is an issue in the parser, fixing it here https://github.com/scalameta/scalameta/pull/2321

kitbellew:
@tgodzik should we close this?

tgodzik:
Another test added https://github.com/scalameta/scalafmt/pull/2509

PR #2509, opened 2021-05-13
Add tests for #2480

tgodzik:
Fixes https://github.com/scalameta/scalafmt/issues/2480

### DIFF
--- a/scalafmt-tests/src/test/resources/scala3/Issues.stat
+++ b/scalafmt-tests/src/test/resources/scala3/Issues.stat
@@ -37,3 +37,15 @@ object Outer:
 
   /** A doc comment */
   object D extends A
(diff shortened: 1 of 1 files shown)


kitbellew (approved):

kitbellew:
@tgodzik this was the last scala3-related issue. want to try for rc3?

tgodzik:
I found another issue in the parser so maybe after that one is merged? Though it's a minor thing which we can add in the final version and in the meantime test out for any other issues.

kitbellew:
let's fix the parser.

tgodzik:
Fixing it here https://github.com/scalameta/scalameta/pull/2341 and I also want to check another issue reported to me internally.

PR #2513, opened 2021-05-16
Fix indentation when using braceless block with single Defn

tgodzik:
Previously, it was not possible to have a single val in a block and not have braces, which is why it didn't work for Scala 3.

Now, we make sure that indentation is added always if we have a single statement, which is a Defn.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -2087,7 +2087,7 @@ class FormatOps(
         val leftOwner = ft.meta.leftOwner
         findTreeWithParentSimple(nft.meta.rightOwner)(_ eq leftOwner) match {
           case Some(t: Term.Block)
-              if t.stats.lengthCompare(1) > 0 && isBlockStart(t, nft) =>
+              if !hasSingleTermStat(t) && isBlockStart(t, nft) =>
             Some(getSplitsMaybeBlock(ft, nft, t, true))
           case _ => None
         }
@@ -2367,7 +2367,8 @@ class FormatOps(
         allowMain: Boolean
     )(implicit line: sourcecode.Line, style: ScalafmtConfig): Seq[Split] = {
       val multiStat = isTreeMultiStatBlock(tree)
-      val forceNL = multiStat || shouldBreakInOptionalBraces(nft)
+      val forceNL =
+        !hasSingleTermStatIfBlock(tree) || shouldBreakInOptionalBraces(nft)
       getSplits(ft, tree, forceNL, allowMain && !multiStat)
     }
 
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
@@ -634,18 +634,26 @@ object TreeOps {
     case _ => false
   }
 
-  def getTermSingleStat(t: Term): Option[Tree] =
+  def getTreeSingleStat(t: Tree): Option[Tree] =
     t match {
       case b: Term.Block => getBlockSingleStat(b)
       case _ => Some(t)
     }
 
-  def getTermLineSpan(b: Tree): Int =
+  def getTreeLineSpan(b: Tree): Int =
     if (b.tokens.isEmpty) 0
     else {
       val pos = b.pos
       pos.endLine - pos.startLine
     }
+
+  def hasSingleTermStat(t: Term.Block): Boolean =
+    getBlockSingleStat(t).exists(_.is[Term])
+
+  def hasSingleTermStatIfBlock(t: Tree): Boolean = t match {
+    case b: Term.Block => hasSingleTermStat(b)
+    case _ => true
+  }
 
   /** In cases like:
     * {{{
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -39,7 +39,7 @@ object RedundantBraces extends Rewrite with FormatTokensRewrite.RuleFactory {
    * "foo { x => y; z }" can't become "foo(x => y; z)" */
   @tailrec
   def canRewriteWithParens(f: Term.Function, nested: Boolean = false): Boolean =
-    !needParensAroundParams(f) && (getTermSingleStat(f.body) match {
+    !needParensAroundParams(f) && (getTreeSingleStat(f.body) match {
       case Some(t: Term.Function) => canRewriteWithParens(t, true)
       case Some(_: Defn) => false
       case x => nested || x.isDefined
@@ -298,13 +298,12 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
 
   private def innerOk(b: Term.Block)(s: Stat): Boolean =
     s match {
-      case _: Term.Function | _: Defn => false
       case t: Term.NewAnonymous =>
         // can't allow: new A with B .foo
         // can allow if: no ".foo", no "with B", or has braces
         !b.parent.exists(_.is[Term.Select]) || t.templ.inits.length <= 1 ||
           t.templ.stats.nonEmpty || t.tokens.last.is[Token.RightBrace]
-      case _ => true
+      case tree => tree.is[Term] && tree.isNot[Term.Function]
     }
 
   private def okToRemoveBlockWithinApply(
@@ -396,16 +395,16 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
       b: Term,
       okIfMultipleStats: => Boolean
   )(implicit style: ScalafmtConfig): Boolean =
-    settings.methodBodies && (getTermSingleStat(b) match {
+    settings.methodBodies && (getTreeSingleStat(b) match {
       case Some(_: Term.PartialFunction) => false
-      case Some(s) => getTermLineSpan(s) <= settings.maxLines
+      case Some(s) => getTreeLineSpan(s) <= settings.maxLines
       case _ => okIfMultipleStats
     })
 
   private def getSingleStatIfLineSpanOk(b: Term.Block)(implicit
       style: ScalafmtConfig
   ): Option[Stat] =
-    getBlockSingleStat(b).filter(getTermLineSpan(_) <= settings.maxLines)
+    getBlockSingleStat(b).filter(getTreeLineSpan(_) <= settings.maxLines)
 
   // special case for Select which might contain a space instead of dot
   private def isPrefixExpr(expr: Tree): Boolean =
--- a/scalafmt-tests/src/test/resources/scala3/Issues.stat
+++ b/scalafmt-tests/src/test/resources/scala3/Issues.stat
@@ -49,3 +49,11 @@ val f: String => PartialFunction[String, Int] = s =>
     5
   case "Goodbye" =>
     0
+<<< single val in block
+object Util:
+  def assertPropPassed(prop: Prop): Unit =
+    val _ = ""
+>>>
+object Util:
+  def assertPropPassed(prop: Prop): Unit =
+    val _ = ""


tgodzik:
I think this is the last issue I found, once we merge this, we can release scalameta and release RC3

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1985,7 +1985,8 @@ class FormatOps(
 
     trait Impl {
       def tryGetSplits(ft: FormatToken, nft: FormatToken)(implicit
-          style: ScalafmtConfig
+          style: ScalafmtConfig,
+          line: sourcecode.Line
```
kitbellew, line 1989: why do you need to pass the line? and if you really do, can that be a separate commit, with item description?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2081,15 +2086,20 @@ class FormatOps(
...
       ): Option[Seq[Split]] = {
         val leftOwner = ft.meta.leftOwner
         findTreeWithParentSimple(nft.meta.rightOwner)(_ eq leftOwner) match {
           case Some(t: Term.Block)
-              if t.stats.lengthCompare(1) > 0 && isBlockStart(t, nft) =>
+              if (t.stats.lengthCompare(1) > 0 ||
```
kitbellew, line 2097: you don't need to check length anymore. just remove this line and replace`exists` with `forall` on next.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2081,15 +2086,20 @@ class FormatOps(
...
+                getBlockSingleStat(t).exists(_.isNot[Term]))
```
kitbellew, line 2098: is it isNot[term], or is[Defn]?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2081,15 +2086,20 @@ class FormatOps(
...
+                && isBlockStart(t, nft) =>
             Some(getSplitsMaybeBlock(ft, nft, t, true))
-          case _ => None
+          case other =>
```
kitbellew, line 2101: must be debug leftover. could you please restore two lines.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -1999,7 +2000,10 @@ class FormatOps(
     // if  match  return  then  throw  try  while  yield
     def unapply(
         ftMeta: FormatToken.Meta
-    )(implicit style: ScalafmtConfig): Option[Seq[Split]] = {
+    )(implicit
+        style: ScalafmtConfig,
+        line: sourcecode.Line
```
kitbellew, line 2005: this line is always the same, right? it's not useful for debugging.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2081,15 +2086,20 @@ class FormatOps(
     }
 
     object BlockImpl extends Impl {
+
```
kitbellew, line 2089: this empty line makes it inconsistent with other Impl implementations?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -634,7 +634,7 @@ object TreeOps {
     case _ => false
   }
 
-  def getTermSingleStat(t: Term): Option[Tree] =
+  def getTermSingleStat(t: Tree): Option[Tree] =
```
kitbellew, line 637: please rename to getTree

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2367,7 +2388,9 @@ class FormatOps(
         allowMain: Boolean
     )(implicit line: sourcecode.Line, style: ScalafmtConfig): Seq[Split] = {
       val multiStat = isTreeMultiStatBlock(tree)
-      val forceNL = multiStat || shouldBreakInOptionalBraces(nft)
+      val singleNoTerms = getTermSingleStat(tree).exists(_.isNot[Term])
```
kitbellew, line 2391: same logic as above. remove multistat, change here exists to forall. also, the check inside repeats, should we create a helper function to share with line 2098 above?

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: I spend a bunch of time trying to figure out why we were adding a newline in line 2084 of Router and it turned out it wasn't in in at all. Maybe we could have the while line plus source file? I can move it to a separate commit.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Ach sorry, that's left over from debugging. Got thrown off by the checks not running and haven't checked for silly changes

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: I think that is safer, but Defn should suffice

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
kitbellew: yeah, i know, I'm used to it by now. your idea of passing the filename is interesting, though, perhaps we should do that later.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
kitbellew: there's a similar logic in RedundantBraces (to avoid removing them), perhaps we should have a shared check.

kitbellew: and if it's safer, let's keep it. you should know better anyway. i still haven't figured out the difference between tree and term.

kitbellew: > there's a similar logic in RedundantBraces (to avoid removing them), perhaps we should have a shared check.

that one checks for Defn or Term.Function (`innerOk`). somewhat different, not much to share. but perhaps we should change `innerOk` to do `tree.isNot[Term] || tree.is[Term.Function]` (similar to this proposal).

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: I wonder how to do it and not have to change every single:
```
implicit  line: sourcecode.Line
```

Into
```
implicit  line: sourcecode.Line,
implicit  file: sourcecode.File
```

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Wait but if I dontt check the length then stats with a single Term will also get caught in this case, no ?

tgodzik: We want either a block with many terms or a block with a single non term. It seems to me exclusive.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: I removed all the added `line`

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Changed it  all a bit, let me know what you think. I think we still need the compare the length.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Changed the logic in RedundantBraces a bit too.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: fixed

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Removed

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Removef

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
tgodzik: Renamed, as well as the method below.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Created a helper function.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -634,19 +634,25 @@ object TreeOps {
...
+  def getTreeLineSpan(b: Tree): Int =
     if (b.tokens.isEmpty) 0
     else {
       val pos = b.pos
       pos.endLine - pos.startLine
     }
 
+  def hasBlockOnlyNonTermStats(t: Tree) =
+    t match {
+      case b: Term.Block => b.stats.forall(_.isNot[Term])
```
kitbellew, line 652: we never really wanted to check all terms, right? just one and only.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: https://github.com/scalameta/scalafmt/pull/2514 <- quick search and replace

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2087,7 +2087,8 @@ class FormatOps(
         val leftOwner = ft.meta.leftOwner
         findTreeWithParentSimple(nft.meta.rightOwner)(_ eq leftOwner) match {
           case Some(t: Term.Block)
-              if t.stats.lengthCompare(1) > 0 && isBlockStart(t, nft) =>
+              if (t.stats.lengthCompare(1) > 0 || hasBlockOnlyNonTermStats(t))
```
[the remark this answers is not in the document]
tgodzik: Och, I missed it is a None in case of multiple stats. Thanks!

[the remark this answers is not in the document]
tgodzik: Fixed!

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -634,19 +634,25 @@ object TreeOps {
...
 
-  def getTermLineSpan(b: Tree): Int =
+  def getTreeLineSpan(b: Tree): Int =
     if (b.tokens.isEmpty) 0
     else {
       val pos = b.pos
       pos.endLine - pos.startLine
     }
 
+  def hasBlockOnlyNonTermStats(t: Tree) =
```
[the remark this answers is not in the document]
tgodzik: Changed!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2367,7 +2368,9 @@ class FormatOps(
         allowMain: Boolean
     )(implicit line: sourcecode.Line, style: ScalafmtConfig): Seq[Split] = {
       val multiStat = isTreeMultiStatBlock(tree)
-      val forceNL = multiStat || shouldBreakInOptionalBraces(nft)
+      val singleNoTerms = hasBlockOnlyNonTermStats(tree)
+      val forceNL =
+        multiStat || singleNoTerms || shouldBreakInOptionalBraces(nft)
```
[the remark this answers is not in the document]
tgodzik: Replaced

kitbellew (approved):

Issue #2512, opened 2021-05-15
"Invariant failed" message on a macro type match (Scala 3)


gaeljw:
- **Version**: 3.0.0-RC2
- **Integration**: sbt
- **Configuration**:
```
version = 3.0.0-RC2
preset = default
runner.dialect = scala3
```

## Steps

Given code like this:
```scala
   def getTypeTreeRec(tpr: TypeRepr)(using Quotes): Expr[TypeTreeTag] = {
      tpr.asType match {
        case '[t] => getTypeTree[t]
      }
    }
```

When I run scalafmt like this:

```bash
sbt scalafmtAll
```

## Problem

Scalafmt fails with:
```
[error] invariant failed:
[error] when verifying scala.meta.classifiers.`package`.XtensionClassifiable[scala.meta.Term](body)(scala.meta.Tree.classifiable[scala.meta.Term]).is[Term.QuotedMacroExpr](Term.this.QuotedMacroExpr.ClassifierClass[scala.meta.Term])
[error] found that scala.meta.classifiers.`package`.XtensionClassifiable[scala.meta.Term](body)(scala.meta.Tree.classifiable[scala.meta.Term]).is[Term.QuotedMacroExpr](Term.this.QuotedMacroExpr.ClassifierClass[scala.meta.Term]) is false
[error] where Term = scala.meta.Term$@5dce6860
[error] where body = '[t]: ...TypeTreeTagMacros.scala

```

## Notes

Seems similar to https://github.com/scalameta/scalafmt/issues/2424 but a slightly different case?

PR #2515, opened 2021-05-17
Fix issues with QuotedMacroType

tgodzik:
Fixes https://github.com/scalameta/scalafmt/issues/2512

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -6,7 +6,7 @@ import org.portablescala.sbtplatformdeps.PlatformDepsPlugin.autoImport._
 
 object Dependencies {
   val metaconfigV = "0.9.10"
-  val scalametaV  = "4.4.17"
+  val scalametaV  = "4.4.18"
   val scalacheckV = "1.15.2"
   val coursier    = "1.0.3"
   val munitV      = "0.7.25"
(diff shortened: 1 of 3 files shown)


project/Dependencies.scala
```diff
@@ -6,7 +6,7 @@ import org.portablescala.sbtplatformdeps.PlatformDepsPlugin.autoImport._
 
 object Dependencies {
   val metaconfigV = "0.9.10"
-  val scalametaV  = "4.4.17"
+  val scalametaV  = "4.4.18"
```
kitbellew, line 9: i'd perhaps recommend upgrading as a separate, explicit PR

kitbellew (approved):

kitbellew (changes requested):
found another pr with the same upgrade...

tgodzik:
> found another pr with the same upgrade...

I can put it into a different one, but I am not sure it's that useful in this case. It doesn't break any tests.

kitbellew:
> > found another pr with the same upgrade...
> 
> I can put it into a different one, but I am not sure it's that useful in this case. It doesn't break any tests.

of course, but in this case `git log` will not tell you when you've upgraded the version (and possibly caused something else).

PR #2516, opened 2021-05-17
Add dialect to allow formatting with -Xsource:3 option

tgodzik:
Since Scala 2.13.6 and 2.12.14 it is possible to use `-Xsource:3`, which adds support for some of the new scala 3 syntax. ~~None of those changes should influence Scalafmt, so I think it would be good to set it as the default.~~

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
@@ -89,6 +89,7 @@ object ScalafmtRunner {
         Scala211,
         scala212,
         scala213,
+        Scala213Source3,
         Sbt0137,
         Sbt1,
         Dotty,
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -101,6 +101,10 @@ enum class Weekday {
 
 Please also see [rewrite rules](#scala3-rewrites) for Scala 3.
 
+Additionally, when using `-Xsource:3` option for Scala 2 you can change the
+dialect to `Scala213Source3`, which will allow to format the new syntax ported
+from Scala 3.
+
 ## Presets
 
 Some sections provide preset values to set multiple parameters at once. These


project/Dependencies.scala
```diff
@@ -6,7 +6,7 @@ import org.portablescala.sbtplatformdeps.PlatformDepsPlugin.autoImport._
 
 object Dependencies {
   val metaconfigV = "0.9.10"
-  val scalametaV  = "4.4.17"
+  val scalametaV  = "4.4.18"
```
kitbellew, line 9: now i'd definitely recommend this upgrade to be done separately

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
```diff
@@ -80,7 +80,8 @@ object ScalafmtRunner {
       .withAllowTrailingCommas(true) // SIP-27, 2.12.2
     val scala213 = Scala213
       .withAllowTrailingCommas(true)
-    val default = scala213
+    val default = Scala213Source3
```
kitbellew, line 83: please don't change the default.

also, you should add the newly supported dialects in the decoder below, otherwise they can't be used.

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
tgodzik: Why would that be a problem? This is just a small subset of syntax changes and people would be super confused if they needed to change the dialect. Scala213 us basically the subset of Scala213Source3, so it only supports more not less.

project/Dependencies.scala
tgodzik: https://github.com/scalameta/scalafmt/pull/2517

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
kitbellew: a dialect dictates how rewrite rules can be applied. so we can't use a superset, or the result might not compile.

tgodzik: ok sure, I haven't thought about it.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
```diff
@@ -89,6 +90,7 @@ object ScalafmtRunner {
         Scala211,
         scala212,
         scala213,
+        source3,
```
kitbellew, line 93: should we just allow explicit `Scala213Source3` instead of the intermediate (and perhaps more ambiguous `source3`)?

tgodzik: Sure, I can add both. I thought that shorter might be better, but maybe that's not the case here.

tgodzik: I also added a small paragraph for it in the docs, what do you think?

kitbellew (approved):
thanks for the doc change!

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
kitbellew: i think it's lovely!

tgodzik:
I think we should be ok to release RC3, I will take a look into it tomorrow!

PR #2504, opened 2021-05-13
Remove optional braces in function blocks

tgodzik:

PR #2520, opened 2021-05-19
Make sure that lambda newlines are preserved under keep

tgodzik:
Addresses https://github.com/scalameta/scalafmt/pull/2504#discussion_r631858751

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -404,12 +404,12 @@ class Router(formatOps: FormatOps) {
             style.newlines.afterCurlyLambdaParams ne
               Newlines.AfterCurlyLambdaParams.squash
 
-          isCurlyLambda && (style.newlines.source match {
+          style.newlines.source match {
             case Newlines.fold => false
-            case Newlines.unfold => noSquash
+            case Newlines.unfold => isCurlyLambda && noSquash
             case Newlines.keep => newlines != 0
-            case Newlines.classic => newlines != 0 && noSquash
-          })
+            case Newlines.classic => isCurlyLambda && newlines != 0 && noSquash
+          }
         }
         val singleLineSplit =
           Split(Space, 0)
--- a/scalafmt-tests/src/test/resources/scala3/OptionalBraces_keep.stat
+++ b/scalafmt-tests/src/test/resources/scala3/OptionalBraces_keep.stat
@@ -1414,7 +1414,8 @@ val f: Int => Int = (x: Int) => {
   x * x
 }
 >>>
-val f: Int => Int = (x: Int) => x * x
+val f: Int => Int = (x: Int) =>
+  x * x
 <<< lambda multiple statements
 rewrite.scala3.removeOptionalBraces = yes
 ===


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -404,12 +404,12 @@ class Router(formatOps: FormatOps) {
             style.newlines.afterCurlyLambdaParams ne
               Newlines.AfterCurlyLambdaParams.squash
 
-          isCurlyLambda && (style.newlines.source match {
+          style.newlines.source match {
```
kitbellew, line 407: i think fold case should replace false with isCurlyLamvda, and keep should become newlines != 0 || isCurlyLambda.

scalafmt-tests/src/test/resources/scala3/OptionalBraces_keep.stat
```diff
@@ -1414,7 +1414,8 @@ val f: Int => Int = (x: Int) => {
   x * x
 }
 >>>
-val f: Int => Int = (x: Int) => x * x
+val f: Int => Int = (x: Int) =>
+  x * x
```
kitbellew, line 1418: do we need to add tests for which isCurlyLambda would be true?

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: That seems to break a bunch of tests:
```
==> X org.scalafmt.FormatTests.newlines/source_fold.stat:1554: preserve ctor body with zero args     |  0.004s munit.ComparisonFailException: /home/tgodzik/Documents/scalafmt/scalafmt-tests/src/test/resources/newlines/source_fold.stat:1554
1553:val foo = new Bar { () => f(42) }
1554:<<< preserve ctor body with zero args
1555:val foo = new Bar { () => f(42) }
values are not the same
=> Obtained
val foo = new Bar { () =>
  f(42)
}

=> Diff (- obtained, + expected)
-val foo = new Bar { () =>
-  f(42)
-}
+val foo = new Bar { () => f(42) }
```

Looks like fold needs to stay false. As for `newlines != 0 || isCurlyLambda.` that breaks:

```
==> X org.scalafmt.FormatTests.newlines/source_keep.stat:1656: keep ctor body with zero args         |  0.027s munit.ComparisonFailException: /home/tgodzik/Documents/scalafmt/scalafmt-tests/src/test/resources/newlines/source_keep.stat:1656
1655:}
1656:<<< keep ctor body with zero args
1657:val foo = new Bar { () => f(42) }
values are not the same
=> Obtained
val foo = new Bar { () =>
  f(42)
}

=> Diff (- obtained, + expected)
-val foo = new Bar { () =>
-  f(42)
-}
+val foo = new Bar { () => f(42) }
```

scalafmt-tests/src/test/resources/scala3/OptionalBraces_keep.stat
tgodzik: there is a bunch of tests which already test it (they broke when I removed it)

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: yes, you're right, i shouldn't review code in the middle of the night. the old logic was isCurlyLambda && false, which is false.

kitbellew (approved):

PR #2528, opened 2021-05-23
ScalafmtConfig: switch to using ConfDecoderEx

### LINKED ISSUE #1490: How to append to default pattern array?
Hello, I'd like to add a pattern to `rewrite.neverInfix.excludeFilters` and I add this in `.scalafmt.conf`
```
rewrite.neverInfix.excludeFilters += "exclude"
```
and apparently that doesn't work as expected as only what is added is considered. So how to append to default pattern array?

kitbellew:
Fixes #1490.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -150,6 +150,21 @@ top-level.
     }
 ```
 
+### Appending to preset collections
+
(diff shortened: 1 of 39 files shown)


olafurpg (approved):
LGTM 👍 It's really nice to finally have a consistent solution to append a base setting

PR #2533, opened 2021-05-24
FormatWriter: insert optional-braces end markers

### LINKED ISSUE #2523: Automatically insert end markers for longer braceless blocks
## Steps

Given code like this:
```scala
def hello = 
  println(1)
  println(2)
  println(3)
  println(4)
```

```bash
scalafmt
```

## Problem

```scala
def hello = 
  println(1)
  println(2)
  println(3)
  println(4)
```

## Expectation

```scala
def hello = 
  println(1)
  println(2)
  println(3)
  println(4)
end hello
```

## Workaround

Manually inserting the markers.

## Notes

This is a follow up from https://contributors.scala-lang.org/t/should-the-compiler-help-with-inserting-end-markes/5074/9

I am not exactly sure how hard it would be to implement @kitbellew might know better. But this would essentially mean inserting a new end stat, since currently `end` is treated as a separate stat by the parser.

kitbellew:
Fixes #2523.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -71,6 +71,13 @@ class FormatWriter(formatOps: FormatOps) {
           sb.append(rewrittenToken)
       }
 
+      location.optionalBraces.toSeq
+        .sortBy { case (indent, _) => -indent }
+        .foreach { case (indent, label) =>
+          val indentStr = getIndentation(indent)
+          sb.append("\n").append(indentStr).append("end ").append(label)
+        }
+
       entry.formatWhitespace
     }
 
@@ -100,6 +107,11 @@ class FormatWriter(formatOps: FormatOps) {
       !initStyle.rewrite.redundantBraces.parensForOneLineApply.contains(false)
     )
       replaceRedundantBraces(result)
+    if (
+      initStyle.runner.dialect.allowSignificantIndentation &&
+      initStyle.rewrite.scala3.insertEndMarkerMinLines > 0
+    )
+      checkOptionalBraces(result)
 
     new FormatLocations(result)
   }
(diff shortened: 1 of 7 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -71,6 +71,12 @@ class FormatWriter(formatOps: FormatOps) {
           sb.append(rewrittenToken)
       }
 
+      location.optionalBraces.toSeq.sortBy(-_._1).foreach {
```
tgodzik, line 74:
```suggestion
      location.optionalBraces.toSeq.sortBy{case (indent, _) => -indent}.foreach {
```
it could be a bit clearer when reading.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -195,6 +206,43 @@ class FormatWriter(formatOps: FormatOps) {
     }
   }
 
+  private def checkOptionalBraces(locations: Array[FormatLocation]): Unit =
+    locations.foreach { floc =>
+      val ob = formatOps.OptionalBraces.get(floc.formatToken)(floc.style)
+      val ownerOpt = ob.flatMap(_.owner).filter {
+        /* if we add the end marker, it might turn a single-stat expression (or
```
tgodzik, line 213: Let's mention it in the docs maybe so that people are not surprised if they set the new option to `1`

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -1416,4 +1465,47 @@ object FormatWriter {
...
+    case _: Term.NewAnonymous => "new"
+    case t: Defn.Class => t.name.value
+    case t: Defn.Object => t.name.value
+    case t: Defn.Trait => t.name.value
+    case t: Defn.Enum => t.name.value
+    case t: Defn.Given => t.name.value
+    case t: Pkg.Object => t.name.value
+    // definitions
+    case t: Defn.Def => t.name.value
+    case t: Defn.Macro => t.name.value
```
tgodzik, line 1480: This is not used in Scala 3, so I would remov eit.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -1416,4 +1465,47 @@ object FormatWriter {
...
+    case t: Defn.GivenAlias => t.name.value
```
tgodzik, line 1481:
```suggestion
    case t: Defn.GivenAlias if t.name.value.nonEmpty => t.name.value
    case t: Defn.GivenAlias => "given"
```

tgodzik (changes requested):
Looks great! Thanks for working on this! I imagine a lot of people will be super happy about it.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: done

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: done

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: done

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: done. perhaps i misread the doc.

kitbellew (commented):
new patch submitted, ptal

tgodzik (approved):
LGTM! Looks awesome. I will take a look at the scalafix issue with scalameta tomorrow and we should be good to release new scalameta, update it here and do the RC4 so that people can experiment with adding the end marker.

PR #2534, opened 2021-05-24
FormatWriter: fix begin location for end marker

kitbellew:
The problem arises when the owner tree refers to the entire source file, meaning there's no previous token.


tgodzik (approved):
LGTM!

PR #2538, opened 2021-05-25
Trees: use .toString to get original input syntax

kitbellew:


tgodzik (approved):
LGTM! I didn't know there was a difference here. Thanks for figuring it out!

Issue #174, opened 2016-04-18
Align types of class arguments


densh:
Diff:
```scala
 final case class Focus(
 -  val blocks:     Seq[Block],
 -  val name:       Local,
 -  val params:     Seq[Val.Local],
 -  val insts:      Seq[Inst],
 -  val value:      Val,
 -  val isComplete: Boolean
 +    val blocks: Seq[Block],
 +    val name: Local,
 +    val params: Seq[Val.Local],
 +    val insts: Seq[Inst],
 +    val value: Val,
 +    val isComplete: Boolean
  )
```

I'd like to preserve alignment of types after column.

Using: 0.2.1

olafurpg:
Work on this issue is well ahead on this branch: https://github.com/olafurpg/scalafmt/tree/align-colon

The main difference is that we insert vertical alignment spaces for `NoSplit`, previously it was only for `Space`. The reason I've been reluctant to merge the branch is that it doesn't play so nicely in cases like this

```scala
case class Bar(i: Int) extends B
case object A extends B
```

In this example we have two potential alignment candidates for Bar, the `:` and the `extends`. There is only one alignment candidate for A, the `extends`. The current implementation only aligns the prefix of alignment candidates that match, which in this case is 0 candidates.

olafurpg:
Closed by #441. The caveats above still apply, but those warrant a separate issue.

PR #2540, opened 2021-05-26
AlignToken: add an optional `parent` pattern

kitbellew:


scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
```diff
@@ -1179,3 +1179,63 @@ libraryDependencies ++= Seq(
...
+  ) {
+    def Foo(
+      firstParam: Int,
+      secondParam: String,
+      thirdParam: Boolean
+    ) = ???
+  }
+}
+>>>
+object a {
```
tgodzik, line 1199: I think I don't really understand this settings, should the `:` be aligned?
```scala
object a {  
  case class Foo(
    firstParam : Int,      
    secondParam: String,      
    thirdParam : Boolean  
  ) { 
      def Foo(        
        firstParam : Int,        
        secondParam: String,        
        thirdParam : Boolean    
      ) = ???  
}}
```

This seems to work in case of `=>`, but in the [examples](https://scalameta.org/scalafmt/docs/configuration.html#aligntokens) it doesn't seem to align anything for `%%` or `%`

tgodzik: But it actually work for `build.sbt` file I tested :thinking:

scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
```diff
@@ -1179,3 +1179,63 @@ libraryDependencies ++= Seq(
...
+  case class Foo(
+      firstParam:  Int,
```
kitbellew, line 1231: the point of the change is to allow further choice. here, in ctor, it's aligned. and below, in def, is not.

scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
tgodzik: Sure I was just thinking based on docs that it would align colons instead of the type after it (it does align `%`, `=>` and not the expression afterwards). If that is the intention then I misunderstood the setting.

As for the lack of `%` align i thought it was connected to the same issue, but seems to be just a problem with the docs themselves

tgodzik (approved):
LGTM!

scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
kitbellew: > Sure I was just thinking based on docs that it would align colons instead of the type after it (it does align `%`, `=>` and not the expression afterwards). If that is the intention then I misunderstood the setting.

the reason is not mentioned in the code but looks like a fix for #174. it only aligns if there's a space, and uses the previous alignment offset if necessary.

Issue #2522, opened 2021-05-21
Scala 3: Reformatting breaks invocation of curried method


djneades:
- **Version**: 3.0.0-RC3
- **Integration**: Sbt via sbt-scalafmt 2.4.2
- **Configuration**:
```
version = 3.0.0-RC3
maxColumn = 80

runner.dialect = scala3
```

## Steps

Given code like this:
```scala
def f(s: String)(t: String) = s"$t: $s"

def g =
  f("Foo")
    ("Bar")
```

When I run scalafmt like this:

```bash
scalafmt
```

## Problem

Scalafmt formats the code like this, which does not compile:
```scala
def f(s: String)(t: String) = s"$t: $s"

def g =
  f("Foo")
  ("Bar")
```

## Expectation

I should like the formatted code to compile.

## Workaround

Scalafmt will not break code where the second opening paren is moved up a line:
 
```scala
def f(s: String)(t: String) = s"$t: $s"

def g =
  f("Foo")(
    "Bar")
```

## Notes

Using the `RedundantParens` rewrite also causes a possibly related problem. With this configuration:

```
version = 3.0.0-RC3
maxColumn = 80

runner.dialect = scala3

rewrite.rules = [ RedundantParens ]
```

Reformatting the original code above results in the following non-compiling code:

```scala
def f(s: String)(t: String) = s"$t: $s"

def g =
  f("Foo") "Bar"
```

tgodzik:
Thanks for reporting! I wasn't aware of this kind of indentation syntax:
```
def g =
  f("Foo")
    ("Bar")
```

I opened up an issue in dotty https://github.com/lampepfl/dotty/issues/12554 since I can't find it actually documented anywhere.

djneades:
@tgodzik Thank you!

djneades:
@tgodzik Would you find this (currently legal with Scala 3.0) indentation syntax unexpected?

```scala
def f(aLongNameToForceWrapping: String)
     (anotherLongNameToForceWrappingOnToTheNextLine: String) =
  "hello"

def g(aLongNameToForceWrapping: String)
     (using anotherLongNameToForceWrappingOnToTheNextLine: String) =
  f("Foo")("Bar")
```

I like the above – it seems much cleaner than the following output from Scalafmt using the first configuration above:

```scala
def f(aLongNameToForceWrapping: String)(
    anotherLongNameToForceWrappingOnToTheNextLine: String
) =
  "hello"

[...]

PR #2553, opened 2021-05-27
[Scala 3] Add tests for indented curried params

tgodzik:
Fixes https://github.com/scalameta/scalafmt/issues/2522


kitbellew (approved):
so we parse it correctly, but don't format it the same way, right?

tgodzik:
> so we parse it correctly, but don't format it the same way, right?

Yeah, we don't have an option for parsing it like that. Unless, 'source.keep' would work here? I can check later on.

Issue #2554, opened 2021-05-28
Failure on macro splices in Scala 3 when RedundantBraces enabled


note:
```
version = 3.0.0-RC3
runner.dialect = scala3
rewrite.rules = [RedundantBraces]
```

## Steps

Given code like this:
```scala
inline def test(inline v: String, inline pred: String): String =
    ${ testCode('v, 'pred)        }

  private def testCode(v: Expr[String], pred: Expr[String])(using Quotes): Expr[String] =
    Expr("abc")
```

When I run scalafmt like this:

```bash
scalafmt
```

## Problem

Scalafmt fails with:
```scala
[info] Formatting 6 Scala sources...
Router:2092 MISSING CASE:
[1229]$                              MacroSplice
[1231]                               Space
[1245]startsWithCode                 Ident
Router:2092 MISSING CASE:
[1529]$                              MacroSplice
[1531]                               Space
[1543]endsWithCode                   Ident
Router:2092 MISSING CASE:
[1798]$                              MacroSplice
[1800]                               Space
[1808]testCode                       Ident
[error] /.../ValidateString.scala:30: error: Unable to format file due to bug in scalafmt
[error]     ${ startsWithCode('v, 'pred)  }
[error]     ^: /.../ValidateString.scala
[error] (Compile / scalafmt) /.../ValidateString.scala:30: error: Unable to format file due to bug in scalafmt
[error]     ${ startsWithCode('v, 'pred)  }
[error]     ^: /.../ValidateString.scala
```

## Expectation

scalafmt does not fail

## Workaround

It works correctly without `RedundantBraces`

## Notes

It looks like `RedundantBraces` incorrectly removes the braces for macro splice and that causes problems

PR #2555, opened 2021-05-28
Don't remove braces of macro splices

note:
Fixes https://github.com/scalameta/scalafmt/issues/2554

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -107,29 +107,29 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
     def replace = replaceToken("(") {
       val rt = ft.right
       new Token.LeftParen(rt.input, rt.dialect, rt.start)
-    }
-    // we will not keep it but will hint to onRight
-    (ft.meta.rightOwner, ft.left) match {
-      case (_, _: Token.MacroSplice) =>
-        null
-      case (t: Term.Function, _) if t.tokens.last.is[Token.RightBrace]  =>
-        if (!okToRemoveFunctionInApplyOrInit(t)) null
-        else if (okToReplaceFunctionInSingleArgApply(t)) replace
-        else removeToken
-      case (t: Term.Block, _) =>
-        if (okToReplaceBlockInSingleArgApply(t)) replace
-        else if (processBlock(t)) removeToken
-        else null
-      case (_: Term.Interpolate, _) =>
-        if (processInterpolation) removeToken else null
-      case (meta.Importer(_, List(x)), _)
-          if !(x.is[Importee.Rename] || x.is[Importee.Unimport]) ||
-            style.runner.dialect.allowAsForImportRename &&
-            (ConvertToNewScala3Syntax.enabled ||
-              !x.tokens.exists(_.is[Token.RightArrow])) =>
-        removeToken
-      case _ => null
-    }
+    } // we will not keep it but will hint to onRight
+    if (ft.left.is[Token.MacroSplice] || ft.left.is[Token.MacroQuote])
+      null
+    else
+      ft.meta.rightOwner match {
+        case t: Term.Function if t.tokens.last.is[Token.RightBrace] =>
+          if (!okToRemoveFunctionInApplyOrInit(t)) null
+          else if (okToReplaceFunctionInSingleArgApply(t)) replace
+          else removeToken
+        case t: Term.Block =>
+          if (okToReplaceBlockInSingleArgApply(t)) replace
+          else if (processBlock(t)) removeToken
+          else null
+        case _: Term.Interpolate =>
+          if (processInterpolation) removeToken else null
+        case meta.Importer(_, List(x))
+            if !(x.is[Importee.Rename] || x.is[Importee.Unimport]) ||
+              style.runner.dialect.allowAsForImportRename &&
+              (ConvertToNewScala3Syntax.enabled ||
+                !x.tokens.exists(_.is[Token.RightArrow])) =>
+          removeToken
+        case _ => null
+      }
   }
 
   private def onRightBrace(
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -107,19 +107,22 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
     def replace = replaceToken("(") {
       val rt = ft.right
       new Token.LeftParen(rt.input, rt.dialect, rt.start)
-    } // we will not keep it but will hint to onRight
-    ft.meta.rightOwner match {
-      case t: Term.Function if t.tokens.last.is[Token.RightBrace] =>
+    }
+    // we will not keep it but will hint to onRight
+    (ft.meta.rightOwner, ft.left) match {
+      case (_, _: Token.MacroSplice) =>
+        null
+      case (t: Term.Function, _) if t.tokens.last.is[Token.RightBrace]  =>
         if (!okToRemoveFunctionInApplyOrInit(t)) null
         else if (okToReplaceFunctionInSingleArgApply(t)) replace
         else removeToken
-      case t: Term.Block =>
+      case (t: Term.Block, _) =>
         if (okToReplaceBlockInSingleArgApply(t)) replace
         else if (processBlock(t)) removeToken
         else null
-      case _: Term.Interpolate =>
+      case (_: Term.Interpolate, _) =>
         if (processInterpolation) removeToken else null
-      case meta.Importer(_, List(x))
+      case (meta.Importer(_, List(x)), _)
           if !(x.is[Importee.Rename] || x.is[Importee.Unimport]) ||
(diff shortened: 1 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -107,19 +107,22 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
...
+    // we will not keep it but will hint to onRight
+    (ft.meta.rightOwner, ft.left) match {
+      case (_, _: Token.MacroSplice) =>
+        null
+      case (t: Term.Function, _) if t.tokens.last.is[Token.RightBrace]  =>
         if (!okToRemoveFunctionInApplyOrInit(t)) null
         else if (okToReplaceFunctionInSingleArgApply(t)) replace
         else removeToken
-      case t: Term.Block =>
+      case (t: Term.Block, _) =>
```
tgodzik, line 119: You can just change one line here:
```suggestion
      case t: Term.Block if ft.left.isNot[Token.MacroSplice] =>
```

tgodzik (changes requested):
Thanks for the contribution! I think we can do it a bit simpler. You should also run `./scalafmt`.

And also, hi @note !

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -107,19 +107,22 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
     def replace = replaceToken("(") {
       val rt = ft.right
       new Token.LeftParen(rt.input, rt.dialect, rt.start)
-    } // we will not keep it but will hint to onRight
-    ft.meta.rightOwner match {
-      case t: Term.Function if t.tokens.last.is[Token.RightBrace] =>
+    }
+    // we will not keep it but will hint to onRight
```
kitbellew, line 111: this comment is to the replaced token above, not to the next line

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -107,19 +107,22 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
...
+    (ft.meta.rightOwner, ft.left) match {
```
kitbellew, line 112: agree with @tgodzik, don't add a second part to match.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
kitbellew: i would even put it after `=>`:

```
  if (ft.left.is[...]) null
  else if ...
```

kitbellew (changes requested):

kitbellew:
@note is macro splice the only case? are we handling interpolation splices (and are they blocks), or other similar cases?

scalafmt-tests/src/test/resources/scala3/Macro.stat
```diff
@@ -31,7 +31,13 @@ throw new AssertionError(s"failed assertion: ${${ showExpr(expr) }}")
 '{   (x: T) => ${ f('x) } }
 >>>
 '{ (x: T) => ${ f('x) } }
-<<< unctions spliced
+<<< functions with RedundantBraces
+rewrite.rules = [RedundantBraces]
+===
+'{   (x: T) => ${ f('x) } }
```
kitbellew, line 37: one other thing: can you create another test case, with extra braces `${ { f('x) } }`?

note:
@kitbellew @tgodzik Thanks a lot for your reviews! It's definitely needed here as I am not familiar with scalafmt architecture so I might not be doing things wrong :)

> @note is macro splice the only case? are we handling interpolation splices (and are they blocks), or other similar cases?

Once you asked I realized there's a similar issue with macro quotes. Added a check for that and relevant tests.

Regarding interpolation, if you meant string interpolation, I run the test for `"abc${var.m}"` and it didn't hit `onLeftBrace`. I don't know how those are represented but seems we don't have to bother about it here

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
note: Was not aware of `is` / `isNot`. I've used both suggestions to some extend but the end result is slightly different

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -108,24 +108,29 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
...
-        removeToken
-      case _ => null
+    if (ft.left.is[Token.MacroSplice] || ft.left.is[Token.MacroQuote]) {
+      null
+    } else {
+      ft.meta.rightOwner match {
+        case t: Term.Function if t.tokens.last.is[Token.RightBrace] =>
+          if (!okToRemoveFunctionInApplyOrInit(t)) {
+            null
+          } else if (okToReplaceFunctionInSingleArgApply(t)) replace
```
kitbellew, line 118: what's changed here? others are simple indentation. if nothing, can we restore?

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -108,24 +108,29 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
...
-      case meta.Importer(_, List(x))
-          if !(x.is[Importee.Rename] || x.is[Importee.Unimport]) ||
-            style.runner.dialect.allowAsForImportRename &&
-            (ConvertToNewScala3Syntax.enabled ||
-              !x.tokens.exists(_.is[Token.RightArrow])) =>
-        removeToken
-      case _ => null
+    if (ft.left.is[Token.MacroSplice] || ft.left.is[Token.MacroQuote]) {
+      null
+    } else {
```
kitbellew, line 113: I'd remove all 4 braces, this repo doesn't use them for single-stat blocks.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -108,24 +108,29 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
...
-      case _: Term.Interpolate =>
-        if (processInterpolation) removeToken else null
-      case meta.Importer(_, List(x))
-          if !(x.is[Importee.Rename] || x.is[Importee.Unimport]) ||
-            style.runner.dialect.allowAsForImportRename &&
-            (ConvertToNewScala3Syntax.enabled ||
-              !x.tokens.exists(_.is[Token.RightArrow])) =>
-        removeToken
-      case _ => null
+    if (ft.left.is[Token.MacroSplice] || ft.left.is[Token.MacroQuote]) {
```
kitbellew, line 111: quick question: is this legal
```
$/* comment */{...}
```

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
note: It compiles with scala 3.0.0 so seems it's legal

It fails with scalafmt (both current release and on my branch) with `Symbol literals are no longer allowed`. Looks like the problem lies in `scala.meta.internal.parsers.ScalametaParser`

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
note: Restored

scalafmt-tests/src/test/resources/scala3/Macro.stat
```diff
@@ -7,6 +7,18 @@ ${ locationImpl() }
 '{   new Location() }
 >>>
 '{ new Location() }
+<<< quote brace with RedundantBraces
```
tgodzik, line 10: Isn't this test case repeated?

note: Good catch, pushing update

tgodzik (approved):
LGTM from my side, thanks!

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
kitbellew: @tgodzik do we want to fix this? if so, `ft.left` should be replaced with `prevNonComment(ft).left`.

tgodzik: I can take a look at it after the weekend and but we can for sure update this to work with comments just in case.

kitbellew: @note thank you for this submission. rather than torture you again, i made the small change to handle comments when the parser is ready (in #2557).

Issue #2511, opened 2021-05-14
Feature request: ability to have multiple blank lines around various kinds of elements


djneades:
- **Version**: 3.0.0-RC2

## Desire

I should like to be able to configure Scalafmt to format code to look like this, with two blank lines surrounding top-level [template definitions (`TmplDef)`](https://dotty.epfl.ch/docs/reference/syntax.html#declarations-and-definitions):

```scala
import scala.collection.mutable.ArrayBuffer


/** 2 blank lines above and below top-level trait definition */
trait A


// 2 blank lines above top-level class definition
// (none below due to end of file)
class B:

  /** 1 blank line above and below method definition */
  def A = ???

  /** 1 blank line above and below nested trait definition */
  trait Bar
      extends A

  /** 1 blank line above (none below due to end of file) */
  lazy val x = 1
```

Some coding styles may prefer something like this, with three blank lines surrounding top-level template definitions and two blank lines surrounding nested method definitions:

```scala
import scala.collection.mutable.ArrayBuffer



/** 3 blank lines above and below top-level trait definition */
trait A:

   // 1 blank line above (2 below due to following def)
   val x = 32
  
  
   /** 2 blank lines above and below nested method definition */
   def A =
      // beginning of a lengthy method body containing a blank line
     
      // continuation of lengthy body following a blank line
      ???


   // 2 blank lines above nested method definition
   // (3 below due following class)
   def B = ???



// 3 blank lines above top-level class definition
// (none below, due to end of file)
class B:


   /** 2 blank lines above and below */
   def A = ???


   /** 2 blank lines above and below nested-trait */
   trait Bar
      extends A


   // 1 blank line normally above val, but 2 enforced due to
   // preceding nested trait (none following due to end of file)
   lazy val x = 1
```

## Rationale

Being able to add more than one blank line before certain major elements such as [template definitions (`TmplDef)`](https://dotty.epfl.ch/docs/reference/syntax.html#declarations-and-definitions) would make it easier to distinguish between the various major and minor elements in a source file. For example, it is a lot easier to see where a class begins and ends if that class is surrounded by more whitespace than the methods and other items _within_ that class. In Scala, where multiple traits, classes, objects, etc. are often defined in a single source file, marking the boundaries of such major elements clearly would be especially useful.

[...]

PR #2558, opened 2021-05-31
FormatWriter: fully determine blank line indices

kitbellew:
Earlier, we would store a potential location and then later check if it should be supplied with an extra blank. Instead, do that at once.

Towards #2511.


tgodzik (approved):
LGTM!

PR #2565, opened 2021-06-01
FormatWriter: additional changes in handling top-level-stat blanks

kitbellew:
Towards #2511.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Newlines.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Newlines.scala
@@ -168,8 +168,8 @@ case class Newlines(
     topLevelStatementsMinBreaks: Int = 1,
     topLevelStatements: Seq[BeforeAfter] = Seq.empty,
     beforeTemplateBodyIfBreakInParentCtors: Boolean = false,
-    templateBodyIfMinStatements: Seq[BeforeAfter] = Seq.empty,
-    templateBodyMinStatements: Int = 2,
+    topLevelBodyIfMinStatements: Seq[BeforeAfter] = Seq.empty,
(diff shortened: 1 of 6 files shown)


kitbellew:
@tgodzik tests passed, one had an internet-access temp fluke.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Newlines.scala
```diff
@@ -168,14 +168,8 @@ case class Newlines(
...
     beforeTemplateBodyIfBreakInParentCtors: Boolean = false,
-    templateBodyIfMinStatements: Seq[BeforeAfter] = Seq.empty,
-    templateBodyMinStatements: Int = 2,
-    @annotation.DeprecatedName(
-      "alwaysBeforeTopLevelStatements",
-      "Use newlines.topLevelStatements instead",
-      "2.5.0"
-    )
-    private val alwaysBeforeTopLevelStatements: Boolean = false,
+    topLevelBodyIfMinStatements: Seq[BeforeAfter] = Seq.empty,
```
tgodzik, line 171: Should we still support `templateBodyIfMinStatements` just in case?

tgodzik (approved):
LGTM, just one question

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Newlines.scala
kitbellew: i assumed that since we added this option recently, people who may have used it with one of the `RC` previews will simply rename it. is there any reason to keep this intermediate name?

tgodzik: Ach ok, I forgot about that!

Issue #2560, opened 2021-05-31
Spurious space inserted in `+_`/`-_` type lambdas


neko-kai:
- **Version**: 3.0.0-RC4
- **Integration**: IntelliJ
- **Configuration**:
```
version = "3.0.0-RC4"
runner.dialect = scala213source3
```

## Steps

Given code like this:

```scala
class x(arg: OptionT[Either[Int, +_]])
class y(arg: OptionT[Either[Int, +*]])
```

## Problem

Scalafmt formats code like this:

```scala
class x(arg: OptionT[Either[Int, + _]]) // space added between `+` & `_` — space is legal but not generally desirable
class y(arg: OptionT[Either[Int, +*]]) // no space added and adding it would make syntax illegal
```

## Expectation

I would like the formatted output to look like this:

```scala
class x(arg: OptionT[Either[Int, +_]])
class y(arg: OptionT[Either[Int, +*]])
```

## Workaround

None

## Notes

Support for `-_`/`+_` as identifiers was added in https://github.com/scalameta/scalameta/pull/2348

PR #2566, opened 2021-06-01
Fix issues with +_ type lambdas

tgodzik:
Fixes https://github.com/scalameta/scalafmt/issues/2560

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1648,6 +1648,11 @@ class Router(formatOps: FormatOps) {
           if isTypeVariant(leftOwner) =>
         Seq(Split(Space(isSymbolicIdent(tok.right)), 0))
 
+      // Kind projector type lambda
+      case FormatToken(T.Ident("+" | "-"), T.Underscore(), _)
+          if leftOwner.is[Type.Name] =>
+        Seq(Split(NoSplit, 0))
+
       // Var args
       case FormatToken(_, T.Ident("*"), _) if rightOwner.is[Type.Repeated] =>
         Seq(
--- a/scalafmt-tests/src/test/resources/unit/Type.stat
+++ b/scalafmt-tests/src/test/resources/unit/Type.stat
@@ -105,3 +105,20 @@ trait Foo[+->[_, _]]
 trait Foo[+ F[_, _]]
 >>>
 trait Foo[+F[_, _]]
+<<< kind projector #2560 
+runner.dialect = scala213source3
+maxColumn = 80
+===
+object O{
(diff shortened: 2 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1648,6 +1648,11 @@ class Router(formatOps: FormatOps) {
           if isTypeVariant(leftOwner) =>
         Seq(Split(Space(isSymbolicIdent(tok.right)), 0))
 
+      // Kind projector type lambda
+      case tok @ FormatToken(T.Ident("+") | T.Ident("-"), T.Underscore(), _)
+          if leftOwner.is[Type.Name] =>
+        Seq(Split(NoSplit, 0))
+
```
kitbellew, line 1655:
```suggestion
      // Kind projector type lambda
      case FormatToken(T.Ident("+" | "-"), _: T.Underscore, _)
          if leftOwner.is[Type.Name] =>
        Seq(Split(NoSplit, 0))

```

kitbellew: above, `tok` is unused, and i think `Ident` can be used once.

scalafmt-tests/src/test/resources/unit/Source3.source
```diff
@@ -0,0 +1,12 @@
+<<< kind projector #2560
```
kitbellew, line 1: optional: we could put it in one of existing scala3 files, or `Dialect.stat`.

scalafmt-tests/src/test/resources/unit/Source3.source
```diff
@@ -0,0 +1,12 @@
+<<< kind projector #2560 
+runner.dialect = scala213source3
+===
+class x(arg: OptionT[Either[Int, +_]])
```
kitbellew, line 4: if you wrap in `object a { ... }` or `package a { ... }`, you can use an existing `.stat` file.

scalafmt-tests/src/test/resources/unit/Source3.source
tgodzik: Since it's more related to Scala 2, I put it into `unit/Type.stat`

scalafmt-tests/src/test/resources/unit/Source3.source
tgodzik: Done

kitbellew (approved):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Fixed, good catch!

PR #2567, opened 2021-06-01
FormatWriter: implement granular blanks control

### LINKED ISSUE #2511: Feature request: ability to have multiple blank lines around various kinds of elements

kitbellew:
Fixes #2511.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -22,6 +22,7 @@ import scala.meta.{
   Case,
   Defn,
   Enumerator,
+  ImportExportStat,
   Importer,
   Lit,
   Pat,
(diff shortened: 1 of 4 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -1085,33 +1087,55 @@ class FormatWriter(formatOps: FormatOps) {
...
+      @tailrec
+      def getNest(tree: Tree, curNest: Int): Int = tree.parent match {
+        case Some(_: Source) | None => curNest
+        case Some(t: Pkg) if !indentedPackage(t) => getNest(t, curNest)
+        case Some(t) => getNest(t, curNest + 1)
+      }
+      def setTopStats(owner: Tree, stats: Seq[Stat], curNest: Int = 1): Unit = {
+        if (stats.isEmpty) return
+        val nest = getNest(owner, curNest)
+        if (nest < 0) return
```
tgodzik, line 1109: When will nest be smaller than 0 ? It seems it's only incremented.

tgodzik (approved):
Just one small question, otherwise LGTM!

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: I used to have a check for `Term.Block` there, which returned -1, but removed it as unnecessary (see comment below). however, in the future, it's possible that we'll add another check which will prohibit nesting by returning a negative value.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -1134,6 +1158,9 @@ class FormatWriter(formatOps: FormatOps) {
         override def apply(tree: Tree): Unit =
           tree match {
             case _: Term.Block =>
+            case t: Source =>
```
kitbellew, line 1161: the Block is skipped just above (in connection with your getNest question).

Issue #2559, opened 2021-05-31
Scala 3 end markers should be inserted at the appropriate place


djneades:
- **Version**: 3.0.0-RC4
- **Integration**: sbt-scalafmt 2.4.2
- **Configuration**:
```
version = 3.0.0-RC4
maxColumn = 80

runner.dialect = scala3
rewrite.scala3.insertEndMarkerMinLines = 2
```

## Steps

Given code like this:
```scala
def f(s: String) =
  val t = "Bar"
  val u = "Baz"
  s"$t $u $s"

/** Some doc comment */
def g = f("Foo")
```

When I run scalafmt like this:

```bash
scalafmt
```

## Problem

Scalafmt formats code like this:
```scala
def f(s: String) =
  val t = "Bar"
  val u = "Baz"
  s"$t $u $s"

/** Some doc comment */
end f
def g = f("Foo")
```

## Expectation

I should like the formatted output to look like this:
```scala
def f(s: String) =
  val t = "Bar"
  val u = "Baz"
  s"$t $u $s"
end f

/** Some doc comment */
def g = f("Foo")
```

kitbellew:
@tgodzik is this comment part of the previous `def` body?

tgodzik:
I think that's the case. Let me see if that is fixable in scalameta.

tgodzik:
Ok, so it looks like the positions are actually correct, but we are using `nextNonComment` which is skipping over the comment in question. Trying to fix it here: https://github.com/scalameta/scalafmt/pull/2568

PR #2568, opened 2021-06-01
Add end marker after the last token of a tree

tgodzik:
Fixes https://github.com/scalameta/scalafmt/issues/2559

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -224,7 +224,7 @@ class FormatWriter(formatOps: FormatOps) {
         }
       }
       ownerOpt.foreach { owner =>
-        val endFt = tokens.tokenAfter(owner)
+        val endFt = tokens.getLast(owner)
         if (!endFt.meta.rightOwner.is[Term.EndMarker]) {
           val end = endFt.meta.idx
           val eLoc = locations(end)
--- a/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
+++ b/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
@@ -1485,3 +1485,23 @@ package a.b.c:
       println("ah")
       println("ha")
 end c
+<<< two consequtive defs #2559
+runner.parser = source
+rewrite.scala3.insertEndMarkerMinLines = 2
+===
+def f(s: String) =
+  val t = "Bar"
+  val u = "Baz"
+  s"$t $u $s"
+
+/** Some doc comment */
+def g = f("Foo")
+>>>
+def f(s: String) =
+   val t = "Bar"
+   val u = "Baz"
+   s"$t $u $s"
+end f
+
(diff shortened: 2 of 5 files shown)


kitbellew (approved):
still, why is the comment part of the `owner`?

tgodzik:
> still, why is the comment part of the `owner`?

From what I checked the positions are correct and the comments is not a part of any of the defs.

Is it possible that `tokenAfter` just gets the next token after the tree? It uses `nextNonComment`, which doesn't iterate over tree tokens anymore.

kitbellew:
@tgodzik we might need a fix in `scalameta` after all. in this code
```
def f(s: String) =
  val t = "Bar"
  val u = "Baz"
  s"$t $u $s"
  // final comment
```
the comment is not part of the body:
```
TYPE: Defn$Def$DefnDefImpl
SOURCE: def f(s: String) =
  val t = "Bar"
  val u = "Baz"
  s"$t $u $s"
STRUCTURE: Defn.Def(Nil, Term.Name("f"), Nil, List(List(Term.Param(Nil, Term.Name("s"), Some(Type.Name("String")), None))), None, Term.Block(List(Defn.Val(Nil, List(Pat.Var(Term.Name("t"))), None, Lit.String("Bar")), Defn.Val(Nil, List(Pat.Var(Term.Name("u"))), None, Lit.String("Baz")), Term.Interpolate(Term.Name("s"), List(Lit.String(""), Lit.String(" "), Lit.String(" "), Lit.String("")), List(Term.Name("t"), Term.Name("u"), Term.Name("s"))))))
TOKENS: def,∙,f,(,s,:,∙,String,),∙,=,¶,∙,∙,val,∙,t,∙,=,∙,"Bar",¶,∙,∙,val,∙,u,∙,=,∙,"Baz",¶,∙,∙,s,",,$,t,,∙,$,u,,∙,$,s,,,"
```
therefore, end marker looks like this:
```
def f(s: String) =
   val t = "Bar"
   val u = "Baz"
   s"$t $u $s"
end f
// final comment
```

kitbellew:
well, and without end marker, same story:
```
def f(s: String) =
   val t = "Bar"
   val u = "Baz"
   s"$t $u $s"
// final comment
```
not sure if we already discussed this, i can't remember.

tgodzik:
So that's kind of expected. Outdent token is inserted before the next non comment token, which means that the comments still belong to the body of the Def.

I will check how it is in Dotty, but I'd don't expect indentation to be influenced by comments

tgodzik:
Ach no, I totally misunderstood what you were saying. But I think we don't really want to have the comment as part of the previous tree. Not sure how we would differentiate:

```scala
def f(s: String) =
   val t = "Bar"
   val u = "Baz"
   s"$t $u $s"
  // final comment

def g = 123
```

from 

```scala
def f(s: String) =
   val t = "Bar"
   val u = "Baz"
   s"$t $u $s"

// comment about g
def g = 123
```

Indentation isn't needed for comments, you can have:
```scala

class A:
  def hello = 1
  def bye = 1
// comment
  def ciao = 2
end A
```
and it still compiles.

I think the current situation is the best we can have.

PR #2569, opened 2021-06-02
FormatWriter: fix bugs around blank lines in top-level stats

### LINKED ISSUE #2511: Feature request: ability to have multiple blank lines around various kinds of elements

kitbellew:
Fixes #2511.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -837,12 +837,23 @@ Each entry on this list consists of the following fields:
 - `blanks`
   - consists of two integer fields, `before` and `after`, specifying how many
     blank lines need to be added in the appropriate place
+  - if omitted while the entry matches, serves to exclude another entry
   - can be specified as a single integer, to set both:
 
 ```
 // these two are equivalent
 newlines.topLevelStatementBlankLines = [{ blanks { before = 1, after = 1 } }]
 newlines.topLevelStatementBlankLines = [{ blanks = 1 }]
+```
+
+If you'd like to override or exclude some cases, add them explicitly:
+
+```
+newlines.topLevelStatementBlankLines = [
+  { maxNest = 0, blanks = 2 } // uses 2 blanks for all unindented statements
+  { regex = "^Import" } // excludes import groups; blanks are not specified
+]
+
 ```
 
 > If multiple entries match a statement, an entry with the lowest `minBreaks`
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -841,7 +841,7 @@ Each entry on this list consists of the following fields:
 
 ```
 // these two are equivalent
-newlines.topLevelStatementBlankLines = [{ blanks { before = 1, after = 1 }]
+newlines.topLevelStatementBlankLines = [{ blanks { before = 1, after = 1 } }]
 newlines.topLevelStatementBlankLines = [{ blanks = 1 }]
 ```
 
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -1324,18 +1324,17 @@ class FormatOps(
 
   // Returns trailing comment, if there exists one, otherwise formatToken
   @inline
-  final def trailingComment(ft: FormatToken, end: Int): FormatToken =
-    if (!ft.right.is[T.Comment]) ft
-    else trailingWithRightComment(ft, end).getOrElse(ft)
-  @tailrec
-  private def trailingWithRightComment(
-      ft: FormatToken,
-      end: Int
-  ): Option[FormatToken] = {
-    val nft = tokens.nextNonCommentSameLine(next(ft))
-    if (nft.hasBlankLine || nft.right.end >= end) Some(nft)
-    else if (nft.noBreak || !nft.right.is[T.Comment]) None
-    else trailingWithRightComment(nft, end)
+  final def trailingComment(ft: FormatToken, end: Int): FormatToken = {
+    @inline
+    def isDone(x: FormatToken) = x.hasBlankLine || x.right.end >= end
+    @tailrec
+    def iter(x: FormatToken): FormatToken = {
+      val nft = tokens.nextNonCommentSameLine(next(x))
+      if (isDone(nft)) nft
+      else if (!nft.right.is[T.Comment]) ft // original
+      else iter(nft)
+    }
+    if (!ft.right.is[T.Comment] || isDone(ft)) ft else iter(ft)
   }
 
   def xmlSpace(owner: Tree): Modification =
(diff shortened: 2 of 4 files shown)


kitbellew:
@djneades this should cover all your cases.

tgodzik (changes requested):
Looks good, but the tests seems to be failing.

kitbellew:
> Looks good, but the tests seems to be failing.

yes, the `bug in trailingComment` fix had a bug :)

kitbellew:
@tgodzik bug fixed.

tgodzik (approved):
LGTM!

djneades:
@kitbellew `3.0.0-RC4+22-19f883af-SNAPSHOT` is looking good to me, thank you.

kitbellew:
@djneades mazal tov!

djneades:
@kitbellew I’ve just tried playing with end markers for the first time, still with `3.0.0-RC4+22-19f883af-SNAPSHOT`. Is it possible to control the number of blank lines around those? I’ve tried using this config:

```
maxColumn = 80

runner.dialect = scala3

newlines {
  topLevelStatementBlankLines = [
      { blanks { before = 3, after = 3 } }
      { regex = "^Import", minBreaks=0, blanks { before = 1, after = 3 } }
      { regex = "^Pkg", minBreaks=0, blanks { before = 0, after = 3 } }
      { regex = "^Defn.Def", blanks { before = 2, after = 2 } }
      {
        regex = "^(Defn.Var|Defn.Val|Defn.Type)"
        blanks { before = 1, after = 1 }
      }
      {
        regex = "^Term.EndMarker", minBreaks=0, blanks { before = 1, after = 1 }
      }
    ]
}
```

I’m not absolutely sure whether I have the `Term.EndMarker` correct, but in any case it is not working for me. This:

```scala
package bug

final class Builder:

  def coalesce(seq: List[String]): Seq[String] =
    ???

end Builder

val ul = Builder()
```

is formatted as this:

```scala
package bug



final class Builder:

  def coalesce(seq: List[String]): Seq[String] =
    ???



end Builder

val ul = Builder()
```

My goal here is to have precisely one blank line above the end marker.

TIA.

kitbellew:
> ```
> newlines {
>   topLevelStatementBlankLines = [
>       { regex = "^Defn.Def", blanks { before = 2, after = 2 } }
>       {
>         regex = "^Term.EndMarker", minBreaks=0, blanks { before = 1, after = 1 }
>       }
>     ]
> }
> ```
> 
> ```scala
>   def coalesce(seq: List[String]): Seq[String] =
>     ???
> 
> end Builder
> ```

the first rule says put 2 blanks after `def`, and the second asks for 1 before `end`. `def` wins, with more blanks.

djneades:
@kitbellew That would make sense. But, with my example, Scalafmt is adding three blank lines above the end marker, not two …

djneades:
@kitbellew Thinking about this a little more, there seems to be a special rule that results in at most one blank line immediately after the beginning of a template (in that there is only one blank line between the `final class Builder` line and the following `def`). I wonder whether there should be a similar rule to limit the blank lines before the end marker?

djneades:
@kitbellew There is also a potential minor issue that end markers inserted with `rewrite.scala3.insertEndMarkerMinLines = 30` have no blank lines above them at all, even with the configuration above.

kitbellew:
> @kitbellew Thinking about this a little more, there seems to be a special rule that results in at most one blank line immediately after the beginning of a template (in that there is only one blank line between the `final class Builder` line and the following `def`). I wonder whether there should be a similar rule to limit the blank lines before the end marker?

first, the end markers are explicitly skipped and never match these rules (i wrote it but forgot).

second, the reason they behave like this is because the way the parser produces them is, they are a completely separate statement following the one they are ending. so the previous gets its blanks, before the end marker is even found. i can fix this, by moving the blanks to after the end marker.

however, what i still can't fix -- and about to give up on, because i also have a day job -- is actually controlling blanks before the end marker. (and this has to do with the rewrite.) so far, i can hardcode no-blanks before any end markers.

djneades:
@kitbellew Thank you. I absolutely understand about the day job and I am grateful for all your work.

PR #2572, opened 2021-06-02
Router: handle `newlines.beforeOpenParenXxxSite`

### LINKED ISSUE #2176: Support newlineBeforeOpenParen
Current version supports [newlineAfterOpenParen](https://scalameta.org/scalafmt/docs/configuration.html#verticalmultilineexcludedanglingparens).
But dangling parens looks unpleasant.
```
def other[F[_]](
  a: String,
  b: String
)( // <- this is nasty //
  implicit  c: String,
  d: String
) = a + b + c
```

I suggest adding option **newlineBeforeOpenParen**.
This option should add a newline before open paren.
So example with the above method will look like:
```
def other[F[_]]
(a: String, b: String)
(implicit c: String, d: String) = a + b + c
```
I guess it should work like 

1. read _**def**_ keyword
2. then the name of method
3. then bounds
4. and finally insert a newline before list param.

At first sight, this also will work with class definitions.
```
case class Foo[F[_]]
(a: String, b: String)
(implicit c: String, d: String) {
...
}
```
The difference in the keyword, instead _**def**_ keyword would be _**class**_

### LINKED ISSUE #2561: Feature request: neat Scala 3 definitions and calls – keep parens on the same line as arguments
- **Version**: 3.0.0-RC4
- **Configuration**:
```
version = 3.0.0-RC4
maxColumn = 40

runner.dialect = scala3

danglingParentheses {
  defnSite = false
}
```

## Steps

Given code like this:
```scala
def notAnInterestingFunction
    (first: String)
    (using fourth: Int)
  (
    "The quick brown fox",
    "jumped over the lazy",
    "dog.",
    "Maybe."
  )

def f = 
  notAnInterestingFunction
    ("first")
    ("second", "third")
    (using 52)
```

```bash
scalafmt
```

## Problem

```scala
def notAnInterestingFunction(
    first: String)(
    second: String,
  (
    "The quick brown fox",
    "jumped over the lazy",
    "dog.",
    "Maybe."
  )

def f =
    "second",
    "third"
  )(using 52)
```

## Expectation

## Notes

kitbellew:
Fixes #2561. Fixes #2176.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1714,6 +1714,82 @@ implicit protected val td: TildeArrow {
   type Out = RouteTestResult } = TildeArrow.injectIntoRoute
 ```
 
+### `newlines.sometimesBeforeColonInMethodReturnType`
+
+If true, in some rare cases a newline could be added before the colon which
+delimits the return type in a method signature.
+
+Normally, this will only happen if formatting otherwise will lead to line
+overflow or the return type spanning too many lines.
+
+Also, see [newlines.beforeOpenParenXxxSite](#newlinesbeforeopenparenxxxsite) below.
+
+```scala mdoc:defaults
+newlines.sometimesBeforeColonInMethodReturnType
+```
+
+```scala mdoc:scalafmt
+maxColumn = 40
+newlines.sometimesBeforeColonInMethodReturnType = true
+---
+def someVeryLongMethodName: Map[String, String] = ???
+```
+
+### `newlines.beforeOpenParenXxxSite`
+
+If true, a parameter group which doesn't fit on the same line will be formatted
+separately, with a newline just before the opening parenthesis.
+
+Specific formatting depends on the value of `newlines.source` (for `keep`, will
+preserve a newline; for `unfold`, will format every parameter group separately
+unless all fit on a single line, etc.).
+
+- `beforeOpenParenDefnSite` applies to definitions (methods, ctors, macros, etc.)
+- `beforeOpenParenCallSite` applies to invocations (method calls) and is only
+  supported for scala3
+
+Additional nuances:
+
+- if `newlines.sometimesBeforeColonInMethodReturnType` is true, a newline will
+  be added before the colon unless the entire signature fits on a line (except
+  when `newlines.source = keep`).
+- if the corresponding `align.openParenXxxSite` is true, multi-line parameters
+  will start on the same line as the opening parenthesis and align; otherwise,
+  formatting will use a newline and an appropriate continuation indent.
+
+```scala mdoc:defaults
+newlines.beforeOpenParenDefnSite
+newlines.beforeOpenParenCallSite
+```
+
+```scala mdoc:scalafmt
+maxColumn = 20
+runner.dialect = scala3 // for CallSite
+align.openParenCallSite = true
+newlines {
+  beforeOpenParenDefnSite = true
+  beforeOpenParenCallSite = true
+  sometimesBeforeColonInMethodReturnType = true
+}
+---
+def fooFunc(foo1: String)(foo2: String, foo3: String): String = ???
+val res = fooFunc("foo1")("foo2", "foo3")
+```
+
+```scala mdoc:scalafmt
+maxColumn = 20
+runner.dialect = scala3 // for CallSite
+align.openParenDefnSite = true
+newlines {
+  beforeOpenParenDefnSite = true
+  beforeOpenParenCallSite = true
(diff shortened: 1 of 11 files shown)


scalafmt-tests/src/test/resources/newlines/source_classic.stat
```diff
@@ -3791,3 +3791,314 @@ type T = A1 & A2 | A3 & A4 | A5 &
...
+    (foo2: String, foo3: String, foo4: Map[String, Map[String, Map[String, String]]]) // this line doesn't fit
+    (foo5: String)
+    (implicit foo6: Int,  foo7: Int,  foo8: Int,  foo9: Int,  foo10: Int,  foo11: Int,  foo12: Int) // this line also doesn't fit
+    : (String, String, String, String) = ???
+}
+>>>
+object a {
+  def fooFunc(foo1: String)
+    (foo2: String)
+    : String = ???
```
kitbellew, line 3822: @tgodzik @poslegm @djneades was debating whether to indent here... opinions?

djneades: @kitbellew I’m not sure I quite understand the question. May I ask what the alternatives would be?

kitbellew: two options:
```
def fooFunc(foo1: String)
  (foo2: String)
  : String = ???
```
or
```
def fooFunc(foo1: String)
  (foo2: String)
: String = ???
```

djneades: Ah, thank you! The former would definitely be my preference.

Note that, with the function body not fitting on the same line as the type signature there, I’d actually want to double-indent the argument blocks and return type, like this:

```scala
def fooFunc
    (foo1: String)
    (foo2: String)
    : String =
  ???
```

kitbellew: increase the `indent.defnSite`.

djneades: Thank you. I do have that in my current `.scalafmt.conf`; I just wanted to be sure it would still work with the new layout options.

docs/configuration.md
```diff
@@ -1704,6 +1704,82 @@ implicit protected val td: TildeArrow {
   type Out = RouteTestResult } = TildeArrow.injectIntoRoute
 ```
 
+### `newlines.sometimesBeforeColonInMethodReturnType`
```
tgodzik, line 1707:
```suggestion
### `newlines.colonInMethodReturnTypeOnOverflow`
```
?

`Sometimes` has weird undermined sound to it

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatTokens.scala
```diff
@@ -135,6 +135,8 @@ class FormatTokens(leftTok2tok: Map[TokenOps.TokenHash, Int])(
   def getLastNonTrivialOpt(tree: Tree): Option[FormatToken] =
     TokenOps.findLastNonTrivialTokenOpt(tree.tokens).map(apply)
 
+  @inline
+  def tokenAfter(token: Token): FormatToken = nextNonComment(before(token))
```
tgodzik, line 139: Does it really look for token after? `before(token)` is the one before no? It's a bit unclear what this function should actually do.

docs/configuration.md
tgodzik: Or is it not a new config?

tgodzik (approved):
Some minor nitpicks, but nothing blocking. Looks good!

docs/configuration.md
kitbellew: this is an old parameter but was never mentioned in the docs. since it's used in the new logic, added it.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatTokens.scala
kitbellew: they both find the token *BUT* if it had been rewritten and is no longer there, `before` returns the token just before it and `after` just after it.

`tokenAfter` returns the first format token such that `right` is to the right of `token` and is not a comment. there are several similar methods below, this one was missing before (while tokenBefore(token) was used already).

kitbellew: added a commit which documents these methods. ptal.

PR #2574, opened 2021-06-03
ExpandImportSelectors: fix check for rename `=>`

### LINKED ISSUE #2571: ExpandImportSelectors rewrite adding braces on imports with special characters
- **Integration**: sbt
- **Configuration**:
```
version = 2.7.5

rewrite.rules = [ExpandImportSelectors]
```

## Steps

Given code like this:
```scala
import zio.prelude.<=>
```

```bash
sbt scalafmt
```

## Problem

```scala
import zio.prelude.{<=>}
```

## Expectation

```scala
import zio.prelude.<=>
```

I looked for possible configuration options to change this behavior but came up empty. This conflicts with the way `scalafix-organize-imports` formats the code, it would be good to align the behavior with some config option. Scalafmt is very hand in that it works fast on-save, so I would not want to disable this feature ideally.

kitbellew:
Fixes #2571.


tgodzik (approved):
LGTM!

PR #2583, opened 2021-06-05
Indents: add a setting for ctrlSite (if/while/for)

### LINKED ISSUE #2573: Feature Request: Allow some additional settings for control sites
The following is our proposed new config for the Scala Native project. We are super excited because we were able to remove most of the configuration items and create a format more aligned with Scala.js and take advantage of all the new features like number formatting. We share code back and forth so having the two repositories similar really helps.

```properties
version = 3.0.0-RC3
docstrings.style = AsteriskSpace
project.git = true
project.excludePaths = [
  "glob:**/scalalib/**"
]
# this creates less of a diff but is not default
# but aligned with Scala.js syntax
newlines.beforeCurlyLambdaParams = multilineWithCaseOnly

# This is not default but avoid ctrl site like if from wrapping
#danglingParentheses.ctrlSite = false
```

## Background

We are attempting to upgrade from version `2.4.2` as shown in this PR: https://github.com/scala-native/scala-native/pull/2315

The first comment includes a link to the unified diff at the bottom. The one thing that stood out to the reviewer was the change in the format of `if` statements when the condition was too long as shown in the following:

```diff
@@ -791,3 +845,5 @@
     }
-    if (uri.scheme == null && scheme != null || uri.scheme != null && scheme == null) {
+    if (
+      uri.scheme == null && scheme != null || uri.scheme != null && scheme == null
+    ) {
       return false
```
We are very interested in maintaining a format that is similar to a Scala 3 optional braces format. Given that the following code has been around for all C, Java, and Scala for a long time, we would like to have something familiar that has the least impact on current code bases.

Using the code from above we would like to see something like the following if the condition is too long:
```scala
if (uri.scheme == null && scheme != null || 
    uri.scheme != null && scheme == null) {
  return false
}
```

If I add `danglingParentheses.ctrlSite = false` and run scalafmt I can get a format like the following:

```scala
if (host.indexOf(':') != -1 && host.indexOf(']') == -1 &&
  host.indexOf('[') == -1) {
  hostVar = "[" + host + "]"
}
```
The above makes it hard to distinguish the condition continuation from the body of the `if` or `while`.

## Proposal

Control sites with wrapped conditions indent 4 spaces by default.

Here are some examples:
```scala
if (host.indexOf(':') != -1 && host.indexOf(']') == -1 &&
    host.indexOf('[') == -1) {
  hostVar = "[" + host + "]"
}

if host.indexOf(':') != -1 && host.indexOf(']') == -1 &&
    host.indexOf('[') == -1 then
  hostVar = "[" + host + "]"

if host.indexOf(':') != -1 && host.indexOf(']') == -1 &&
    host.indexOf('[') == -1
then
  hostVar = "[" + host + "]"

```
I think this should be the default without the `danglingParentheses.ctrlSite = false` setting as it is the most like current code bases and what people are used to seeing.

Some other options that could be configured if needed?

```scala
if (host.indexOf(':') != -1 && host.indexOf(']') == -1 &&
    host.indexOf('[') == -1
) {
  hostVar = "[" + host + "]"
}

if (
  host.indexOf(':') != -1 && host.indexOf(']') == -1 &&
    host.indexOf('[') == -1
) {
  hostVar = "[" + host + "]"
}
// both of the above would work with 2 space indentation on the condition continuation as well.
```

kitbellew:
Fixes #2573.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1915,6 +1915,9 @@ object a {
 Forces dangling on open/close parens around control structures (`if`, `while`,
 `for`) when line breaks must occur.
 
+For optional braces in scala3, this parameter also controls whether to break
+before `then` or `do` in a multi-line condition.
+
 ```scala mdoc:defaults
 danglingParentheses.ctrlSite
 ```
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -235,6 +235,36 @@ argument1, // indented by 2
 )
 ```
 
+### `indent.ctrlSite`
+
+This parameter controls indentation within control expressions (`if/while/etc`).
+If not set, the value of `indent.callSite` applies.
+
+> Since v3.0.0.
+
+```scala mdoc:scalafmt
+indent.callSite = 2
+indent.ctrlSite = 4
+---
+if (
+ foo && // indented by 4
+ bar
+) {
+  baz
+}
+```
+
+```scala mdoc:scalafmt
+indent.callSite = 2
+---
+if (
+ foo && // indented by 2
+ bar
+) {
+  baz
+}
+```
+
 ### `indent.defnSite`
 
 ```scala mdoc:defaults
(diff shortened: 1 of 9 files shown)


tgodzik (approved):
LGTM

ekrich:
Wow, that is really comprehensive. Thanks so much for all the work.

PR #2585, opened 2021-06-05
Newlines: beforeOpenParenXxx take fold/unfold/keep

### LINKED ISSUE #2561: Feature request: neat Scala 3 definitions and calls – keep parens on the same line as arguments

kitbellew:
Fixes #2561.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1810,12 +1810,19 @@ def someVeryLongMethodName: Map[String, String] = ???
 
 ### `newlines.beforeOpenParenXxxSite`
 
-If true, a parameter group which doesn't fit on the same line will be formatted
-separately, with a newline just before the opening parenthesis.
+This parameter, if enabled, causes a parameter group which doesn't fit on the
+same line to be formatted separately, with a newline just before the opening
+parenthesis.
 
-Specific formatting depends on the value of `newlines.source` (for `keep`, will
-preserve a newline; for `unfold`, will format every parameter group separately
-unless all fit on a single line, etc.).
+> Since v3.0.0.
+
+Use `null` (default) to disable it. Otherwise, takes the same values as
+[newlines.source](#newlinessource) (or explicit `source` to mirror the current
+value of `newlines.source`).
+
+Specific formatting depends on the value: for `keep`, will preserve a newline;
+for `unfold`, will format every parameter group separately unless all fit on a
+single line; `fold` will use a more compact formatting.
 
 - `beforeOpenParenDefnSite` applies to definitions (methods, ctors, macros, etc.)
 - `beforeOpenParenCallSite` applies to invocations (method calls) and is only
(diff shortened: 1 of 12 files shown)


scalafmt-tests/src/test/resources/newlines/source_keep.stat
```diff
@@ -3984,7 +3984,7 @@ object a {
 align.preset = none
 align.openParenDefnSite = true
 runner.dialect = scala3
-newlines.beforeOpenParenDefnSite = true
+newlines.beforeOpenParenDefnSite = source
```
tgodzik, line 3987: If the whole test is about keep, why do we set some of it to `source` and some to `keep`? Could we keep it all `source` or all `keep`? I think this brings in some unintended confusion.

docs/configuration.md
```diff
@@ -1810,12 +1810,19 @@ def someVeryLongMethodName: Map[String, String] = ???
...
+This parameter, if enabled, causes a parameter group which doesn't fit on the
+same line to be formatted separately, with a newline just before the opening
+parenthesis.
 
-Specific formatting depends on the value of `newlines.source` (for `keep`, will
-preserve a newline; for `unfold`, will format every parameter group separately
-unless all fit on a single line, etc.).
+> Since v3.0.0.
+
+Use `null` (default) to disable it. Otherwise, takes the same values as
```
tgodzik, line 1819:
```suggestion
Use `classic` (default) to disable it. Otherwise, takes the same values as
```
to be in line with the other names?

tgodzik (changes requested):
Looks good, but I left some minor comments.

scalafmt-tests/src/test/resources/newlines/source_keep.stat
kitbellew: i wanted to test that both `source` and `keep` work. just minor variation in test cases, to prevent regression.

docs/configuration.md
kitbellew: `classic` cannot be specified explicitly (for `newlines.source` or any other which uses the same type). i specifically made it hidden so that the only way to get the original behaviour is by _not_ specifying anything.

even if `classic` were available, it would mean _enable with classic behaviour`.

the only reason to mention it here is if people want to specify `unfold` but then disable for a subset of files via `fileOverride`.

tgodzik (approved):
LGTM!

PR #2590, opened 2021-06-06
Router: use extra indent with beforeOpenParenDefnSite

### LINKED ISSUE #2561: Feature request: neat Scala 3 definitions and calls – keep parens on the same line as arguments

kitbellew:
Fixes #2561.

### DIFF
--- a/scalafmt-tests/src/test/resources/newlines/source_unfold.stat
+++ b/scalafmt-tests/src/test/resources/newlines/source_unfold.stat
@@ -4203,11 +4203,11 @@ class fooClass
     (foo1: String)
     (foo2: String): String = ???
   def fooFunc
-    (foo1: String)
-    (foo2: String)
-    (foo3: String)
-    (foo4: String)
-    (foo5: String): (
+     (foo1: String)
+     (foo2: String)
+     (foo3: String)
+     (foo4: String)
+     (foo5: String): (
       String,
       String,
       String,
--- a/scalafmt-tests/src/test/resources/newlines/source_unfold.stat
+++ b/scalafmt-tests/src/test/resources/newlines/source_unfold.stat
@@ -3935,6 +3935,7 @@ type T =
 <<< #2561 scala 2: defn, no align
 align.preset = none
 newlines.beforeOpenParenDefnSite = source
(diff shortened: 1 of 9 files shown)


scalafmt-tests/src/test/resources/newlines/source_unfold.stat
```diff
@@ -4350,7 +4437,7 @@ enum Namespace(val uri: String | Null):
 package dev.muve.vdom
 
 enum Namespace
-  (val uri: String | Null):
+    (val uri: String | Null):
```
tgodzik, line 4440:
```suggestion
  (val uri: String | Null):
```
we already have an indentation of two here and we don't set the extra indentation. Why did this change?

tgodzik: So it's `indent.significant + extra`? I would say 0 is a better default. It might be hard for people to figure out where they are getting the additional 2

kitbellew: will fix shortly

tgodzik: thanks!

kitbellew: done, ptal

tgodzik (approved):
LGTM!

PR #2594, opened 2021-06-10
[Scala 3] Add support for indented block after if keyword

tgodzik:
If conditions can now be included within an indented block after the if keyword.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -2037,6 +2037,7 @@ class FormatOps(
           case _: T.KwFinally => FinallyImpl
           case _: T.KwMatch => MatchImpl
           case _: T.KwThen => ThenImpl
+          case _: T.KwIf => IfImpl
           case _: T.KwElse => ElseImpl
           case _: T.KwReturn | _: T.ContextArrow | _: T.LeftArrow |
               _: T.KwThrow | _: T.KwWhile | _: T.KwYield =>
@@ -2361,6 +2362,28 @@ class FormatOps(
             })
           case _ => None
         }
+    }
+
+    private object IfImpl extends Factory {
+      def create(ft: FormatToken, nft: FormatToken)(implicit
+          style: ScalafmtConfig
+      ): Option[OptionalBracesRegion] = {
+        ft.meta.leftOwner match {
+          case t: Term.If =>
+            t.cond match {
+              case b: Term.Block =>
+                Some(
+                  new OptionalBracesRegion {
+                    def owner = Some(t)
+                    def splits = Some(getSplits(ft, b, true))
+                    def rightBrace = blockLast(b)
+                  }
+                )
+              case _ => None
+            }
+          case _ => None
+        }
+      }
     }
 
     private object ElseImpl extends Factory {
--- a/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
+++ b/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
@@ -1557,3 +1557,79 @@ val (length, width) =
      (hourMarkLength, hourMarkWidth)
   else
      (secondMarkLength, secondMarkWidth)
+<<< if indentation
+if // comm
+    val cond1 = true
+    (a1 && a2) == cond1
+  then
+    val exitCode = 1
+    exitCode
+    
+>>>
+if // comm
+   val cond1 = true
+   (a1 && a2) == cond1
+then
+   val exitCode = 1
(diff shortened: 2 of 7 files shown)


scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -1557,3 +1557,50 @@ val (length, width) =
      (hourMarkLength, hourMarkWidth)
   else
      (secondMarkLength, secondMarkWidth)
+<<< if indentation
+if // comm
+    val cond1 = true
+    (a1 && a2) == cond1
+  then
```
kitbellew, line 1564: here and in other places: can you add some tests with `danglingParentheses.ctrlSite = false`? when this setting is `true`, formatter forces a break before `then`; otherwise, it doesn't, and I would like to make sure scalameta can still parse it. otherwise, we might need to change the logic around `danglingParentheses.ctrlSite = false` when `if` condition was a block.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2363,6 +2364,27 @@ class FormatOps(
...
+            (t.cond match {
+              case b: Term.Block => Some(true)
+              case _ => None
+            }).map { forceNL =>
+              new OptionalBracesRegion {
+                def owner = Some(t)
+                def splits = Some(getSplits(ft, t.cond, forceNL))
+                def rightBrace = blockLast(t.cond)
+              }
+            }
```
kitbellew, line 2382:
```suggestion
            t.cond match {
              case b: Term.Block =>
                new OptionalBracesRegion {
                  def owner = Some(t)
                  def splits = Some(getSplits(ft, b, true))
                  def rightBrace = blockLast(b)
                }
              case _ => None
            }
```

kitbellew (changes requested):

kitbellew:
by the way, don't remember: are `while` conditions already handled similarly?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
kitbellew: do we need to check that the block contains more than one statement?

tgodzik: We don't treat it as block in the parser if it's a single indented statement. That's in line how Scala 3 parser works.

tgodzik:
> by the way, don't remember: are `while` conditions already handled similarly?

While seems to be handled using `BlockImpl` already.

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
tgodzik: This works with a block, but seems to break with a single statement if we have a comment:
```
danglingParentheses.ctrlSite = false
===
if // comm
    (a1 && a2) == cond1
  then
    val exitCode = 1
    exitCode
```

gets formatted to 
```scala
if // comm
   (a1 && a2) == cond1 then
```

but that is correct compiler wise, just doesn't work in the parser yet. I will add the tests.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Changed, but added `Some()` around.

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
tgodzik: Added tests for `danglingParentheses`, let me know if anything is missing.

kitbellew: so does the danglingParens setting affect formatting? the idea was that `false` should lead to no break before `then`, if compiler allows, of course.

tgodzik: `danglingParens.ctrlSite` seem to only affect the formatting if the condition starts with `(` there is no special handling for optional braces.

We could do it later one after https://github.com/scalameta/scalameta/pull/2384 is released.

tgodzik: What do you think @kitbellew ? Merge this and work on `danglingParens.ctrlSite` later or wait for the release and try to handle it all here?

kitbellew: let's merge and then follow up.

kitbellew (approved):

PR #2598, opened 2021-06-22
Update Scalameta to 4.4.22

tgodzik:


scalafmt-tests/src/main/scala/org/scalafmt/util/ScalaFile.scala
```diff
@@ -76,9 +75,10 @@ object ScalaFile {
   }
 
   private def downloadReposTar(destination: File): Unit = {
+    import sys.process._
     val fileToDownload = new URL(reposTarballUrl)
     println(s"Downloading $reposTarballUrl...")
-    FileUtils.copyURLToFile(fileToDownload, destination)
```
tgodzik, line 81: I removed apache commons from scalameta, which were brought in transitively from the testkit. Replaced it with the same workaround from `sys.process`.

kitbellew (approved):

PR #2601, opened 2021-06-23
Router: minor optimization for apply

kitbellew:
SingleLineBlock leads to fewer paths to explore so use it instead of PenalizeNewlines when in reality the latter should lead to equivalent formatting.

In this case, a 10000 penalty on any newlines without any exclude ranges with an available lower-cost alternative will in most cases be the same as using single-line format.

Exceptions include cases with multiline constants as arguments, and some rare cases of previously suboptimal formatting, due to complexity of preceding code.


tgodzik (approved):
LGTM

PR #2602, opened 2021-06-24
[Scala 3] Allow `then` and `if` at the end of indentation under danglingParentheses = false

tgodzik:
This is a follow up from https://github.com/scalameta/scalafmt/pull/2594#pullrequestreview-685371567

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -2052,7 +2052,8 @@ class FormatOps(
         ft: FormatToken,
         tree: Tree,
         forceNL: Boolean = false,
-        useMain: Boolean = false
+        useMain: Boolean = false,
+        danglingKeyword: Boolean = true
     )(implicit fileLine: FileLine, style: ScalafmtConfig): Seq[Split] = {
       val end = tokens.getLast(tree)
       val slbExpire = nextNonCommentSameLine(end).left
@@ -2061,7 +2062,9 @@ class FormatOps(
         else
           getClosingIfEnclosedInMatching(tree).map(tokens.tokenBefore(_).left)
       def nlPolicy(implicit fileLine: FileLine) =
-        decideNewlinesOnlyAfterClose(closeOpt.getOrElse(slbExpire))
+        if (danglingKeyword)
+          decideNewlinesOnlyAfterClose(closeOpt.getOrElse(slbExpire))
+        else NoPolicy
       val indentLen =
         if (useMain) style.indent.main else style.indent.getSignificant
       val indent =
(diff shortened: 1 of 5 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2052,16 +2052,20 @@ class FormatOps(
...
     )(implicit fileLine: FileLine, style: ScalafmtConfig): Seq[Split] = {
       val end = tokens.getLast(tree)
       val slbExpire = nextNonCommentSameLine(end).left
       val closeOpt =
         if (isTuple(tree)) None
         else
           getClosingIfEnclosedInMatching(tree).map(tokens.tokenBefore(_).left)
-      def nlPolicy(implicit fileLine: FileLine) =
-        decideNewlinesOnlyAfterClose(closeOpt.getOrElse(slbExpire))
+      def nlPolicy(implicit fileLine: FileLine) = {
```
kitbellew, line 2064: braces are not needed

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2113,7 +2117,16 @@ class FormatOps(
               if !hasSingleTermStat(t) && isBlockStart(t, nft) =>
             Some(new OptionalBracesRegion {
               def owner = t.parent
-              def splits = Some(getSplitsMaybeBlock(ft, nft, t, true))
+              def splits = Some(
+                getSplitsMaybeBlock(
+                  ft,
+                  nft,
+                  t,
+                  true,
```
kitbellew, line 2125: since you are already unfolding the parameters, could you explicitly name this boolean, too,

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2113,7 +2117,16 @@ class FormatOps(
...
+                  danglingKeyword = style.danglingParentheses.ctrlSite &&
+                    ft.left.is[Token.KwWhile]
```
kitbellew, line 2127: should `while` have its own implementation class, like `if` and `do` do?

kitbellew (changes requested):
thank you for remembering this

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: I would say no, since it seems to work fine currently and this is the only difference between normal block and a while. I would hold off on splitting it until more differences pop up.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Named, and also named the other one.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
tgodzik: Removed

kitbellew (approved):

PR #2619, opened 2021-06-30
Fixing ScalaDoc heading output.

pgarrity-rally:
Prior to this fix, ScalaDoc headings were rewritten to numeric values if `docstrings.wrap` was set to `yes`. The specific version I was using locally was `3.0.0-RC5`. Examples of what I was seeing:

- `==Foo==` became `122Foo122`
- `===Foo===` became `183Foo183`

The delimiter calculation was multiplying a char by a number, which resulted in a numeric value.

Haven't contributed here before - sorry if I've missed anything! I added a test which showed the old behavior and verified the fix, ran all of the tests afterwards, ran scalafmt on my changeset.

### DIFF
--- a/scalafmt-tests/src/test/scala/org/scalafmt/CommentTest.scala
+++ b/scalafmt-tests/src/test/scala/org/scalafmt/CommentTest.scala
@@ -12,9 +12,11 @@ class CommentTest extends FunSuite {
     )
 
   private val wrapStyle: ScalafmtConfig =
-    ScalafmtConfig.default.copy(docstrings =
-      ScalafmtConfig.default.docstrings
-        .copy(wrap = Docstrings.Wrap.yes)
+    ScalafmtConfig.default.copy(
+      maxColumn = 20,
+      docstrings = ScalafmtConfig.default.docstrings.copy(
+        wrap = Docstrings.Wrap.yes
+      )
     )
 
   test("remove trailing space in comments") {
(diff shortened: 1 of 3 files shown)


scalafmt-tests/src/test/scala/org/scalafmt/CommentTest.scala
```diff
@@ -124,4 +130,19 @@ class CommentTest extends FunSuite {
...
+    val original = """
+      |/** =Heading1=
+      |  *
+      |  * ==Heading2==
+      |  *
+      |  * ===Heading3===
+      |  */
+      |""".stripMargin
+
+    val expected = original
```
kitbellew, line 144: perhaps i'd request original to be different so that it's clear the formatter actually does something. that's why your `expected = original` looks out of place. :)

kitbellew (changes requested):
thanks for the change, omission on our part!

scalafmt-tests/src/test/scala/org/scalafmt/CommentTest.scala
pgarrity-rally: That's a fair point, but I'm wondering if you have any ideas that would be relevant to this change? That's what I got stuck on when drafting the test.

So... since this depends on wrap being enabled, I should do things that'll get impacted by wrap and show that the headings are preserved even under those conditions.

Thanks, I'll hopefully have this PR updated tomorrow.

pgarrity-rally: Actually I was thinking about it, so just took a swag at this tonight. Happy to adjust it more (or draft some entirely different test altogether) if this update is insufficient.

kitbellew: patrick, i am sorry, i don't know why i blanked out. this file `CommentTest` is specially crafted for comments and trailing space, something that cannot be done otherwise.

can you please put your test in `scalafmt-tests/src/test/resources/test/JavaDoc.stat` instead? this is where all the "normal" tests reside.

pgarrity-rally: Can-do -- pushed! That section of the documentation kinda confused me so thanks for the pointer to the correct file to place the test.

kitbellew (approved):
many thanks!

pgarrity-rally:
Thanks for reviewing and pointing me in the right direction for tests

PR #2627, opened 2021-07-01
CliOptions: store paths as File, resolve on demand

### LINKED ISSUE #2616: Scalafmt GraalVM prebuilt images RC2÷RC5 not working
- **Version**: scalafmt 3.0.0-RC2÷RC5
- **Integration**: Bash Cli (Ubuntu)

Given code like this:
```bash
> scalafmt-native del.scala
```
i'm getting errors like:
```
Exception in thread "main" java.nio.file.NoSuchFileException: /home/runner/.../del.scala
```
None, of the downloaded `scalafmt-native 3.0.0-RC2÷RC5` native images work.

The last working image though is `scalafmt-native 3.0.0-RC1`

kitbellew:
This might also help with obscure file resolution issues around our native builds. Fixes #2616.

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
@@ -4,7 +4,6 @@ import java.io.File
 import java.util.Date
 
 import org.scalafmt.Versions
-import org.scalafmt.util.AbsoluteFile
 import scopt.OptionParser
 
 object CliArgParser {
@@ -41,15 +40,6 @@ object CliArgParser {
 
       private def readConfig(contents: String, c: CliOptions): CliOptions = {
         c.copy(configStr = Some(contents))
-      }
-
-      private def readConfigFromFile(
-          file: String,
-          c: CliOptions
-      ): CliOptions = {
-        val configFile =
-          AbsoluteFile.fromFile(new File(file), c.common.workingDirectory)
-        c.copy(config = Some(configFile.jfile.toPath))
       }
 
       head("scalafmt", Versions.nightly)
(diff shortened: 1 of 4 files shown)


scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
```diff
@@ -99,8 +89,8 @@ object CliArgParser {
         .text(
           "use project filters even when specific files to format are provided"
         )
-      opt[String]('c', "config")
-        .action(readConfigFromFile)
+      opt[File]('c', "config")
```
olafurpg, line 92: Do you know how this change affects parsing of relative paths? Does it absolutize eagerly? Ideally, relative filenames should be kept relative so that we can resolve them from the working directory that nailgun provides.

olafurpg (commented):
The graalvm issue should be fixable by avoiding initialization of the working directory at build time. I don’t have the native image flag name at the top of my head but it’s possible to force certain classes to initialize at runtime.

kitbellew:
> The graalvm issue should be fixable by avoiding initialization of the working directory at build time. I don’t have the native image flag name at the top of my head but it’s possible to force certain classes to initialize at runtime.

i don't really understand how this works, but i experimented locally, and if we access the working directory in the cli parser, it points to the build directory, but if it is read later, when the formatter starts (like in this commit), it uses the right one.

scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
kitbellew: `File` keeps them relative. `CommonOptions.cwd` will contain the nailgun directory, and lazy `CommonOptions.workingDirectory` is initialized from `cwd` or else from `user.dir`.

olafurpg (approved):
I took a stab at tweaking the native-image generation in  https://github.com/scalameta/scalafmt/pull/2629  and it seems to fix https://github.com/scalameta/scalafmt/issues/2616

I think both PRs are complementary so we could merge them both.

PR #2629, opened 2021-07-02
Use sbt-native-image plugin to build native image.

olafurpg:
Previously, the build used sbt-assembly to generate the native-image
binary. This made it necessary to start sbt with a pre-installed GraalVM
version. With sbt-native-image, the correct GraalVM version is installed
automatically.

Also, this commit tweaks the GraalVM native-image linking options to
force the `AbsoluteFile` class to initialize a runtime. This should fix
a bug when passing in relative paths to native-image binaries.

Alternative to #2627


kitbellew (approved):
@olafurpg so what is your hypothesis regarding what had been happening? was the user.val property getting read at compile time? i need to read how initializing at build time works, and why it affected the cli opt parser class...

olafurpg:
My hypothesis is that native-image evaluated `AbsoluteFile.userDir` at build time instead of runtime

https://github.com/scalameta/scalafmt/blob/029d9228f4ddeeba29204c12eb05c57d37699847/scalafmt-core/shared/src/main/scala/org/scalafmt/util/AbsoluteFile.scala#L30

The native-image optimizer can evaluate code during link time and hardcode the computed result. This behavior is configurable through the `--initialize-at-build-time` and `--initialize-at-run-time` flags. Scalafmt initializes everything by default at build time, as configured here

https://github.com/scalameta/scalafmt/blob/029d9228f4ddeeba29204c12eb05c57d37699847/scalafmt-cli/src/main/resources/META-INF/native-image/org.scalafmt/scalafmt-cli/native-image.properties#L8

PR #2632, opened 2021-07-13
Update Scalameta to 4.4.24

tgodzik:
I think most of the bigger issues were fixed, there still some things leftover, but nothing that should influence Scalafmt.

I think we should be good to do `3.0.0`


tgodzik:
Damn, it seems I broke something with the positions.

tgodzik:
The change actually seems valid, for `while (a > 0) do `  `(a > 0)` belongs to `Term.ApplyInfix`, which makes sense. Should we make `RedundantParens` work here?

PR #2635, opened 2021-07-17
FormatWriter: fix `given` end marker label

### LINKED ISSUE #2634: [Scala 3] anonymous given instances get end-markers that don't compile
- **Version**: 3.0.0-RC6
- **Integration**: Metals (VS Code) and CLI
- **Configuration**:
```conf
version = "3.0.0-RC6"
runner.dialect = Scala3
rewrite.scala3.insertEndMarkerMinLines = 2
```

## Steps

Given code like this:
```scala
given Foo[String] with
  def foo(s: String) =
    true
```

## Problem

```scala
given Foo[String] with
  def foo(s: String) =
    true
end
```

And when running scalafmt again:

```scala
given Foo[String] with
  def foo(s: String) =
    true
end 
end
```

And so on.

Even with 1 `end` marker, the code doesn't compile. 

## Expectation

```scala
given Foo[String] with
  def foo(s: String) =
    true
end given
```

## Workaround

Giving the given instance a name will add the name to the end marker, which compiles. Manually adding `end given` as the marker also works

kitbellew:
Fixes #2634.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -1583,7 +1583,9 @@ object FormatWriter {
     case t: Defn.Object => t.name.value
     case t: Defn.Trait => t.name.value
     case t: Defn.Enum => t.name.value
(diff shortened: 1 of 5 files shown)


tgodzik (approved):
LGTM!

PR #2636, opened 2021-07-17
Scalafmt: parse configuration directly from file

### LINKED ISSUE #2631: Wrong error message with invalid config
- **Version**: 3.0.0-RC6
- **Integration**: Metals
- **Configuration**:
```
version = 3.0.0-RC6
align=does-not-exist
```

## Steps

When run currently:
```bash
Invalid config: <input>:3:0 error: Type mismatch;
  found    : String (value: "does-not-exist")
  expected : Object
    "align" : "does-not-exist",
^
: /home/tgodzik/Documents/metals/.scalafmt.conf
```

but previously it would be:

```bash
Invalid config: error: Type mismatch;
  found    : String (value: "does-not-exist")
  expected : Object
    "align" : "does-not-exist",
^
: /home/tgodzik/Documents/metals/.scalafmt.conf
```

## Problem

Not sure where `<input>:3:0` comes from and can't reproduce it in tests.

## Expectation

`<input>:3:0` would not be in the error message

## Workaround

Ignore the changed message

## Notes

Might be caused by the recent changes to metaconfig?

This popped up in Metals tests: https://github.com/scalameta/metals/pull/2914

This is a very minor thing, so it's not really blocking `3.0.0`, but might be good to fix and I can figure out where it's happening.

kitbellew:
Previously, we'd parse the configuration file into a `typesafe.config` object, then convert it to a HOCON string and invoke the formatter config parser on it. If there's a parsing error with that string, the formatter wouldn't know where it came from and refer to `<input>` as the source of the configuration.

Instead, pass the path to the config file directly to the new method allowing parsing configuration from a file. Fixes #2631.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
@@ -1,5 +1,7 @@
 package org.scalafmt
 
+import java.nio.file.Path
+
 import metaconfig.Configured
 import scala.annotation.tailrec
 import scala.meta.Dialect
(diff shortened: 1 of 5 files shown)


tgodzik (approved):
Thanks for looking into this! LGTM

tgodzik:
@kitbellew should we release the final 3.0.0 and announce it on twitter? I think the biggest scalameta issues are fixed and if anything new pops up (it most certainly will) we can release bugfix versions later on. What do you think?

kitbellew:
@tgodzik i have a pair of things in the works, give me a few days (and your review) 😀

tgodzik:
Sure! Will try to review everything as soon as possible!

Issue #2633, opened 2021-07-13
Potential bug running against Scala.js


ekrich:
When trying to run `scalafmt` on the Scala.js code base as seen here:
https://github.com/scala-js/scala-js/pull/4522

Config: https://github.com/scala-js/scala-js/pull/4522/commits/f02a0886a203f338705b1131bf59da4a6567d417

## Steps

Given code like this:
https://github.com/ekrich/scala-js/tree/topic/scalafmt

When I run scalafmt like this:

```bash
scripts/scalafmt --test 2> diff.txt
```

## Problem

Scalafmt formats code like this:
```scala
org.scalafmt.cli.FailedToFormat: /Users/eric/workspace/scala-js/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/NameGen.scala
Caused by: org.scalafmt.Error$PreciseIncomplete: /Users/eric/workspace/scala-js/linker/shared/src/main/scala/org/scalajs/linker/backend/emitter/NameGen.scala:72: error: Unable to format file due to bug in scalafmt
    cache.getOrElseUpdate(name, {
                              ^
	at org.scalafmt.Scalafmt$.$anonfun$doFormatOne$6(Scalafmt.scala:133)
	at scala.util.Success.flatMap(Try.scala:258)
	at org.scalafmt.Scalafmt$.$anonfun$doFormatOne$3(Scalafmt.scala:128)
	at scala.meta.parsers.Parsed.fold(Errors.scala:12)
	at scala.meta.parsers.Parsed.fold$(Errors.scala:11)
	at scala.meta.parsers.Parsed$Success.fold(Errors.scala:26)
	at org.scalafmt.Scalafmt$.doFormatOne(Scalafmt.scala:124)
	at org.scalafmt.Scalafmt$.doFormat(Scalafmt.scala:102)
	at org.scalafmt.Scalafmt$.formatCode(Scalafmt.scala:71)
	at org.scalafmt.cli.ScalafmtCoreRunner$.unsafeHandleFile(ScalafmtCoreRunner.scala:78)
	at org.scalafmt.cli.ScalafmtCoreRunner$.handleFile(ScalafmtCoreRunner.scala:60)
	at org.scalafmt.cli.ScalafmtCoreRunner$.$anonfun$run$5(ScalafmtCoreRunner.scala:40)
	at org.scalafmt.cli.ScalafmtCoreRunner$.$anonfun$run$5$adapted(ScalafmtCoreRunner.scala:39)
[...]

PR #2642, opened 2021-07-24
Router: apply `, {` rule to binPack case as well

kitbellew:
Helps with #2633.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -780,10 +780,14 @@ class Router(formatOps: FormatOps) {
         val noNL = style.newlines.sourceIgnored || formatToken.noBreak
         Seq(Split(NoSplit.orNL(noNL), 0))
 
-      case tok @ FormatToken(open @ LeftParenOrBracket(), right, _)
-          if !isSuperfluousParenthesis(formatToken.left, leftOwner) &&
-            (!style.binPack.unsafeCallSite && isCallSite(leftOwner)) ||
-            (!style.binPack.unsafeDefnSite && isDefnSite(leftOwner)) =>
+      case tok @ FormatToken(open @ LeftParenOrBracket(), right, _) if {
+            if (isCallSite(leftOwner))
+              !style.binPack.unsafeCallSite &&
+              !isSuperfluousParenthesis(formatToken.left, leftOwner)
+            else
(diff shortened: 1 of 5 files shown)


scalafmt-tests/src/test/resources/default/Apply.stat
```diff
@@ -1,4 +1,4 @@
-
+#ONLY #
```
tgodzik, line 1:
```suggestion
```
?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1214,30 +1218,40 @@ class Router(formatOps: FormatOps) {
         )
       // non-statement starting curly brace
       case FormatToken(_: T.Comma, open: T.LeftBrace, _)
-          if !style.poorMansTrailingCommasInConfigStyle && {
-            if (isCallSite(leftOwner)) !style.binPack.unsafeCallSite
-            else isDefnSite(leftOwner) && !style.binPack.unsafeDefnSite
-          } =>
+          if !style.poorMansTrailingCommasInConfigStyle &&
```
tgodzik, line 1221: So `isDefnSite` is no longer handled in this case and should be handled above in 783, right?

tgodzik (changes requested):
Looks good! Just one small comment and a question from me.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: actually, the thought was that defn site looks impossible for this rule. comma is followed by a parameter name which can't start with a brace... can it?

scalafmt-tests/src/test/resources/default/Apply.stat
kitbellew: ah, thank you, debugging leftovers..

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Ach yeah! Makes sense, I can't find any scenario that this would fit in.

scalafmt-tests/src/test/resources/default/Apply.stat
kitbellew: done

tgodzik (approved):
LGTM!

PR #2653, opened 2021-07-28
Support Mdoc fence formatting

swoogles:
This came out of a discussion with @olafurpg  here: https://github.com/scalameta/mdoc/issues/537

There are still a few open questions near the end of the thread, but I think he is busy, so I'm submitting to the main repo in case other maintainers can answer.

1. Is it acceptable to add a new top-level `project.includeMarkdown` config to toggle this behavior? Or should it be incorporated into existing configs somehow?
2. His first draft only targets `scala mdoc` fences, but it seems to be purely a scalafmt behavior. Could we widen it to target to all `scala` snippets in the affected markdown files?

I have run this against a Scala book that I'm working on and it performed perfectly.
Would love to see it make it into the mainline :)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -3184,6 +3184,13 @@ project {
 }
 ```
 
+### `project.includeMarkdown`
+```scala mdoc:defaults
+project.includeMarkdown = true
+```
+If true, format triple-backticked `scala mdoc` snippets in Markdown files in the project.
+This will not affect any of the markdown text outside the snippets.
+
 ### `fileOverride`
 
 > Since v2.5.0.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -3184,13 +3184,6 @@ project {
 }
 ```
 
-### `project.includeMarkdown`
-```scala mdoc:defaults
-project.includeMarkdown = true
-```
-If true, format triple-backticked `scala mdoc` snippets in Markdown files in the project.
-This will not affect any of the markdown text outside the snippets.
-
 ### `fileOverride`
 
 > Since v2.5.0.
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
@@ -113,19 +113,23 @@ object Scalafmt {
         flatMapAll(chunks.iterator)(doFormatOne(_, style, file))
           .map(_.mkString("\n@\n"))
     } else if (FileOps.isMarkdown(file)) {
-      val markdown = MarkdownFile.parse(Input.VirtualFile(file, code))
-      flatMapAll(markdown.parts.iterator) {
-        case text: Text => Success(text.value)
-        case fence: CodeFence =>
-          if (fence.info.value.startsWith("scala mdoc")) {
-            doFormatOne(fence.body.value, style, file).map { formatted =>
-              fence.newBody = Some(formatted.trim)
+      if (code.contains("```scala mdoc")) {
+        val markdown = MarkdownFile.parse(Input.VirtualFile(file, code))
+        flatMapAll(markdown.parts.iterator) {
+          case text: Text => Success(text.value)
+          case fence: CodeFence =>
+            if (fence.info.value.startsWith("scala mdoc")) {
+              doFormatOne(fence.body.value, style, file).map { formatted =>
+                fence.newBody = Some(formatted.trim)
+              }
+            } else {
+              Success("")
             }
-          } else {
-            Success("")
-          }
-      }.map { _ =>
-        markdown.renderToString
+        }.map { _ =>
+          markdown.renderToString
+        }
+      } else {
+        Success(code)
       }
     } else {
       doFormatOne(code, style, file, range)
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
@@ -1,8 +1,8 @@
 package org.scalafmt
 
 import java.nio.file.Path
-import metaconfig.Configured
 
+import metaconfig.Configured
 import scala.annotation.tailrec
 import scala.meta.Dialect
 import scala.meta.Input
@@ -10,6 +10,7 @@ import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
 import scala.util.matching.Regex
+
 import org.scalafmt.config.Config
 import org.scalafmt.Error.PreciseIncomplete
 import org.scalafmt.config.FormatEvent.CreateFormatOps
@@ -112,22 +113,21 @@ object Scalafmt {
       val markdown = MarkdownFile.parse(Input.VirtualFile(file, code))
 
       val resultIterator: Iterator[Try[String]] =
-        markdown.parts.iterator.collect {
-          case fence: MarkdownPart.CodeFence
-              if fence.info.value.startsWith("scala mdoc") =>
-            val res = doFormatOne(fence.body.value, style, file)
-            res.foreach { formatted =>
-              fence.newBody = Some(formatted.trim)
-            }
-            res
-        }
+        markdown.parts.iterator
+          .collect {
+            case fence: MarkdownPart.CodeFence
+                if fence.info.startsWith("scala mdoc") =>
+              val res = doFormatOne(fence.body, style, file)
+              res.foreach { formatted =>
+                fence.newBody = Some(formatted.trim)
+              }
+              res
+          }
       if (resultIterator.isEmpty) Success(code)
       else {
-        val fRes: Option[Try[String]] =
-          resultIterator
-            .collectFirst { case f: Failure[_] => f }
-        fRes
-          .getOrElse(Success[String](markdown.renderToString))
+        resultIterator
+          .find(_.isFailure)
+          .getOrElse(Success(markdown.renderToString))
       }
     } else {
       doFormatOne(code, style, file, range)
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
@@ -1,8 +1,8 @@
 package org.scalafmt
 
 import java.nio.file.Path
-
 import metaconfig.Configured
+
 import scala.annotation.tailrec
 import scala.meta.Dialect
 import scala.meta.Input
@@ -10,7 +10,6 @@ import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
 import scala.util.matching.Regex
-
 import org.scalafmt.config.Config
 import org.scalafmt.Error.PreciseIncomplete
 import org.scalafmt.config.FormatEvent.CreateFormatOps
@@ -20,10 +19,7 @@ import org.scalafmt.internal.BestFirstSearch
 import org.scalafmt.internal.FormatOps
 import org.scalafmt.internal.FormatWriter
 import org.scalafmt.rewrite.Rewrite
-import org.scalafmt.util.FileOps
-import org.scalafmt.util.CodeFence
-import org.scalafmt.util.Text
-import org.scalafmt.util.MarkdownFile
+import org.scalafmt.util.{FileOps, MarkdownFile, MarkdownPart}
 
 /** WARNING. This API is discouraged when integrating with Scalafmt from a build tool
   * or editor plugin. It is recommended to use the `scalafmt-dynamic` module instead.
@@ -113,23 +109,25 @@ object Scalafmt {
         flatMapAll(chunks.iterator)(doFormatOne(_, style, file))
           .map(_.mkString("\n@\n"))
     } else if (FileOps.isMarkdown(file)) {
-      if (code.contains("```scala mdoc")) {
-        val markdown = MarkdownFile.parse(Input.VirtualFile(file, code))
-        flatMapAll(markdown.parts.iterator) {
-          case text: Text => Success(text.value)
-          case fence: CodeFence =>
-            if (fence.info.value.startsWith("scala mdoc")) {
-              doFormatOne(fence.body.value, style, file).map { formatted =>
-                fence.newBody = Some(formatted.trim)
-              }
-            } else {
-              Success("")
+      val markdown = MarkdownFile.parse(Input.VirtualFile(file, code))
+
+      val resultIterator: Iterator[Try[String]] =
+        markdown.parts.iterator.collect {
+          case fence: MarkdownPart.CodeFence
+              if fence.info.value.startsWith("scala mdoc") =>
+            val res = doFormatOne(fence.body.value, style, file)
+            res.foreach { formatted =>
+              fence.newBody = Some(formatted.trim)
             }
-        }.map { _ =>
-          markdown.renderToString
+            res
         }
-      } else {
-        Success(code)
+      if (resultIterator.isEmpty) Success(code)
+      else {
+        val fRes: Option[Try[String]] =
+          resultIterator
+            .collectFirst { case f: Failure[_] => f }
+        fRes
+          .getOrElse(Success[String](markdown.renderToString))
       }
     } else {
       doFormatOne(code, style, file, range)
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
@@ -21,6 +21,9 @@ import org.scalafmt.internal.FormatOps
 import org.scalafmt.internal.FormatWriter
 import org.scalafmt.rewrite.Rewrite
 import org.scalafmt.util.FileOps
+import org.scalafmt.util.CodeFence
+import org.scalafmt.util.Text
+import org.scalafmt.util.MarkdownFile
 
 /** WARNING. This API is discouraged when integrating with Scalafmt from a build tool
   * or editor plugin. It is recommended to use the `scalafmt-dynamic` module instead.
@@ -64,7 +67,8 @@ object Scalafmt {
       if (filename == defaultFilename) baseStyle
       else { // might throw for invalid conf
         val style = baseStyle.getConfigFor(filename)
-        val isSbt = FileOps.isAmmonite(filename) || FileOps.isSbt(filename)
+        val isSbt = FileOps.isAmmonite(filename) || FileOps.isSbt(filename) ||
+          FileOps.isMarkdown(filename)
         if (isSbt) style.forSbt else style
       }
     val isWin = code.contains(WinLineEnding)
@@ -101,14 +105,30 @@ object Scalafmt {
       file: String,
       range: Set[Range] = Set.empty
   ): Try[String] =
-    if (!FileOps.isAmmonite(file)) doFormatOne(code, style, file, range)
-    else {
+    if (FileOps.isAmmonite(file)) {
       // XXX: we won't support ranges as we don't keep track of lines
       val chunks = ammonitePattern.split(code)
       if (chunks.length <= 1) doFormatOne(code, style, file, range)
       else
         flatMapAll(chunks.iterator)(doFormatOne(_, style, file))
           .map(_.mkString("\n@\n"))
+    } else if (FileOps.isMarkdown(file)) {
+      val markdown = MarkdownFile.parse(Input.VirtualFile(file, code))
+      flatMapAll(markdown.parts.iterator) {
+        case text: Text => Success(text.value)
+        case fence: CodeFence =>
+          if (fence.info.value.startsWith("scala mdoc")) {
+            doFormatOne(fence.body.value, style, file).map { formatted =>
+              fence.newBody = Some(formatted.trim)
+            }
+          } else {
+            Success("")
+          }
+      }.map { _ =>
+        markdown.renderToString
+      }
+    } else {
+      doFormatOne(code, style, file, range)
     }
 
   private def doFormatOne(
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
@@ -113,25 +113,22 @@ object Scalafmt {
       val markdown = MarkdownFile.parse(Input.VirtualFile(file, code))
 
       val resultIterator: Iterator[Try[String]] =
-        markdown.parts.iterator
-          .collect {
-            case fence: MarkdownPart.CodeFence
-                if fence.info.startsWith("scala mdoc") =>
-              val res = doFormatOne(fence.body, style, file)
-              res.foreach { formatted =>
-                fence.newBody = Some(formatted.trim)
-              }
-              res
-          }
+        markdown.parts.iterator.collect {
+          case fence: MarkdownPart.CodeFence
+              if fence.info.startsWith("scala mdoc") =>
+            val res = doFormatOne(fence.body, style, file)
+            res.foreach { formatted =>
+              fence.newBody = Some(formatted.trim)
+            }
+            res
+        }
       if (resultIterator.isEmpty) Success(code)
-      else {
+      else
         resultIterator
           .find(_.isFailure)
           .getOrElse(Success(markdown.renderToString))
-      }
-    } else {
+    } else
       doFormatOne(code, style, file, range)
-    }
 
   private def doFormatOne(
       code: String,
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
@@ -32,7 +32,7 @@ import metaconfig.generic.Surface
   *
   *               and want to add one extra token (for example "|>") to align by, write
   *
-  *               align.tokens."+" = [ "|> ]
+  *               align.tokens."+" = [ "|>" ]
   *
   *               NOTE. Adding more alignment tokens may potentially decrease the
   *               vertical alignment in formatted output. Customize at your own
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
@@ -32,7 +32,7 @@ import metaconfig.generic.Surface
   *
   *               and want to add one extra token (for example "|>") to align by, write
   *
-  *               align.tokens."+" = [ "|>" ]
+  *               align.tokens."+" = [ "|> ]
   *
   *               NOTE. Adding more alignment tokens may potentially decrease the
   *               vertical alignment in formatted output. Customize at your own
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
@@ -38,7 +38,8 @@ object ProjectFiles {
   private implicit val fs: file.FileSystem = file.FileSystems.getDefault
 
   val defaultIncludePaths =
-    Seq("glob:**.scala", "glob:**.sbt", "glob:**.sc")
+    // TODO: figure out how to make `*.md` pattern optional
+    Seq("glob:**.scala", "glob:**.sbt", "glob:**.sc", "glob:**.md")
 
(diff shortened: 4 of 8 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
```diff
@@ -101,14 +105,34 @@ object Scalafmt {
...
-    else {
+    if (FileOps.isAmmonite(file)) {
       // XXX: we won't support ranges as we don't keep track of lines
       val chunks = ammonitePattern.split(code)
       if (chunks.length <= 1) doFormatOne(code, style, file, range)
       else
         flatMapAll(chunks.iterator)(doFormatOne(_, style, file))
           .map(_.mkString("\n@\n"))
+    } else if (FileOps.isMarkdown(file)) {
+      if (code.contains("```scala mdoc")) {
```
swoogles, line 116: This is a small performance-related piece that Olafurpg sugested (If I understood him correctly), to bypass parsing and rendering markdown files that don't have any affected fenced blocks.

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
```diff
@@ -1030,3 +1030,103 @@ class CliTest extends AbstractCliTest with CliTestBehavior {
   }
 
 }
+
+class CliTestSimp extends AbstractCliTest {
```
swoogles, line 1034: I'm not familiar with the full test setup of this project.
These tests pass over here, but fail if pulled into the class above.

scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
```diff
@@ -101,14 +105,34 @@ object Scalafmt {
...
+            if (fence.info.value.startsWith("scala mdoc")) {
+              doFormatOne(fence.body.value, style, file).map { formatted =>
+                fence.newBody = Some(formatted.trim)
+              }
+            } else {
+              Success("")
+            }
+        }.map { _ =>
+          markdown.renderToString
+        }
```
kitbellew, line 130: you don't need `flatMapAll`. why not

```
  val resultIterator = markdown.parts.iterator.collect {
    case fence: CodeFence if fence.info.value.startsWith("scala mdoc") =>
      val res = doFormatOne(fence.body.value, style, file)
      res.foreach { formatted =>
        // XXX if you need to trim, why not trim fence.body before formatting?
        fence.newBody = Some(formatted.trim)
      }
      res
  }
  if (resultIterator.isEmpty) Success(code)
  else
    resultIterator
      .collectFirst { case f: Failure[_] => f }
      .getOrElse(Success(markdown.renderToString))
```

docs/configuration.md
```diff
@@ -3184,6 +3184,13 @@ project {
 }
 ```
 
+### `project.includeMarkdown`
```
kitbellew, line 3187: i don't think you need this. users can explicitly add a glob for `.md` files, see https://scalameta.org/scalafmt/docs/configuration.html#appending-to-preset-collections

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
```diff
@@ -53,8 +54,10 @@ object ProjectFiles {
       val useIncludePaths =
         pf.includePaths.ne(defaultIncludePaths) || pf.includeFilters.isEmpty
       val includePaths = if (useIncludePaths) pf.includePaths else Seq.empty
+      val markdownPaths =
```
kitbellew, line 57: remove. see comment at the top.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,133 @@
+package org.scalafmt.util
+
+import scala.meta.inputs.Position
+import scala.meta.inputs.Input
+import scala.collection.mutable
+
+final case class MarkdownFile(input: Input, parts: List[MarkdownPart]) {
+  private val appends = mutable.ListBuffer.empty[String]
+  def appendText(text: String): Unit = {
```
kitbellew, line 9: this method is not used in this patch, nor is `appends`. remove?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,133 @@
+package org.scalafmt.util
+
+import scala.meta.inputs.Position
+import scala.meta.inputs.Input
+import scala.collection.mutable
+
+final case class MarkdownFile(input: Input, parts: List[MarkdownPart]) {
```
kitbellew, line 7: input is not used anywhere, remove?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,133 @@
...
+final case class CodeFence(
+    openBackticks: Text,
+    info: Text,
+    body: Text,
+    closeBackticks: Text
+) extends MarkdownPart {
+  var newPart = Option.empty[String]
+  var newInfo = Option.empty[String]
+  var newBody = Option.empty[String]
+}
```
kitbellew, line 133: please put them inside `object MarkdownPart`

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,133 @@
...
+    }
+    def acceptParts(): List[MarkdownPart] = {
+      var state: State = State.Text
+      val parts = mutable.ListBuffer.empty[MarkdownPart]
+      var curr = 0
+      text.linesWithSeparators.foreach { line =>
+        val end = curr + line.length()
+        state match {
+          case State.Text =>
+            if (line.startsWith("```")) {
```
kitbellew, line 59: what if the fence is not at the beginning of the line?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,133 @@
...
+              val backticks = line.takeWhile(_ == '`')
+              val info = line.substring(backticks.length())
+              state = State.CodeFence(curr, backticks, info)
+            } else {
+              parts += newText(curr, end)
+            }
+          case s: State.CodeFence =>
+            if (
+              line.startsWith(s.backticks) &&
+              line.forall(ch => ch == '`' || ch.isWhitespace)
```
kitbellew, line 69: isWhiteSpace or isSpaceChar?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,133 @@
...
+            if (line.startsWith("```")) {
+              val backticks = line.takeWhile(_ == '`')
+              val info = line.substring(backticks.length())
+              state = State.CodeFence(curr, backticks, info)
+            } else {
+              parts += newText(curr, end)
+            }
+          case s: State.CodeFence =>
+            if (
+              line.startsWith(s.backticks) &&
```
kitbellew, line 68: this is, by the way, an example where looking for ` ```scala mdoc ` would be incorrect, since it could be nested inside some other ` ``` ` fence.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,133 @@
...
+              line.forall(ch => ch == '`' || ch.isWhitespace)
+            ) {
+              parts += newCodeFence(s, curr, end)
+              state = State.Text
+            }
+        }
+        curr += line.length()
```
kitbellew, line 75: or `curr = end`

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,133 @@
...
+final case class Text(value: String) extends MarkdownPart
+final case class CodeFence(
+    openBackticks: Text,
+    info: Text,
+    body: Text,
+    closeBackticks: Text
+) extends MarkdownPart {
+  var newPart = Option.empty[String]
+  var newInfo = Option.empty[String]
+  var newBody = Option.empty[String]
```
kitbellew, line 132: these three are never set. remove?

kitbellew (changes requested):
sorry, didn't read through entire thing, let's remove unnnecessary code, will be easier to review what remains.

scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
swoogles: Thanks, done.

scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
swoogles: Thanks! Using this now.
I looked into pre-trimming per your note, but it ends up leaving an extra blank line at the end of the output and failing the current tests.

docs/configuration.md
swoogles: Ah-ha! I just knew something like this must be possible.
Thanks, fixed this and removed the new/unnecessary config.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
swoogles: Removed

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: Sounds good. I just assumed Olafurpg had done it for some reason I didn't understand :)
Removed

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: Removed

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: Yep, that's better, changed.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: I'm going to address the simpler points that you made and then target nested fences in a follow-up commit, since I really have no idea how hard it will be to handle those properly. 
I'll make sure to create the relevant test cases as part of that work.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: Yeah...I don't quite see how he intended these to be used.
I'll remove them for now, and change the sections that match on their values above just unconditionally use the `None` case.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: Done

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: Update - `newPart` and `newInfo` are never written to, but `newBody` is over at `Scalafmt:121`, and the formatting does fail if we don't match on that.
I'm going to leave it in for now

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: Looks like `isSpaceChar` is more appropriate here, changed.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: I added some tests that confirm that this is able to handle (Ignore) nested fences.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: I struggled with this for a few hours and was unable to come up with the correct combination of trimming and character counting to get it to format indented fences.

So at this point, I just want to seek clarification-Is this a requirement? 

I just tried indenting my `scala mdoc` fences in my book project, and that breaks mdoc already.  Or more accurately, mdoc ignores indented fences. Since that's the case, I'm wondering if trying to format indented `scala mdoc` fences is a valid goal.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
```diff
@@ -32,7 +32,7 @@ import metaconfig.generic.Surface
   *
   *               and want to add one extra token (for example "|>") to align by, write
   *
-  *               align.tokens."+" = [ "|> ]
+  *               align.tokens."+" = [ "|>" ]
```
swoogles, line 35: Unrelated to this work, but I just noticed the missing closing `"` when updating the Markdown config to use this syntax.

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
swoogles: Figured this out

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,123 @@
...
+      var curr = 0
+      text.linesWithSeparators
+        .map(line =>
+          if (line.dropWhile(_.isWhitespace).startsWith("```"))
+            line.dropWhile(_.isWhitespace)
+          else
+            line
+        )
+        .foreach { line =>
+          val end = curr + line.length()
```
kitbellew, line 58: since you truncated the line above, this calculation is incorrect. suggest merging the map logic into this foreach:
```
        .foreach { line =>
          val trimmedLine = line.trim()
          val end = curr + line.length()
          state match {
            case State.Text =>
              if (trimmedLine.startsWith("```")) {
                val index = trimmedLine.indexWhere(_ != '`', 3)
                val (backticks, info) =
                  if (index < 0) (trimmedLine, String.empty) else trimmedLine.partition(index)
                state = State.CodeFence(curr, backticks, info)
              } else {
                parts += newText(curr, end)
              }
            case s: State.CodeFence =>
              if (
                trimmedLine.lengthCompare(s.backticks.length()) >= 0 &&
                trimmedLine.forall(_ == '`')
              ) {
                parts += newCodeFence(s, curr, end)
                state = State.Text
              }
          }
          curr = end
        }
```

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
```diff
@@ -101,14 +101,36 @@ object Scalafmt {
...
+            }
+            res
+        }
+      if (resultIterator.isEmpty) Success(code)
+      else {
+        val fRes: Option[Try[String]] =
+          resultIterator
+            .collectFirst { case f: Failure[_] => f }
+        fRes
+          .getOrElse(Success[String](markdown.renderToString))
```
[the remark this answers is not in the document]
swoogles: Yep, that works, thanks for the alternative

scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
```diff
@@ -1,16 +1,15 @@
...
 import scala.annotation.tailrec
 import scala.meta.Dialect
 import scala.meta.Input
 import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
 import scala.util.matching.Regex
-
 import org.scalafmt.config.Config
 import org.scalafmt.Error.PreciseIncomplete
```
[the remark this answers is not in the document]
swoogles: The newlines? I'm not sure what introduced them.
I will remove them.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,123 @@
...
+
+import scala.meta.inputs.Position
+import scala.meta.inputs.Input
+import scala.collection.mutable
+
+final case class MarkdownFile(parts: List[MarkdownPart]) {
+  def renderToString: String = {
+    val out = new StringBuilder()
+    parts.foreach(_.renderToString(out))
+    out.toString()
```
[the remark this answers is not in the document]
swoogles: Fixed.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,123 @@
...
+      part
+    }
+    def acceptParts(): List[MarkdownPart] = {
+      var state: State = State.Text
+      val parts = mutable.ListBuffer.empty[MarkdownPart]
+      var curr = 0
+      text.linesWithSeparators
+        .map(line =>
+          if (line.dropWhile(_.isWhitespace).startsWith("```"))
+            line.dropWhile(_.isWhitespace)
```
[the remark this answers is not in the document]
swoogles: Meaning the second `.dropWhile` ?
I was trying to keep things as a single expression in the map. I will pull it into a separate variable to only do the `.dropWhile` a single time.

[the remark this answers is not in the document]
swoogles: Actually, this was just for the attempted parsing of indented fences. 
I will remove this whole `map` section if that's not a requirement.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,123 @@
...
+          case Some(newBody) =>
+            out.append(newBody)
+        }
+        fence.closeBackticks.renderToString(out)
+    }
+}
+object MarkdownPart {
+  final case class Text(value: String) extends MarkdownPart
+  final case class CodeFence(
+      openBackticks: Text,
```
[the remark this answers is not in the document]
swoogles: I just took Olafur's code as-is here. Will downgrade them to plain `String`s and see what other changes that requires.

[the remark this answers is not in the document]
swoogles: Okay, this ended up letting me removing several things, including `MarkdownPart.pos`.

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
```diff
@@ -91,6 +91,8 @@ abstract class AbstractCliTest extends FunSuite {
       val exit = Cli.run(config)
       assertExit(exit)
       val obtained = dir2string(input)
+      println("Obtained: ")
+      println(obtained)
```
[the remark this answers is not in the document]
swoogles: Yep, sorry, removed.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,123 @@
...
+  }
+  def parse(input: Input): MarkdownFile = {
+    val parser = new Parser(input)
+    val parts = parser.acceptParts()
+    MarkdownFile(parts)
+  }
+}
+
+sealed abstract class MarkdownPart {
+  var pos: Position = Position.None
```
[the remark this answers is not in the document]
swoogles: Olafur was tracking it in a few places, but never acting on it. Removed.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: I got rid of the line truncating, since it wasn't actually enabling parsing of indented fences.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/Align.scala
swoogles: Removed from this changeset.

swoogles:
Investigating the test failures.
Not sure what's different about those environments. 
Running `ci-test` locally passes cleanly :slightly_frowning_face:

swoogles:
Looks like the `TEST="2.12"` version of the tests are failing locally, so I can try and get those resolved locally and re-push.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: Actually, this was the cause of the latest breakage. 
Restored `isWhiteSpace` and now everything passes.
Executed both of these locally:
`TEST="2.12" sbt clean ci-test`
and
`TEST="2.13" sbt clean ci-test`

Should be good to go, unless you see some new things @kitbellew !

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
```diff
@@ -1029,4 +1029,196 @@ class CliTest extends AbstractCliTest with CliTestBehavior {
...
+  test(s"ignores .md files if not present in includePaths") {
+    val input = string2dir(
+      s"""|/foobar.md
+        |# Hello
+        |Example usage 1
+        |```scala mdoc
+        |val  x   =   42
+        |```
+        |/.scalafmt.conf
+        |version = ${Versions.version}
```
kitbellew, line 1086: why are you specifying the version here? the versions are only used for dynamic loading of actually released versions, and there's no published version which supports this new functionality.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,119 @@
...
+          case Some(newBody) =>
+            out.append(newBody)
+        }
+        out.append(fence.closeBackticks)
+    }
+}
+object MarkdownPart {
+  final case class Text(value: String) extends MarkdownPart
+  final case class CodeFence(
+      openBackticks: String,
```
kitbellew, line 111: are open backticks ever going to be different from close?

scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
```diff
@@ -101,14 +102,35 @@ object Scalafmt {
...
+            case fence: MarkdownPart.CodeFence
+                if fence.info.startsWith("scala mdoc") =>
+              val res = doFormatOne(fence.body, style, file)
+              res.foreach { formatted =>
+                fence.newBody = Some(formatted.trim)
+              }
+              res
+          }
+      if (resultIterator.isEmpty) Success(code)
+      else {
```
kitbellew, line 127: unnecessary curly braces

scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
```diff
@@ -101,14 +102,35 @@ object Scalafmt {
...
       if (chunks.length <= 1) doFormatOne(code, style, file, range)
       else
         flatMapAll(chunks.iterator)(doFormatOne(_, style, file))
           .map(_.mkString("\n@\n"))
+    } else if (FileOps.isMarkdown(file)) {
+      val markdown = MarkdownFile.parse(Input.VirtualFile(file, code))
+
+      val resultIterator: Iterator[Try[String]] =
+        markdown.parts.iterator
+          .collect {
```
kitbellew, line 117: unnecessary line break

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
```diff
@@ -0,0 +1,119 @@
...
+
+}
+
+sealed abstract class MarkdownPart {
+  final def renderToString(out: StringBuilder): Unit =
+    this match {
+      case MarkdownPart.Text(value) =>
+        out.append(value)
+      case fence: MarkdownPart.CodeFence =>
+        out.append(fence.openBackticks)
```
kitbellew, line 97: a bit unclear... at which point are we adding newlines? I'd expect after info, then after body, then after closing backticks.

kitbellew (changes requested):
thank you for being extremely patient and bearing with me, I'm reviewing, probably incompletely, on the mobile phone while on vacation...

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
swoogles: Ah, I had just cargo-culted here, thinking that we always need to specify a version in the tests.
I wonder if *this* is what's killing the CI process...
If it's grabbing something that I've published locally, but is unavailable anywhere else.
Removing this.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: It sounds like that's against spec, so not defined behavior.
https://meta.stackoverflow.com/a/409129/1128869
We could do a few different things

1. Ignore the discrepancy, and just copy the front backticks onto the end
2. Preserve the unbalanced backticks
3. Error out if unbalanced

Currently we're doing # 2. I don't have any strong feelings about this.

scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
swoogles: Removed

scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
swoogles: Removed

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: Looks like that's handled via `text.linesWithSeparators` on line 56.
Everything we're appending here is going to include its trailing line break.
From the `linesWithSeparators` docs-
```
  /** Return an iterator of all lines embedded in this string,
   *  including trailing line separator characters.
   * /
```

scalafmt-core/shared/src/main/scala/org/scalafmt/util/MarkdownFile.scala
swoogles: Added a test that codifies the current behavior.
Will update it if a different choice is made.

swoogles:
> thank you for being extremely patient and bearing with me, I'm reviewing, probably incompletely, on the mobile phone while on vacation...

Oh shoot, I don't want to impose on your vacation!

I just wish I could get the build to run when I push changes, so that I could make sure this was in a green state whenever you're more available for reviewing.

No rush from my end though, it's going to be a long while before we're trying to publish our book :)

kitbellew (approved):

swoogles:
:fireworks: 
Thanks @kitbellew !
Any rough idea when the next RC will go out?

kitbellew:
@swoogles can you please send another change, describing this behaviour in `configuration.md`?

swoogles:
> @swoogles can you please send another change, describing this behaviour in `configuration.md`?

Yep, will do.

Issue #2640, opened 2021-07-19
Scalafmt 3.0.0-RC6 with Scala3 dialect has issues with implicit inline def


mdedetrich:
This template is a guideline, not a strict requirement.
- **Version**: 3.0.0-RC6
- **Integration**: sbt-scalafmt
- **Configuration**:
```
runner.dialect = scala3
version=3.0.0-RC6
maxColumn=120
project.git = true
```

## Steps

Given code like this:
```scala
implicit inline def qm: QueryMeta[PersonName, String] = { //hellooo
  queryMeta[PersonName, String](
    quote {
      (q: Query[PersonName]) => q.map(p => p.name)
    }
  )((name: String) => PersonName(name))
}
```

When I run scalafmt like this:

```bash
sbt test:scalafmt
```

## Problem

Scalafmt format fails with the following error
```
[error] /home/mdedetrich/github/protoquill/quill-sql/src/test/scala/io/getquill/QueryMetaTest.scala:66: error: => expected but def found
[error]       implicit inline def qm: QueryMeta[PersonName, String] = { //hellooo
[error]                       ^: /home/mdedetrich/github/protoquill/quill-sql/src/test/scala/io/getquill/QueryMetaTest.scala
```

## Expectation

scalafmt to format code correctly without any errors

## Workaround

None apart from deleting the code

## Notes

Here is the original code https://github.com/getquill/protoquill/blob/d095b54d4e267ed0f2a1e6a60b6f0aa16b122d99/quill-sql/src/test/scala/io/getquill/QueryMetaTest.scala#L66-L72

This seems to be just a case where the scala3 dialect isn't aware of a new case that is present in Scala3, it may also be an issue with scalameta not being up to date enough to handle this case?

tgodzik:
Thanks for reporting! This is a parser issue, fixing it here https://github.com/scalameta/scalameta/pull/2428

mdedetrich:
Awesome thanks! Would this be released in an upcoming `RC7`?

tgodzik:
Most likely, next will be `3.0.0`, I think we fixed most of the bigger bugs already and any outliers can be fixed later.

mdedetrich:
Cool, thanks for pushing this through do quickly

kitbellew:
@mdedetrich surprisingly, i cannot reproduce this problem, hence i can't test whether the new scalameta release fixed it. cc @tgodzik

mdedetrich:
[...]

PR #2655, opened 2021-07-30
Scalameta: upgrade to v4.4.27

kitbellew:


tgodzik:
We could try updating to scalameta SNAPSHOT first next time to figure out if anything needs to be fixed still. We wouldn't need to do a rerelease then.

kitbellew:
> We could try updating to scalameta SNAPSHOT first next time to figure out if anything needs to be fixed still. We wouldn't need to do a rerelease then.

1. where do we get a scalameta snapshot?
2. should we somehow include testing of scalafmt before merging scalameta commits? a bit drastic but it looks like scalafmt has more unit tests (and out of more than 3.5k tests, only two failed, so...)

tgodzik:
> > We could try updating to scalameta SNAPSHOT first next time to figure out if anything needs to be fixed still. We wouldn't need to do a rerelease then.
> 
> 1. where do we get a scalameta snapshot?

I usually check the last release job output for the version https://github.com/scalameta/scalameta/actions/workflows/release.yml

It should also be available in sonatype (I think, but I would need to search for the link)

> 2. should we somehow include testing of scalafmt before merging scalameta commits? a bit drastic but it looks like scalafmt has more unit tests (and out of more than 3.5k tests, only two failed, so...)

I think testing manually before the release should really be enough. We already test a number of Scala 2 and Scala 3 sources, so anything big should pop up in the PRs. I usually do it just in case before the release.

kitbellew:
@tgodzik found a published recent snapshot, build passes. now we need to wait until 4.4.26 is released.

tgodzik:
I can release it tomorrow!

tgodzik:
new release is on it's way! I made a mistake and released `2.2.26` instead of `4.4.26`, but it should be republished soon:
https://github.com/scalameta/scalameta/actions/runs/1102300362

tgodzik:
The release is out, it should also close https://github.com/scalameta/scalafmt/issues/2640

kitbellew:
@tgodzik PTAL

tgodzik (approved):
LGTM!

PR #2656, opened 2021-07-30
BinPack: implement logic for unsafeXxxSite=Oneline

kitbellew:
Mentioned in #2633.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -782,10 +782,10 @@ class Router(formatOps: FormatOps) {
 
       case tok @ FormatToken(open @ LeftParenOrBracket(), right, _) if {
             if (isCallSite(leftOwner))
-              !style.binPack.unsafeCallSite &&
+              style.binPack.unsafeCallSite.isNever &&
               !isSuperfluousParenthesis(formatToken.left, leftOwner)
             else
-              !style.binPack.unsafeDefnSite &&
+              style.binPack.unsafeDefnSite.isNever &&
               isDefnSite(leftOwner)
           } =>
         val close = matching(open)
@@ -1033,7 +1033,7 @@ class Router(formatOps: FormatOps) {
         splitsNoNL ++ splitsNL ++ splitsForAssign.getOrElse(Seq.empty)
 
       case FormatToken(open @ LeftParenOrBracket(), right, between)
-          if style.binPack.unsafeDefnSite && isDefnSite(leftOwner) =>
+          if !style.binPack.unsafeDefnSite.isNever && isDefnSite(leftOwner) =>
         val close = matching(open)
         val isBracket = open.is[T.LeftBracket]
         val indent = Num(style.indent.getDefnSite(leftOwner))
@@ -1088,7 +1088,7 @@ class Router(formatOps: FormatOps) {
         }
 
       case FormatToken(LeftParenOrBracket(), _, _)
-          if style.binPack.unsafeCallSite && isCallSite(leftOwner) =>
+          if !style.binPack.unsafeCallSite.isNever && isCallSite(leftOwner) =>
         val open = formatToken.left
         val close = matching(open)
         val indent = Num(style.indent.callSite)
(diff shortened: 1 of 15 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -865,11 +865,11 @@ object TreeOps {
     if (groups.isEmpty) None else Some(groups)
   }
 
-  def isBinPack(owner: Tree, enabled: Boolean)(implicit
+  def isBinPackNever(owner: Tree, isNever: Boolean)(implicit
```
tgodzik, line 868: Is this used?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1103,25 +1113,43 @@ class Router(formatOps: FormatOps) {
...
           mustUseConfigStyle(formatToken, !opensLiteralArgumentList)
 
+        val argsOpt = (leftOwner match {
+          case SplitCallIntoParts(_, args) =>
+            args.fold(Some(_), formatOps.findArgsFor(open, _))
+          case _ => None
+        }).filter(_.nonEmpty)
+        val isSingleArg = argsOpt.exists(_.lengthCompare(1) == 0)
+
+        def findComma(ft: FormatToken) =
```
tgodzik, line 1123: Should we extract it? It's also used above. The logic overall seems very similar.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1103,25 +1113,43 @@ class Router(formatOps: FormatOps) {
...
+          findFirst(ft, close) {
+            case FormatToken(_, _: T.Comma, _) => true
+            case _ => false
+          }.map(_.right)
+
+        val oneline = style.binPack.unsafeCallSite == BinPack.Unsafe.Oneline
```
tgodzik, line 1129:
```suggestion
        val oneline = style.binPack.unsafeCallSite.isOneline
```
would it make sense to add additional methods?

scalafmt-core/shared/src/main/scala/org/scalafmt/util/StyleMap.scala
```diff
@@ -36,17 +37,18 @@ class StyleMap(
...
-      def changeStyle(style: ScalafmtConfig): Boolean = {
+      def changeStyle(style: ScalafmtConfig): Option[ScalafmtConfig] = {
         val changing = curr != style
-        if (changing) {
+        if (!changing) None
+        else {
           startBuilder += tok.left.start
           styleBuilder += style
-          curr = style
+          try Some(curr)
```
tgodzik, line 49: Could we write this explicitely?
```scala
val changed = Some(curr)
curr = style
changed
```
since `Some(curr)` doesn't throw an exception this seems a weird usage of try and might be unclear to what's going on.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1281,24 +1310,56 @@ class Router(formatOps: FormatOps) {
         }
       case FormatToken(_: T.Comma, right, _) if leftOwner.isNot[Template] =>
         // TODO(olafur) DRY, see OneArgOneLine.
```
tgodzik, line 1312: Does the comment still make sense? Without an issue connected to it it's hard to figure out what this means exactly.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Also this seems to be a very similar logic repeated 3 times, would it be possible reduce at least some parts of it?

tgodzik (changes requested):
I haven't found anything major, but I think we could improve it a bit? So mostly nitpicks.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: done

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: removed

tgodzik (approved):
LGTM!

PR #2657, opened 2021-07-30
FormatOps: fix bug with vertical multiline

kitbellew:
The rule mistakenly matched the wrong `(` and led to a binpack-like formatting of default parameters containing an Apply.


tgodzik (approved):
LGTM!

PR #2670, opened 2021-08-08
Scalameta: upgrade to v4.4.26

### LINKED ISSUE #2640: Scalafmt 3.0.0-RC6 with Scala3 dialect has issues with implicit inline def

kitbellew:
Fixes #2640.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -6,7 +6,7 @@ import org.portablescala.sbtplatformdeps.PlatformDepsPlugin.autoImport._
 
 object Dependencies {
   val metaconfigV = "0.9.14"
-  val scalametaV  = "4.4.24"
+  val scalametaV  = "4.4.26"
   val scalacheckV = "1.15.4"
   val coursier    = "1.0.3"
   val munitV      = "0.7.27"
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM!

PR #2671, opened 2021-08-08
Add documentation for Markdown formatting

swoogles:
I tried for a while to use the `mdoc:scalafmt` approach to document this, but it was a snake eating its tail, so I resorted to this basic version for the sake of getting _something_ out there for users.


kitbellew (approved):

kitbellew:
@swoogles thank you!

PR #2686, opened 2021-08-15
FormatWriter: print markdown code blocks relative

kitbellew:
- The regular code blocks contain full leading whitespace while markdown block are relative to the opening fence thus handle them differently.
- Fix an off-by-1 margin calculation applicable to all code blocks.
- Allow top-level terms in scaladoc.


tgodzik (approved):
LGTM

PR #2691, opened 2021-08-21
FormatOps: keep NL for opt braces in partial func

### LINKED ISSUE #2687: Scalafmt 3.0.0 fails on Scala3 code starting with case statement
- **Version**: 3.0.0
- **Integration**: scalafmt-native
- **Configuration**:
```
runner.dialect = scala3
version=3.0.0
maxColumn=120
project.git = true
```

## Steps

Given code like this:
```scala
  given liftPropertyAlias: NiceLiftable[PropertyAlias] with {
    def lift = case PropertyAlias(a, b) => '{ PropertyAlias(${ a.expr }, ${ b.expr }) }
  }
```

```bash
scalafmt
```

## Problem

Scalafmt errors out with the following

```
/home/mdedetrich/github/protoquill/quill-sql/src/main/scala/io/getquill/parser/Lifter.scala:110: error: illegal start of simple expression
    def lift = case PropertyAlias(a, b) => '{ PropertyAlias(${ a.expr }, ${ b.expr }) }
```

## Expectation

I would like for scalafmt to not error out

## Workaround

None as far as I can tell, not even using `// format: off` and `// format: on` seems to work

## Notes

Original code where it fails https://github.com/getquill/protoquill/blob/master/quill-sql/src/main/scala/io/getquill/parser/Lifter.scala#L106-L109

kitbellew:
Fixes #2687.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -2240,7 +2240,7 @@ class FormatOps(
           case t @ SplitAssignIntoParts((x: Term.PartialFunction, _)) =>
             Some(new OptionalBracesRegion {
               def owner = Some(t)
-              def splits = getSplitsForStats(ft, nft, x.cases, false)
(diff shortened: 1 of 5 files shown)


tgodzik (approved):
LGTM!

PR #2694, opened 2021-08-22
FormatWriter: add parameter to set number of blanks before or after all statements

kitbellew:


scalafmt-tests/src/test/resources/newlines/alwaysBeforeTopLevelStatements.source
```diff
@@ -709,3 +709,119 @@ end Builder
...
+object Foo {
+  def apply[V <: VAType](someProperty: String): SomePropertyBuilder[V] =
+    new SomePropertyBuilder(someProperty)
+
+}
+<<< before negative, topLevelBodyIfMinStatements = before
+newlines.topLevelBodyMinStatements = 1
+newlines.topLevelBodyIfMinStatements = [before]
+newlines.topLevelStatementBlankLines = [
+  { blanks = { before = -1, after = 1 } }
```
tgodzik, line 753: So here, `before = -1` is being ignored due to `newlines.topLevelBodyIfMinStatements = [before]`

tgodzik (approved):
Looks good, though I am not sure I understand what is the intended behaviour. Left a question about it.

scalafmt-tests/src/test/resources/newlines/alwaysBeforeTopLevelStatements.source
kitbellew: correct, that's the test, the higher number wins.

PR #2696, opened 2021-08-24
PreferCurlyFors: count non-Guards, not Generators

### LINKED ISSUE #2695: PreferCurlyFors generates illegal code (and it isn't curly either)
- **Version**: 3.0.0
- **Integration**: Coursier launcher
- **Configuration**:
```
version = 3.0.0
rewrite.rules = [ PreferCurlyFors ]
```

## Steps

Given code like this:
```scala
object O {
  for (pos <- "a->1|b->2".split('|'); c = pos.split("->"))
  yield ()
}
```

```bash
scalafmt --config scalafmt.conf S.scala
```

## Problem

```scala
object O {
  for (
    pos <- "a->1|b->2".split('|')
    c = pos.split("->")
  )
    yield ()
}
```

which the compiler then actually rejects:

```scala
% scalac S.scala
S.scala:4: error: value c is not a member of Array[String]
possible cause: maybe a semicolon is missing before `value c`?
    c = pos.split("->")
    ^
1 error
```

## Expectation

```scala
object O {
  for {
    pos <- "a->1|b->2".split('|')
    c = pos.split("->")
  }
    yield ()
}
```

## Notes

this came up over at https://github.com/playframework/twirl/pull/437

kitbellew:
Fixes #2695.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/PreferCurlyFors.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/PreferCurlyFors.scala
@@ -17,8 +17,10 @@ object PreferCurlyFors extends Rewrite with FormatTokensRewrite.RuleFactory {
   override def create(ftoks: FormatTokens): FormatTokensRewrite.Rule =
     new PreferCurlyFors(ftoks)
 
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM

Issue #2712, opened 2021-08-30
Formatting issue with scalafmt 3.0.1


github-julien-pie:
Hey there 🤙

- **Version**: 3.0.1
- **Integration**: IntelliJ
- **Configuration**:
```
version = 3.0.1

preset = defaultWithAlign
maxColumn = 100
importSelectors = singleLine
assumeStandardLibraryStripMargin = true
rewrite.rules = [RedundantBraces, RedundantParens]
rewrite.redundantBraces.stringInterpolation = true
project.git = true
optIn.breaksInsideChains = true
lineEndings = preserve
trailingCommas = preserve
newlines.implicitParamListModifierPrefer = before

rewriteTokens = {
  "⇒": "=>"
  "←": "<-"
  "→": "->"
}
```

## Steps

Given code like this:
```scala
class Toto() {}

private def myScalaFmtTest(): String = {
  val test = new Toto
  ("test")
}
```

When I run scalafmt like this:

```bash
scalafmt
```

## Problem

Scalafmt formats code like this:
```scala
class Toto() {}

private def myScalaFmtTest(): String = {
  val test = new Toto "test"
}
```

## Expectation

I would like the formatted output to look like this:
```scala
class Toto() {}

private def myScalaFmtTest(): String = {
  val test = new Toto
  "test"
}
```

## Workaround

When reformating the code manually, I can bypass this issue.

## Notes

I know the code is not that nice, but in our original use case we returned something else than a `String` and got the exact same behaviour.

tgodzik:
Thanks for reporting! Trying to fix it in https://github.com/scalameta/scalafmt/pull/2713

github-julien-pie:
Thank a lot guys for this quick fix 

![giphy](https://user-images.githubusercontent.com/87750553/131466567-b78bc78d-e402-4f5d-9a17-ef5ce7fc83bc.gif)

PR #2713, opened 2021-08-30
Add newline instead of space if contructor is followed by a literal

tgodzik:
Fixes https://github.com/scalameta/scalafmt/issues/2712

### DIFF
--- a/scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat
+++ b/scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat
@@ -745,3 +745,19 @@ object Application {
 
   def main(args: Array[String]): Unit = app.run(args)
 }
+<<< constructor issue
+rewrite.rules = [ RedundantParens]
+===
+class Toto() {
+private def myScalaFmtTest(): String = {
+  val test = new Toto
+  ("test")
+}
+}
+>>>
+class Toto() {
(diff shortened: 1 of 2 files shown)


scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat
```diff
@@ -745,3 +745,19 @@ object Application {
 
   def main(args: Array[String]): Unit = app.run(args)
 }
+<<< constructor issue
+rewrite.rules = [ RedundantParens]
+===
+class Toto() {
+private def myScalaFmtTest(): String = {
+  val test = new Toto
+  ("test")
```
kitbellew, line 754: i suspect the problem here might be a little different. if you prepend ONLY to the test title, it will print both formatting details, and the rules invoked should be the same.

perhaps because of redundant paren rewrite, the statement start logic missed this case? can you add another test where the expression in parens doesn't start with a literal? a method call?

kitbellew (changes requested):

scalafmt-tests/src/test/resources/rewrite/RedundantParens.stat
tgodzik: Ach, you are right, it's ` redundant paren rewrite`. Do we update statement starts with that rule?

PR #2715, opened 2021-08-30
FormatOps: find stmt starts after redundant parens

### LINKED ISSUE #2712: Formatting issue with scalafmt 3.0.1

kitbellew:
Fixes #2712.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -90,7 +90,7 @@ class FormatOps(
 
   private[internal] val soft = new SoftKeywordClasses(dialect)
   private[internal] val statementStarts =
-    getStatementStarts(topSourceTree, tokens(_).left, soft)
+    getStatementStarts(topSourceTree, tokens.after(_).left, soft)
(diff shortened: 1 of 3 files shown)


kitbellew:
@tgodzik replace #2713.

tgodzik (approved):
LGTM!

PR #2719, opened 2021-09-01
Imports: implement contiguousGroups logic

### LINKED ISSUE #2718: Sort imports with groups can reorder program execution
- **Version**: 3.0.1
- **Integration**: IntelliJ
- **Configuration**:
```
version = "3.0.1"
rewrite.rules = [Imports]
rewrite.imports.groups = [["""scala\..*"""]]
```

## Steps

Given code like this:
```scala
object Main {
  import Foo._
  val bar: Bar = getBar
  import bar.ec

  def getBar: Foo.Bar = ???
}
object Foo {
  trait Bar {
    implicit def ec: ExecutionContext
  }
}
```

IntelliJ keybind

## Problem

```scala
object Main {
  import Foo._
  import bar.ec
  val bar: Bar = getBar

  def getBar: Foo.Bar = ???
}
object Foo {
  trait Bar {
    implicit def ec: ExecutionContext
  }
}
```

## Expectation

```scala
object Main {
  import Foo._
  val bar: Bar = getBar
  import bar.ec

  def getBar: Foo.Bar = ???
}
object Foo {
  trait Bar {
    implicit def ec: ExecutionContext
  }
}
```

## Workaround
Keeping the first import outside the class it works fine.

## Notes

See also..

kitbellew:
Fixes #2718.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2650,15 +2650,6 @@ import foo._
 
 #### Imports: `groups`
 
-This rule will separate all import statements into groups. If sorting is
-enabled (i.e., not `none`), imports will also be sorted within each group.
-
-The `groups` parameter defines several sets of regular expressions; each
-set defines a single group, and the groups are output in the order they
-are configured (imports not matching any of the regexes will form their own
-group at the end). Regular expressions are applied to the entire parent domain
-of the import statement, up to and including the final dot.
-
 > Keep in mind that this functionality should be used very carefully if
 > hierarchical (relative) imports are allowed in your codebase. Groups
 > should only refer to typical top-level domains such as `java`, `org`,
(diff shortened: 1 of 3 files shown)


kitbellew:
@tgodzik thoughts on this?

docs/configuration.md
```diff
@@ -2668,6 +2659,21 @@ of the import statement, up to and including the final dot.
...
+
+The rule accepts the following parameters:
+
+- `rewrite.imports.groups`: defines several sets of regular expressions; each
+  set defines a single group, and the groups are output in the order they
+  are configured (imports not matching any of the regexes will form their own
+  group at the end). Regular expressions are applied to the entire parent domain
+  of the import statement, up to and including the final dot.
+- `rewrite.imports.contiguousGroups` (since v3.0.2):
+  - if `no` (default), grouping will happen on all imports within the same
```
tgodzik, line 2673: I would honestly make it a default and not even add a setting for this. There is a high chance that `val` or something else between imports is there on purpose.

kitbellew: i can make it default but in our codebase, we prefer to take imports spread out around the file and put them in one place. users can define their vals and then import everything.

tgodzik: Ach, ok, it makes sense. But I think the default should be to not touch it, that's the safer side.

kitbellew: changed the default

tgodzik: Awesome thanks!

tgodzik (approved):

Issue #2725, opened 2021-09-06
Keep new line for a single extension method for Scala 3


danicheg:
That perhaps not a bug, rather a 'help wanted' issue.

- **Scala version**: 3.0.2
- **Scalafmt version**: 3.0.2
- **Integration**: sbt
- **Configuration**:
```
version = 3.0.2
maxColumn = 100
continuationIndent.callSite = 2
newlines {
  sometimesBeforeColonInMethodReturnType = false
  source=keep
}
align {
  arrowEnumeratorGenerator = false
  ifWhileOpenParen = false
  openParenCallSite = false
  openParenDefnSite = false
  tokens = ["%", "%%"]
}
docstrings.style = Asterisk
rewrite {
  rules = [SortImports, RedundantBraces]
  redundantBraces.maxLines = 1
}
fileOverride {
  "glob:**/scala-3.x/src/main/scala/**" {
    runner.dialect = scala3
  }
}
```

## Steps

Given code like this:
```scala
extension (a: All) 
  def getAll: Boolean = a
```

When I run scalafmt like this:

```bash
scalafmt
```

## Problem

Scalafmt formats code like this:
```scala
extension (a: All) def getAll: Boolean = a
```

## Expectation

I would like the formatted output to look like this:
```scala
extension (a: All) 
  def getAll: Boolean = a
```

tgodzik:
Thanks for reporting! I think it's rather a bug since you have `source=keep` set it should leave the newline as is.

Let me see if I can fix it quickly.

danicheg:
Thanks to @dos65 for the initial check on this.

PR #2726, opened 2021-09-06
Respect newlines parameter in extension methods with a single param

tgodzik:
Fixes https://github.com/scalameta/scalafmt/issues/2725

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -499,12 +499,19 @@ class Router(formatOps: FormatOps) {
           if leftOwner.is[Defn.ExtensionGroup] &&
             nextNonComment(formatToken).right.isNot[LeftParenOrBrace] =>
         val expireToken = getLastToken(leftOwner)
-        // Force a newline with the body of a Defn.ExtensionGroup if the
-        // method doesn't fit on a single line.
-        Seq(
-          Split(Space, 0).withSingleLine(expireToken),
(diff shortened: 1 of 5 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -499,12 +499,19 @@ class Router(formatOps: FormatOps) {
           if leftOwner.is[Defn.ExtensionGroup] &&
             nextNonComment(formatToken).right.isNot[LeftParenOrBrace] =>
         val expireToken = getLastToken(leftOwner)
-        // Force a newline with the body of a Defn.ExtensionGroup if the
-        // method doesn't fit on a single line.
-        Seq(
-          Split(Space, 0).withSingleLine(expireToken),
-          Split(Newline, 1).withIndent(style.indent.main, expireToken, After)
-        )
+        def baseSplit(cost: Int = 0) = Split(Newline, cost)
```
kitbellew, line 502: 1. I'd rename it `nlSplit`
2. lets add implicit `FileLine`, so we know where the split was defined and the ordering of splits is preserved (before, Newline was after Space, and now it is before, which might change formatting in some cases).

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: 1. Renamed.
2. Added the implicit `FileLine`

Thanks for reviewing!

kitbellew (approved):

PR #2728, opened 2021-09-07
ScalafmtRunner: warn if default dialect is used

### LINKED ISSUE #2723: Suggestion - warn about default `runner.dialect`
I think it will save time for both users and mantainers  if there would be an option to discover `runner.dialect` setting just by interacting with scalafmt. I think this one looks more like required one that optional and it correlates with amount of issues about that.

So, what if in case of any parsing error there will be an additional warning like:
```
[warn] runner.dialect isn't configured - the default `scala213` was used. 
Please check [link to documentation] and specify it in configuration
```

That would allow to do a soft transition of this setting from an optional to required and increase the amount of people who know about it.

@tgodzik @kitbellew wdyt?

kitbellew:
Fixes #2723.


scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtRunner.scala
```diff
@@ -40,6 +40,14 @@ case class ScalafmtRunner(
   def parse(input: meta.inputs.Input): Parsed[_ <: Tree] =
     dialect(input).parse(parser.parse)
 
+  def isDefaultDialect = dialect eq ScalafmtRunner.Dialect.default
+  def warnDefault: Unit =
+    if (isDefaultDialect) {
+      val msg = ScalafmtRunner.Dialect.known
+        .map(_.source)
+        .mkString("Default dialect is deprecated; use explicit: [", ",", "]")
```
dos65, line 48: Maybe also add a link to documentation?

dos65 (approved):
LGTM!

PR #2731, opened 2021-09-08
ConvertToNewScala3Syntax: guard on dialect flags

### LINKED ISSUE #2730: Feature request: convertToNewSyntax for Scala213Source3
- **Version**: 3.0.2
- **Configuration**:
```
runner.dialect = Scala213Source3
```

## Notes

Scala 2 with `-Xsource:3` supports imports with star instead of underscore, the new varargs syntax and renaming imports with `as` but it does not support the new control flow syntax so the rewrite rules would have to be split up.

kitbellew:
Remove the overall guard on significant indentation. Fixes #2730.


tgodzik (approved):
LGTM!

PR #2749, opened 2021-09-18
TreeOps: classify Type.Apply only as isCallSite

kitbellew:
It was also present in isDefnSite.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -827,7 +827,7 @@ class Router(formatOps: FormatOps) {
         val defnSite = !tupleSite && anyDefnSite
 
         val indent =
-          if (anyDefnSite && !leftOwner.isInstanceOf[Type.Apply])
+          if (anyDefnSite)
             Num(style.indent.getDefnSite(leftOwner))
           else
             Num(style.indent.callSite)
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
@@ -399,9 +399,9 @@ object TreeOps {
     tree match {
       case _: Decl.Def | _: Defn.Def | _: Defn.Macro | _: Defn.Class |
           _: Defn.Trait | _: Ctor.Secondary | _: Decl.Type | _: Defn.Type |
-          _: Type.Apply | _: Type.Param | _: Type.Tuple | _: Defn.Enum |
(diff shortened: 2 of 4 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -827,7 +827,7 @@ class Router(formatOps: FormatOps) {
         val defnSite = !tupleSite && anyDefnSite
 
         val indent =
-          if (anyDefnSite && !leftOwner.isInstanceOf[Type.Apply])
```
tgodzik, line 830: Wasn't this because `Type.Apply` is actually only present in definitions?

It also seems that we will get a lot of changes if people are using default configuration.

tgodzik (commented):
I wouldn't change this currently as it seems it could influence a lot of people's codebases and cause quite difficult to explain changes.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -399,9 +399,9 @@ object TreeOps {
     tree match {
       case _: Decl.Def | _: Defn.Def | _: Defn.Macro | _: Defn.Class |
           _: Defn.Trait | _: Ctor.Secondary | _: Decl.Type | _: Defn.Type |
-          _: Type.Apply | _: Type.Param | _: Type.Tuple | _: Defn.Enum |
```
kitbellew, line 402: take a look at line 440 below. `Type.Apply` is present in both call site and defn site, but it can't be both.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
kitbellew: fist, this specific line doesn't change anything, it just removes the no longer necessary check because Type.Apply is not a defnSite anymore.

also, `Type.Apply` can be present in both definitions and not, but _itself_ it is *not* a definition site.

```
  // used within a call site
  val a = foo(bar: Map[String, String])

  // used within a defn site but on the rhs, hence not a defn
  type a = Map[String, String]
```

kitbellew (commented):
as you can see from the tests, the only place where this change affects the result is when binpacking is enabled. however, several changes we've merged over the last week (including #2751 an hour ago) have already changed how binpacking behaves, so by comparison this is small potatoes.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
tgodzik: Cool! Makes sense.

tgodzik (approved):
Makes sense, thanks for the explnations! LGTM

PR #2753, opened 2021-09-20
FormatWriter: fix alignment of multiline defns

### LINKED ISSUE #2750: Blank Newlines between `def`s aren't enough to break alignment (3.0.0)
- **Version**: 3.0.0
- **Integration**: command line
- **Configuration**:
```
version = 3.0.0

maxColumn = 160
align.preset = most

align.tokens."+" = [
  {
    code = "=",
    owners=[{regex="Term.Param|Defn|Term.Assign"}]
  },
  {
    code = ":",
    owners = [{regex="Term.Param|Decl.Def"}]
  },
]

```

## Steps

Given code like this:
```scala
object Thing {

  def apply(foo: FooBaz.booberry, someInt: Int, booControl: Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz =
    apply(policy, Some(beatBytes), booControl)

  def apply(foo: FooBaz.booberry, someInt: Int)(implicit p: Parameters): FooBaz =
    apply(policy, Some(beatBytes), booControl = None)

  def apply(fooBaz: Int, someBooberry: Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz =
    apply(policy = FooBaz.booberry, Some(someInt), booControl)

  def apply(fooBaz: Int)(implicit p: Parameters): TLNode =
    apply(policy = FooBaz.booberry, Some(someInt), booControl = None)
}
```

When I run scalafmt like this, where SmallTest.scala contains the code above

```bash
scalafmt --config <config file shown above> SmallTest.scala
```

## Problem

Scalafmt formats code like this: (note added whitespace in the `def` lines)
```scala
object Thing {

  def apply(foo: FooBaz.booberry, someInt: Int, booControl: Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz =
    apply(policy, Some(beatBytes), booControl)

  def apply(foo:    FooBaz.booberry, someInt: Int)(implicit p:                            Parameters): FooBaz =
    apply(policy, Some(beatBytes), booControl = None)

  def apply(fooBaz: Int, someBooberry:        Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz =
    apply(policy = FooBaz.booberry, Some(someInt), booControl)

  def apply(fooBaz: Int)(implicit p:          Parameters): TLNode =
    apply(policy = FooBaz.booberry, Some(someInt), booControl = None)
}
```

## Expectation

** I expect that blank lines between defs, regardless of multiline options, would be enough to "stop" the alignment.**
```scala
object Thing {

  def apply(foo: FooBaz.booberry, someInt: Int, booControl: Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz =
    apply(policy, Some(beatBytes), booControl)

  def apply(foo: FooBaz.booberry, someInt: Int)(implicit p: Parameters): FooBaz =
    apply(policy, Some(beatBytes), booControl = None)

  def apply(fooBaz: Int, someBooberry: Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz =
    apply(policy = FooBaz.booberry, Some(someInt), booControl)

  def apply(fooBaz: Int)(implicit p: Parameters): TLNode =
    apply(policy = FooBaz.booberry, Some(someInt), booControl = None)
}
```

## Workaround

I've found that by putting the def body in `{...}` it does do the alignment properly:

```
object Thing {

  def apply(foo: FooBaz.booberry, someInt: Int, booControl: Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz = {
    apply(policy, Some(beatBytes), booControl)
  }

  def apply(foo: FooBaz.booberry, someInt: Int)(implicit p: Parameters): FooBaz = {
    apply(policy, Some(beatBytes), booControl = None)
  }

  def apply(fooBaz: Int, someBooberry: Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz = {
    apply(policy = FooBaz.booberry, Some(someInt), booControl)
  }

  def apply(fooBaz: Int)(implicit p: Parameters): TLNode = {
    apply(policy = FooBaz.booberry, Some(someInt), booControl = None)
  }
}
```

And by adding the `align.mutliline = false` alignment it also does it "properly" as expected above, but I actually do want multiline alignment if there is no blank newline.

Also, If I put the input `def`s and bodies on one line and get `scalafmt` to pick where the line will be too long and needs to be broken down to the next line, things are "as I expect" though not optimal from an overall code appearance:

```
object Thing {

  def apply(foo: FooBaz.booberry, someInt: Int, booControl: Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz =
    apply(policy, Some(beatBytes), booControl)

  def apply(foo: FooBaz.booberry, someInt: Int)(implicit p: Parameters): FooBaz = apply(policy, Some(beatBytes), booControl = None)

  def apply(fooBaz: Int, someBooberry: Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz =
    apply(policy = FooBaz.booberry, Some(someInt), booControl)

  def apply(fooBaz: Int)(implicit p:   Parameters): TLNode = apply(policy = FooBaz.booberry, Some(someInt), booControl = None)

}
```
## Notes

This seems somewhat related to the issue I reported in #2703 which was dismissed as user error, but I am still surprised that blank newlines "cancel" alignment in some scenarios but not all.

kitbellew:
Fixes #2750.

### DIFF
--- a/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
+++ b/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
@@ -1414,13 +1414,13 @@ object Thing {
   def apply(foo: FooBaz.booberry, someInt: Int, booControl: Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz =
     apply(policy, Some(beatBytes), booControl)
 
-  def apply(foo:    FooBaz.booberry, someInt: Int)(implicit p:                            Parameters): FooBaz =
+  def apply(foo: FooBaz.booberry, someInt: Int)(implicit p: Parameters): FooBaz =
     apply(policy, Some(beatBytes), booControl = None)
 
-  def apply(fooBaz: Int, someBooberry:        Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz =
+  def apply(fooBaz: Int, someBooberry: Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz =
     apply(policy = FooBaz.booberry, Some(someInt), booControl)
 
-  def apply(fooBaz: Int)(implicit p:          Parameters): TLNode =
+  def apply(fooBaz: Int)(implicit p: Parameters): TLNode =
     apply(policy = FooBaz.booberry, Some(someInt), booControl = None)
 }
 <<< #2750 2 no blocks, no blanks
--- a/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
+++ b/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
@@ -1445,11 +1445,11 @@ object Thing {
 object Thing {
   def apply(foo: FooBaz.booberry, someInt: Int, booControl: Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz =
     apply(policy, Some(beatBytes), booControl)
-  def apply(foo: FooBaz.booberry, someInt: Int)(implicit p: Parameters): FooBaz =
+  def apply(foo: FooBaz.booberry, someInt: Int)(implicit p: Parameters):                                             FooBaz =
     apply(policy, Some(beatBytes), booControl = None)
-  def apply(fooBaz: Int, someBooberry: Option[OptionalBooberryParams])(implicit p: Parameters): FooBaz =
+  def apply(fooBaz: Int, someBooberry: Option[OptionalBooberryParams])(implicit p: Parameters):                      FooBaz =
     apply(policy = FooBaz.booberry, Some(someInt), booControl)
-  def apply(fooBaz: Int)(implicit p: Parameters): TLNode =
+  def apply(fooBaz: Int)(implicit p: Parameters):                                                                    TLNode =
     apply(policy = FooBaz.booberry, Some(someInt), booControl = None)
 }
 <<< #2750 3 blocks, blanks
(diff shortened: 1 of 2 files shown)


scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
```diff
@@ -918,15 +918,15 @@ object a {
 }
 >>>
 object a {
-  def foo: Int = {
+  def foo: Int       = {
```
tgodzik, line 921: Should we have additional align here and in the two examples below?

If I understand the intention, we should not align if the statements are not just below one another. That's what seems to be happening in `#2750 4 blocks, no blanks` test case.

kitbellew: one of things highlighted in the submitted issue is the inconsistent handling of multiline definitions with and without curly braces (blocks vs no blocks). so that was one of the fixes.

both here and below, `align.preset = most` is used and it enables multiline alignment. since these methods are not separated by blank lines (those should interrupt alignment), they are aligned.

the difference with the case below is that includes alignment of `:` (and they do not align, hence `=` we don't get to `=`), while here there are only equals.

tgodzik: > and they do not align

Is it an issue that there is multiple `:` on the line? We do not try to align in that case?

kitbellew: the way I've implemented it _now_ is that parameter colons should only align with parameters from the same definition (that is, the align owner is the definition) whereas the align owner for the equals is the template which contains the definition.

you can only have one align owner for all tokens on a given line, so it's either colons or equals but not both.

might need to adjust the logic to figure out which should be used when both are present. give me some time, please.

kitbellew: ptal, added a commit to choose the largest align container.

tgodzik (approved):
LGTM!

PR #2756, opened 2021-09-20
RedundantBraces: restore stringInterpolation flag

### LINKED ISSUE #2754: rewrite.redundantBraces.stringInterpolation is ignored
- **Version**: 3.0.4 (observed starting with 3.0.0)
- **Integration**: CLI
- **Configuration**:

```
version = 3.0.4
# rewrite.redundantBraces.stringInterpolation = false  # should be default
```

## Steps

Given code like this:

```scala
object Test {
  val foo = 5
  val bar = true
  s"${foo}_${bar}"
}
```

```bash
scalafmt
```

## Problem

```scala
object Test {
  val foo = 5
  val bar = true
  s"${foo}_$bar"
}
```

## Expectation

```scala
object Test {
  val foo = 5
  val bar = true
  s"${foo}_${bar}"
}
```

## Notes

[Documentation says](https://scalameta.org/scalafmt/docs/configuration.html#redundantbraces):

> Default: rewrite.redundantBraces.stringInterpolation = false

But it behaves like this setting is ignored. Even if set explicitly to `false`, it still removes braces in the string interpolation.

kitbellew:
Fixes #2754.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
@@ -129,8 +129,10 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
         else if (okToReplaceBlockInSingleArgApply(t)) replace
         else if (processBlock(t)) removeToken
         else null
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM!

laughedelic:
Thanks!

PR #2757, opened 2021-09-20
Router: allowFolding for trailingCommas != never

### LINKED ISSUE #2755: Control newlines with trailing commas
The Dart code formatter `dartfmt` has a very useful feature, which I'm missing in `scalafmt` now:

If there is a trailing comma, the formatter adds line-breaks after every comma. If there isn't, it tries to put all elements into one line and wrap after `max_columns`.
(https://flutter.dev/docs/development/tools/formatting#using-trailing-commas)

You can try this behavior online here:
https://dartpad.dev/

With the following code:
```dart
// try adding and removing the trailing commas, then press the "Format" button

int multiply(
  int x,
  int y,
) =>
    x * y;

int longFunctionName(int a, int b, int c, int d, int e, int f, int g, int h,
        int i, int j, int k, int l) =>
    a + b;
```

kitbellew:
Fixes #2755.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -155,19 +155,19 @@ class Router(formatOps: FormatOps) {
         val newlineBeforeClosingCurly = decideNewlinesOnlyBeforeClose(close)
 
         val newlinePolicy = style.importSelectors match {
-          case ImportSelectors.noBinPack =>
-            newlineBeforeClosingCurly & splitOneArgOneLine(close, leftOwner)
+          case ImportSelectors.singleLine =>
+            NoPolicy
           case ImportSelectors.binPack =>
             newlineBeforeClosingCurly
-          case ImportSelectors.singleLine =>
-            SingleLineBlock(close)
+          case _ =>
+            newlineBeforeClosingCurly & splitOneArgOneLine(close, leftOwner)
         }
 
         Seq(
           Split(Space(style.spaces.inImportCurlyBraces), 0)
             .withPolicy(policy),
           Split(Newline, 1)
-            .onlyIf(style.importSelectors != ImportSelectors.singleLine)
+            .onlyIf(newlinePolicy ne NoPolicy)
             .withPolicy(newlinePolicy)
             .withIndent(style.indent.main, close, Before)
         )
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -154,8 +154,9 @@ class Router(formatOps: FormatOps) {
         )
         val newlineBeforeClosingCurly = decideNewlinesOnlyBeforeClose(close)
 
+        val mustDangleForTrailingCommas = getMustDangleForTrailingCommas(close)
         val newlinePolicy = style.importSelectors match {
-          case ImportSelectors.singleLine =>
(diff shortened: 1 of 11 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1373,6 +1384,11 @@ class Router(formatOps: FormatOps) {
               // Do whatever the user did if infix.
               Split(Space.orNL(newlines == 0), 0)
             )
+          else if (
+            style.trailingCommas.eq(TrailingCommas.keep) &&
```
tgodzik, line 1388: What about `TrailingCommas.always` shouldn't the behaviour be the same?

kitbellew: then we need to add another label. in reality, what they all mean is, if there's a newline after the last param/arg, what should be done. none of them forced that newline, before this change.

tgodzik: But currently with `TrailingCommas.always` we will remove the comma for anything shorter and add it for anything longer than a single line, which is unexpected. We should at least not remove the comma in that case.

tgodzik: I am not even talking about when it's on the same line.

Let's say we have:
```scala
apply(a, b,
)
```

with `keep` we will get:
```
apply(
  a,
  b,
)
```

but with always we will get in rewritten to:

```
apply(a, b)
```

This changes if the line is longer so at the same moment another normal method might be rewritten from:
```scala
veryLongMethodName(a, b)
```

to 

```scala
veryLongMethodName(
  a,
  b,
)
```

This can even happen in the same file, two operations doing the apposite things.

kitbellew: when i talked about "same line", i meant _after_ scalafmt chose the splits (not in the source code). if we want to force a trailing comma, we also need to force a newline, and i am going to add an explicit parameter for that.

tgodzik (approved):
LGTM!

kitbellew:
@tgodzik adding support for forcing a comma for `always` and `multiple` turned out to be a bit complex, with 10 commits. i have split them up into several pull requests (see #2758, #2759, #2760).

will not merge this one now, will update it after the other ones are merged.

kitbellew:
@tgodzik rebased on the PRs merged earlier today, implementing the logic via the `allowFolding` parameter and not just for `keep`.

scalafmt-tests/src/test/resources/trailing-commas/trailingCommasPreserve.stat
```diff
@@ -63,6 +162,61 @@ def a(b: Int, c: Int,
...
+>>>
+def a(
+    b: Int,
+    c: Int, /* comment */
+)
+<<< keep, do not fold, with spaces and comment before break, binpack
+spaces.inParentheses = true
+maxColumn = 40
+binPack.preset = true
+importSelectors = binPack
```
tgodzik, line 210:
```suggestion
```

This is probably not neccessary.

tgodzik (approved):
LGTM, just one minor comment, can be ignored.

scalafmt-tests/src/test/resources/trailing-commas/trailingCommasPreserve.stat
kitbellew: you're right, there are no imports but added anyway for consistency, in case the test is amended in the future, as the title of the test says "binpack".

PR #2758, opened 2021-09-23
Router: rule for comment left before comment right

kitbellew:
This simplifies the case when both left and right are comments. Helps with #2755.


tgodzik (approved):
LGTM!

PR #2759, opened 2021-09-23
RewriteSettings: add a trailing commas section

kitbellew:
Deprecate and copy existing value into that section. Helps with #2755.


tgodzik (approved):
LGTM

PR #2760, opened 2021-09-23
FormatWriter: fix NL search after trailing comma

kitbellew:
Helps with #2755.

### DIFF
--- a/scalafmt-tests/src/test/resources/trailing-commas/trailingCommasMultiple.stat
+++ b/scalafmt-tests/src/test/resources/trailing-commas/trailingCommasMultiple.stat
@@ -147,3 +147,93 @@ object Foo {
       foo())
   }
 }
+<<< #2755 one, comma
+maxColumn = 80
+runner.parser = source
+===
+import foo.{
+  b,
+}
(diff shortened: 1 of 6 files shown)


scalafmt-tests/src/test/resources/trailing-commas/trailingCommasMultiple.stat
```diff
@@ -147,3 +147,93 @@ object Foo {
...
+  def method(a: String,
+    b: String,
+  ) = ???
+  method(a,
+    b,
+  )
+}
+>>>
+import foo.{a, b}
+class foo(a: String, b: String) {
```
tgodzik, line 198: Shouldn't this actually have `,`  ? That seems to be happening with https://scalameta.org/scalafmt/docs/configuration.html#trailing-commas-multiple

kitbellew: if you take a look at the second commit separately, you'll see the test that is fixed by this pr.

tgodzik (approved):

PR #2789, opened 2021-10-01
FormatWriter: other counting of end marker lines

### LINKED ISSUE #2786: Line calculation rules in `rewrite.scala3.insertEndMarkersMinLines`
- **Version**: 3.0.5
- **Configuration**:
```
runner.dialect = scala3
rewrite.scala3.convertToNewSyntax = yes
rewrite.scala3.removeOptionalBraces = yes
rewrite.scala3.insertEndMarkerMinLines = 5
```

Is more close to discussion rather to the issue. 
After [trying this setting](https://github.com/scalameta/metals/pull/3165/files) on metals I found some places where end marker inserting doesn't look great.

- Could we ignore non-body lines in the following samples?
  - Method
    ```scala
    def foo(
        a1: Int,
        a2: Int
    ): Int =
      println()
      42 
    // <- end marker is added but body is 2 lines only
    ```
  - Class
    ```scala
    case class X(
      a1: Int,
      a2: Int
    ):
      def foo: Int = 42
    // <- end marker is added but body is 1 line only
    ```
- Could end marker for  `if`-`else` being omitted if the last else block is less than min lines:
  ```scala
  def foo2: Int =
    fx()
    if true then
      println()
      println()
      println()
      println()
      println()
      println()
      println()
      32
    else 42 
   // <- end marker is added
  end foo2
  ```

kitbellew:
Fixes #2786.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2953,6 +2953,22 @@ the end marker will be deleted.
 > - there are comments before the end marker, as without the end marker they
 >   would be treated as outside of the optional-braces region.
 
+### `rewrite.scala3.countEndMarkerLines`
+
+> Since v3.0.6.
+
+This flag dictates which part of the expression terminated by the end marker
+is used to calculate the span for the purposes of applying
+[`insertEndMarkerMinLines`](#rewritescala3insertendmarkerminlines) and
+[`removeEndMarkerMaxLines`](#rewritescala3removeendmarkermaxlines).
+
+- `all` (default): the entire expression
+- `lastBlockOnly`: only the last block with significant indentation relative to
+  the start of the said expression (as a replacement for the closing curly brace
+  which would have been used otherwise); for instance:
+  - in case of a class, this would be the body of the class
+  - but for an if-else, this would be just the `else` part
+
 ## Vertical Multiline
 
 Since: v1.6.0.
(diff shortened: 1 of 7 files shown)


docs/configuration.md
```diff
@@ -2953,6 +2953,22 @@ the end marker will be deleted.
...
+
+> Since v3.0.6.
+
+This flag dictates which part of the expression terminated by the end marker
+is used to calculate the span for the purposes of applying
+[`insertEndMarkerMinLines`](#rewritescala3insertendmarkerminlines) and
+[`removeEndMarkerMaxLines`](#rewritescala3removeendmarkermaxlines).
+
+- `all` (default): the entire expression
+- `lastBlockOnly`: only the last block with significant indentation relative to
```
dos65, line 2966: Wdyt about making `lastBlockOnly` a default instead of `all`?

I understand that it changes the previous behavior but in this case there aren't a lot of codebases that will be affected.

dos65 (approved):
:fire:  Thx for this fix!

docs/configuration.md
kitbellew: why would we do that? I'm introducing this option just for you but it's not the most logical option.

kitbellew:
@tgodzik ptal

docs/configuration.md
dos65: I might be wrong, but I have a strong feeling that this new one is more close to expected one from user perspective.
Anyway feel free to ignore this comment.

The only thing I'm worried is that if there will be more people like me it will save some time from questions/issues on this topic.

tgodzik:
> @tgodzik ptal

I will take a look later today!

tgodzik (approved):
Looks good!

Issue #2790, opened 2021-10-05
for() loops incorrectly idented for scala 3 and no brackets


kostaskougios:
This template is a guideline, not a strict requirement.
- **Integration**: IntelliJ,sbt
- **Configuration**:
```
version = "3.0.6"
runner.dialect = scala3
align.openParenCallSite = false
align.openParenDefnSite = false
maxColumn = 160
```

## Steps

Given code like this:
```scala
  def update(tpf: Float): Unit =
    val s = stateModifier(currentState, tpf)
    currentState = s.state
    processCommands(s.commands)
    for (scene <- scenes.all) 
      scene.updateLogicalState(tpf)
      scene.updateGeometricState()
```

When I run scalafmt like this:

```bash
sbt scalafmt
```

## Problem

Scalafmt formats code like this:
```scala
  def update(tpf: Float): Unit =
    val s = stateModifier(currentState, tpf)
    currentState = s.state
    processCommands(s.commands)
    for (scene <- scenes.all)
    scene.updateLogicalState(tpf)
    scene.updateGeometricState()

```

## Expectation
The for() loop shouldn't change 

## Workaround

adding brackets:
```scala
    for (scene <- scenes.all) {
      scene.updateLogicalState(tpf)
      scene.updateGeometricState()
    }
```

kitbellew:
i think scalameta does not support old syntax with optional braces, only `for ... do`. @tgodzik ?

kostaskougios:
thanks, yes `for scene <- scenes.all do` works

tgodzik:
@kitbellew I just checked and we support `for(...)<indentation>`, so most likely it's an issue in scalafmt.

PR #2791, opened 2021-10-06
Allow optional braces after old style while and for

tgodzik:
Fixes https://github.com/scalameta/scalafmt/issues/2790

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -2123,19 +2123,19 @@ class FormatOps(
               def splits = Some(getSplitsForIf(ft, nft, t))
               def rightBrace = blockLast(thenp)
             })
-          case t: Term.For if !nft.right.is[T.KwDo] =>
-            // unsupported except for right brace
+          case f @ Term.For(_, t) if !nft.right.is[T.KwDo] =>
             Some(new OptionalBracesRegion {
-              def owner = None
-              def splits = None
-              def rightBrace = blockLast(t.body)
+              def owner = Some(f)
+              def splits =
+                Some(getSplitsMaybeBlock(ft, nft, t))
+              def rightBrace = blockLast(t)
             })
-          case t: Term.While if !nft.right.is[T.KwDo] =>
-            // unsupported except for right brace
+          case w @ Term.While(_, t) if !nft.right.is[T.KwDo] =>
             Some(new OptionalBracesRegion {
-              def owner = None
-              def splits = None
-              def rightBrace = blockLast(t.body)
+              def owner = Some(w)
+              def splits =
+                Some(getSplitsMaybeBlock(ft, nft, t))
+              def rightBrace = blockLast(t)
             })
           case _ => None
         }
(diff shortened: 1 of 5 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2133,19 +2133,21 @@ class FormatOps(
               def splits = Some(getSplitsForIf(ft, nft, t))
               def rightBrace = blockLast(t.thenp)
             })
-          case t: Term.For if !nft.right.is[T.KwDo] =>
-            // unsupported except for right brace
+          case Term.For(_, t: Term.Block)
```
kitbellew, line 2136: looks like we now have extra conditions. is that equivalent? keep in mind that right brace method may have been used not just for formatting but for rewrite, maybe end marker, and possibly something else. can we do a quick search on usage of `rightBrace` and see if there are cases we'd forgotten to cover with tests?

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2133,19 +2133,21 @@ class FormatOps(
...
+              if !nft.right.is[T.KwDo] && isBlockStart(t, nft) =>
             Some(new OptionalBracesRegion {
-              def owner = None
-              def splits = None
```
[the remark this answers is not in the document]
kitbellew: in #2420, the last comment talks about following up with "for" and "while" after fixing in the parser, perhaps we simply never did.

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -2208,3 +2208,35 @@ def foo2: Int =
       44
    end if
 end foo2
+<<< 2790 for loops indentation
+def update(tpf: Float): Unit =
+  val s = stateModifier(currentState, tpf)
+  currentState = s.state
+  processCommands(s.commands)
+  for (scene <- scenes.all)
```
kitbellew, line 2216: can you add another loop with a single statement in the body? it looks like plain old scala2 code and should not use significant indentation. same with while.

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -2208,3 +2208,35 @@ def foo2: Int =
       44
    end if
 end foo2
+<<< 2790 for loops indentation
+def update(tpf: Float): Unit =
+  val s = stateModifier(currentState, tpf)
```
kitbellew, line 2213: optional: i don't think we need these three lines. just the loops is probably enough.

kitbellew (changes requested):

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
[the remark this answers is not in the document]
tgodzik: Och, that might be right :sweat_smile:

kitbellew:
Submitted in #2803.

PR #2800, opened 2021-10-15
Correcting dockerize job to run when a release is created

jmherbst:
I believe this step has been broken for a while. The latest tagged docker container on hub is `v2.7.5`

Looking at the actions log from latest release commit https://github.com/scalameta/scalafmt/runs/3777211161

It shows that the artifact was uploaded in the CI:release run, but the `dockerize` job was not executed.

The `dockerize` job was executed in the CI:Push run, but the artifact isn't uploaded in that run, so the `dockerize` step fails

I think this change will correct this behavior, so that when future `v` tag'd releases are created, it will upload the artifact, and then dockerize will run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
           java-version: ${{ matrix.java }}
       - run: git fetch --tags -f
       - run:
-          # for GitOps tests	
+          # for GitOps tests
           git config --global user.email "scalafmt@scalameta.org" && git config --global user.name "scalafmt"
       - run: TEST="2.12" sbt ci-test
         shell: bash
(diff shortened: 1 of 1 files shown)


kitbellew:
@tgodzik @poslegm can you please take a look?

tgodzik:
I can take a look at it next week, but I am not too familiar with the docker stuff. The change does look sensible though.

kitbellew:
> Awesome, thanks @kitbellew ! Looking forward to keeping an eye on the next release to see how this does.
> 
> P.S.: Any chance you'd be able to apply the `hacktoberfest-accepted` label to this PR for me to participate in that event?

how do i apply the label? :)

kitbellew:
@jmherbst could you please take a look:

- i pushed a new v3.0.7 tag; this resulted in two builds:
  - CI (failed): https://github.com/scalameta/scalafmt/actions/runs/1364303766 (failed)
  - Release: https://github.com/scalameta/scalafmt/actions/runs/1364303764
- then edited a release via the ui; this resulted in one build:
  - CI: https://github.com/scalameta/scalafmt/actions/runs/1364306322

So, comparing to what happened with v3.0.6, I don't see a difference. Dockerize still failed.

PR #2835, opened 2021-10-30
Scalafmt: make explicit dialect a required field

kitbellew:


tgodzik (approved):
Looks like a good idea!

PR #2841, opened 2021-11-01
IndentOperator: rename `include` to `includeRegex`

### LINKED ISSUE #2840: Cannot use include inside {} of indentOperator
- **Version**: 3.0.8
- **Integration**: IntelliJ
- **Configuration**:
```
version = 3.0.8

runner.dialect = "scala212"

maxColumn = 100

indentOperator {
    topLevelOnly = true
    exclude = """^(&&|\|\|)$"""
    include = """^.*$"""
}
```

## Steps

Format any file.

## Problem

An error is thrown:
```
Failed to load scalafmt config:
parse error: String: 12: include keyword is not followed by a quoted string, but by: '='
```

## Expectation

No error.

---

It seems like config parser ignores the fact that `include` keyword is inside `indentOperator` braces.

kitbellew:
Otherwise, it conflicts with HOCON reserved keyword. Fixes #2840.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -539,23 +539,28 @@ function {
 }
 ```
 
-#### `indentOperator.exclude`
+#### `indentOperator.excludeRegex`
 
 Defines a regular expression for excluded infix operators. If an eligible
 operator matches, it will not be indented.
 
+In v3.1.0, this parameter was renamed from `indentOperator.exclude`.
+
 ```scala mdoc:defaults
-indentOperator.exclude
+indentOperator.excludeRegex
 ```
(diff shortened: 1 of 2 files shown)


docs/configuration.md
```diff
@@ -541,23 +541,23 @@ function {
 }
 ```
 
-#### `indentOperator.exclude`
+#### `indentOperator.excludeRegex`
```
tgodzik, line 544: Although should we mention since when it's available? And the previous name?

tgodzik: Thanks!

PR #2843, opened 2021-11-02
ScalafmtConfig: require `version` in a config file

kitbellew:


scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
```diff
@@ -15,8 +15,8 @@ import org.scalafmt.util.ValidationOps
 /** Configuration options for scalafmt.
   *
   * @param version
-  *   The version of scalafmt to use for this project. Currently not used, the
-  *   plan is to use this field for the IntelliJ+sbt integrations.
+  *   The version of scalafmt to use for this project. Must match the current
+  *   binary version.
```
tgodzik, line 19: What do you mean by binary version? Scala binary version?

scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamicError.scala
```diff
@@ -48,4 +48,8 @@ object ScalafmtDynamicError {
 
   case class UnknownError(cause: Throwable)
       extends ScalafmtDynamicError("unknown error", cause)
+
+  object ConfigVersionRequired
+      extends Error("Version will be required, can't replace with default")
```
tgodzik, line 53:
```suggestion
      extends Error("Version is required, it can't be replaced with default")
```
?

tgodzik (approved):
Looks good, just two quick questions.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
kitbellew: how should i rephrase it? i meant the version of scalafmt that is running (aka, no dynamic resolution at this stage).

scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtDynamicError.scala
kitbellew: there's a nuance here. the configuration file might contain a version so the user could find this error message misleading.

what i'd like to say is, we can't use `withRespectVersion` or `withDefaultVersion` (whether or not version is present in the configuration file) because they would allow a version to be missing.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ScalafmtConfig.scala
tgodzik: just maybe `Must match the current version`?

PR #2884, opened 2021-11-11
scalafmt-dynamic: move tests into the module

kitbellew:
The tests module has extra dependencies which masked a bug introduced in the dynamic module (see #2883).


kitbellew:
@tgodzik test fail because of a bug, it is fixed in #2883. i intentionally uploaded the change before that pr is merged, to demonstrate the problem. cc @poslegm (affects sbt-scalafmt)

PR #2918, opened 2021-11-23
ScalafmtDynamic: fix handling of .md files

kitbellew:
When handling of `.md` files was added to `Scalafmt.formatCode` method (see #2563), `ScalafmtDynamic.tryFormat` should have added support for versions before 3.0.0-rc7 (which came out with that change), instead of 2.6.3 which added support for `.sbt` and `.sc`.

The logic in `ScalafmtDynamic` needs to complement (rather than mirror) the logic in `Scalafmt.formatCode`.


tgodzik (approved):
LGTM!

I think you wanted to reference https://github.com/scalameta/scalafmt/pull/2653/files instead.

PR #2920, opened 2021-11-25
ScalafmtReporter: use ScalafmtException to wrap

kitbellew:
For cases when the tool is invoked inside an IDE, this provides a better context for the error than a vanilla runtime exception.


kitbellew:
@unkarjedy borrowed this from your code

tgodzik (approved):
:rocket:

PR #2919, opened 2021-11-23
ScalafmtReflectConfig: add access to indents

kitbellew:

PR #2925, opened 2021-11-26
ScalafmtDynamic: use dependency injection

kitbellew:


tgodzik (approved):
Looks good, but could you add a bit of background on why it's needed?

kitbellew:
@tgodzik initially, attempted to follow @unkarjedy suggestion, made in #2919, to pave the way for intellij to use scalafmt-dynamic.

one small part of this commit (extracted dependencies) has helped achieve that goal, but i found that intellij uses a vastly different approach to referring to files (uses virtual files which might not be represented by an nio.Path), which for now limits the ability to adopt the rest.

wonder how scalafmt is invoked in metals, could you help me find that, @tgodzik?

tgodzik:
> wonder how scalafmt is invoked in metals, could you help me find that, @tgodzik?

It's being done here: https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala#L551

PR #2935, opened 2021-11-29
BatchFileMode: move file search logic from GitOps

kitbellew:
Since not all means to locate files actually use git, it seemed like a
suboptimal location.


tgodzik (approved):
:rocket:

PR #2962, opened 2021-12-14
Add `cross` (from SBT) to the exclude infix list

alejandrohdezma:
Example:

```sbt
addCompilerPlugin("org.typelevel" % "kind-projector" % "0.13.2" cross CrossVersion.full)
```


kitbellew:
@alejandrohdezma the default set contains a few very common operators from scala library and the rest refers to scalatest matchers. we shouldn't add sbt-specified operators here.

you can add to your config:
```
rewrite.neverInfix.excludeFilters."+" = [ "cross" ]
```

PR #3015, opened 2021-12-29
Router: fix for-do like while-do before

### LINKED ISSUE #3013: Inconsistent formatting of for/do
This Scala 3 code is formatted as follows:
```scala
class Format:
   def f =
      val received = mutable.Buffer.empty[String]
      for (m <- 0 until 10) do
         val msg = m.toString
         received.addOne(msg)

   def g =
      val received = mutable.Buffer.empty[String]
      for (m <- 0 until 10)
         do
            val msg = m.toString
            received += msg
```
There is a newline inserted before the second `do` but not before the first, and I don't understand why the behavior is not the same for both.  (If I manually insert a newline before the first `do`, Scalafmt keeps it, but what I really want is to avoid the newline before the second `do`.)  Unless I'm unaware of some arcane configuration, this seems inconsistent.

I'm running Scalafmt through IntelliJ, and my configuration is below:
```
version = 3.3.0
runner.dialect = scala3

fileOverride {
  "glob:**/*.sbt" {
    runner.dialect = sbt1
  }
}

preset = defaultWithAlign
maxColumn = 100
indent.main = 3

indentYieldKeyword = false
trailingCommas = preserve
includeNoParensInSelectChains = false
optIn.breaksInsideChains = false

binPack.literalsIncludeSimpleExpr = true

newlines {
  alwaysBeforeElseAfterCurlyIf = true
  beforeMultiline = keep
  beforeCurlyLambdaParams = multilineWithCaseOnly
  usingParamListModifierPrefer = before
  avoidForSimpleOverflow = [tooLong,punct,slc]
  avoidInResultType = true
}

spaces {
  beforeContextBoundColon = Always
  inImportCurlyBraces = true
  afterSymbolicDefs = true
}

literals {
  long=Upper
  float=Lower
  double=Lower
  hexPrefix=Lower
  scientific=Upper
}
```

kitbellew:
Follow-on to #2415. Fixes #3013.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1876,9 +1876,9 @@ class Router(formatOps: FormatOps) {
         Seq(Split(Space(style.spaces.isSpaceAfterKeyword(right)), 0))
       case FormatToken(close: T.RightParen, right, _) if (leftOwner match {
             case _: Term.If => !nextNonComment(formatToken).right.is[T.KwThen]
-            case _: Term.For => true
             case _: Term.ForYield => style.indentYieldKeyword
-            case _: Term.While => !nextNonComment(formatToken).right.is[T.KwDo]
+            case _: Term.While | _: Term.For =>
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM

PR #3016, opened 2021-12-29
Router: treat multi-line comment before NL as SLC

kitbellew:
Since we now can rewrite single-line comments as multi-line, the latter need to be treated the same, or idempotence will arise.


tgodzik (approved):
:rocket:

PR #3018, opened 2021-12-30
FormatOps: consider infix case body as top level

kitbellew:
If the case body contains more than one statement, with infix being the last one, then infix is considered to be part of a top-level Block.

Hence, we need to treat the standalone infix the same, for consistency.


tgodzik (approved):
LGTM!

PR #3031, opened 2022-01-05
RedundantBraces: add noParams option to defnBodies

### LINKED ISSUE #3026: RedundantBraces for `val`/`var`/`type` cannot be configured after #3022
```
version = 3.3.1

rewrite.redundantBraces.methodBodies = false // remove braces only in interpolations
rewrite.redundantBraces.generalExpressions = false // remove braces only in interpolations
rewrite.redundantBraces.parensForOneLineApply = true
```

## Problem

The intention of the above configuration was to remove braces only in interpolations and single-line applications, however after #3022 `val`/`var`/`type` defs braces are also modified - but without an option to disable it as for `def` defs.

## Expectation

Expected `rewrite.redundantBraces.methodBodies` to also affect `val`/`var`. Often the choice between `val` & `def` affects only performance, not semantics, so it's unexpected for braces to change when switching between them.

## Proposal

However, `methodBodies` option cannot logically affect `type` definitions. The solutions, in order of increasing complexity, could be:

* Simplistic: Rename `methodBodies` option to `definitions`. Deprecate & alias old name.

* Simple: Add `defnBodies` option alongside `methodBodies`.

* Complex: Remodel option into an array similar to `danglingParentheses.exclude`:
    ```
    rewrite.redundantBraces.exclude = [def, val, var, type, macro]
    // default: rewrite.redundantBraces.exclude = []
    ```

kitbellew:
Fixes #3026.


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -309,16 +308,21 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
...
         !isProcedureSyntax(d) &&
         !disqualifiedByUnit
 
-      case d: Defn.Var => d.rhs.exists(checkBlockAsBody(b, _))
-      case d: Defn.Val => checkBlockAsBody(b, d.rhs)
-      case d: Defn.Type => checkBlockAsBody(b, d.body)
-      case d: Defn.Macro => checkBlockAsBody(b, d.body)
-      case d: Defn.GivenAlias => checkBlockAsBody(b, d.body)
+      case d: Defn.Var =>
+        defnBodies(true) && d.rhs.exists(checkBlockAsBody(b, _))
```
tgodzik, line 317:
```suggestion
        defnBodies(noParams = true) && d.rhs.exists(checkBlockAsBody(b, _))
```
Otherwise it's a bit unclear

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -309,16 +308,21 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
...
+      case d: Defn.Val =>
+        defnBodies(true) && checkBlockAsBody(b, d.rhs)
```
tgodzik, line 319:
```suggestion
        defnBodies(noParams = true) && checkBlockAsBody(b, d.rhs)
```

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -340,6 +344,25 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
   ): Boolean =
     rhs.eq(b) && getSingleStatIfLineSpanOk(b).exists(innerOk(b))
 
+  private def defnBodies(
```
tgodzik, line 347:
```suggestion
  private def shouldRemoveDefnBraces(
```
maybe a bit better explaining ?

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -340,6 +344,25 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
...
+      noParams: => Boolean
+  )(implicit style: ScalafmtConfig): Boolean =
+    settings.defnBodies match {
+      case RedundantBracesSettings.DefnBodies.all => true
+      case RedundantBracesSettings.DefnBodies.none => false
+      case RedundantBracesSettings.DefnBodies.noParams => noParams
+    }
+
+  private def defnBodies(
```
tgodzik, line 356:
```suggestion
  private def isNoDefnParams(
```

tgodzik (approved):
Looks good! Added only some smaleer comments about the naming,

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
kitbellew: will do

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
kitbellew: ok, will explain

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -149,7 +149,7 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
               !x.tokens.exists(_.is[Token.RightArrow])) =>
         removeToken
       case t: Ctor.Secondary
-          if t.stats.isEmpty && style.rewrite.redundantBraces.defnBodies =>
+          if t.stats.isEmpty && isDefnBodiesEnabled(noParams = false) =>
```
kitbellew, line 152: added explicit `noParams =` here and everywhere else.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantBraces.scala
```diff
@@ -340,6 +348,25 @@ class RedundantBraces(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
   ): Boolean =
     rhs.eq(b) && getSingleStatIfLineSpanOk(b).exists(innerOk(b))
 
+  private def isDefnBodiesEnabled(
```
kitbellew, line 351: renamed this method from `defnBodies`, should be better explained.

i used the word `Enabled` to stress that this is a configuration check primarily, and not a complete determination that it's ok to remove braces here (since we are also calling `checkBlockAsBody`).

tgodzik (approved):
LGTM again, thanks!

PR #3036, opened 2022-01-07
FormatWriter: refactor scaladoc formatting, escape text a-la code fence

kitbellew:
Prepare for scalameta upgrade to next version (v4.4.32) which supports nested terms in scaladoc lists and tags.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -661,7 +661,8 @@ class FormatWriter(formatOps: FormatOps) {
 
         protected class WordFormatter(
             appendLineBreak: () => Unit,
-            extraMargin: String = " "
+            extraMargin: String = " ",
+            prefixFirstWord: String => String = _ => ""
         ) {
           final def apply(
               iter: WordIter,
@@ -687,9 +688,11 @@ class FormatWriter(formatOps: FormatOps) {
             val word = iter.next()
             var lines = linesSoFar
             var nextLineBeg = lineBeg
+            def firstWordPrefix = prefixFirstWord(word)
             def nextLineLength = 1 + word.length + sb.length - lineBeg
             if (atLineBeg) {
               if (needSpaceIfAtLineBeg) sb.append(' ')
+              sb.append(firstWordPrefix)
             } else if (nextLineLength <= maxLength) {
               sb.append(' ')
             } else {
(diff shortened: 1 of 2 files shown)


dos65 (commented):
Could you check the following case?
```
object a {
  /** 
   * - abc
   *   ```scala
   *   println("!!!")
   *   ```
   * - asdasd
   *
   */
}
```

Locally, after format I'm getting:
```
object a {

  /**
   *    - abc
   *  ```scala
   *  println("!!!")
   *  ```
   *    - asdasd
   */
}
```

Where `MdCodeBlock` should be indented by 3 space

kitbellew:
> Could you check the following case?
> 
> ```
> object a {
>   /** 
>    * - abc
>    *   ```scala
>    *   println("!!!")
>    *   ```
>    * - asdasd
>    *
>    */
> }
> ```
> 
> Locally, after format I'm getting:
> 
> ```
> object a {
> 
>   /**
>    *    - abc
>    *  ```scala
>    *  println("!!!")
>    *  ```
>    *    - asdasd
>    */
> }
> ```
> 
> Where `MdCodeBlock` should be indented by 3 space

@dos65 this change doesn't implement nesting yet, for that we need to release scalameta and upgrade this dependency here.

this change implements escaping of **text** which _looks_ like a code fence but isn't.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -941,10 +967,17 @@ class FormatWriter(formatOps: FormatOps) {
             text: Scaladoc.Text,
             termIndent: String
         ): Unit = {
-          // remove space as iterWords adds it
-          sb.setLength(sb.length - 1)
+          def prefixFirstWord(word: String): String = {
+            def checkFence(head: Char) =
+              (head == '`' || head == '~') && word.indexWhere(_ != head, 1) < 0
+            if (word.length >= 3 && checkFence(word(0)))
```
dos65, line 973: Maybe just `if (word.startsWith("```") || word.startsWith("~~~"))`?

dos65 (approved):
LGTM!

Just one minor comment

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: i did it because the fence should consist of all characters being the same.

but now i am thinking that you might be right; we are trying to prevent a text to be _interpreted_ as a fence, and text like `~~~scala` will definitely be parsed into fence and info string.

PR #3037, opened 2022-01-08
Router: handle then and do similar to else/yield

### LINKED ISSUE #3033: Scala 3: `newlines.source = keep` doesn't preserve line break before `then`
- **Version**: 3.2.1
- **Configuration**:
```
version = 3.2.1
runner.dialect = scala3
newlines.source = keep
```

## Steps

Given code like this:
```scala
class C:
  if true
  then 3
  else 4
```

```bash
scalafmt .
```

## Problem

```scala
class C:
  if true then 3
  else 4
```

## Expectation

I would like the formatted output to be exactly the same as the input.

### LINKED ISSUE #3034: Scala 3: `newlines.source = keep` doesn't preserve line break before `do`
- **Version**: 3.2.1
- **Configuration**:
```
version = 3.2.1
runner.dialect = scala3
newlines.source = keep
```

## Steps

Given code like this:
```scala
for x <- range
do matrix(x)(0) += matrix(x - 1)(0)
```

```bash
scalafmt .
```

## Problem

```scala
for x <- range do matrix(x)(0) += matrix(x - 1)(0)
```

## Expectation

I would like the formatted output to be the same as the input.

kitbellew:
Fixes #3033. Fixes #3034.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -1932,6 +1932,15 @@ class Router(formatOps: FormatOps) {
             .withSingleLineNoOptimal(expire, exclude, noSyntaxNL = noSyntaxNL),
           Split(Newline, 1)
         )
+      case ft @ FormatToken(_, _: T.KwThen | _: T.KwDo, _) =>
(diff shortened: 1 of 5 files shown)


tgodzik (approved):

SethTisue:
thank you!

PR #3047, opened 2022-01-09
FormatOps: unfold yield correctly and consistently

kitbellew:
We were handling it differently in some cases, and inconsistent with `while` and other `for`.


tgodzik (approved):
:rocket:

PR #3054, opened 2022-01-12
Scalameta: upgrade to 4.4.32 (scaladoc fixes)

kitbellew:


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -931,18 +931,10 @@ class FormatWriter(formatOps: FormatOps) {
...
-                  val words = x.part.iterator.map(_.syntax)
-                  words.foreach(sb.append(' ').append(_))
-                  appendBreak()
-                case Some(x: Scaladoc.Text) =>
-                  appendBreak()
-                  val tagMargin = getIndentation(2 + termIndent.length)
-                  sb.append(tagMargin)
-                  formatTextAfterMargin(x, tagMargin)
-                case _ => appendBreak()
+              appendBreak()
```
tgodzik, line 934: Will this not break for `@usecase` ?

tgodzik (approved):
Looks good, just a single question from me.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
kitbellew: it will not, @dos65 and i changed tags with a label and no description to behave like usecase and take the entire line rather than single word.

PR #3080, opened 2022-01-24
Router: implement newlines.inInterpolation logic

### LINKED ISSUE #3079: Starting with v3.3.1 scalafmt formats boilerplate code in a weird way
See typelevel/cats#4116

**Configuration**:
```
version=3.3.3
align.openParenCallSite = true
maxColumn = 120
danglingParentheses.preset = true
rewrite.rules = [AvoidInfix, SortImports, RedundantParens, SortModifiers]
docstrings = JavaDoc
newlines.afterCurlyLambda = preserve
docstrings.style = keep
docstrings.oneline = unfold

runner.dialect = scala213source3
fileOverride {
  "glob:**/*.sbt" {
    runner.dialect = sbt1
  }
  "glob:**/src/{main,test}/scala-2.12/**" {
    runner.dialect = scala212source3
  }
  "glob:**/src/{main,test}/scala-3.x/**" {
    runner.dialect = scala3
  }
}
project.excludeFilters = [
  "scalafix/*"
]

```

## Problem

Scalafmt formats code in boilerplate blocks which is not expected to happen.
Furthermore, the formatted results look pretty ugly: the formatting breaks block margins:
https://github.com/typelevel/cats/pull/4116/files?diff=unified&w=0#diff-dedd06aeff4542c5b99c68ad8ee7fecde2588c7adba68944521cda08e709aa83L153-R154

## Expectation

The boilerplate code blocks should not be formatted by default since the result code can be very different from the source template.

Or at least there should be an option to disable it in configs.

kitbellew:
Fixes #3079.

### DIFF
--- a/scalafmt-tests/src/test/resources/default/String.stat
+++ b/scalafmt-tests/src/test/resources/default/String.stat
@@ -377,13 +377,10 @@ object a {
 object a {
   s"foo1 ${bar} foo2 ${baz(bar)} foo3 ${qux(baz, bar)} foo4"
   s"""
-    |foo1 ${bar} foo2 ${baz(bar)} foo3 ${qux(
-      baz,
-      bar)} foo4
+    |foo1 ${bar} foo2 ${baz(bar)} foo3 ${qux(baz, bar)} foo4
     |""".stripMargin
   s"""
-    |fo1 ${br} fo2 ${bz(br)} fo3 ${qux(bz,
-                                       br)} fo4
+    |fo1 ${br} fo2 ${bz(br)} fo3 ${qux(bz, br)} fo4
     |""".stripMargin
 }
 <<< #3079 allow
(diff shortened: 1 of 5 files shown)


scalafmt-tests/src/test/resources/default/String.stat
```diff
@@ -360,3 +360,90 @@ object X {
...
+>>>
+object a {
+  s"foo1 ${
+            bar
+          } foo2 ${baz(bar)} foo3 ${
+                                     qux(baz, bar)
+                                   } foo4"
+  s"""
+    |foo1 ${bar} foo2 ${baz(bar)} foo3 ${qux(baz,
+                                             bar)} foo4
```
tgodzik, line 447: Shouldn't there be an additional break here? `foo4` seems to reach maxColumn (51) if I calculated correctly.

tgodzik (approved):
Looks good, though I am not sure about one example.

scalafmt-tests/src/test/resources/default/String.stat
kitbellew: ah... let me check

kitbellew: you were right. added another option, oneline, which is least likely to overflow, and mentioned in the docs the caveat about the others. please take another look.

tgodzik (approved):
LGTM!

satorg:
@kitbellew thank you for looking into it!

PR #3082, opened 2022-01-26
FormatOps: implement indentOperator.exemptScope

kitbellew:
current `topLevelOnly` was likely intended to implement https://github.com/scala-js/scala-js/blob/main/CODINGSTYLE.md#long-expressions-with-binary-operators but the result is not as described.

the new `enclosedAlone` option attempts to do that instead, but needs confirmation from scalajs developers.

### DIFF
--- a/scalafmt-tests/src/test/resources/test/IndentOperator.stat
+++ b/scalafmt-tests/src/test/resources/test/IndentOperator.stat
@@ -2956,7 +2956,7 @@ object a {
     b
   } else
     a &&
-    b
+      b
   if (
     a &&
     b
@@ -2965,12 +2965,12 @@ object a {
     b
   } else
     a &&
-    b
+      b
   if a &&
     b
   then
     a &&
-    b
+      b
   else {
     a &&
     b
@@ -2980,7 +2980,7 @@ object a {
     b
   then
     a &&
-    b
+      b
   else {
     a &&
     b
@@ -3193,24 +3193,24 @@ object a {
     b
   )
     a &&
-    b
+      b
   while (
     a &&
     b
   )
     a &&
-    b
+      b
   while a &&
     b
   do
     a &&
-    b
+      b
   while
     a &&
     b
   do
     a &&
-    b
+      b
 }
 <<< aloneArgOrBody, while
 runner.dialect = scala3
@@ -3375,7 +3375,7 @@ object a {
 object a {
   foo(
     a &&
-      b
+    b
   )
   foo {
     a &&
@@ -3389,10 +3389,10 @@ object a {
   )
   foo(
     a &&
-      b
+    b
   )(
     a &&
-      b
+    b
   )
 }
 <<< aloneArgOrBody, apply
@@ -3426,7 +3426,7 @@ object a {
 object a {
   foo(
     a &&
-      b
+    b
   )
   foo {
     a &&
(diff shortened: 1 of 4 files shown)


kitbellew:
@ekrich @sjrd please comment

scalafmt-tests/src/test/resources/test/IndentOperator.stat
```diff
@@ -2906,3 +2906,159 @@ object a {
...
+  foo {
+    a &&
+    b
+  }
+  foo(
+    a &&
+      b,
+    a &&
+      b
+  )
```
sjrd, line 3010: Theses would also be
```suggestion
  foo(
    a &&
    b,
    a &&
    b
  )
```
Even when there are multiple parameters, the binary operator chain is alone *in their actual parameter*.

sjrd (commented):
There is one thing below that is off compared to our style guide. Otherwise it looks perfect. :)

scalafmt-tests/src/test/resources/test/IndentOperator.stat
kitbellew: what's an actual parameter?

kitbellew: and if you don't need them to be enclosed (which one parameter gives you but multiple do not), why is being alone in the case body (which is what started this conversation) not included?

sjrd: An actual parameter is an argument at call site. As opposed to a formal parameter, which is the parameter at definition site.

> and if you don't need them to be enclosed (which one parameter gives you but multiple do not), why is being alone in the case body (which is what started this conversation) not included?

Hum actually being alone in a `case` body should be included as well, yes.

The "alone" condition is not really about being *enclosed*. It's about avoiding things like
```scala
val foo = {
  doSomethingFirst()
  thenReturnThis ||
  that
}
```
because there it's extremely confusing.

kitbellew: ok, now I'm confused. @ekrich, didn't you say that the recent change to infix as entire case body was not welcomed?

@sjrd any other cases? so it's not about having delimiters but being alone? then what about
```
if (...)
  a &&
  b
```
indent or not? it's alone in `then` body. or method body? or yield body?

or
```
// infix retrieves a lambda, and it's alone
(foo get
  bar)(arg)
```

sjrd: > @sjrd any other cases? so it's not about having delimiters but being alone? then what about
> 
> ```
> if (...)
>   a &&
>   b
> ```
> 
> indent or not? it's alone in `then` body. or method body? or yield body?

Our coding style would mandate `{}` around the then branch (because it has more than one line; yes I know scalafmt won't do this for me ;-) ), so this is a case that cannot happen for us. If I go beyond our own style, and we accept this as valid, I think it should be considered alone in the `then` body, yes. I think this is only way that it can decently generalize to braceless syntax anyway.

> 
> or
> 
> ```
> // infix retrieves a lambda, and it's alone
> (foo get
>   bar)(arg)
> ```

Good question. I don't think I would get in that situation, given my coding habits. I don't have a good answer for this one. I would say do whatever is simpler to implement. 🤷‍♂️

kitbellew: > Our coding style would mandate `{}` around the then branch (because it has more than one line; yes I know scalafmt won't do this for me ;-) )...

i implemented this rewrite recently: https://scalameta.org/scalafmt/docs/configuration.html#inserting-braces

tgodzik (approved):
Looks good from my perspective, though I have no stake at any particular style here.

PR #3084, opened 2022-01-27
Preserve indentation inside codeblocks

dos65:
Previsoly in some cases there were additional unwated spaces inside code
lines.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
@@ -992,21 +992,29 @@ class FormatWriter(formatOps: FormatOps) {
             termIndent: String,
             isRelative: Boolean
         ): Unit = {
+          val offsetOpt =
+            if (isRelative) None
+            else {
+              val minSpaces = code.foldLeft(Int.MaxValue) { (res, x) =>
+                val matcher = docstringLeadingSpace.matcher(x)
+                if (matcher.lookingAt()) math.min(res, matcher.end())
+                else res
+              }
+              if (minSpaces < Int.MaxValue) {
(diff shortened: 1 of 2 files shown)


scalafmt-tests/src/test/resources/test/JavaDoc.stat
```diff
@@ -3218,3 +3218,33 @@ object a {
    */
   val a = 1
 }
+<<< preserve indentation in code blocks
+docstrings.style = Asterisk
```
kitbellew, line 3222: you need to do this for all styles, otherwise it's not fully tested. AsteriskSpace is the tricky one. i would also recommend switching styles; that is, write the original in, say, SpaceAsterisk but ask to reformat as AsteriskSpace. and make sure it still works and doesn't lead to non-idempotence.

kitbellew (changes requested):

scalafmt-tests/src/test/resources/test/JavaDoc.stat
dos65: good point.
added

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatWriter.scala
```diff
@@ -992,21 +992,40 @@ class FormatWriter(formatOps: FormatOps) {
...
+              if (spaces.nonEmpty) {
+                val minSpaces = spaces.min
+                val offset = minSpaces - termIndent.length()
+                val out =
+                  if (offset > 0)
+                    (getIndentation(offset), minSpaces)
+                  else
+                    ("", minSpaces)
+                Some(out)
+              } else None
```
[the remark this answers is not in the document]
dos65: thx!

PR #3091, opened 2022-01-31
Fix early initializers

dos65:


scalafmt-tests/src/test/resources/default/Class.stat
```diff
@@ -551,3 +551,31 @@ object a {
...
+  val u: ru.type = ru
+  val mirror: u.Mirror = u.runtimeMirror(classOf[scala.meta.Tree].getClassLoader)
+} with scala.meta.internal.trees.Reflection with scala.meta.internal.tokens.Reflection
+>>>
+object CoreReflection extends {
+  val u: ru.type = ru
+  val mirror: u.Mirror =
+    u.runtimeMirror(classOf[scala.meta.Tree].getClassLoader)
+} with scala.meta.internal.trees.Reflection
+  with scala.meta.internal.tokens.Reflection
```
dos65, line 565: Without this fix it's formatted as:
```scala
  object CoreReflection
      extends {
        val u: ru.type = ru
        val mirror: u.Mirror = u.runtimeMirror(classOf[scala.meta.Tree].getClassLoader)
      }
      with scala.meta.internal.trees.Reflection
      with scala.meta.internal.tokens.Reflection
```

kitbellew: actually, i would think this new formatting is NOT desirable. why is `extends` treated differently than `with`?

how is this formatted:
```
object CoreReflection
      extends Foo(
        bar,
        baz
      )
      with scala.meta.internal.trees.Reflection
      with scala.meta.internal.tokens.Reflection
```

kitbellew (changes requested):
-1

kitbellew:
also, even if we consider this, shouldn't we use `binPack.parentConstructors` to decide whether this is appropriate.

scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -856,7 +856,9 @@ object TreeOps {
     }
 
   def getStartOfTemplateBody(template: Template): Option[Token] =
-    template.self.tokens.headOption
+    template.early.headOption
```
kitbellew, line 859: this function must return the token which starts the body of the class/trait/object, not before that

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -1884,7 +1884,18 @@ class Router(formatOps: FormatOps) {
             )
           // trait A extends B with C with D with E
           case template: Template =>
-            typeTemplateSplits(template, style.indent.withSiteRelativeToExtends)
+            if (template.early.nonEmpty) {
+              val expire = template.inits.last.tokens.last
+              Seq(
+                Split(Space, 0).withIndent(1, expire, ExpiresOn.After),
```
kitbellew, line 1890: indent 1? what does this even do?

scalafmt-tests/src/test/resources/default/Class.stat
dos65: From what I see, currently the logic is designed for the case of `extends-with` chain and then template body.
The early-init is a different case - there is a body between `extends` and `with` chain.

I think it would be better to support this construction if the required amount of changes is similar to what I proposed. That's the style how they were used + Scalafmt 2 worked worked well with it + migration to 3 triggers unwanted reformat.

kitbellew: see #3094, it does something similar but consistent with configuration. the difference is indentation.

PR #3092, opened 2022-02-01
Router: set interpolation indents correctly

kitbellew:
It was not apparent from tests because they use continuation indent of 4 in reality, but previously looked as if they had used indent 2.


tgodzik (approved):
:rocket:

PR #3094, opened 2022-02-01
FormatOps: improve binpacking parentConstructors

kitbellew:
Alternative to #3091.

### DIFF
--- a/scalafmt-tests/src/test/resources/binPack/ParentConstructors.stat
+++ b/scalafmt-tests/src/test/resources/binPack/ParentConstructors.stat
@@ -114,3 +114,144 @@ def method[T <: ColType](): DTDataset[
     with FCol with GCol with HCol
     with ICol
 ]
+<<< #3091 always
+maxColumn = 100
+runner.parser = source
+binPack.parentConstructors = always
+===
+object CoreReflection extends {
+    val u: ru.type = ru
+    val mirror: u.Mirror = u.runtimeMirror(classOf[scala.meta.Tree].getClassLoader)
+  }
+  with scala.meta.internal.trees.Reflection(
+    u,
+    mirror
+  )
+  with scala.meta.internal.tokens.Reflection
+object CoreReflection extends {
+    val u: ru.type = ru
+    val mirror: u.Mirror = u.runtimeMirror(classOf[scala.meta.Tree].getClassLoader)
+  }
+  with scala.meta.internal.trees.Reflection
+>>>
+object CoreReflection
+    extends {
+      val u: ru.type = ru
+      val mirror: u.Mirror = u.runtimeMirror(classOf[scala.meta.Tree].getClassLoader)
+    } with scala.meta.internal.trees.Reflection(
+      u,
+      mirror
+    ) with scala.meta.internal.tokens.Reflection
+object CoreReflection
+    extends {
+      val u: ru.type = ru
+      val mirror: u.Mirror = u.runtimeMirror(classOf[scala.meta.Tree].getClassLoader)
+    } with scala.meta.internal.trees.Reflection
+<<< #3091 never
+maxColumn = 100
+runner.parser = source
+binPack.parentConstructors = never
+===
+object CoreReflection extends {
+    val u: ru.type = ru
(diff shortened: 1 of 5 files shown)


scalafmt-tests/src/test/resources/binPack/ParentConstructors.stat
```diff
@@ -114,3 +114,71 @@ def method[T <: ColType](): DTDataset[
...
+    val u: ru.type = ru
+    val mirror: u.Mirror = u.runtimeMirror(classOf[scala.meta.Tree].getClassLoader)
+  }
+  with scala.meta.internal.trees.Reflection
+  with scala.meta.internal.tokens.Reflection
+>>>
+object CoreReflection extends {
+      val u: ru.type = ru
+      val mirror: u.Mirror = u.runtimeMirror(classOf[scala.meta.Tree].getClassLoader)
+    } with scala.meta.internal.trees.Reflection with scala.meta.internal.tokens.Reflection
```
dos65, line 131: it only fixes `extends` position.
early stats and `} with ..` shouldn't be indented

kitbellew: why shouldn't they be indented? everything between the beginning of a template and the opening brace of its body should be indented.

in your case,
```
... {
      val u: ru.type = ru
      val mirror: u.Mirror = u.runtimeMirror(classOf[scala.meta.Tree].getClassLoader)
    }
```
is not the body of `CoreReflection` but of some other (parent) construct. definitely has to be indented.

kitbellew: here's another example, existing behaviour:

https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/resources/binPack/ParentConstructors.stat#L49

your proposal is contrary to that.

scalafmt-tests/src/test/resources/scalajs/Class.stat
```diff
@@ -54,8 +54,7 @@ private class Encoder extends CharsetEncoder(
     }
 >>>
 object a {
-  private class Encoder
-      extends CharsetEncoder(UTF_16_Common.this, 2.0f, 2.0f,
+  private class Encoder extends CharsetEncoder(UTF_16_Common.this, 2.0f, 2.0f,
```
tgodzik, line 57: Is this change intended?

scalafmt-tests/src/test/resources/scala3/Enum.stat
```diff
@@ -129,9 +129,8 @@ enum Foo(val x: Int) {
 }
 >>>
 enum Foo(val x: Int) {
-  case Bar
-      extends Foo(1) with X
-      with Y with Z
+  case Bar extends Foo(1)
```
tgodzik, line 132: This seems wrong now. With `binPack.parentConstructors = Always` wasn't the previous state correct?

tgodzik (changes requested):
The changes to early initializers look good, but I wonder if anything should have been also updated?

scalafmt-tests/src/test/resources/scalajs/Class.stat
kitbellew: yes, same explanation is above. `always` attempts to binpack as much as possible unless there was an explicit break.

scalafmt-tests/src/test/resources/binPack/ParentConstructors.stat
dos65: If we treat that `parentConstructor` doesn't affect early init body and work only for `extends`  and with chain it wouldn't be contrary.

Let me try to fix it it properly, I didn't noticed that there is such setting at the moment when I was preparing a fix.

kitbellew: can you explain how init body is different? other than indentation, what else would you like to do?

kitbellew:
just for visual effect, here's a version which avoids indenting early ctors: in my view, the formatting is not especially readable or pleasing, the dissonance with subsequent parents is prominent:

https://github.com/kitbellew/scalafmt/commit/ce42c5e343eae5d03f8ec7958cc6ed24550792b4

dos65 (approved):
:+1:

PR #3095, opened 2022-02-01
State: correctly compute line widths in strings and interpolations with strip margin

### LINKED ISSUE #3090: Scalafmt is not idempotent
Hey there 👋 

First of all, thanks for your work, `scalafmt` is great tool ❤️ 

I've just discovered a case where `scalafmt` is not idempotent:

- **Version**: 3.4.0
- **Integration**: CLI
- **Configuration**:
```
version = 3.4.0

runner.dialect = scala213
preset = defaultWithAlign
maxColumn = 100
importSelectors = singleLine
project.git = true
optIn.breaksInsideChains = true
lineEndings = preserve
trailingCommas = preserve
```

## Steps

Given code like this:
```scala
class Test {
  val test = foo.bar.baz
    .map(thisvariablenameisreallylong =>
      fr"""UPDATE mytable
          |SET
          | column = $thisvariablenameisreallylong
          |WHERE column2 = ${foo.bar}""".stripMargin.update.run
    ).getOrElse {
      ().pure[ConnectionIO]
    }
}
```

```bash
scalafmt --debug ./Test.scala
```

## Problem

```scala
class Test {
  val test = foo.bar.baz
    .map(thisvariablenameisreallylong => fr"""UPDATE mytable
                                             |SET
                                             | column = $thisvariablenameisreallylong
                                             |WHERE column2 = ${foo.bar}""".stripMargin.update.run).getOrElse {
      ().pure[ConnectionIO]
    }
}
```

But when I run scalafmt once again:
```bash
scalafmt --debug ./Test.scala
```

```
class Test {
  val test = foo.bar.baz
    .map(thisvariablenameisreallylong =>
      fr"""UPDATE mytable
          |SET
          | column = $thisvariablenameisreallylong
          |WHERE column2 = ${foo.bar}""".stripMargin.update.run
    ).getOrElse {
      ().pure[ConnectionIO]
    }
}
``` 

## Expectation

Scalafmt should be idempotent. Meaning it could be applied multiple times without changing the result beyond the initial application. 

## Workaround
I can deactivate the scalafmt using https://scalameta.org/scalafmt/docs/configuration.html#-format-off

## Note
I just noticed this issue https://github.com/scalameta/scalafmt/issues/339. Maybe there's a fix for my particular case?

kitbellew:
Fixes #3090.
- For strings, didn't use the parameter to `.stripMargin` call to detect the right strip margin character.
- For interpolations, didn't handle strip margin at all

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/State.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/State.scala
@@ -355,6 +355,14 @@ object State {
           } else identity
           val pipe = getStripMarginChar(ft.meta.rightOwner)
           getColumnsWithStripMargin(pipe, syntax, firstNL, margin, firstLength)
+        case _: Token.Interpolation.Part =>
+          val margin: Int => Int = if (style.assumeStandardLibraryStripMargin) {
+            // 1 for '|'
+            val adjusted = 1 + indent
+            _ => adjusted
+          } else identity
+          val pipe = getStripMarginCharForInterpolate(ft.meta.rightOwner)
+          getColumnsWithStripMargin(pipe, syntax, firstNL, margin, firstLength)
         case _ =>
           val lastNewline = syntax.length - syntax.lastIndexOf('\n') - 1
           (firstLength, lastNewline)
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM!

PR #3104, opened 2022-02-03
Implement changes from Intellij Code Inspection

kitbellew:


scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -69,20 +69,20 @@ object Cli {
...
-      else if (atFile == "-") Some(Source.stdin.getLines())
-      else if (!Files.isRegularFile(Paths.get(atFile))) None
-      else Some(Source.fromFile(atFile).getLines())
+      if (atFile == arg) builder += arg // doesn't start with @
+      else if (atFile == "-") builder ++= Source.stdin.getLines()
+      else if (!Files.isRegularFile(Paths.get(atFile))) builder += arg
+      else {
+        val src = Source.fromFile(atFile)
+        try builder ++= src.getLines()
+        finally src.close()
```
tgodzik, line 82: You could also use `scala.util.Using` 
```suggestion
        builder ++= Using.resource(Source.fromFile(atFile))(_.getLines())
```

scalafmt-sysops/shared/src/main/scala/org/scalafmt/sysops/FileOps.scala
```diff
@@ -80,8 +80,11 @@ object FileOps {
   }
 
   @inline
-  private[sysops] def readAsURL(url: URL)(implicit codec: Codec): String =
-    scala.io.Source.fromURL(url).getLines().mkString("", "\n", "\n")
+  private[sysops] def readAsURL(url: URL)(implicit codec: Codec): String = {
+    val src = scala.io.Source.fromURL(url)
+    try src.getLines().mkString("", "\n", "\n")
+    finally src.close()
```
tgodzik, line 86:
```suggestion
    builder ++= Using.resource(scala.io.Source.fromURL(url))(_.getLines().mkString("", "\n", "\n"))
```
or something along these lines

tgodzik (approved):
LGTM! I just added some comments. For some of the unused stuff we could add compiler options and verify that on CI with scalafix. We do that in Metals via:
- https://github.com/scalameta/metals/blob/main/.scalafix.conf
- and https://github.com/scalameta/metals/blob/main/build.sbt#L32

scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
kitbellew: done, thank you, didn't know about this

scalafmt-sysops/shared/src/main/scala/org/scalafmt/sysops/FileOps.scala
kitbellew: done as well

kitbellew (commented):
> ... For some of the unused stuff we could add compiler options ...
> * and https://github.com/scalameta/metals/blob/main/build.sbt#L32

thank you, added the compiler options. let's do scalafix later, if you don't mind; not very experienced with it.

scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
kitbellew: it's only available for scala 2.13. added this package for 2.12:

https://index.scala-lang.org/bigwheel/util-backports/util-backports/2.1

PR #3115, opened 2022-02-08
Router: no `inInterpolation=allow` if scala3 `if`

### LINKED ISSUE #3113: Illegal Scala 3 code is produced when `align.inInterpolation = true`
- **Version**: 3.4.2
- **Integration**: VSCode & Metals
- **Configuration**:
```
runner.dialect = scala3
version = "3.4.2"
maxColumn = 80
align.inInterpolation = true
```

## Steps

Given code like this:
```scala

val s = raw"""
          ${
            if true then
              println()
              "1"
            else
              "2"
          }
          ${if false then "3" else "4"}
        """
```

When I run scalafmt from `Format Document` command in VSCode:
## Problem

```scala

val s = raw"""
          ${if true then
               println()
               "1"
             else "2"}
          ${if false then "3" else "4"}
        """
```

and the following compile error is raised by the Scala 3 compiler:

```
end of statement expected but 'else' found
```

## Expectation

I would like the formatted output to compile

kitbellew:
Fixes #3113.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -217,11 +217,16 @@ class Router(formatOps: FormatOps) {
             .withIndents(alignIndents.getOrElse(mainIndents))
             .withPolicy(decideNewlinesOnlyBeforeClose(close))
         }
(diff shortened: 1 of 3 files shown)


tgodzik (approved):
:rocket:

Atry:
Thank you!

PR #3120, opened 2022-02-10
Scalafmt: use ammonite parser to handle scripts

kitbellew:


scalafmt-tests/src/test/resources/test/Dialect.source
```diff
@@ -74,6 +74,7 @@ import mill._
 interp.repositories() =
   interp.repositories() ++ Seq(coursier.MavenRepository("https://jitpack.io"))
 
+@
```
tgodzik, line 77: Why the double `@` ?

tgodzik (approved):
Looks good, just one comment.

scalafmt-tests/src/test/resources/test/Dialect.source
kitbellew: because there are two `@` tokens in the source. old approach was splitting on one or more such tokens as delimiter, and outputting one. new one, obviously, doesn't remove tokens.

this change will not affect existing code. if it had already been handled by scalafmt, then it should have only one already.

PR #3131, opened 2022-02-19
Rewrite: use scalameta PlaceholderChecks

kitbellew:


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantParens.scala
```diff
@@ -60,9 +60,9 @@ class RedundantParens(ftoks: FormatTokens) extends FormatTokensRewrite.Rule {
 
       case _: Lit | _: Name | _: Term.Interpolate => true
 
-      case Term.Apply(f, List(b @ (_: Term.Block | _: Term.PartialFunction)))
+      case Term.Apply(_, List(b @ (_: Term.Block | _: Term.PartialFunction)))
           if b.tokens.headOption.exists(_.is[Token.LeftBrace]) =>
-        !RewriteCtx.hasPlaceholder(f)
```
tgodzik, line 65: It's actually much nicer not to need to care about the palceholder :sweat_smile:  I need to add it in Metals.

tgodzik (approved):
LGTM!

PR #3152, opened 2022-03-07
AvoidInfix: implement allowInfixInPlaceholderArg

kitbellew:


docs/configuration.md
```diff
@@ -2523,9 +2523,15 @@ This rule replaces infix expressions `a op b` with proper method calls `a.op(b)`
...
 > those which end in `:`) which would have had to be rewritten as `b.op(a)`.
 
-The rule takes the `rewrite.neverInfix` parameter group which consists of
-`includeFilters` and `excludeFilters`, two lists of regular expressions,
-which determine which operators are eligible for this rewrite.
+The rule takes the following parameters:
+
+- `rewrite.neverInfix` parameter group which consists of `includeFilters` and `excludeFilters`,
+  two lists of regular expressions, which determine which operators are eligible for this rewrite
+- (since 3.4.4) `rewrite.allowInfixPlaceholderArg` (default: `true`) will not rewrite infix
```
tgodzik, line 2530: Why would we want to special case this?

tgodzik (commented):
Not sure I understand the need for this new param. Could you link to some context?

docs/configuration.md
kitbellew: looks like historically, `_ compareTo _` was not rewritten by the `AvoidInfix` rule. probably because many cases with a placeholder can't be rewritten (because they can't be put in parentheses without changing the meaning), but a simple solo placeholder can be.

we can't force the rewrite now because it will lead to a bunch of changes, and most people don't like that. so we can either keep this case as infix (and abandon this change), or add a parameter to force the rewrite.

tgodzik (approved):
Thanks for the explanation! LGTM

PR #3163, opened 2022-03-22
Router: check formatInfix using the infix tree

kitbellew:
To upgrade to scalameta v4.5.1 (which deprecates Type.{And,Or} trees in favour of regular Type.ApplyInfix trees), we will need to add special handling of Type.ApplyInfix trees.

For that, checking newlines.afterInfix will need to be done selectively, based on the tree, hence we need to pass the tree parameter around.


tgodzik (approved):
LGTM

PR #3164, opened 2022-03-23
Scalameta: upgrade to v4.5.1

### LINKED ISSUE #3153: `indentOperator` doesn't apply to union and intersection type expressions
## Configuration ##
```
version = "3.4.3"
runner.dialect = scala3

newlines.source = keep

indentOperator.excludeRegex = "^(&|\\||&&|\\|\\|)$"
indentOperator.exemptScope = aloneArgOrBody
```

## Problem ##
Given the above configuration and given this code:

```scala
val x =
  1 |
  2 |
  3 |
  4

type T =
  Byte |
  Short |
  Int |
  Long |
  Double
```

when I run scalafmt via CLI like this: `scalafmt test.scala --stdout`

The output is:

```scala
val x =
  1 |
  2 |
  3 |
  4

type T =
  Byte |
    Short |
    Int |
    Long |
    Double
```

## Expectation
I would like the formatted output to keep its original indentation:
```scala
val x =
  1 |
  2 |
  3 |
  4

type T =
  Byte |
  Short |
  Int |
  Long |
  Double
```

However, it seems the `indentOperator` configuration only applies to value expressions and not type expressions (e.g., union types).

## Workaround
None.

kitbellew:
Also, remove handling of deprecated Type.{And,Or} trees and instead switch scala3 type infixes to regular infix handling. Fixes #3153.


scalafmt-tests/src/test/resources/newlines/source_keep.stat
```diff
@@ -4192,27 +4200,28 @@ type T = A1 & A2 | A3 & A4 | A5 & A6 /* c1 */ | /* c2 */ A7 & A8 | A9 & A10[T10]
...
-  A46 /* c1 */ | /* c2 */ A47 & A48 |
-  A49 & A50[T50] |
-  A51 & A52 | A53 & A54 | A55 &
-  A56 /* c1 */ | /* c2 */ A57 & A58 |
-  A59 & A60[T60] |
-  A61 & A62 | A63 & A64 | A65 &
-  A66 /* c1 */ | /* c2 */ A67 & A68 |
-  A69 & A70[T70]
+type T =
+  A1 & A2 | A3 & A4 | A5 & A6 /* c1 */ | /* c2 */ A7 & A8 | A9 & A10[
```
kitbellew, line 4204: while this might look bad, in reality it's because in the original code snippet, there were no breaks between infix types. for any existing code which had previously been formatted using source=keep and Type.Or trees, the modified code wouldn't change formatting because it would still preserve breaks.

scalafmt-tests/src/test/resources/newlines/source_unfold.stat
```diff
@@ -4215,76 +4215,20 @@ type T = A1 & A2 | A3 & A4 | A5 & A6 | A7 & A8 | A9 & A10[T10] |
...
-    A62 |
-    A63 &
-    A64 |
-    A65 &
-    A66 |
-    A67 &
-    A68 |
-    A69 &
-    A70[T70]
+  A1 & A2 | A3 & A4 | A5 & A6 |
```
tgodzik, line 4218: Shouldn't unfold try to break it to separate lines as previously? This looks ok, but I am not sure about the expected bahaviour in case of long infix types

scalafmt-tests/src/test/resources/scala3/UnionIntersectionType.stat
```diff
@@ -52,11 +52,9 @@ def hello(): FirstTypeInTheInterSectionType & SecondTypeInTheInterSectionType &
     val a  : FirstTypeInTheInterSectionType & SecondTypeInTheInterSectionType 
 }
 >>>
-def hello(): FirstTypeInTheInterSectionType &
-  SecondTypeInTheInterSectionType &
-  ThirdTypeInTheInterSectionType = {
-  val a: FirstTypeInTheInterSectionType &
-    SecondTypeInTheInterSectionType
+def hello()
+    : FirstTypeInTheInterSectionType & SecondTypeInTheInterSectionType & ThirdTypeInTheInterSectionType = {
```
tgodzik, line 56: This seems to now exceed 30 column limit, is this correct?

scalafmt-tests/src/test/resources/newlines/source_unfold.stat
kitbellew: no, `unfold` uses `afterInfix=many` which breaks a little more often than `fold`s `afterInfix=some` but not on every infix.

perhaps it could be improved but that would introduce changes to existing code which used `{Term,Type}.ApplyInfix` in other situations, not just `Type.{And,Or}`...

scalafmt-tests/src/test/resources/scala3/UnionIntersectionType.stat
kitbellew: correct, but the configuration uses `keep` here.

kitbellew: not for infix. infix formatting rules were added about a year ago, they are somewhat complex, and before that the formatter always kept breaks around infix operators as is, regardless of line length.

PR #3175, opened 2022-03-25
Don't throw no files error if project filters are on 

reimai:
Respecting project filters might result in an empty file set to check


kitbellew:
@reimai use `--mode any` to accomplish this (or `anygit`).

reimai:
> @reimai use --mode any to accomplish this (or anygit).

I'm using scalafmt with a list of files to check (my project is not under git version system), and it's possible that the list is entirely made of files filtered out by project filter.

kitbellew:
> > @reimai use --mode any to accomplish this (or anygit).
> 
> I'm using scalafmt with a list of files to check (my project is not under git version system), and it's possible that the list is entirely made of files filtered out by project filter.

i don't understand how that relates to my suggestion.

reimai:
oh, mode any in that case would not affect the way files are selected
well, thank you, that worked

PR #3176, opened 2022-03-25
FormatOps: handle `if` as if no `else` when `keep`

### LINKED ISSUE #3173: Newline before multi-line can't be "kept" as in source
```
version = 3.4.3

runner.dialect = scala212
docstrings.style = keep
maxColumn = 100

newlines.source = keep
newlines.beforeMultiline = keep
# newlines.forceBeforeMultilineAssign = never
```

When I run scalafmt via CLI like this: `-c scalafmt.conf test-scalafmt.scala`

## Steps

Given code like this:
```scala
class Foo {
  def foo(): Unit = {
    // foo
    val foo = if (true) {
        ""
      } else {
        "a"
      }
  }
}
```

## Problem

```scala
class Foo {
  def foo(): Unit = {
    // foo
    val foo =
      if (true) {
        ""
      } else {
        "a"
      }
  }
}
```

## Expectation

I would like the formatted output to look like this (not change the val line):
```scala
class Foo {
  def foo(): Unit = {
    // foo
    val foo = if (true) {
      ""
    } else {
      "a"
    }
  }
}

```

## Workaround

No workaround found after fiddling with settings for several hours :(

kitbellew:
Fixes #3173.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -1689,6 +1689,7 @@ class FormatOps(
     def foldedNonEmptyNonComment(
         body: Tree,
         nlSplitFunc: Int => Split,
+        isKeep: Boolean,
         spaceIndents: Seq[Indent] = Seq.empty
     ): Seq[Split] = {
       def bheadFT = tokens.getHead(body)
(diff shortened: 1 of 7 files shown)


iuliand-db:
Thanks a lot for doing this! for what it's worth, LGTM!

iuliand-db:
Is there anything we could help to get this in? I could test it locally, but I'd prefer to stick to an official pre-release version than to fork this repo.

tgodzik (approved):
I am off on vacation, but the pr looks good after a quick look!

kitbellew:
> I am off on vacation, but the pr looks good after a quick look!

thank you! where are you vacationing?

tgodzik:
> > I am off on vacation, but the pr looks good after a quick look!
> 
> thank you! where are you vacationing?

Madrid and Barcelona, unfortunately it's much colder than expected 😅

PR #3199, opened 2022-04-30
Router: for scala3, move NL before macro to after

### LINKED ISSUE #3188: Scalafmt 3.5.2 with scala3 dialect formats code that is not valid Scala 3
Related to https://github.com/scalameta/scalameta/issues/2727

We can see the issue happening here: https://github.com/scalameta/munit/pull/518

(I'm really sorry if my previous issue, linked above, didn't have enough details and inevitably caused us to end up here. :grimacing: ) 

```
version = "3.5.2"
runner.dialect = scala3
```

When I run scalafmt via CLI like this: `scalafmt -c .scalafmt.conf test.scala`

## Steps

Given code like this:
```scala
package munit.internal

import munit.Clue
import scala.quoted._
import scala.language.experimental.macros

object MacroCompat {

  trait ClueMacro {
    inline implicit def generate[T](value: T): Clue[T] = ${ clueImpl('value) }
    implicit def generate[T](value: T): Clue[T] = macro MacroCompatScala2.clueImpl
  }

}
```

## Problem

```scala
package munit.internal

import munit.Clue
import scala.quoted._
import scala.language.experimental.macros

object MacroCompat {

  trait ClueMacro {
    inline implicit def generate[T](value: T): Clue[T] = ${ clueImpl('value) }
    implicit def generate[T](value: T): Clue[T] =
      macro MacroCompatScala2.clueImpl
  }

}
```
Note the `macro MacroCompatScala2.clueImpl` on a new line.

Once formatted, scala 3.1.2 will not parse this code, yielding:
```
[error] -- [E103] Syntax Error: /home/andrew/src/scalafmttest/test.scala:12:6 ----------
[error] 12 |      macro MacroCompatScala2.clueImpl
[error]    |      ^^^^^
[error]    |      Illegal start of statement
```

## Expectation

The output format should be able to parsed by Scala 3

kitbellew:
Fixes #3188.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -2599,6 +2599,13 @@ class Router(formatOps: FormatOps) {
       case ft @ FormatToken(_: T.Comment, _, _) if ft.hasBreak =>
         splitsAsNewlines(splits)
       case ft if ft.meta.formatOff && ft.hasBreak => splitsAsNewlines(splits)
+      case FormatToken(_: T.Equals, r: T.KwMacro, _)
+          if dialect.allowSignificantIndentation =>
+        // scala3 compiler doesn't allow newline before `macro`
+        splits.map { s =>
+          if (!s.isNL) s
+          else s.withMod(Space).andPolicy(decideNewlinesOnlyAfterClose(r))
+        }
       case _ => splits
     }
   }
--- a/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
+++ b/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
@@ -3388,3 +3388,32 @@ type T =
   Short |
   Int | Long |
   Double
+<<< #3188
+maxColumn = 60
+===
+object a {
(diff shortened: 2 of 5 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -2599,6 +2599,13 @@ class Router(formatOps: FormatOps) {
       case ft @ FormatToken(_: T.Comment, _, _) if ft.hasBreak =>
         splitsAsNewlines(splits)
       case ft if ft.meta.formatOff && ft.hasBreak => splitsAsNewlines(splits)
+      case FormatToken(_: T.Equals, r: T.KwMacro, _)
+          if dialect.allowInlineMods =>
```
kitbellew, line 2603: @tgodzik not sure if this is the right way to identify scala 3, perhaps as good as any

tgodzik: I would use `allowSignificantIndentation`, because it's the cause of the issue. Without significant indentation there would be no issues

tgodzik (approved):
Just a small comment, otherwise LGTM

kitbellew:
> Just a small comment, otherwise LGTM

Done.

PR #3214, opened 2022-05-14
FormatTokens: implement isEnclosedInParens

kitbellew:
This method will apply only to parens, not just any matching delimiters, and will in the future also include those just outside the tree.


tgodzik (approved):
The change looks good! Is it only a refactoring or doe sit fix something?

kitbellew:
> The change looks good! Is it only a refactoring or doe sit fix something?

needed for upgrade to next parser version, which excludes parens from tree positions. oh, that simple change is really hard on the formatter.

PR #3230, opened 2022-05-23
RedundantBraces: handle blocks carefully

kitbellew:


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RedundantParens.scala
```diff
@@ -60,13 +60,24 @@ object RedundantParens extends Rewrite with FormatTokensRewrite.RuleFactory {
...
+            _: Term.SplicedMacroExpr | _: Term.Block =>
           Some(true)
         case _ => None
       }
     }
   }
 
+  object IsInBraces {
+    def unapply(t: Tree): Option[Boolean] = t match {
+      case _: Term.Block | _: Term.PartialFunction =>
```
tgodzik, line 72: Will this work with optional braces?

kitbellew: can you have parens around optional braces?

tgodzik: Ach, it cant's from what I know, so this is just the situation with `({ ... })` ? In that case it's fine

tgodzik (approved):

PR #1516, opened 2019-10-04
Add a space between constructor annotations and their parameter lists

dominics:
Fixes #618 which, although closed due to inactivity, describes an improvement to spacing after constructor annotations - specifically, it'd be nice to retain a space between the constructor annotation and the first parameter list of the constructor. The desired spacing looks like:

```scala
class Foo @Inject() (bar: Bar) {
```

Whereas, currently, we'd remove the space, producing:

```scala
class Foo @Inject()(bar: Bar) {
```

PR #3261, opened 2022-06-07
Config to control if a space is added after Constructor Annotation

### LINKED ISSUE #3260: Feature Request: Remove space between Annotation and Constructor Parameters

```
version = 3.5.8
runner.dialect = scala213

#preset = IntelliJ

maxColumn = 150
docstrings.style = keep
#docstrings.style = SpaceAsterisk
docstrings.wrap = no
docstrings.blankFirstLine = yes
docstrings.forceBlankLineBefore = false

importSelectors = singleLine

align.closeParenSite = true

newlines.source = keep
newlines.beforeMultiline = keep
newlines.topLevelStatementBlankLines = [
  {
    blanks { before = 0 }
  }
]

rewrite.rules = [AvoidInfix]
rewrite.neverInfix.excludeFilters."+" = [ "map", "forward", "publish" ]

align.preset = none
```

## Steps

Given code like this:
```scala
class MyClass @Inject()(parameter1: String, parameter2: String)
```

## Problem

```scala
class MyClass @Inject() (parameter1: String, parameter2: String)
```

## Expectation

```scala
class MyClass @Inject()(parameter1: String, parameter2: String)
```

## Notes

This appears to be a deliberate decision from the project maintainers (see https://github.com/scalameta/scalafmt/pull/1516, and the discussion on https://github.com/scalameta/scalafmt/issues/618), however, this goes against the current rules of my project, and I would very much like to apply Scalafmt without having to change this particular convention.

If the project maintainers would be amenable, I'll create a PR to add a configuration setting for this particular feature.

Rocketeer007:
Fixes #3260, allowing users to opt-out from the change made by #1516 for #618

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4035,6 +4035,18 @@ spaces.afterSymbolicDefs=true
 def +++(a: A): F[A]
 ```
 
+### `spaces.afterConstructorAnnotation`
+
+```scala mdoc:defaults
+spaces.afterConstructorAnnotation
+```
+
+```scala mdoc:scalafmt
+spaces.afterConstructorAnnotation=true
+---
+class A @Inject() (param1: B)
+```
+
 ## Literals
 
 > Since v2.5.0.
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Spaces.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Spaces.scala
@@ -43,6 +43,9 @@ import metaconfig._
   * @param afterSymbolicDefs
   *   If true, adds a single space after an operator method. For example:
   *   {{{def <=> [T](that: T): Boolean}}}
+  * @param afterConstructorAnnotation
+  *   If true, adds a single space after a constructor annotation. For example:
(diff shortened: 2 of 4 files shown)


kitbellew:
> This feature request is not just about avoiding the git-blame confusion. It's also about being able to enforce the currently agreed coding style.

the formatter, unfortunately, is unable to support all possible coding styles; teams deciding to adopt it usually make some changes to their agreements.

> Looking at the Scalafmt project, my assumption was that this is intended as a configurable formatter, rather than an "opinionated" one that will only enforce the coding styles endorsed by its creators.

It was conceived as opinionated but has become more flexible with time. However, the parameters we add are those which receive widespread interest -- and which convey to the users that this parameter will result in unambiguous, readable code. this last part, however, might not be true in this case.

> Please don't get me wrong - "opinionated" is a good thing, if that's your intent. Unfortunately, for me, this particular choice will currently block me from using Scalafmt on my project.

we are sorry to hear that.

@Rocketeer007 i am very, very grateful to you for figuring out how to make this modification in this codebase, and how to test it properly (in the past 2.5 years that i have been involved, this is probably only the second time this happened), so kudos to you. it makes it even harder not to accept this submission, i apologize for that, but it's.

PR #3278, opened 2022-07-21
Rewrite: check BOF/EOF when looking for LF

### LINKED ISSUE #3275: ExpandImportSelectors add newline under the first import if there's no package

```
version = 3.5.8
runner.dialect = scala3

rewrite.rules = [ExpandImportSelectors]
```

```
❯ scalafmt --version
scalafmt 3.5.8
```

When I run scalafmt via CLI like this: `scalafmt foo.scala` (where `foo.scala` is the following code in `Steps` section).

## Steps

Given code like this: `foo.scala`
```scala
import scala.collection.mutable
import scala.collection.mutable.AbstractMap
```

## Problem

```scala
import scala.collection.mutable

import scala.collection.mutable.AbstractMap
```

## Expectation

```scala
import scala.collection.mutable
import scala.collection.mutable.AbstractMap
```

do not add newline

## Workaround

I found if there's package declaration before import, there won't be a newline.

```scala
package foo
import scala.collection.mutable
import scala.collection.mutable.AbstractMap
```

## Notes

This behavior would be a problem if we use `scalafix-organize-import` as this contradicts with that.

kitbellew:
Fixes #3275.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Imports.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/Imports.scala
@@ -420,7 +420,7 @@ object Imports extends RewriteFactory {
 
     protected val groups = Array.fill(settings.numGroups + 1)(new Grouping)
 
-    protected def addToGroup(
+    protected final def addToGroup(
         group: Grouping,
         kw: String,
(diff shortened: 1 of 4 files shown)


tgodzik (approved):
LGTM!

PR #3279, opened 2022-07-25
FormatOps: check for left brace in multistat block

### LINKED ISSUE #3276: Conditional statements can get a different format when moving to scala3 dialect

```
version = "3.5.8"
runner.dialect = scala3
maxColumn = 120
align.preset = some
align.stripMargin = true
docstrings.style = Asterisk
docstrings.wrap = no
newlines.inInterpolation = avoid
lineEndings = preserve
```

scalafmt.bat FormatTest.scala

## Steps

Given code like this:
```scala
class Foo() {

  def ok: Boolean =
    if (1 == 1) {
      true
    } else
      false

  def notOK: Boolean =
    if (1 == 1) {
      println("Some code")      
      true
    } else
      false
}
```

## Problem

```scala
class Foo() {

  def ok: Boolean =
    if (1 == 1) {
      true
    } else
      false

  def notOK: Boolean =
    if (1 == 1) {
      println("Some code")
      true
    } else false
}

```

## Expectation

I would like the formatted output to be unchanged.  Note Scalafmt has moved the `false` in `notOK`

## Workaround

I've found that by going back to dialect Scala213 this is not an issue

kitbellew:
Depending on how scalameta parses the code, the left brace could be just before the multistat block (previously covered) or at the head of it. Fixes #3276.

### DIFF
--- a/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
+++ b/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
@@ -3443,3 +3443,63 @@ object a:
       val x = 4
       4 == 3
    do 5
+<<< #3276
+class Foo() {
+  def ok: Boolean =
+    if (1 == 1) {
+      true
+    } else
+      false
+
+  def notOK: Boolean =
+    if (1 == 1) {
+      println("Some code")
+      true
+    } else
+      false
+}
+>>>
+class Foo() {
+  def ok: Boolean =
+    if (1 == 1) {
+      true
+    } else
+      false
+
+  def notOK: Boolean =
+    if (1 == 1) {
+      println("Some code")
+      true
+    } else false
+}
+<<< #3276 2
+class Foo() {
+  def notOK: Boolean =
+    if (1 == 1)
+      {c; a + b}.foo
+    else
+      false
+}
+>>>
+class Foo() {
+  def notOK: Boolean =
+    if (1 == 1) { c; a + b }.foo
+    else
+      false
+}
+<<< #3276 3
+class Foo() {
+  def notOK: Boolean =
+    if (1 == 1)
+      {a + b}.foo
+      c
+    else
+      false
+}
+>>>
+test does not parse
+    if (1 == 1)
+    { a + b }.foo
+    c
+    else false
+  ^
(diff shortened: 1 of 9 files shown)


scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -3443,3 +3443,65 @@ object a:
...
+      {c; a + b}.foo
+    else
+      false
+}
+>>>
+class Foo() {
+  def notOK: Boolean =
+    if (1 == 1) { c; a + b }.foo
+    else
+      false
```
tgodzik, line 3489:
```suggestion
  def notOK: Boolean =
    if (1 == 1) { c; a + b }.foo
    else false
```
why only fold thenp ?

scalafmt-tests/src/test/resources/scala3/OptionalBraces_keep.stat
```diff
@@ -3424,3 +3424,66 @@ object a:
...
+  def notOK: Boolean =
+    if (1 == 1)
+      {c; a + b}.foo
+    else
+      false
+}
+>>>
+class Foo() {
+  def notOK: Boolean =
+    if (1 == 1) { c; a + b }.foo
```
tgodzik, line 3468: Shouldn't we keep the newline here ?

tgodzik (changes requested):

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
kitbellew: `else` part uses the existing newline. and the `then` part uses an old rule about `{`. i will copy the same tests into scala2, they are formatted exactly the same.

i will try to address that rule separately.

scalafmt-tests/src/test/resources/scala3/OptionalBraces_keep.stat
kitbellew: same comment about an old rule as above.

tgodzik (approved):
LGTM then!

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
kitbellew: @tgodzik addressed in #3282, ptal.

PR #3284, opened 2022-08-10
Newlines: add forceBeforeAssign

### LINKED ISSUE #3271: Feature Request: newline before single-statement method body
I would like my method signatures to break the line after `=` when the body is not a code block, even if it fits.

Examples:

```scala
// maxColumn = 80


def sum(a: Int, b: Int): Int =
  a + b
  
def greatest(a: Int, b: Int): Int =
  if (a > b) a
  else b
```

kitbellew:
Fixes #3271.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1746,6 +1746,19 @@ class A {
 }
 ```
 
+### `newlines.forceBeforeAssign`
+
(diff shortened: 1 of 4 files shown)


tgodzik (approved):
LGTM

PR #3299, opened 2022-08-17
FormatOps: bugfix, check else for opening brace

### LINKED ISSUE #3295: Conditional statements can get a different format when moving to scala3 dialect (mk II)
This is very similar to #3276 but the brackets are now in the second part of the conditional statement

```
version = "3.5.9"
runner.dialect = scala3
maxColumn = 120
align.preset = some
align.stripMargin = true
docstrings.style = Asterisk
docstrings.wrap = no
newlines.inInterpolation = avoid
lineEndings = preserve
```

scalafmt.bat FormatTest.scala

## Steps

Given code like this:
```scala
class Foo() {
  def ok: Boolean =
    if (1 == 1)
      true
    else {
      false
    }

  def notOK: Boolean =
    if (1 == 1)
      true
    else {
      println("Some code")
      false
    }
}
```

## Problem

```scala
class Foo() {
  def ok: Boolean =
    if (1 == 1)
      true
    else {
      false
    }

  def notOK: Boolean =
    if (1 == 1) true
    else {
      println("Some code")
      false
    }
}
```

## Expectation

I would like the formatted output to be unchanged. Note Scalafmt has moved the `true` in `notOK`

## Workaround

I've found that by going back to dialect Scala213 this is not an issue

kitbellew:
Fixes #3295.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -2162,8 +2162,9 @@ class FormatOps(
               def rightBrace = blockLast(b)
             })
           case t @ Term.If(_, thenp, _) if !nft.right.is[T.KwThen] && {
-                isTreeMultiStatBlock(thenp) || isElsePWithOptionalBraces(t) ||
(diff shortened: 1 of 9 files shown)


tgodzik (approved):
LGTM!

PR #3300, opened 2022-08-17
ScalafmtRunner: add dialectOverride

kitbellew:


scalafmt-tests/src/test/scala/org/scalafmt/config/ScalafmtConfigTest.scala
```diff
@@ -67,4 +67,20 @@ class ScalafmtConfigTest extends FunSuite {
...
+  test("dialect override") {
+    val config1 = Config
+      .fromHoconString("""
+        |runner.dialect = scala213
+        |""".stripMargin)
+      .get
+    assert(!config1.runner.getDialect.allowToplevelTerms)
+    val config2 = Config
+      .fromHoconString("""
+        |runner.dialectOverride.allowToplevelTerms = true
```
tgodzik, line 79: Is it only needed for toplevel terms ? Maybe we should have a dialect for scripts instead?

tgodzik (commented):
Looks good, but I do have a question about the usability. Isn't it harder for the user to try to figure this out instead of having relevant predefined dialects?

scalafmt-tests/src/test/scala/org/scalafmt/config/ScalafmtConfigTest.scala
kitbellew: not at all, this is just a test to show that override works.

kitbellew:
> Looks good, but I do have a question about the usability. Isn't it harder for the user to try to figure this out instead of having relevant predefined dialects?

while having predefined dialects works well, we have had complaints in the past when some change in scala3 was backported to Scala 2, or kind projector plugin had a new feature, and rather than us trying to keep up, this will allow people to enable features themselves.

tgodzik (approved):

PR #3303, opened 2022-08-19
ConvertToNewScala3Syntax bugfix: `_` only in types

### LINKED ISSUE #3302: Wildcard `_` get replaced by invalid `?` in match types

```
version = "3.5.9"
runner.dialect = "scala3"
rewrite.scala3.convertToNewSyntax = yes
rewrite.scala3.removeOptionalBraces = yes
```

```
scalafmt --config .scalafmt.conf
```

## Steps

Given code like this:
```scala
type Inner[A] = A match
  case Seq[x] => x
  case _ => A
```

Note: the problem occurs with any match type with a default/wildcard case.

## Problem

```scala
type Inner[A] = A match
  case Seq[x] => x
  case ? => A //Which doesn't compile
```

## Expectation

```scala
type Inner[A] = A match
  case Seq[x] => x
  case _ => A
```

## Workaround

Disabling placeholder rewrite should probably do the trick (until this bug is fixed) but I struggle to find the right parameter.

kitbellew:
Fixes #3302.


dos65 (approved):
:+1:

PR #3310, opened 2022-08-31
FormatOps: force config style on token, not tree

### LINKED ISSUE #3309: A little regression of undesirable line break on upgrading 3.5.8 -> 3.5.9
In the blaze, we faced an [adding a line break](https://github.com/http4s/blaze/pull/723/files#diff-75708f00aaaed216a2d15f536422faace89e41db2ada2bf708411f8cd9ae77f2R151-R153) when upgrading Scalafmt 3.5.8 -> 3.5.9. 

## Configuration ##
The config remained [the same](https://github.com/http4s/blaze/blob/main/.scalafmt.conf) (besides bumping the version).

## Steps
A simplified example:
```scala
def this(
      responseHeaderTimeout: Duration,
      idleTimeout: Duration,
  ) = this(
    responseHeaderTimeout = responseHeaderTimeout,
    idleTimeout = idleTimeout
  )(F)
```

## Problem
```scala
def this(
      responseHeaderTimeout: Duration,
      idleTimeout: Duration,
  ) = this(
    responseHeaderTimeout = responseHeaderTimeout,
    idleTimeout = idleTimeout
  )(
    F
  )
```

## Expectation
That code should remain unchanged.

Can you please confirm – this behavior is expected or is it a regression?

kitbellew:
Some trees (like Init) have multiple sets of arguments, and we need to identify the right group -- and the open paren/bracket for it. Fixes #3309.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -969,7 +969,7 @@ class FormatOps(
       ft: FormatToken,
       allowForce: => Boolean = true
   )(implicit style: ScalafmtConfig): Boolean =
-    opensConfigStyle(ft) || allowForce && forceConfigStyle(ft.meta.leftOwner)
+    opensConfigStyle(ft) || allowForce && forceConfigStyle(hash(ft.left))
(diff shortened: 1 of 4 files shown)


tgodzik (approved):
LGTM

PR #3330, opened 2022-09-30
TreeOps: improve implicit `implicit` search

kitbellew:
In scalameta v4.6.0, it will no longer be empty but point to the token which comes before the parameter starts.


tgodzik (approved):
LGTM!

Issue #3326, opened 2022-09-28
Rewrite imports with asterix to wildcard imports (or vice versa)


mlachkar:
After enabling -Xsource:3, I have started to see a mix of imports with wildcards and other with asterix. Is there already a configuration to enforce one style only ? Thank you ?

tgodzik:
Thanks for reporting! Did you try https://scalameta.org/scalafmt/docs/configuration.html#rewritescala3converttonewsyntax?

I think it might work for Scala 2 with dialect `Scala213Source3`

mlachkar-da:
Hello! I have tried it , but it only works on one direction from `_` to `*` and it also adds more rewrite which are not compatible with scala213.

tgodzik:
Which ones? We check for the specific dialect options before applying any of the rewrites. Do you have an example? That might be something to fix.

kitbellew:
> Hello! I have tried it , but it only works on one direction from `_` to `*` and it also adds more rewrite which are not compatible with scala213.

it's true that the rewrite is one direction only (from scala2 to scala3) but if the rewrites are incompatible with your given dialect, in the next release you will be able to modify the dialect parameters directly and achieve the right combination. Please also see #3332.

mlachkar-da:
@tgodzik I will take some time and post the errors I have when I run the rewrite rule. 

@kitbellew Thank you, I have seen the dialect overwrite. I think it's nice feature, but I m using `3.0.6` and upgrading would break the current style

tgodzik:
> @tgodzik I will take some time and post the errors I have when I run the rewrite rule.
> 
> @kitbellew Thank you, I have seen the dialect overwrite. I think it's nice feature, but I m using `3.0.6` and upgrading would break the current style

Actually, that's also a bug. Upgrading scalafmt shouldn't break the current style. @kitbellew takes great care to not change the default behaviour. Do you have an example of what would change?

mlachkar-da:
yes.
those are the annoying changes. 
```diff 
    logger.info(s"Auto-connecting local participants ${connectParticipants
-      .map(_._1.unwrap)} to local domains ${activeDomains.map(_.name.unwrap)}")
+        .map(_._1.unwrap)} to local domains ${activeDomains.map(_.name.unwrap)}")
```

```diff
-    )(consoleEnvironment.environment.executionContext)
+    )(
+      consoleEnvironment.environment.executionContext
+    )

[...]

PR #3338, opened 2022-10-06
fix: enable scalafmt-dynamic to download snapshot versions

tanishiking:
see: https://github.com/scalameta/scalafmt/issues/3326#issuecomment-1269587121

scalafmt failed to parse version like "3.5.9+41-60b84efb-SNAPSHOT" because when we implemented this logic back then,
we expect scalafmt to version something like 3.5.9-RC1-SNAPSHOT (not expected hash between RC version and snapshot)

This commit enables scalafmt to download snapshot version sonatype's snapshot repo is added by default.
https://github.com/scalameta/scalafmt/blob/60b84efb6ee1f424ccadd2b132677f5fd1724a7e/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/CoursierDependencyDownloader.scala#L30-L33

### DIFF
--- a/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtVersion.scala
+++ b/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtVersion.scala
@@ -7,6 +7,7 @@ case class ScalafmtVersion(
     minor: Int,
     patch: Int,
     rc: Int = 0,
+    hash: String = "",
     snapshot: String = ""
 ) extends Comparable[ScalafmtVersion] {
   private val integerRepr: Int = {
@@ -22,6 +23,7 @@ case class ScalafmtVersion(
   override def toString: String =
     s"$major.$minor.$patch" +
       (if (rc > 0) s"-RC$rc" else "") +
+      (if (hash.nonEmpty) hash else "") +
       snapshot
 
   override def compareTo(o: ScalafmtVersion): Int = {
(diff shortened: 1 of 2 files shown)


scalafmt-dynamic/src/test/scala/org/scalafmt/dynamic/ScalafmtVersionSuite.scala
```diff
@@ -22,28 +22,36 @@ class ScalafmtVersionSuite extends FunSuite {
...
-      Some(ScalafmtVersion(99, 99, 99, 99, "-SNAPSHOT"))
+      Some(ScalafmtVersion(99, 99, 99, 99, "", "-SNAPSHOT"))
     )
     assertEquals(
       ScalafmtVersion.parse("99.99.99-RC99+foobar"),
-      Some(ScalafmtVersion(99, 99, 99, 99, "+foobar"))
+      Some(ScalafmtVersion(99, 99, 99, 99, "+foobar", ""))
+    )
+    assertEquals(
+      ScalafmtVersion.parse("99.99.99+foobar"),
```
kitbellew, line 44: is this a duplicate test? (see next one)

scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtVersion.scala
```diff
@@ -34,17 +36,18 @@ case class ScalafmtVersion(
 object ScalafmtVersion {
 
   private val versionRegex =
-    """(\d{1,2})\.(\d{1,2})\.(\d{1,2})(?:-RC(\d{1,2}))?([+].*|-SNAPSHOT)?""".r
```
kitbellew, line 37: wouldn't this original regex parse the hashes as well, but simply put them in snapshot? i.e., using test below:
```
   assertEquals(
      ScalafmtVersion.parse("3.5.9+41-60b84efb-SNAPSHOT"),
      Some(ScalafmtVersion(3, 5, 9, 0, "+41-60b84efb-SNAPSHOT"))
    )
```
and if so, what does this change enable if parsing worked already?

scalafmt-dynamic/src/test/scala/org/scalafmt/dynamic/ScalafmtVersionSuite.scala
tanishiking: Oh, right, I'll remove one :+1:

scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/ScalafmtVersion.scala
tanishiking: Ahhh, you're right, it should work without my change...

tanishiking: Maybe @mlachkar uses older version of scalafmt? (older than https://github.com/scalameta/scalafmt/commit/f7568e9410efa4dff9280ef3fcd27f796af524dd)

tanishiking:
anyway closing as we found this change isn't needed 😅

PR #3343, opened 2022-10-07
ConvertToNewScala3Syntax: only simple Repeated

### LINKED ISSUE #3337: rewrite rules for scala 3 breaks compilation

```
version = 3.5.9

project.git = true

align.preset = none # never align to make the fmt more diff friendly
maxColumn = 100
runner.fatalWarnings = true
trailingCommas = multiple
# Disable scala doc wrapping (behavior changed in v3.0.0).
docstrings.wrap = no
newlines.topLevelStatementBlankLines = [
  { blanks { after = 1 }, regex = "^Import" }
]

runner.dialect = "Scala213Source3"

...
```

NB: before submitting, please confirm that the problem is observed in the
*latest published* version of the formatter! We don't publish hotfixes for older
versions, and the problem you have observed in an older version may have already
been fixed.

When I run scalafmt via CLI like this: `<command-line parameter>`

If you were not using CLI initially:
- please check using `scalafmt` CLI before submitting here (required)
- if the problem does not present via CLI, DO NOT SUBMIT
  here: this is NOT a formatter issue
  - it could be an issue with your integration (sbt-scalafmt, metals, intellij etc.)
  - instead please submit the issue in the appropriate tool's repository
  - for instance, if the tool is `sbt` with the `sbt-scalafmt` plugin: please
    submit in `sbt-scalafmt` and include scalafmt parameters from `build.sbt`
    and the `sbt` command used
  - conversely, do not submit formatting issues into other tools; you must first
    verify using `scalafmt` CLI

## Steps

Given code like this:
```scala
    Lifecycle.close(
      tokenProvider +:
        channelPerEndpoint.toList.map { case (endpoint, channel) =>
          new CloseableChannel(channel, logger, s"grpc-client-auth-$endpoint")
        }: _*
    )(logger)
```

## Problem

```scala
    Lifecycle.close(
      tokenProvider +:
        channelPerEndpoint.toList.map { case (endpoint, channel) =>
          new CloseableChannel(channel, logger, s"grpc-client-auth-$endpoint")
        }*
    )(logger)
```

## Expectation
I expect the code to compile.

kitbellew:
Fixes #3337.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/ConvertToNewScala3Syntax.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/ConvertToNewScala3Syntax.scala
@@ -1,9 +1,6 @@
 package org.scalafmt.rewrite
 
-import scala.meta.Importee
-import scala.meta.Pat
-import scala.meta.Term
-import scala.meta.Type
+import scala.meta.{Importee, Pat, Term, Type}
 import scala.meta.tokens.Token
 
 import org.scalafmt.config.ScalafmtConfig
@@ -54,7 +51,7 @@ private class ConvertToNewScala3Syntax(ftoks: FormatTokens)
 
       case _: Token.Colon if dialect.allowPostfixStarVarargSplices =>
         ft.meta.rightOwner match {
-          case _: Term.Repeated =>
+          case t: Term.Repeated if isSimpleRepeated(t) =>
             removeToken // trick: to get "*", just remove ":" and "_"
           case _ => null
         }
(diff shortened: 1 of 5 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/ConvertToNewScala3Syntax.scala
```diff
@@ -18,6 +15,7 @@ object ConvertToNewScala3Syntax extends FormatTokensRewrite.RuleFactory {
   override def create(ftoks: FormatTokens): FormatTokensRewrite.Rule =
     new ConvertToNewScala3Syntax(ftoks)
 
+  private def isSimpleRepeated(t: Term.Repeated): Boolean = t.expr.is[Term.Name]
```
tgodzik, line 18:
```suggestion
  private def isSimpleRepeated(t: Term.Repeated): Boolean = t.expr.isNot[Term.ApplyInfix]
```
?

tgodzik (changes requested):
Maybe it's enough to check if it's not an infix expression?

kitbellew:
> Maybe it's enough to check if it's not an infix expression?

ah, i didn't know. i thought `Seq("")*` was also not allowed. is it?

tgodzik:
> > Maybe it's enough to check if it's not an infix expression?
> 
> ah, i didn't know. i thought `Seq("")*` was also not allowed. is it?

It does compiler for me. With infix it's because of the operator precendence I think (not sure)

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/ConvertToNewScala3Syntax.scala
kitbellew: thank you, applied -- and added the check for `isEnclosedInParens` as well.

tgodzik (approved):
LGTM!

Issue #3344, opened 2022-10-11
Feature: Implement git ratcheting


mdedetrich:
After going through and updating various Scala OS projects old version of Scalafmt (typically the 2.x.x series) to the newest versions one of the problems that invariably happens is dealing with bag bang formats which generate massive diff's. This typically creates understandable friction amongst the OS projects, the most recent one I am dealing with is updating the Scalafmt version of an Akka fork.

For Java projects this problem is solved via a technique called ratcheting (see https://github.com/diffplug/spotless/tree/main/plugin-gradle#ratchet). The basic premise is that rather than formatting the entire codebase you instead only format the files that have changed from a specific git reference, typically this is the main upstream tracked branch (i.e. `origin/main` or `origin/master`). What this means is that you can do significant changes in the `.scalafmt.conf` and only the current file being edited/committed will be the one picked up by Scalafmt. In summary this has the following advantages

- Avoids the situation where you have a giant commit from a single person. Note only does this create misleading insights about collaboration's but also messes with tools like git blame. With git ratcheting, only the person editing the file will get triggered with a scalafmt diff, not the entire project
- Greatly increases buy in for updating between significant versions of Scalafmt for various projects. A large portion of Scala OS projects (especially the big ones) still tend to use older versions of Scalafmt and due to the nature of Scalafmt changes over time practically speaking its quite common to not be able to exactly replicate the older Scalafmt configuration in the latest version of Scalafmt.
- Decreases the hesitance in experimenting with different Scalafmt options. Due to the fact that changing Scalafmt can create massive project wide diffs, most people tend to just pick a Scalafmt configuration and stick with it for a while. Not only that, it also exacerbates the problem of projects sticking to old scalafmt versions for long periods of time.
  - This is further amplified due to how variable Scala syntax is and how configurable Scalafmt in general is (at least compared to Java)
  
## Design
On the top of I can imagine the design of such feature is to reuse the existing `project` configuration. Assuming you have `project.git = true` in your `.scalafmt.conf`, you can then enable your ratcheting by having something like.

```
project.ratchet.gitref = "origin/main" // default value null, errors if project.git != true
```

This would instruct scalafmt to calculate the files that have been changed between the current commit + unchanged changes and the `gitref` and apply that to the final file filter. If projects have various branches on remote then one can just have a different value for `project.ratchet.gitref` in those different branches.

tgodzik:
Thanks for reporting!

[...]

PR #3345, opened 2022-10-11
Update diff-branch arg saying it supports valid git ref

mdedetrich:
This is just a nice to have PR resulting from the discussion at https://github.com/scalameta/scalafmt/issues/3344. When I first read the help output for the cli arg I naively assumed it only worked with local branches, this makes it more clear and should avoid silly questions.

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
@@ -140,7 +140,8 @@ object CliArgParser {
       opt[String]("diff-branch")
         .action((branch, c) => c.copy(mode = Option(DiffFiles(branch))))
         .text(
-          "If set, only format edited files in git diff against provided branch. Has no effect if mode set to `changed`."
+          "If set, only format edited files in git diff against provided valid git reference. Has no effect if mode " +
+            "set to `changed`."
         )
       opt[Unit]("build-info")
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
LGTM! Thanks!

scalafmt-cli/src/main/scala/org/scalafmt/cli/CliArgParser.scala
```diff
@@ -140,7 +140,8 @@ object CliArgParser {
       opt[String]("diff-branch")
         .action((branch, c) => c.copy(mode = Option(DiffFiles(branch))))
         .text(
-          "If set, only format edited files in git diff against provided branch. Has no effect if mode set to `changed`."
+          """If set, only format edited files in git diff against provided branch. Has no effect if mode set to `changed`;
+            |supports both local and remote branches.""".stripMargin
```
kitbellew, line 144: it doesn't support local or remote branches, it supports any valid git reference -- because it invokes an actual `git` binary. so this clarification is no less inaccurate.

i'd either change `against provided branch` to `against provided valid git reference`, or abandon this commit.

kitbellew (changes requested):

kitbellew:
@mdedetrich i realized that there are additional confusing aspects (for instance, `-diff` option should NOT be used along with `-diff-branch`, as they conflict with each other). let me take this over and send a more comprehensive commit. thanks for your help!

PR #3351, opened 2022-10-24
PolicyOps: unindent correctly in binpacking

kitbellew:
Apply it only if there are no breaks before start of exclude range.


kitbellew:
@mdedetrich fixes inconsistent indentation shown here: https://github.com/mdedetrich/akka-apache/compare/a42307861c9e37a604243435c321680dba9d33d9..95aec89cd3ae00e55a053cd38fb3efe3e34239a0#diff-4d5075641c71bbb5c06c066e2c568cad848505c358e7867a0f52bb96fb943c29

the fix is not to reduce the indentation of `Behaviours` but _not_ to reduce the indentation of the subsequent block.

mdedetrich:
Thanks for looking into this!

tgodzik (approved):
LGTM!

PR #3353, opened 2022-10-28
Router: always skip shebang line

### LINKED ISSUE #3267: Support shebang in .scala files with the Scala 3 dialect
```
version = 3.5.8
runner.dialect = scala3
```

## Steps

Given code like this:
```scala
#!/usr/bin/env scala

@main def run() = println("hello world!")
```

## Problem

```
.../main.scala:1: error: [dialect scala3] illegal start of definition identifier
#!/usr/bin/env scala
^
```

## Expectation

The file is legal scala3 (can be run with `scala main.scala` and compiled with `scalac main.scala`), and should be formatted while ignoring the first line.

## Notes

There seems to be supported for ammonite scripts, or rather anything with a `.sc` file ending

kitbellew:
Fixes #3267.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -6,7 +6,6 @@ import org.scalafmt.config.{ImportSelectors, Newlines, ScalafmtConfig, Spaces}
 import org.scalafmt.internal.ExpiresOn.{After, Before}
 import org.scalafmt.internal.Length.{Num, StateColumn}
 import org.scalafmt.internal.Policy.NoPolicy
-import org.scalafmt.sysops.FileOps
 import org.scalafmt.util._
 import org.scalameta.FileLine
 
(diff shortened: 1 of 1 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
```diff
@@ -67,7 +67,7 @@ object Scalafmt {
   ): Formatted.Result = {
     def getStyleByFile(style: ScalafmtConfig) = {
       val isSbt = FileOps.isAmmonite(filename) || FileOps.isSbt(filename) ||
-        FileOps.isMarkdown(filename)
+        FileOps.isMarkdown(filename) || code.startsWith("#!")
```
tgodzik, line 70: Is this the line that would allow toplevel statements like in sbt? Though Scala 3 only allows toplevel definitions and not statements, I don't think we need to change anything since it's included in the dialect.

kitbellew: yes, that's the line. it doesn't change the dialect to "sbt" but simply enables allowTopLevelStatements.

on a different note: perhaps the formatter should probably always allow top level since its job is not to validate code but to format it 😀

tgodzik: Maybe? That shouldn't influence the formatting at all. We can use a more inclusive dialect.

tgodzik: But anyway, I don't think we need to enable toplevel statements when shebang is present. I doesn't really change the Scala 3 behaviour. That's just 5 five cents 😅

kitbellew: ok, let me split these two up.

tgodzik (approved):
LGTM, thanks!

PR #3355, opened 2022-10-30
Router: always add space between `(` and comment

kitbellew:
That is done in some places and missed in others; and there's always a space between a comment and a `)`.


tgodzik (approved):
LGTM!

PR #3395, opened 2022-11-28
FormatOps: enclosing parens owned by enclosed tree

kitbellew:
Regardless of how `scalameta` treats enclosing parens (whether owned by the tree they enclose, or by its parent), let's assign the enclosed tree as the owner. Exceptions include cases where parens are an integral part of the parent (tuple, if/while conditions etc.)


tgodzik (approved):
LGTM!

PR #3398, opened 2022-12-04
TreeOps: consistency for apply/infix lookalikes

kitbellew:
Use an equivalent rule for Pat.ExtractInfix as for Term.ApplyInfix; similarly, treat arguments to Init the same as Term.Apply.

### DIFF
--- a/scalafmt-tests/src/test/resources/newlines/source_keep.stat
+++ b/scalafmt-tests/src/test/resources/newlines/source_keep.stat
@@ -1263,7 +1263,8 @@ object WrapperToHaveStatTestCaseParserWorking {
       someInt = 5
     ) @annot3
     @deprecated(value =
-      "bar") def bar2 = 1
+      "bar"
+    ) def bar2 = 1
   }
 
   @annot @annot2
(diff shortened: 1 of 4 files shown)


scalafmt-tests/src/test/resources/scalajs/Apply.stat
```diff
@@ -925,7 +925,8 @@ object a {
 >>>
 object a {
   a match {
-    case foo bar (baz @ qux) =>
+    case foo bar (
```
kitbellew, line 928: the break here is needed, as previous formatting exceeded `maxColumn`

scalafmt-tests/src/test/resources/newlines/source_keep.stat
```diff
@@ -1327,7 +1328,8 @@ object WrapperToHaveStatTestCaseParserWorking {
       someInt = 5
     ) @annot3
     @deprecated(value =
-      "bar") def bar2 = 1
+      "bar"
```
kitbellew, line 1331: break here is needed because of dangling parentheses setting, and consistent with the expression just above.

tgodzik (approved):
LGTM!

PR #3405, opened 2022-12-10
Native: use 22.3.0, switch to setup-graalvm

kitbellew:


kitbellew:
PTAL, an attempt to fix recent build failures blocking any commits to this repo.

@olafurpg the sbt-native-image plugin might need to be updated to support `musl` since the earlier approach no longer works.

I don't know if the proposed change still works in standalone mode, but for `isCI` case the `setup-graalvm` step is used to download and set up the tooling.

olafurpg (approved):

PR #3426, opened 2022-12-27
Don't use deprecated scalameta methods

kitbellew:


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2073,14 +2073,16 @@ class FormatOps(
           style: ScalafmtConfig
       ): Option[OptionalBracesRegion] =
         ft.meta.leftOwner match {
-          case ParamClauseParent(t @ Defn.ExtensionGroup(_, _, b: Term.Block))
-              if isBlockStart(b, nft) =>
+          case ParamClauseParent(
+                t @ Defn.ExtensionGroup.internal.Impl(_, b: Term.Block)
+              ) if isBlockStart(b, nft) =>
```
tgodzik, line 2078:
```suggestion
          case ParamClauseParent(
                t : Defn.ExtensionGroup
              ) if isBlockStart(t.body, nft) =>
```
I would avoid using the internal Impl classes since they might not be binary compatible.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2073,14 +2073,16 @@ class FormatOps(
...
             Some(new OptionalBracesRegion {
               def owner = Some(t)
               def splits = Some(getSplitsMaybeBlock(ft, nft, b))
               def rightBrace = blockLast(b)
             })
-          case t @ Term.If(_, thenp, _) if !nft.right.is[T.KwThen] && {
+          case t @ Term.If.internal.Impl(_, thenp, _, _)
```
tgodzik, line 2084:
```suggestion
          case t : Term.If
```
and then just use t.thenp

tgodzik, line 2084:
```suggestion
          case t : Term.If
```
and then just use t.thenp.

Also in other places in the file

tgodzik (changes requested):
Could we not use `.internal.Impl` especially when it can be avoided.

scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
kitbellew: will do. however, why would we need (or expect) binary compatibility here? for every upgrade of the parser, we explicitly modify the formatter. the whole point to add those `internal.Impl.unapply` methods was for a case like this.

tgodzik: Hmm... binary compat might not be a problem yeah. Maybe I went overeager, but I still think it's a bit easier to just annotate type

tgodzik (approved):
LGTM

PR #3429, opened 2023-01-05
Documentation: Remove troubleshooting from FAQ title

drozd-A:
This page seems more like an FAQ page rather than troubleshooting. This PR renames the FAQ / Troubleshooting page to only FAQ.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,6 +1,6 @@
 ---
 id: faq
-title: FAQ / Troubleshooting
+title: FAQ
 ---
 
 ## Why not Scalariform?
--- a/website/i18n/en.json
+++ b/website/i18n/en.json
@@ -18,7 +18,7 @@
         "title": "Contributing to the website"
       },
       "faq": {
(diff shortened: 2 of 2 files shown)


docs/faq.md
```diff
@@ -1,6 +1,6 @@
 ---
 id: faq
-title: FAQ / Troubleshooting
+title: FAQ
```
kitbellew, line 3: i am not sure why @gabro chose this title. perhaps faq is about "how does the formatter do something?" whereas this document also contains a lot of "why does the formatter do something?". it's possible that people who have this kind of question might not naturally select the faq page to look for an answer. just a theory.

perhaps the original author could help review.

gabro: That particular title was migrated from the old website (see https://web.archive.org/web/20170320124508/https://scalameta.org/scalafmt) so the original author would be @olafurpg.
That said, I don't have any particular argument against the rename

kitbellew:
let's reopen if consensus emerges

PR #3446, opened 2023-01-21
FormatOps: handle fewer braces and refined types

### LINKED ISSUE #3447: Incorrect code generated when formatting valid code using fewer braces syntax
## Configuration ##
```
version = 3.7.0
runner.dialect = scala3
```

When I run scalafmt via CLI like this: `scalafmt --stdin`

## Steps

Given code like this:
```scala
def sum(xs: List[Int]): Int =
  xs.foldLeft(0):
    case (a, b) => a + b
```

## Problem

```scala
def sum(xs: List[Int]): Int =
  xs.foldLeft(0):
  case (a, b) => a + b
```

## Expectation

```scala
def sum(xs: List[Int]): Int =
  xs.foldLeft(0):
    case (a, b) => a + b
```

## Workaround

Use braces instead of indentation for the last parameter to `foldLeft`:
```scala
def sum(xs: List[Int]): Int =
  xs.foldLeft(0) {
    case (a, b) => a + b
  }
```

## Notes

Only fails on Scala 3.3 onwards (I used 3.3.0-RC1-bin-20230118-9337bd6-NIGHTLY) since the original syntax is only valid with the [Fewer Braces-feature](https://docs.scala-lang.org/sips/fewer-braces.html) (which has been [merged](https://github.com/lampepfl/dotty/pull/16297) into Scala 3.3)

kitbellew:
Fixes #3447.

### DIFF
--- a/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
+++ b/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
@@ -3598,3 +3598,189 @@ object a {
   val a = test("" +: Seq("").map { _ + "" }: _*)
   val a = test(("" +: Seq("").map { _ + "" })*)
 }
+<<< coloneol in fewer braces 1
+maxColumn = 80
+===
+object a:
+  def f(): Unit =
+    List(1, 2, 3).foldLeft(1):
+      case (a, b) => a + b
+>>>
+test does not parse
+object a:
+   def f(): Unit =
+     List(1, 2, 3).foldLeft(1):
+     case (a, b) => a + b
+     ^
+<<< coloneol in fewer braces 1, main > sig
+maxColumn = 80
+indent.main = 4
+===
+object a:
+  def f(): Unit =
+    List(1, 2, 3).foldLeft(1):
+      case (a, b) => a + b
+>>>
+test does not parse
+object a:
+   def f(): Unit =
+       List(1, 2, 3).foldLeft(1):
+       case (a, b) => a + b
+       ^
+<<< coloneol in fewer braces 2
+maxColumn = 80
+===
+object a:
+  def f(): Unit =
+    List(1, 2, 3).foo: a =>
+      a + 2
+    .foo: a =>
+      2 + a
+    .apply:
+      12 + 3
+>>>
+test does not parse
+object a:
+   def f(): Unit =
+     List(1, 2, 3)
+       .foo: a => a + 2
+               ^
+<<< coloneol in fewer braces 2, main > sig
+maxColumn = 80
+indent.main = 4
+===
+object a:
+  def f(): Unit =
+    List(1, 2, 3).foo: a =>
+      a + 2
+    .foo: a =>
+      2 + a
+    .apply:
+      12 + 3
+>>>
+test does not parse
+object a:
+   def f(): Unit =
+       List(1, 2, 3)
+           .foo: a => a + 2
+                   ^
+<<< coloneol in fewer braces 3
+maxColumn = 80
+===
+object a:
+  def f(): Unit =
+    List(1, 2, 3).foo:
+       case a : Int =>
+       case _ =>
+    otherTerm()
+>>>
+test does not parse
+object a:
+   def f(): Unit =
+      List(1, 2, 3).foo:
+      case a: Int =>
+      ^
+<<< coloneol in fewer braces 3, main > sig
+maxColumn = 80
+indent.main = 4
+===
+object a:
+  def f(): Unit =
+    List(1, 2, 3).foo:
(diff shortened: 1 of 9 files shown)


scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -3598,3 +3598,35 @@ object a {
   val a = test("" +: Seq("").map { _ + "" }: _*)
   val a = test(("" +: Seq("").map { _ + "" })*)
 }
+<<< coloneol in fewer braces
+maxColumn = 80
+===
+object a:
```
tgodzik, line 3604: Could you also maybe add two more tests such as:
```scala
object a:
  def f(): Unit =
    List(1, 2, 3).foo: a =>
      a + 2
    .apply:
      12 + 2
```

and maybe:
```scala
object a:
  def f(): Unit =
    List(1, 2, 3).foo:
       case a : Int =>
       case _ =>
   otherTerm()
      
```

kitbellew: thanks for the suggestion! also handled `.match<outdent>`.

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -3598,3 +3598,84 @@ object a {
...
+      case _ => a + 2
+    .apply:
+      12 + 3
+>>>
+object a:
+   def f(): Unit =
+     List(1, 2, 3).match
+        case _ => a + 2
+       .apply:
+          12 + 3
```
tgodzik, line 3661: Shouldn't it be:
```suggestion
     List(1, 2, 3).match
        case _ => a + 2
     .apply:
          12 + 3
```

scalafmt-tests/src/test/resources/scala3/OptionalBraces_keep.stat
```diff
@@ -3580,3 +3580,83 @@ object a {
...
+      a + 2
+    .apply:
+      12 + 3
+>>>
+object a:
+   def f(): Unit =
+     List(1, 2, 3).foo: a =>
+        a + 2
+       .apply:
+          12 + 3
```
tgodzik, line 3610:
```suggestion
   def f(): Unit =
     List(1, 2, 3).foo: a =>
        a + 2
     .apply:
          12 + 3
```
?

scalafmt-tests/src/test/resources/scala3/OptionalBraces_keep.stat
```diff
@@ -3580,3 +3580,83 @@ object a {
...
+      case _ => a + 2
+    .apply:
+      12 + 3
+>>>
+object a:
+   def f(): Unit =
+     List(1, 2, 3).match
+        case _ => a + 2
+       .apply:
+          12 + 3
```
tgodzik, line 3642:
```suggestion
   def f(): Unit =
     List(1, 2, 3).match
        case _ => a + 2
     .apply:
          12 + 3
```

scalafmt-tests/src/test/resources/scala3/OptionalBraces_keep.stat
kitbellew: it comes from the rule which adds indent to first NL before a dot (so that it's indented relative to the beginning of the full expression), and your proposal will be inconsistent with the formatting so far:
```
   // short maxColumn
   def f(): Unit =
     List(1, 2, 3).foo(a =>
       a + 2)
       .apply(
         12 + 3)
   // longer maxColumn
   def f(): Unit =
     List(1, 2, 3).foo(a => a + 2)
       .apply(12 + 3)
   // there's no NL before a dot
   def f(): Unit =
     List(1, 2, 3).foo { a =>
       a + 2
     }.apply {
       12 + 3
     }
   // there's a NL before a dot
   def f(): Unit =
     List(1, 2, 3).foo { a =>
       a + 2
     } //
       .apply {
         12 + 3
       }
```
i didn't like this formatting for `fold` (no proper unindent before select) and forced NL before the first dot if there's a second select. i could do the same for `classic` (but not `keep`).

i could try formatting it like this if you prefer:
```
   // short maxColumn
   def f(): Unit =
     List(1, 2, 3).foo: a =>
         a + 2
       .apply:
         12 + 3)
   // but, should we still do this?
   def f(): Unit =
     List(1, 2, 3).foo: a =>
       a + 2
```

tgodzik: Ach right, I didn't realize it.

tgodzik (approved):
LGTM!

PR #3450, opened 2023-02-06
docs: add a example of how to use sbt scalafmtOnly

robsonpeixoto:
The documentation do not keep clear that the `sbt` parameter must be a single string.

ref: https://github.com/scalameta/sbt-scalafmt/issues/253#issuecomment-1419413773


kitbellew:
@robsonpeixoto thank you for this submission. however, since this applies to absolutely every `sbt` command which takes parameters (for instance, `testOnly`) and is not limited to this plugin, I will close it.

PR #3451, opened 2023-02-09
FormatOps: add extra indent to fewer-braces apply

### LINKED ISSUE #3448: `newlines.source = keep` (Scalafmt 3.7.1) and fewerBraces cause incorrect and non-idempotent formatting
example code
```scala
//> using scala "3.3.0-RC2"

def Block(f: => Int): Int = f

def test(): String =
  Block:
    2 + 2
  .toString
```

config 
```
version = 3.7.1
runner.dialect = scala3
newlines.source = keep
```

format once
```scala
def test(): String =
  Block:
    2 + 2
    .toString
```

format twice:
```scala
def test(): String =
  Block:
    2 + 2
      .toString
```

`newlines.source = fold,unfold` are also broken with different results

Edit: my full example also seems to have the problem without setting `newlines.source`, but the minimal reproduction here does not seem to exhibit that 🤔

kitbellew:
Fixes #3448.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -1418,19 +1418,20 @@ class FormatOps(
     } || ft.right.is[soft.KwUsing]
 
   @tailrec
-  final def findPrevSelect(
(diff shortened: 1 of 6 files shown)


tgodzik (approved):
LGTM

Issue #3466, opened 2023-02-16
Handle interleaved definitions post-scalameta-4.7.4


kitbellew:

SethTisue:
just a note that I'm interested in seeing scalafmt move to current Scalameta, since in the Scala 2 community build we can only have a single Scalameta version, and metals and scalafix have moved on to newer versions.

Thus, once Scalafmt gets on 4.7, then I'll be able to test newer metals changes in the community build.

PR #3478, opened 2023-02-20
FormatOps vertical multiline: explicit implicits

kitbellew:
Specifically, implicitParamListModifierForce=after should not apply when `implicit` is explicit rather than applies once to the entire group. Helps with #3466.

### DIFF
--- a/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
+++ b/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
@@ -4277,17 +4277,14 @@ object a:
      )(b: B,
        bs: B*
      )(implicit
-       implicit
-       c: C) {}
+       implicit c: C) {}
    class F(
        a: A,
        as: A*
      )(b: B,
        bs: B*
-     )(implicit
-       c: C,
-       implicit
-       d: D) {}
+     )(implicit c: C,
+       implicit d: D) {}
    class F(
        a: A,
        as: A*
@@ -4296,8 +4293,7 @@ object a:
      )(implicit
        c: C,
        d: D,
-       implicit
-       e: E) {}
+       implicit e: E) {}
 <<< interleaved, vertical multiline, implicitParamListModifierForce = [before]
 maxColumn = 40
 verticalMultiline.atDefnSite = true
@@ -4381,18 +4377,15 @@ object a:
        bs: B*
      )(
        implicit
-       implicit
-       c: C) {}
+       implicit c: C) {}
    class F(
        a: A,
        as: A*
      )(b: B,
        bs: B*
      )(
-       implicit
-       c: C,
-       implicit
-       d: D) {}
+       implicit c: C,
+       implicit d: D) {}
    class F(
        a: A,
        as: A*
@@ -4402,8 +4395,7 @@ object a:
        implicit
        c: C,
        d: D,
-       implicit
-       e: E) {}
+       implicit e: E) {}
 <<< interleaved, short
 maxColumn = 23
 ===
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -1233,9 +1233,10 @@ class FormatOps(
       case Decision(FormatToken(LeftParenOrBracket(), _, m), ss)
           if allParenOwners.contains(m.leftOwner) =>
         ss.filter(!_.isActiveFor(SplitTag.VerticalMultilineSingleLine))
-      case Decision(ftd @ FormatToken(soft.ImplicitOrUsing(), _, _), _)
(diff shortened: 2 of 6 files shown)


scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -4277,17 +4277,14 @@ object a:
      )(b: B,
        bs: B*
      )(implicit
-       implicit
-       c: C) {}
+       implicit c: C) {}
```
tgodzik, line 4280: Actually, how does it parse? Two implicits in a row doesn't parse for me :thinking:

kitbellew: well, then we should change our scalameta parser :) or the grammar:

under https://dotty.epfl.ch/docs/internals/syntax.html#type-and-value-parameters
```
ClsParamClauses   ::=  {ClsParamClause} [[nl] ‘(’ [‘implicit’] ClsParams ‘)’]
...
ClsParam          ::=  ...
                       [{Modifier} (‘val’ | ‘var’) | ‘inline’] Param
```
and under https://dotty.epfl.ch/docs/internals/syntax.html#bindings-and-imports
```
Modifier          ::=  LocalModifier
...
LocalModifier     ::=  ...
                    |  ‘implicit’
```
so nothing really prevents multiple implicits. I don't know if `{...}` means multiple _different_ ones, but even in that case the first parameter is definitely covered.

tgodzik: I think the grammar then. Both the compiler and scalameta fail to parse it.

kitbellew: wait, if scalameta did fail to parse it, then scalafmt wouldn't be able to format it. perhaps a different dialect?

kitbellew: in any case, this specific change doesn't really touch the multiple implicits logic, it's just _some_ of the tests which are affected. unless, of course, the compiler doesn't allow explicit implicits all of a sudden, like:
```
...(
  implicit a: A,
  implicit b: B
)
```

tgodzik: Ach wait, that's my bad. It is possible for classes and not for defs, which I didn't realize. But it fails for both classes and defs when trying to compiler, so this seems like an issue in scalameta parser only.

But yeah, it's not strictly related to this PR I was just confused how this even parsed.

tgodzik (approved):

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
kitbellew: should we open an issue in the parser and describe how it works? does it allow explicit implicits, does it require unique modifiers per parameter, does it only check consecutive ones? are there modifiers which aren't allowed at all?

ideally, they'd revise the grammar...

tgodzik: I think this is an omission really and we can open an issue in the parser, but most other modifiers should be handled properly
just implicit/using are a weird case

kitbellew: @tgodzik could you please check this:
```
class A(implicit implicit val a: String):
  // foo
```
in my Intellij IDE it works (and without `val` doesn't).

tgodzik: There is an error from the compiler: `Repeated modifier implicit`

agilesteel:
Given
```hocon
version = "3.7.1"

runner.dialect = scala213

maxColumn = 100

newlines.implicitParamListModifierForce = [after]

verticalMultiline {
  arityThreshold = 3
  atDefnSite = yes
  newlineAfterOpenParen = yes
}
```
this code
```scala
object Foo {
  def bar(a: Int)(implicit b: Int): Int =
    a + b
}
```
remains unformatted. However with `3.7.2` and `3.7.3` it is formatted as:
```scala
object Foo {
  def bar(
      a: Int
    )(implicit
      b: Int
    ): Int =
    a + b
}
```
Is this the intended behavior? I've had this config for years and only now it started unfolding everything as soon as it sees the `implicit` keyword.

Let me know if it's a bug and I'll open an issue.

kitbellew:
> Given
> ...
> Is this the intended behavior? I've had this config for years and only now it started unfolding everything as soon as it sees the `implicit` keyword.

thank you for sharing. are you saying that what you observe is somehow related to this specific pull request? if so, how?

if not, could you please verify your case using the most recent published version of the `scalafmt` binary and submit a bug report if necessary?

agilesteel:
I saw you mentioned `implicitParamListModifierForce=after` in the first comment of this PR so I assumed it's related.

I'll open an issue if you think that it's a bug. I'll do it tomorrow though. It's getting late over here.

PR #3501, opened 2023-03-11
FormatOps: force break on multi-stat try blocks

### LINKED ISSUE #3492: Reformat causes parse error when using semicolon in try
The following code formats fine:

```
object Test:
  def test =
    try
      1
      2
    catch
      case _: RuntimeException => 2
      case _: Exception        => 3
```

But if I change it to:

```
object Test:
  def test =
    try
      1; 2
    catch
      case _: RuntimeException => 2
      case _: Exception        => 3
```

This is reformatted as:

```
object Test:
  def test =
    try 1; 2
    catch
      case _: RuntimeException => 2
      case _: Exception        => 3
```

Which then causes a parse error in scalafmt.

My scalafmt conf file:

```
version = 3.7.2
maxColumn = 140
runner.dialect = scala3
```

kitbellew:
Fixes #3492.


tgodzik (approved):
:rocket:

Issue #3489, opened 2023-03-05
Inconsistent `fewerBraces` indentation


soronpo:
## Configuration (required) ##

Please paste the contents of your `.scalafmt.conf` file here:
```
version = 3.7.2
runner.dialect = scala3
```

## Command-line parameters (required) ##

When I run scalafmt via CLI like this: `scalafmt a.scala`

## Steps

Given code like this:
```scala
class Top extends EDDesign:
  process(clk, rst):
    val c = Bits(16) const all(0)
    if (rst)
      y :== c
    else if (clk.rising)
      y :== x
  val myblock = process(all):
    val my_var = Bits(16) <> VAR
    my_var := x
    y :== my_var
  process.forever:
    z :== x
    y :== z
end Top
```

## Problem

Scalafmt formats code like this:
```scala
class Top extends EDDesign:
  process(clk, rst):
      val c = Bits(16) const all(0)
      if (rst)
        y :== c
      else if (clk.rising)
        y :== x
  val myblock = process(all):
      val my_var = Bits(16) <> VAR
      my_var := x
      y :== my_var
  process.forever:
    z :== x
    y :== z
end Top
```

## Expectation

I would like the formatted output to look like this (same as the input, in this case):
```scala
class Top extends EDDesign:
  process(clk, rst):
    val c = Bits(16) const all(0)
    if (rst)
      y :== c
    else if (clk.rising)
      y :== x
  val myblock = process(all):
    val my_var = Bits(16) <> VAR
    my_var := x
    y :== my_var
  process.forever:
    z :== x
    y :== z
end Top
```

## Workaround

None

## Notes

I think that it's worth considering adding an `indent.fewerBracesSite` option that will control the indentation for fewer braces.

kitbellew:
@soronpo since you saw the other thread, you should understand the reason, and the consistency behind it.

since we must indent chained methods (`.select`), including after `foo:` and after `foo.bar:`, what do you propose to do? apply the extra indent only if there's a chained select, and shift everything right when it is added? that would make a lot of people unhappy...

kitbellew:
> Scalafmt formats code like this:
> 
> ```scala
> ...
>   val myblock = process(all):
>       val my_var = Bits(16) <> VAR
>       my_var := x
>       y :== my_var
>   process.forever:
>     z :== x
>     y :== z
> end Top
> ```

[...]

PR #3514, opened 2023-03-16
Router: indent trailing fewer braces, too

kitbellew:
Previously, we applied a space indent when fewer braces were followed by another select. This takes care of the case when there are no more selects.

Towards #3489.

### DIFF
--- a/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
+++ b/scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
@@ -3609,7 +3609,7 @@ object a:
 object a:
    def f(): Unit =
      List(1, 2, 3).foldLeft(1):
-        case (a, b) => a + b
+          case (a, b) => a + b
 <<< coloneol in fewer braces 1, main > sig
 maxColumn = 80
 indent.main = 4
@@ -3617,12 +3617,12 @@ indent.main = 4
 object a:
   def f(): Unit =
     List(1, 2, 3).foldLeft(1):
-      case (a, b) => a + b
+          case (a, b) => a + b
 >>>
 object a:
    def f(): Unit =
        List(1, 2, 3).foldLeft(1):
-          case (a, b) => a + b
+              case (a, b) => a + b
 <<< coloneol in fewer braces 2
 maxColumn = 80
 ===
@@ -3679,8 +3679,8 @@ object a:
 object a:
    def f(): Unit =
       List(1, 2, 3).foo:
-         case a: Int =>
-         case _      =>
+           case a: Int =>
+           case _      =>
       otherTerm()
 <<< coloneol in fewer braces 3, main > sig
 maxColumn = 80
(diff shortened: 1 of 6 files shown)


soronpo:
@kitbellew Thank you so much for taking care of this! I really appreciate it.

scalafmt-tests/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -3609,20 +3609,20 @@ object a:
...
   def f(): Unit =
     List(1, 2, 3).foldLeft(1):
-      case (a, b) => a + b
+          case (a, b) => a + b
 >>>
 object a:
    def f(): Unit =
        List(1, 2, 3).foldLeft(1):
-          case (a, b) => a + b
+              case (a, b) => a + b
```
tgodzik, line 3625: This looks like way too much indentation. Looks like it's applied 3 times, wouldn't really one be enough? The alternative with {}  will have single indentation:
```
  object a:
    def f(): Unit =
      List(1, 2, 3).foldLeft(1) {
        case (a, b) => a + b
        case (a, b) => a + b
      }
```

kitbellew: this example uses 4 as the main indent. so there's only one additional level added, to make way for a potential select continuation after it.

tgodzik: We haven't yet agreed on the dotty PR, no? I having doubts about whether we should even merge it. If we don't does this change still make sense?

kitbellew: i don't know. it fixes an inconsistency but you're right that both are not yet supported.

but if the dotty change isn't merged, we'll have a problem...

are you opposed to the change martin proposed?

tgodzik: I think that change might hard to nail especially with the wierd `1-space` rule.

kitbellew: i asked about that space as it seemed unexpected. no documentation mentions that one can be within one space of a significant indentation rather than matching it exactly... does the code not match it exactly?

if he agrees with the half-open interval you're proposing, would that unblock it?

tgodzik: > if he agrees with the half-open interval you're proposing, would that unblock it?

I think that would be the most logical yeah.

kitbellew: @tgodzik i would like to merge this change anyway. i have another pr ready, which makes this configurable, so if the compiler change is made, this is ready, and if not, there's a currently compatible alternative. wdyt?

tgodzik (approved):
It doesn't seem to break anything, so it should be ok to merge. Sorry for not responding earlier, my last week was super hectic.

PR #3518, opened 2023-03-24
Router: don't add extra indent for fewer braces

### LINKED ISSUE #3489: Inconsistent `fewerBraces` indentation

kitbellew:
And don't indent select after them, as that's not yet supported by the compiler. Fixes #3489.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -2076,28 +2076,17 @@ class FormatOps(
                 .withSingleLine(opt)
                 .andPolicy(decideNewlinesOnlyAfterToken(opt))
             }
-            val indent = t.parent match {
-              case Some(p: Term.Apply) =>
-                @tailrec
-                def isSelect(ma: Member.Apply): Boolean = ma.fun match {
-                  case x: Member.Apply => isSelect(x)
-                  case x => x.is[Term.Select]
-                }
-                if (isSelect(p)) None // select is taken care off elsewhere
-                else Some(style.indent.main + style.indent.getSignificant)
-              case _ => None
-            }
             Some(new OptionalBracesRegion {
               def owner = t.parent
               def splits = Some(t.values match {
                 case (tf: Term.FunctionTerm) :: Nil
                     if !style.newlines.alwaysBeforeCurlyLambdaParams &&
                       t.parent.exists(_.is[Term.Apply]) =>
-                  getSplits(ft, t, forceNL = false, indentOpt = indent) match {
+                  getSplits(ft, t, forceNL = false) match {
                     case s +: rs if !s.isNL => funcSplit(tf)(s.fileLine) +: rs
                     case ss => ss
                   }
-                case _ => getSplits(ft, t, forceNL = true, indentOpt = indent)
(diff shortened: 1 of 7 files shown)


docs/configuration.md
```diff
@@ -589,6 +589,81 @@ foo // c1
...
+- `always`: always applies extra indent
+- `never`: doesn't apply any extra indent
+- `beforeSelect`: applies extra indent only to fewer-braces
+  expressions followed by a `.select`
+
+```scala mdoc:defaults
+indent.fewerBraces
+```
+
+> In Scala 3.3.0-RC3, only `never` provides compiler-compatible code.
```
tgodzik, line 607: Should we introduce a setting based on something that might be fixed? I would hold off in that case

kitbellew: could you explain it, please? if this is fixed in the compiler, some people might still prefer different options. and if it's not, then the current implementation doesn't work.

by the way, are you and Martin close to an agreement on that fix? :)

tgodzik: `only `never` provides compiler-compatible code.`

That's what worries me, if user sets something wrong it breaks their compilation, no?

> by the way, are you and Martin close to an agreement on that fix? :)

I need to come back to it, don't remember what's the current status :O

kitbellew: the only other suggestion i might have is

- add a new dialect field in scalameta to say select can be indented after fewer braces
- set that flag in scala33
- add a new intermediate scala330rc2 dialect where this flag is not set
- move fewer braces to that dialect as well, as it hadn't been available before that.

tgodzik: > so what's the solution, then? right now, it will always break their compilation...

Couldn't we make `never` the default without the new setting for now?

kitbellew: > > so what's the solution, then? right now, it will always break their compilation...
> 
> Couldn't we make `never` the default without the new setting for now?

done

tgodzik (approved):
Thanks!

Issue #3520, opened 2023-03-30
Curried functions with implicit parameter lists are always unfolded even though the verticalMultiline.arityThreshold is not reached.


agilesteel:
## Configuration ##

```hocon
version = "3.7.3"

runner.dialect = scala213

maxColumn = 100

newlines.implicitParamListModifierForce = [after]

verticalMultiline {
  arityThreshold = 3
  atDefnSite = yes
  newlineAfterOpenParen = yes
}
```

## Steps

Given code like this:
```scala
object Foo {
  def bar(a: Int)(implicit b: Int): Int =
    a + b
}
```

## Problem

Scalafmt formats code like this:
```scala
object Foo {
  def bar(
      a: Int
    )(implicit
      b: Int
    ): Int =
    a + b
}
```

## Expectation

I would like the formatted output to look like this:
```scala
object Foo {
  def bar(a: Int)(implicit b: Int): Int =
    a + b
}
```

## Workaround

I've found that by switching back all the way to `3.7.1` fixes the problem.

## Notes

I think it's related to https://github.com/scalameta/scalafmt/pull/3478

kitbellew:
@agilesteel perhaps it's because 3.7.1 and before handled this incorrectly. the parameter is named `implicitParamListModifierForce`, hence the newline is **forced**, it cannot be omitted. if you try to format without vertical multiline, it will introduce a newline as well.

kitbellew:
> Maybe this should be a feature request then? I think it makes sense to respect the threshold.

The threshold forces the format, but it's not the only way; if the signature can't be formatted on one line (such as when a newline is forced), that will cause this formatting as well.

> ```scala
> object Foo {
>   def bar(
>       a: Int
>     )(implicit b: Int // b: Int should be on the next line
>     ): Int =
>     a + b
> }
> ```

Why did you expect `b: Int` to be on the next line? The entire `implicit b: Int` fits on one line so `newlines.implicitParamListModifierPrefer` should not apply, according to the documentation.

Also, if dangling parens parameter is set appropriately, this will also happen:
> ```scala
> object Foo {
>   def bar(
>       a: Int
>     )(b: Int
>     ): Int =
>     a + b
> }
> ```
The reason is that the formatter doesn't force a newline when indentation would be the same or larger, compare with:
> ```scala
> object Foo {
>   def bar(
>       a: Int
>     )(
>       b: Int
>     ): Int =
>     a + b
>
>   def bar(
>       a: Int
>     )(
[...]

PR #3521, opened 2023-03-31
TreeOps: get the whole implicit clause, not values

kitbellew:
Towards #3520.


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
```diff
@@ -2432,7 +2432,7 @@ class Router(formatOps: FormatOps) {
         val spaceSplit = Split(Space, 0)
           .notIf(style.newlines.forceAfterImplicitParamListModifier)
           .withPolicy(
-            SingleLineBlock(params.last.tokens.last),
+            SingleLineBlock(params.tokens.last),
```
kitbellew, line 2435: here, includes the closing `)` now.

tgodzik (approved):

Issue #3543, opened 2023-05-21
Consider changing the default behavior when no version is set


ckipp01:
## Configuration (required) ##

You can mimic what I'm about to show by just having an empty `.scalafmt.conf` file.

The version of the scalafmt cli that I'm using to test this is `3.7.3`.

## Steps

1. Have an empty `.scalafmt.conf` file in a workspace
2. Use `scalafmt .` to attempt to format that workspace
3. You'll see the error below:

```
error: missing version (current 3.7.3)
```

## Problem

I ironically hit on this about a week ago and knew that I screwed up and must have left my version empty. However recently I came across [this thread](https://fosstodon.org/@mo8it/110396291266757088) where a newcomer to the language hit on this and it wasn't clear at all what they should do about it. For those familiar to scalafmt, it may be super obvious that you need to add the version to the `.scalafmt.conf` file, but how would a newcomer know that?

## Expectation

I'd expect at the very least maybe a more detailed message like:

```
error: missing version (current 3.7.3)

To fix this, you should add the following to your .scalafmt.conf file:
version = 3.7.3
```

Another option would be to just add it for them. Since they don't currently have one, maybe just

1. assume the latest and format with it
2. add that version to the `.scalafmt.conf` file
3. give an info message to the user that you did that

kitbellew:
i can add an extra sentence to the error message.

however, i don't think we should manipulate the configuration file.

i also don't think we should assume any defaults, this was the case before and we intentionally removed this. multiple users of the codebase might have a different version of the formatter installed, and each of them might end up with a slightly different result which would then be inadvertently and intentionally sent for review, or simply merged.

kitbellew:
also, version is not the only required option. scala dialect must also be specified, and there's no default.

ckipp01:
> however, i don't think we should manipulate the configuration file.

Do you maybe have some reasoning as to why? Scalafmt is in the perfect to position to know _what_ is missing. I'm assuming in this case where there is no version set, setting a version would be pretty harmless, and also ensure that it's consistently being used in the future. Maybe if the worry is changing it without really giving the user a choice, we could do that? Is it possible with the CLI library to prompt a use with something like?

```
You have no version set for scalafmt. The latest is 3.7.3. Would you like us to update your .scalafmt.conf?

1. yes
2. no
```

No idea if that's possible.

> i also don't think we should assume any defaults, this was the case before and we intentionally removed this. 

[...]

PR #3555, opened 2023-06-06
Fix proposed version syntax

JD557:
Follow up to https://github.com/scalameta/scalafmt/pull/3544, the proposed version string looks something like:
`version = '3.7.4'`

While scalafmt only accepts strings formatted like:
`version = 3.7.4`

Curiously enough, the original issue (https://github.com/scalameta/scalafmt/issues/3543) had it right 😅

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
@@ -96,7 +96,7 @@ object Cli {
     "true" == System.getProperty("scalafmt.native-image", "false")
 
   private def getProposedConfigVersion(options: CliOptions): String =
-    s"version = '$stableVersion'"
+    s"version = $stableVersion"
 
   private def findRunner(
       options: CliOptions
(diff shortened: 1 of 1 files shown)


scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
```diff
@@ -96,7 +96,7 @@ object Cli {
     "true" == System.getProperty("scalafmt.native-image", "false")
 
   private def getProposedConfigVersion(options: CliOptions): String =
-    s"version = '$stableVersion'"
+    s"version = $stableVersion"
```
kitbellew, line 99: could you make the same change on line 137 below (where it was copied from)?

kitbellew (changes requested):

scalafmt-cli/src/main/scala/org/scalafmt/cli/Cli.scala
JD557: Done, thanks for the heads up 👍

PR #3557, opened 2023-06-12
Router: handle parameter `indent.fewerBraces`

kitbellew:
Now that dotty compiler supports indented fewer-braces chains since v3.3.1-RC1, let's implement this formatting. For #3489.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -589,6 +589,82 @@ foo // c1
 }
 ```
 
+#### `indent.fewerBraces`
+
+This parameter controls whether extra `indent.main` is added
+to the sole argument of a method call using the "fewer braces"
+syntax. The following values are supported:
+
+- `always`: always applies extra indent
+- `never`: doesn't apply any extra indent
+- `beforeSelect`: applies extra indent only to fewer-braces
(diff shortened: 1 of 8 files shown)


tgodzik (approved):
LGTM!

xerial:
@kitbellew It looks like a bug introduced with this change that fails to insert a new line after extension method:

For example:
```scala
    given IntervalLike[MyClass] with
      extension (a: MyClass)
        def start: Long = a.createdAt
        def end: Long   = a.endAt
```

This code will be formatted as follows:
```scala
    given IntervalLike[MyClass] with
      extension (a: MyClass) def start: Long = a.createdAt   //  causes compilation error with missing new line
      def end: Long                          = a.endAt
```

kitbellew:
@xerial thank you for reporting, could you perhaps create a separate issue next time?

as to this example, the bug was indeed introduced in the latest version but not via this change, it came from scalameta. someone has also reported it, and the fix will be in the next version.

xerial:
Thanks. Will do. I was overwhelmed by the number of items I need to fill in the issue template 😅

PR #3573, opened 2023-07-04
RemoveScala3OptionalBraces: not match if infix lhs

kitbellew:
Addresses rewrite failure found by @lenguyenthanh in lichess-org/lila#13160.

PR #3576, opened 2023-07-06
RemoveScala3OptionalBraces: handle infix on rbrace

kitbellew:
Instead of handling a specific owner of the left brace (previously, it was a Term.Match only), let's instead look for infix when we encounter the right brace instead, and ensure that the next token is not the op of an infix.

This also handles the case when Term.Match was not an LHS but in the middle of an infix chain.

Follow-on to #3573.


kitbellew:
> hey @kitbellew, it seems you fixed some (or all) of the issues I mentioned in [discord](https://discord.com/channels/632642981228314653/632665341864181780/1126264341235245066). Do I need create those issues any more?

if nothing remains, then no.

tgodzik (approved):

PR #3581, opened 2023-07-07
Router: use template to break before `extends`

### LINKED ISSUE #3580: Unit test failures on 2.13.12
Unit tests fail on the Scala 2 community build.

HEAD

## Problem

```scala
[scalafmt] [error] Failed tests:
[scalafmt] [error] 	org.scalafmt.FidelityTest
[scalafmt] [error] 	org.scalafmt.FormatTests
```

## Notes

It looks like `Tokens#last` doesn't throw on empty `Tokens`. Scalafmt should handle empty tokens:
```
defnBeforeTemplate(leftOwner).filter(_.tokens.nonEmpty).map { x =>
  val policyEnd = Policy.End.On(x.tokens.last)
  delayedBreakPolicy(policyEnd)(forceNewlineBeforeExtends)
}
```
(I didn't pursue a fix.)

Noticed at https://github.com/scala/community-build/issues/1680#issuecomment-1625131437

```
==> X org.scalafmt.FormatTests.unit/Annotations.stat:59: #1640                                       |  0.001s org.scalafmt.util.FormatException: java.util.NoSuchElementException: last of empty IndexedSeq
    at org.scalafmt.FormatTests.run(FormatTests.scala:56)
    at org.scalafmt.FormatTests.$anonfun$new$3(FormatTests.scala:33)
    at org.scalafmt.FormatTests.$anonfun$new$3$adapted(FormatTests.scala:33)
    at org.scalafmt.util.CanRunTests.$anonfun$runTest$2(CanRunTests.scala:19)
Caused by: java.util.NoSuchElementException: last of empty IndexedSeq
    at scala.collection.IndexedSeqOps.last(IndexedSeq.scala:110)
    at scala.collection.IndexedSeqOps.last$(IndexedSeq.scala:105)
    at scala.meta.tokens.Tokens.last(Tokens.scala:29)
    at org.scalafmt.internal.Router.$anonfun$getSplitsImpl$118(Router.scala:838)
    at scala.Option.map(Option.scala:242)
    at org.scalafmt.internal.Router.getSplitsImpl(Router.scala:837)
```

kitbellew:
In older versions of scalameta, `extends` was not part of template, and we had to identify the tree just preceding it.

Also, make sure to handle cases when the template is empty (in that case there's no reason to break before a non-existent `extends`).

Fixes #3580.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -374,6 +374,13 @@ class FormatOps(
     new Policy.Delay(policy, Policy.End.Before(from))
   }
 
+  val WithTemplateOnLeft = new ExtractFromMeta(ft =>
+    ft.meta.leftOwner match {
+      case lo: Stat.WithTemplate => Some(lo.templ)
+      case _ => None
+    }
+  )
+
   def templateCurlyFt(template: Template): Option[FormatToken] =
     getStartOfTemplateBody(template).map(tokenBefore)
 
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/Router.scala
@@ -817,14 +817,10 @@ class Router(formatOps: FormatOps) {
       case FormatToken(
             _: T.KwObject | _: T.KwClass | _: T.KwTrait | _: T.KwEnum,
             _,
-            _
+            WithTemplateOnLeft(template)
           ) =>
-        def expire = defnTemplate(leftOwner)
-          .flatMap {
-            getTemplateGroups(_).flatMap(
-              _.lastOption.flatMap(_.headOption.flatMap(_.tokens.headOption))
-            )
-          }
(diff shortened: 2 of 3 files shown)


tgodzik (approved):

SethTisue:
@kitbellew do you think it would be appropriate for us to say something about this in the [Scala 2.13.12 release notes](https://github.com/scala/scala-dev/pull/850)? and if so, can you suggest a wording?

kitbellew:
@SethTisue oh, what did I miss? how's this related to that scala release?

SethTisue:
https://github.com/scalameta/scalafmt/issues/3580 was reported against a nightly build of Scala 2.13.12, after https://github.com/scala/scala/pull/10392 was merged. I'm asking if we have a situation where users who upgrade Scala 2.13.12 will also need to upgrade their Scalafmt version in order to format their code.

(It's possible I have some misconception.)

kitbellew:
@SethTisue I think the problem was because scalafmt was _rebuilt_ and re-tested (and failed) with the new scala version, as part of that community build. but the released version of scalafmt hadn't been built with that version (and the released version of scalafmt hadn't failed the same tests), so it doesn't seem like it would have real-world consequences. @tgodzik what do you think?

PR #3612, opened 2023-08-13
Router: implement `spaces.beforeXxxArgInParens`

### LINKED ISSUE #3385: feat: add more options for DSL
Hello,

## Configuration

This issue is about options available as of https://scalameta.org/scalafmt/docs/configuration.html and is not specific to a currently released version of `scalafmt`.

It suggests more options and describe how they would behave. These options do not follow the [Scala Style Guide](https://docs.scala-lang.org/style/) as it is for DSL (examples here are for SpinalHDL, a DSL to describe digital hardware).

*Note: examples provided here are minimal to show the syntax, they are not real-case logic.*

## Suggestion 1: Add DSL keywords to add space after them

The idea is to get [`spaces.afterKeywordBeforeParen`](https://scalameta.org/scalafmt/docs/configuration.html#spacesafterkeywordbeforeparen) working for custom keywords

For instance, a configuration could be:
```scala
spaces.afterWordsBeforeParen = ["when", "switch", "is"]
```

Given code like this (formatted):
```scala
when(a === b) {
  c := d
}
switch(c) {
  is(False) {
    e := False
  }
  is(True) {
    e := a
  }
}
```

I would like the formatted output to look like this, with an extra space before the opening brace:
```scala
when (a === b) {
  c := d
}
switch (c) {
  is (False) {
    e := False
  }
  is (True) {
    e := a
  }
}
```

## Suggestion 2: special management of infix operators

SpinalHDL provides modifiers to act on hardware types (representing wires).

For instance, a configuration could be:
```scala
spaces.afterAlnumInfix = false
```

It would remove the space after an infix operator followed by a `(`, but only if this operator is only alphanumeric.

Another (maybe better) configuration could be:

```scala
spaces.removeAfterInfix = ["init"]
```

It would be useful to split the modifier (method) and the path to the wire acted on, by using dots for the path and not for the modifier.

Given code like this (formatted):
```scala
val x = Reg(RGB())
x.r init (0)
x.g init (0)
x.b init (0)
```

*Note: another option is `x.b.init(0)` but the path is mixed with the modifier*

I would like the formatted output to look like this, with an extra space before the opening brace:
```scala
val x = Reg(RGB())
x.r init(0)
x.g init(0)
x.b init(0)
```

Here I see I act on the signal `x.b` to set it's `init` value to 0.

### LINKED ISSUE #3607: Do not insert space after arity-0 infix operator, before parens

```
version = 3.7.12
runner.dialect = scala212
maxColumn = 120
```

When I run scalafmt via CLI like this: `coursier launch scalafmt`

## Steps

Given code like this:
```scala
val logger = mock[Logger]
logger.warn(any) shouldDoNothing()
```

## Problem

```scala
val logger = mock[Logger]
logger.warn(any) shouldDoNothing () // Note the extra space after `shouldDoNothing`
```

## Expectation

```scala

val logger = mock[Logger]
logger.warn(any) shouldDoNothing()
```

## Workaround

–

## Notes

Note that `shouldDoNothing` is an arity-0 macro provided by Mockito Scala:

```scala
def shouldDoNothing(): Unit = macro DoSomethingMacro.doesNothing
```

kitbellew:
Fixes #3385. Fixes #3607.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4300,6 +4300,71 @@ spaces.afterSymbolicDefs=true
 def +++(a: A): F[A]
 ```
 
+### `spaces.beforeXxxArgInParens`
+
+> Since v3.7.13.
+
+These parameters control whether a space should be added before an argument of
+a function call or infix expression, if the argument is enclosed in parentheses.
+
+They take the following values:
+
+- `Never`: no space is added
+- `Always`: space is added
+- `AfterSymbolic`: space is added if the infix operator or function name is
+  symbolic (doesn't start with a letter or underscore)
+
+Please note that these parameters will not affect spacing after an
+[unary operator](https://docs.scala-lang.org/scala3/reference/changed-features/operators.html#unary-operators)
+(i.e., one of `+`, `-`, `!`, `~`), as it's neither a function call nor an infix.
+
+Also, `spaces.beforeApplyArgInParens` generalizes the special-case parameter
+`spaces.afterTripleEquals` which only applies to a `===` function call.
+
+```scala mdoc:defaults
+spaces.beforeApplyArgInParens
+spaces.beforeInfixArgInParens
+```
+
+```scala mdoc:scalafmt
+spaces.beforeApplyArgInParens = Always
+spaces.beforeInfixArgInParens = Always
+---
++(baz)
+===(baz)
+bar(baz)
+
+foo +(baz)
+foo ===(baz)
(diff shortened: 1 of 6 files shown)


kitbellew:
@tgodzik any thoughts on this?

tgodzik (approved):
I was wondering if we needed a new setting (as always), but I guess this sounds resonable.

PR #3613, opened 2023-08-18
PolicyOps: use `newlines.ignoreInSyntax`

### LINKED ISSUE #3608: Insert newline / align closing paren after method call with single multiline string

```
version = 3.7.12
runner.dialect = scala212
maxColumn = 120
```

When I run scalafmt via CLI like this: `coursier launch scalafmt`

## Steps

Given code like this:
```scala
println("""
  |A very long string
  |with multiple lines
  |""".stripMargin
)
```

## Problem

```scala
println("""
  |A very long string
  |with multiple lines
  |""".stripMargin) // No newline before closing paren, not aligned with beginning of statement
```

## Expectation

```scala
println("""
  |A very long string
  |with multiple lines
  |""".stripMargin
)
```

## Workaround

–

## Notes

It works fine when the method is called with multiple arguments:

```scala
println(
  """
  |A very long string
  |with multiple lines
  |""".stripMargin,
  "ok"
)
```

I'm assuming that scalafmt (kind of correctly) considers a multiline string to be a single argument so no newline is inserted before the closing paren (same as `println("ok")`). My rationale is that it would be more consistent if a multiline string actually spanning multiple lines would be considered multiple arguments, in which case scalafmt can be configured to insert a newline before the closing paren.

The behavior for strings using the multiline syntax but actually consisting of a single line is fine:

```scala
println("""A pretty short string""".stripMargin)
```

kitbellew:
Specifically, for PenalizeAllNewlines and SingleLineBlock policies. Fixes #3608.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2576,6 +2576,44 @@ newlines.inInterpolation
 - `avoid`: attemps to avoid breaks within the spliced code, regardless of line overflow
 - `oneline`: formats the splice on a single line, or breaks after `${` if overflows
 
+### `newlines.ignoreInSyntax`
+
+The formatter frequently chooses between adding a newline and continuing the
+same line but either prohibiting or heavily discouraging subsequent newlines
+_between_ tokens, to fit the rest of the expression on the same line.
+
+However, in many cases and, for historical reasons, intentionally, newlines
+_within_ tokens have been frequently ignored, leading to "single-line" blocks
+which actually span multiple lines.
+
+This boolean parameter now allows controlling whether to ignore newlines found
+in syntax of strings or other possibly multi-line tokens when newlines are
+otherwise prohibited or undesirable (such as for single-line formatting).
+
+```scala mdoc:defaults
+newlines.ignoreInSyntax
+```
+
+> Since v3.7.13. Prior to that, this behaviour was always enabled.
+
+```scala mdoc:scalafmt
+newlines.ignoreInSyntax = true
+---
+// ignores newline in string, pretends everything fits on one line
+println(s"""${1}
+    """.stripMargin
+)
+```
+
+```scala mdoc:scalafmt
+newlines.ignoreInSyntax = false
+---
+// detects newline in string, forces proper multi-line formatting
+println(s"""${1}
+    """.stripMargin
+)
+```
+
 ### `optIn.annotationNewlines`
 
 This boolean parameter controls newlines after annotations.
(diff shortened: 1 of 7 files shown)


docs/configuration.md
```diff
@@ -2531,6 +2531,36 @@ newlines.inInterpolation
 - `avoid`: attemps to avoid breaks within the spliced code, regardless of line overflow
 - `oneline`: formats the splice on a single line, or breaks after `${` if overflows
 
+### `newlines.ignoreInSyntax`
+
+This boolean parameter controls whether to ignore newlines in strings (or other
```
tgodzik, line 2536: Shouldn't this only be used for strings? It's a bit unclear what `other possibly multi-line tokens` could entail. Overall, would that be a setting for that one user?

kitbellew: 1. the code doesn't really specify which tokens could have embedded newlines; other possible options are comment and xml.
2. as to the target audience: i would definitely use it in our codebase, not a big fan of ignored embedded newlines myself.

kitbellew: @tgodzik any additional thoughts on this?

tgodzik: I think I am missing the exact meaning here. `whether to ignore newlines in strings` yet the newlines are outside the strings,  so this only affects strings if they are inside some other srtucture?

kitbellew: hmm... let me add a longer explanation :)

docs/configuration.md
```diff
@@ -2576,6 +2576,44 @@ newlines.inInterpolation
...
+same line but either prohibiting or heavily discouraging subsequent newlines
+_between_ tokens, to fit the rest of the expression on the same line.
+
+However, in many cases and, for historical reasons, intentionally, newlines
+_within_ tokens have been frequently ignored, leading to "single-line" blocks
+which actually span multiple lines.
+
+This boolean parameter now allows controlling whether to ignore newlines found
+in syntax of strings or other possibly multi-line tokens when newlines are
+otherwise prohibited or undesirable (such as for single-line formatting).
```
kitbellew, line 2591: @tgodzik do you think this explanation makes it a bit clearer?

tgodzik: `leading to "single-line" blocks
which actually span multiple lines.` is a bit unclear to me, though I guess the example here would need to be the best explanation we can get

tgodzik (approved):
Thanks for expanding the explanation!

docs/configuration.md
```diff
@@ -2576,6 +2576,44 @@ newlines.inInterpolation
...
+newlines.ignoreInSyntax
+```
+
+> Since v3.7.13. Prior to that, this behaviour was always enabled.
+
+```scala mdoc:scalafmt
+newlines.ignoreInSyntax = true
+---
+// ignores newline in string, pretends everything fits on one line
+println(s"""${1}
```
kitbellew, line 2603: this is an example. contains newline but will be formatted without breaks around open and close parentheses.

PR #3616, opened 2023-08-18
FormatWriter: implement `align.delayUntilSpace`

### LINKED ISSUE #3614: Tabular Alignment

```
#.scalafmt.conf
version = 3.7.12
runner.dialect = scala213

align.tokens = [{
  code = ")"
  owners = [{
    regex = "Defn\\.Def"
  }]
}]
```

I did not attempt this using the CLI (since this is probably not a bug)

## Steps

I want to end up with code like this:
```scala
object Test {

  def meeethod1(pram1      : AnyRef): Any = ???
  def methd2   (paaaaaram2 : Any   ): Any = ???
  def methh3   (param333333: Any   ): Any = ???
  def md4      (param4     : Any   ): Any = ???

}
```

## Expectation
Something, anything would align in any way.

## Notes

I don't think this is an issue with scalafmt per se, but I don't know whether parentheses are a valid alignment token. I am not well versed in scalafmt, but as far as I can tell, if I set a value for `align.token`, the `regex` section should enclose the area I want to align over, and these will be aligned over matching parents. For a colon I can make sense of it, Term.Param owns the colon, and the token there-after is aligned for every parent (Defn.Def). 

```
align.tokens = [{
  code = ":"
  owners = [{
    regex = "Term\\.Param"
    parents = [ "Defn\\.Def" ]
  }]
}]
```
===>
```
object Test {

  def meeethod1(pram1:    AnyRef): Any = ???
  def methd2(paaaaaram2:  Any): Any = ???
  def methh3(param333333: Any): Any = ???
  def md4(param4:         Any): Any = ???

}
```
Thus, all the values to the right of the colon are aligned.  No matter what I do, however, I cannot coax it into aligning over parentheses per the desired example. I've played with the AST explorer and tried a great many combinations of tokens to try achieve something, set values are wide and specific as I can - even inspected the source code a little (to no avail). 

My best guess is that I would have to be able to set the alignment token to be a member of the AST itself to achieve what I'm trying to achieve, which does not seem to be a supported behaviour.

kitbellew:
Fixes #3614.

### DIFF
--- a/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
+++ b/scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
@@ -2054,8 +2054,27 @@ object a {
   def meth3(     param333333: Any    ): Any = ???
   def md4(       param4:      Any    ): Any = ???
 }
-<<< #3614 !spaces.inParentheses
+<<< #3614 !spaces.inParentheses align.delayUntilSpace
 spaces.inParentheses = false
+align.delayUntilSpace = true
+align.tokens."+" = [ { code = ":" }, { code = "(" }, { code = ")" }, { code = "=" } ]
+===
+object a {
+  def meeethod1(pram1: AnyRef): Any = ???
+  def methd2(paaaaaram2: Any): Any = ???
+  def meth3(param333333: Any): Any = ???
+  def md4(param4: Any): Any = ???
+}
+>>>
+object a {
+  def meeethod1(pram1:       AnyRef): Any = ???
+  def methd2(paaaaaram2:     Any):    Any = ???
+  def meth3(param333333:     Any):    Any = ???
+  def md4(param4:            Any):    Any = ???
+}
+<<< #3614 !spaces.inParentheses !align.delayUntilSpace
+spaces.inParentheses = false
+align.delayUntilSpace = false
 align.tokens."+" = [ { code = ":" }, { code = "(" }, { code = ")" }, { code = "=" } ]
 ===
 object a {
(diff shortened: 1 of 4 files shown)


treetime:
Should this interact with Align.multiline? 
For
```
spaces.inParentheses = false
align.delayUntilSpace = false
align.tokens."+" = [ { code = ":" }, { code = "(" }, { code = ")" }, { code = "=" } ]
```

Given:
```
object potato {
   def meeethod1(pram1:   AnyRef): Any = ???

   def methd2(paaaaaram2: Any):    Any = ???
   def meth3(param33333:  Any):    Any = ???
   def md4(param4:        Any):    Any = ???
}
```
yields:
```
object potato {
  def meeethod1(pram1: AnyRef): Any = ???

  def methd2(paaaaaram2: Any): Any = ???
  def meth3 (param33333: Any): Any = ???
  def md4   (param4    : Any): Any = ???
}
```

Adding Align.multiline yields the same; I would've expected alignment to happen despite the newline.

kitbellew:
> Should this interact with Align.multiline? For

this is orthogonal to multiline; also, in your example, none of the expressions are occupying multiple lines.

keep in mind that alignment applies to successive lines only, and there's a __blank line__, the current alignment block is flushed.

scalafmt-tests/src/test/resources/align/DefaultWithAlign.stat
```diff
@@ -2071,3 +2072,21 @@ object a {
...
+align.tokens."+" = [ { code = ":" }, { code = "(" }, { code = ")" }, { code = "=" } ]
+===
+object a {
+  def meeethod1(pram1: AnyRef): Any = ???
+  def methd2(paaaaaram2: Any): Any = ???
+  def meth3(param333333: Any): Any = ???
+  def md4(param4: Any): Any = ???
+}
+>>>
+object a {
```
tgodzik, line 2087: Actually, I would have thought it would work like this always with alignment, though I don't use so never tested. I guess since this changes the default, we need the additional flag.

tgodzik (approved):
LGTM

PR #3621, opened 2023-08-28
Installation: document COURSIER_REPOSITORIES

### LINKED ISSUE #3316: Scalafmt seems to ignore COURSIER_REPOSITORIES
Our environment does not support egress so we use Nexus repository manager. While we are able to use coursier to fetch what we want by specifying `COURSIER_REPOSITORIES` AND `COURSIER_CREDENTIALS`, `scalafmt` seems to hit `oss.sonatype.org` while resolving the version specified in the `.scalafmt.conf` file. If the former env variable isn't set, it straight away goes to maven. Am I correct in assuming that `scalafmt` supports `COURSIER_REPOSITORIES`, [ref](https://github.com/scalameta/scalafmt/blob/47650706f50b1a36ed71c505658044f71252a826/scalafmt-dynamic/src/main/scala/org/scalafmt/dynamic/CoursierDependencyDownloader.scala#L29)? If it does indeed support it, shouldn't it hit our Nexus instead?

Version: 3.5.9

Additional helpful information:
I've tried running `coursier` to fetch something which hit our Nexus and was successful. The next immediate command that I ran was `scalafmt` as `scalafmt Foo.scala --debug` which hit some other endpoint/host instead of our Nexus. And because the egress is blocked, it simply sat there with the progress bar at 0%. I've double checked that the env variables are correct. Coursier being able to pull what I've asked for also shows that the configuration is correct.

## Configuration ##

```
version = 3.5.9
...
```

## Command-line parameters ##

`scalafmt Main.scala --debug`

## Expectation

I'd expect that `scalafmt` format the code as per its functionality.

kitbellew:
Fixes #3316.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -91,6 +91,9 @@ formatter, the correct release will be downloaded dynamically.
 
 > Since v3.1.0, the `version` parameter is required to be specified explicitly.
 
+Since that requires access to artifact repositories, please see more on that on the
+[Installation](installation.md#using-custom-repositories-with-cli) page.
+
 ## Scala Dialects
 
 The formatter supports various dialects defined and supported by `scalameta` parser. However,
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
:rocket:

PR #3657, opened 2023-10-03
Scalafmt: don't parse shebang line as scala

### LINKED ISSUE #3642: scalafmt --check fails with an error on scripts with valid `hashbang` lines
```
version = 3.7.14
preset = default
align.preset = more
maxColumn = 200
align.stripMargin = true
runner.dialect = Scala3
fileOverride {
  "glob:**/scala-3.3.1/**" {
    runner.dialect = scala3
  }
}
```

NB: before submitting, please confirm that the problem is observed in the
*latest published* version of the formatter! We don't publish hotfixes for older
versions, and the problem you have observed in an older version may have already
been fixed.

When I run scalafmt via CLI like this: `scalafmt --check`

## Steps

Any directory with scripts having the following `hashbang` line causes problems:
```scala
#!/usr/bin/env -S scala @classpathAtfile
def main(args: Array[String]): Unit =
   println("hello")
```

## Problem
Scalafmt rejects the hashbang line, complaining about the @ character:
```scala
<input>:1: error: [dialect scala3] ; expected but @ found
#!/usr/bin/env -S scala @classpathAtfile
                        ^
<input>:1: error: [dialect scala3] ; expected but @ found
#!/usr/bin/env -S scala @classpathAtfile
                        ^
<input>:1: error: [dialect scala3] ; expected but @ found
#!/usr/bin/env -S scala @classpathAtfile
                        ^
<input>:1: error: [dialect scala3] ; expected but @ found
#!/usr/bin/env -S scala @classpathAtfile
                        ^
<input>:1: error: [dialect scala3] ; expected but @ found
#!/usr/bin/env -S scala @classpathAtfile
                        ^
<input>:1: error: [dialect scala3] ; expected but @ found
#!/usr/bin/env -S scala @classpathAtfile
                        ^
error: --test failed
```

Another possibly related rejection is the following script `hashbang`:
```
scalafmt --check ./biz.sc
<input>:1: error: [dialect scala3] ; expected but double constant found
#!/usr/bin/env -S scala -cp target/scala-3.3.1/classes

error: ParseError=2                                            ^
```

philwalk@d5 MINGW64 ~/work
## Expectation
If the file `./classpathAtfile` is not found, a warning is printed, and in any case, as long as `scala` in the `PATH` is `scala 3`, the script prints `hello`.

It's also necessary that `/usr/bin/env` accept the `-S` option, as is the case for recent versions.
The result is equivalent to invoking the script via the following command line:
```bash
scala @classpathAtfile biz.sc
```
In this simple `hello world` example, the file `classpathAtfile` can be empty, and need not exist.  In general, it's a way to set the classpath.

## Workaround
On the command line, pass a list of all directories other than those with the offending script `hashbang` lines.
A better workaround would be to be able to exclude a directory in `.scalafmt.conf`, or similar.

kitbellew:
Fixes #3642.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
@@ -79,6 +79,19 @@ object Scalafmt {
     )
   }
 
+  private[scalafmt] def splitCodePrefix(input: String): (String, String) =
+    if (!input.startsWith("#!")) ("", input)
+    else {
+      val beforeNL = input.indexOf(UnixLineEnding, 2)
+      if (beforeNL < 0) (input + UnixLineEnding, "")
+      else {
+        val afterNL = beforeNL + UnixLineEnding.length
+        val hasBlank = input.startsWith(UnixLineEnding, afterNL)
+        val idx = if (hasBlank) afterNL + UnixLineEnding.length else afterNL
+        (input.substring(0, idx), input.substring(idx))
+      }
+    }
+
   private def formatCodeWithStyle(
       code: String,
       style: ScalafmtConfig,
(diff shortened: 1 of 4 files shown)


tgodzik (approved):
LGTM!

PR #3658, opened 2023-10-03
FormatOps: fix Defn as optional braces block

### LINKED ISSUE #3653: Improper format of try / catch without braces in Scala3

```
version = 3.7.14
runner.dialect = scala3
```

When I run scalafmt via CLI like this: `scalafmt foo.scala`

## Steps

Given try / catch without braces
```scala
//> using scala 3.3.1
@main def foo(): Unit =
  try
    val x = 3 / 0
  catch
    case e: Throwable => println(s"Caught: ${e.getMessage}")
```

## Problem

Scalafmt removes the newlines / indent after `try` and `catch`:
```scala
//> using scala 3.3.1
@main def foo(): Unit =
  try val x = 3 / 0
  catch case e: Throwable => println(s"Caught: ${e.getMessage}")
```
This doesn't compile anymore.

## Expectation

I would like the formatted output to be unchanged:
```scala
//> using scala 3.3.1
@main def foo(): Unit =
  try
    val x = 3 / 0
  catch
    case e: Throwable => println(s"Caught: ${e.getMessage}")

```
as this compiles and run fine.

kitbellew:
Fixes #3653.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
@@ -2174,7 +2174,7 @@ class FormatOps(
               def rightBrace = blockLast(t.body)
             })
           case t: Term.If if !nft.right.is[T.KwThen] && {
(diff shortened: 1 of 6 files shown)


tgodzik (approved):
LGTM!

PR #3670, opened 2023-10-24
Emit possible solutions in SearchStateExploded msg

rpiaggio:
Emit more information when a `SearchStateExploded` happens, pointing to the docs.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/Error.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/Error.scala
@@ -80,7 +80,7 @@ object Error {
   ) extends Error({
         val tok = LoggerOps.log2(ft)
         val line = ft.left.pos.endLine
-        s"Search state exploded on '$tok', line $line"
+        s"Search state exploded on '$tok', line $line.\nConsider increasing runner.optimizer.maxVisitsPerToken (default 10000) or even runner.maxStateVisits (default 1000000).\nSee https://scalameta.org/scalafmt/docs/configuration.html#search-state-exploded. "
       }) {
     def line: Int = ft.left.pos.endLine
   }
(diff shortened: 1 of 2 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/Error.scala
```diff
@@ -80,7 +80,7 @@ object Error {
   ) extends Error({
         val tok = LoggerOps.log2(ft)
         val line = ft.left.pos.endLine
-        s"Search state exploded on '$tok', line $line"
+        s"Search state exploded on '$tok', line $line.\nConsider increasing runner.optimizer.maxVisitsPerToken (default 10000) or even runner.maxStateVisits (default 1000000).\nSee https://scalameta.org/scalafmt/docs/configuration.html#search-state-exploded. "
```
kitbellew, line 83: please include only the doc link, there's no need to repeat the recommendation.

rpiaggio: After following the link, I had to look around the documentation to find the default values of these properties in order to know to what value to increase them. I think it's more useful for a user that has to deal with this exception to provide the recommendation and the default values right away in the error message.

If it's all the same to you, I sent another commit removing the link in order not to have duplicate information.

kitbellew (changes requested):
either link, or variable names but no defaults. better link. yes, people need to expect to spend some time to figure it out, you can't spoon-feed.

when you get an out of memory exception, you don't submit a bug report, you go read what that is about -- and allocate more memory. why not the same here?

rpiaggio:
Sorry, but I disagree. I think error messages should be as helpful as possible to users in order to solve the problem as quickly as possible.

I sent this PR to save time for future users who stumble upon this error. I still don't know what either of these configuration properties are.

If you are worried about duplicate logic by having the defaults here, they can be moved to a constant from where they are referenced by the code and the error message.

kitbellew:
defaults don't matter, your current settings matter more. since you can't show them, don't show anything.

following a link doesn't take too much time, but being able to explain the problem in more detail is possible on the page but not in the error message.

i appreciate your help and i think it's a good idea to provide a pointer on what to do. link to the web page is compact, requires no duplication of information or synchronization, so it's a good compromise.

PR #3684, opened 2023-11-07
MarkdownParser: add formatter using mdoc parser

### LINKED ISSUE #3672: Indented code examples in markdown are not checked
I have markdown documentation with examples which I want to check for code formatting.

My examples are often indented, and it seems that this indentation makes them ignored.

## Configuration ##

`.scalafmt.conf` reproducing the issue:

```
version = 3.7.15
#project.git = true
project.includePaths."+" = ["glob:**.md"]
maxColumn = 120
runner.dialect = Scala213Source3
fileOverride {
  "glob:**/src/main/scala-3/**" { runner.dialect = scala3 }
  "glob:**/src/test/scala-3/**" { runner.dialect = scala3 }
}

align.preset = some

rewrite.rules = [Imports, RedundantBraces, SortModifiers]
rewrite.imports.sort = scalastyle

rewrite.scala3.removeOptionalBraces = false

docstrings.blankFirstLine = no
docstrings.style = SpaceAsterisk
docstrings.wrap = no

newlines.sometimesBeforeColonInMethodReturnType = true
lineEndings=unix
```

## Command-line parameters ##

```
scalafmt --check
```

## Steps

I attempted to check formatting for:

    ```scala mdoc
    //> using dep io.scalaland::chimney::{{ git.tag or local.tag }}
    import io.scalaland.chimney.Transformer
    
    case class Foo   (a    : String)
    ```
    
    !!! example
    
        ```scala mdoc
        //> using dep io.scalaland::chimney::{{ git.tag or local.tag }}
        import io.scalaland.chimney.Transformer
    
        case class Foo   (a   : String)
        ```

While `!!! example` can be considered non-standard Markdown extension, the following is pretty standard but also doesn't work:

    ```scala mdoc
    //> using dep io.scalaland::chimney::{{ git.tag or local.tag }}
    import io.scalaland.chimney.Transformer
    
    case class Foo   (a    : String)
    ```
    
      * example inside a bullet point list:
    
        ```scala mdoc
        //> using dep io.scalaland::chimney::{{ git.tag or local.tag }}
        import io.scalaland.chimney.Transformer
    
        case class Foo   (a   : String)
        ```

## Problem

Output is:

    --- a/Users/dev/Workspaces/GitHub/kanapeczka/test.md
    +++ b/Users/dev/Workspaces/GitHub/kanapeczka/test.md
    @@ -4,3 +4,3 @@
    
    -case class Foo   (a    : String)
    +case class Foo(a: String)
     ```
    error: --test failed

It checks only non-indented code example, and skips the indented one.

## Expectation

    --- a/Users/dev/Workspaces/GitHub/kanapeczka/test.md
    +++ b/Users/dev/Workspaces/GitHub/kanapeczka/test.md
    @@ -4,3 +4,3 @@
    
    -case class Foo   (a    : String)
    +case class Foo(a: String)
     ```
    --- a/Users/dev/Workspaces/GitHub/kanapeczka/test.md
    +++ b/Users/dev/Workspaces/GitHub/kanapeczka/test.md
    @@ -14,3 +14,3 @@
    -    case class Foo   (a   : String)
    +.   case class Foo(a: String)
     ```

    error: --test failed

## Workaround

I don't have one if I want to keep my current formatting (which would degrade the UX).

## Note

I haven't checked, but probably the same situation would be with blockquites. This 
    
    example
    
    > example
    >
    > ```scala mdoc
    > type X = Foo Either Bar
    > ```

would also be valid markdown.

kitbellew:
Fixes #3672.


kitbellew:
java8 build fails, probably because mdoc-parser is not built with that version of java (according to your recent discussion on discord), so I proposed #3685 to address that.

tgodzik (approved):

PR #3685, opened 2023-11-09
YML: Make java11 primary, remove java8

kitbellew:


tgodzik (approved):
LGTM!

lefou:
Nice move. Releasing it as a patch release was not.

kitbellew:
> Nice move. Releasing it as a patch release was not.

@lefou should we release the next version as minor? cc @tgodzik

tgodzik:
That's my bad, we should have released as a minor bump. We should do that for the next one then.

lefou:
> > Nice move. Releasing it as a patch release was not.
> 
> @lefou should we release the next version as minor? cc @tgodzik

It won't hurt. But unless, you plan to roll back this change in the `3.7.x` line, it won't change much for user.

Issue #3689, opened 2023-11-11
org.scalafmt.sbt.ScalafmtSbtReporter$ScalafmtSbtError: scalafmt: [v3.7.16] corrupted class path


armanbilge:
For example: https://github.com/typelevel/toolkit/actions/runs/6830956374/job/18579714429?pr=122#step:6:25

But this seems to be widespread 🤔

seveneves:
The `org.scalafmt.interfaces.ScalafmtException: [v3.7.16] corrupted class path:` exception is caused by following when running `scalafmt` cli tool with scalafmt config set to `3.7.16`

```
Caused by: java.lang.ClassNotFoundException: org.scalafmt.config.Config
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:445)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:592)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:525)
	at org.scalafmt.dynamic.ScalafmtReflect.<init>(ScalafmtReflect.scala:21)
	at org.scalafmt.dynamic.ScalafmtModuleLoader$.$anonfun$loadClassPath$1(ScalafmtModuleLoader.scala:73)
	at scala.util.Try$.apply(Try.scala:210)
	at org.scalafmt.dynamic.ScalafmtModuleLoader$.org$scalafmt$dynamic$ScalafmtModuleLoader$$loadClassPath(ScalafmtModuleLoader.scala:71)
	at org.scalafmt.dynamic.ScalafmtModuleLoader$WithDownloader.$anonfun$load$2(ScalafmtModuleLoader.scala:37)
	at scala.util.Success.fold(Try.scala:281)
	at org.scalafmt.dynamic.ScalafmtModuleLoader$WithDownloader.load(ScalafmtModuleLoader.scala:37)
	at org.scalafmt.dynamic.ScalafmtModuleLoader$CachedProxy.load$1(ScalafmtModuleLoader.scala:56)
	at org.scalafmt.dynamic.ScalafmtModuleLoader$CachedProxy.$anonfun$load$3(ScalafmtModuleLoader.scala:57)
	at scala.util.Try$.apply(Try.scala:210)
	at org.scalafmt.dynamic.utils.ReentrantCache.getOrAddToCache(ReentrantCache.scala:41)
[...]

PR #3690, opened 2023-11-12
ScalafmtReflect: lazily load obsolete class

kitbellew:
Also, restore that same class in the repository so older versions of the dynamic ScalafmtReflect will continue functioning even if they won't be using the class. Helps with #3689.


kitbellew:
after the first commit, test start failing with the same errors as what everyone has seen (corrupted class path etc.); not because of the contents of the commit (running the tests as of the release triggers that) but just to separate from the second commit which introduces the fix.

PR #3698, opened 2023-11-22
Changed example `maxColumn` to 120, which is more Scala conventional

samthebest:
Intellij default has always been 120.  A quick search also says about 170 OS repos use 120, only 30 use 100.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -10,7 +10,7 @@ Here is an example `.scalafmt.conf`:
 
 ```scala config
 align.preset = more    // For pretty alignment.
-maxColumn = 100 // For my wide 30" display.
+maxColumn = 120 // For my wide 30" display.
 ```
 
 ## Most popular


kitbellew:
@samthebest what's the point of this change?

samthebest:
@kitbellew  When developers create a new `scalafmt` file they often copy and paste the example here, which is not consistent with the Intellij default visual aid (grey vertical line signifying the line limit), nor the majority of existing Scala repositories.  The inconsistency consequently requires every developer in that team to (potentially frequently) have to change their Intellij settings so that the visual aid readjusts.  Alternatively the team can change the scalafmt file after the initial creation, which in some teams gets contentious if there exists a developer who is particularly concerned about git history authorship preservation.

So this PR will save future developers that hassle and overhead to their work, and/or any potentially unnecessary large commits that change the authorship of multiple lines of code.

For a 1 character change to a `.md` file that seems like a pretty good trade off! 😄

kitbellew:
a quick search for "scala style guide" on Google produces the following at the top:

- https://github.com/databricks/scala-style-guide#linelength
- https://github.com/databricks/scala-style-guide#linelength
- https://flink.apache.org/how-to-contribute/code-style-and-quality-scala/ (at the bottom of the page)

not a single source refers to 120. so, it's debatable and hence makes no sense to change the example.

samthebest:
@kitbellew  

There are roughly 170 OS repositories that use 120 given by this approximate search: https://github.com/search?q=maxColumn+%3D+120&type=commits

Then about 33 for length 100 https://github.com/search?q=maxColumn+%3D+100&type=commits

Of private source industrial repositories in my personal experience I have encountered only 1 repo (of say 100) that uses a line limit of 100, and when I asked the developers "why did you pick 100 when 120 has been the ubiquitously accepted convention"? They said "because they copy and pasted the example in scalafmt docs".  I'm sure it might be possible to pay a consultancy to get broader insight into the actual line limits used in private source repositories, if you happen to have such evidence and it is contrary to my claim then I'd be happy to see it.

The reason why 120 has long been the ubiquitously accepted convention is because Intellij IDE has been the longest ubiquitously accepted IDE for Scala, which came prior to scalafmt and which picks a default of 120 characters.

Of course it's debatable, but to say it makes "no sense" I think is a little unfair to the thrust of my argument.  If every single PR that had some aspect of it that was moderately debatable was dismissed as making "no sense" then the software development industry to would slow down somewhat.

You have to be charitable to the notion that saving developers time, removing inconvenience and history authorship upheaval is a noble cause and that some trade offs, like accepting a 1 char PR, is reasonable.  Also the numbers are fairly solid here.  It is a fairly well known fact that Intellij is the dominant Scala IDE, I guess I could try to prove that with numbers also if you wish?

tgodzik:
I mean, we could change the example, but you can always find people who will shout that 100 is better or even 80. This might be subject to quite a lot of bikeshedding. Also if you look at the exact string `maxColumn = 100` and by code the difference is around 2x for using 120.

This is just an example and it's set once per repo. I don't think it would influence any remotely great gains in productivity.

tgodzik:
Anyway, thanks for the suggestion, but I don't think we need to discuss it as it will take away more time from all of us than needed, which we could spend on more fruitful collaboration.

samthebest:
@tgodzik You are actually in a position where you have pretty good oversight.  Ask your (esteemed, and highly regarded) colleagues Zbigniew or Mikolaj.  They will see a lot of repos and will have a pretty good idea of the actually used line limits used in industry.  Tell them Sam asked you to change the example to 120 ;)

@tgodzik What evidence would change your mind? What if I did actually prove that 99% of private industry repositories use 120 line limit for Scala and say 99% of Scala developers use Intellij?

If I proved that say over the course of Scala's lifetime that this MR would save a total of say 100 hours of developers time?

What threshold would convince you?

tgodzik:
I believe you might be right that a lot of repositories use 120 limit, though it's also possible that some people use the default or other numbers. My main point is that I feel we all spend more time on discussing it than most people when changing the default in their repository. You don't actually need to set it, and if you do you do it on purpose. 

Should we also remove `align.preset = more` or change that? This all becomes much more complicated. We were talking at some point about a blessed configuration that would be suggested by most people but that would require a lot of discussion which I am not sure anybody is ready to take on.

>  say 99% of Scala developers use Intellij

Well I know that is not true from their developer's survey :smile: 

And I don't believe this small change will have any influence whatsoever. What could change things is people agreeing on certain configuration or removing configuration options at all, so that this never needs to be discussed. The other less nuclear option is for the tools like Intellij to generate scalafmt with suggested maxColumn right from the start, they might be open to that.

samthebest:
> tools like Intellij to generate scalafmt with suggested maxColumn right from the start

Now that's a good idea, how about we do a deal - you reopen and merge this PR, and in the meantime I'll create a PR for the community Intellij to do exactly that. If I don't manage to implement that feature by say next March, then you can put the example back to your preferred value of 100.

@tgodzik The reason we are spending time discussing it is because you are ignoring my unassailable arguments 😆 !!  Lots of discussion is not needed, 120 is the Intellij default, there is no further discussion - why deviate from that ubiquitously held convention?

> Should we also remove align.preset = more or change that?

My PR does not concern that, there are 1000s of lines of code in this repo which we need not discuss, this PR concerns just a single character.

samthebest:
As the example currently stands it did cause a productivity hit at my current company, because our automatic scalafmt CI check kept failing PRs until we realised that it's because Intellij settings differ from our scalafmt file.  Ok it wasn't hours and hours of time lost, but still, if we can avoid it happening to others in future it seems like an altruistic  no brainer!

kitbellew:
@samthebest give up. this is an example, not a style guide. fix your tooling.

samthebest:
@kitbellew  So you would prefer that the 10s of 1000s of Scala developers across the world each update their tooling rather than you clicking a couple of buttons in Github to merge a PR?  The time savings we could make when at that scale must be many orders of magnitude multiplier of merging the PR.

Intellij has existed since around 2001 and I'm sure the default has always been 120.  Scalafmt AFAIK has existed around 2014, so Intellij predates scalafmt by 13 years.   In total the majority of Scala developers, which number in the 10s of 1000s have for about 22 years been using a line limit of 120 characters.

> Give up

This is a curious statement, it implies that you somehow see this discourse as a battle of wills, as though preserving an entrenched clearly incorrect analysis somehow threatens your sense of being a separate identity, and this takes precedence over the purely rational and scientific domain of software engineering, and even the altruistic productivity gains.

If that really is the case, and your energy for civil debate is waning, I would kindly, with the greatest of intentions, suggest that instead of sending short stabby comments, which provide no new information or reasoning approaches, that you go for a walk or do some leisure activity - allow your subconscious to ponder the problem for you.  When refreshed, ask yourself the question "what are the objective trade offs", what objectively would be the harm to you or others if an example of a line limit scalafmt happened to match and agree with Intellij's default?  Is there actually any logical or scientific basis to your objection? Perhaps such an objection may exist, and I would be happy to hear it - my position can be easily updated if presented new and relevant facts.

tgodzik:
Thanks @samthebest for your efforts, but I really believe this discussion leads to nowhere, starting to get heated and the change wouldn't make a big difference. Locking the discussion for now, sorry about that, but open source is already hard as it is and we do need to spend out energy elsewhere

PR #3700, opened 2023-11-23
Exclude "of" from default AvoidInfix configuration (used in ScalaTest)

### LINKED ISSUE #3699: Allow usage of "behavior of" (used in ScalaTest) by default when AvoidInfix is enabled
## Configuration ##

```
version = 3.7.17

rewrite {
  rules = [AvoidInfix]
}
```

## Steps

Given code like this:
```scala
behavior of "XYZ"
```

## Problem

```scala
behavior.of("XYZ")
```

## Expectation

```scala
behavior of "XYZ"
```

## Workaround

A workaround is to add the following line to the configuration:

```
rewrite.neverInfix.exclude."+" = "of"
```

However it seems to be an unnecessary tediousness and complexity of configuration especially given that most other ScalaTest infix keywords are already excluded from `AvoidInfix` rule by default.

## Proposed solution

I am willing to submit a pull request (PR) that adds "of" to the default `neverInfix.exclude` list.

However, it's worth noting that the "behavior of" example is used in tests as [a manual token append demonstration](https://github.com/scalameta/scalafmt/blob/master/scalafmt-tests/src/test/resources/rewrite/AvoidInfix2.stat#L18-L35). We could potentially choose a different keyword in the test scenario to make the change.
I would like to confirm this before contributing.

piter75:
Fixes #3699

It also:
 - makes the `more ScalaTest` formatting test a bit more realistic
 - uses some arbitrary infix call in excludeFilters list append test

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Pattern.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/Pattern.scala
@@ -37,6 +37,7 @@ object Pattern {
       "in",
       "ignore",
       "be",
+      "of",
       "taggedAs",
       "thrownBy",
       "synchronized",
--- a/scalafmt-tests/src/test/resources/rewrite/AvoidInfix.stat
+++ b/scalafmt-tests/src/test/resources/rewrite/AvoidInfix.stat
@@ -71,9 +71,19 @@ a banana walked have ordered a martini
 >>>
 (a.banana(walked) have ordered).a(martini)
 <<< more scalatest
-cp should not contain (jar3.getPath())
+class SomeScalaTest {
+  behavior of "XYZ"
+  it should "pass" in {
+    cp should not contain (jar3.getPath())
+  }
+}
 >>>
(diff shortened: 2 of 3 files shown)


kitbellew:
@piter75 thank you for this submission. there's an alternative, #3711, which takes a slightly more conservative approach of excluding only `behavior.of`, as opposed to `*.of` which might be a little too wide.

kitbellew:
> @kitbellew The "wideness" of the exclusion crossed my mind but I decided it's ok if we already exclude both `in` and `be`.

is there a way to limit `in` or `be`?

piter75:
IMHO `be` always comes after `should` or `must` in assertions but I would have to do some research to be certain

`in` is used in both `FlatSpec` and `FreeSpec` [styles](https://www.scalatest.org/user_guide/selecting_a_style) which implies that it comes after a String "literal".
Maybe we could narrow it down to `" in`.
Someone may probably store the preceding string in a variable but that would (most probably) be rare.

Those changes could potentially be backwards incompatible so maybe it's not worth changing it.

PR #3705, opened 2023-11-24
NeverInfixPattern: allow to match lhs as well

kitbellew:
Helps with #3699.

### DIFF
--- a/scalafmt-tests/src/test/resources/rewrite/AvoidInfix2.stat
+++ b/scalafmt-tests/src/test/resources/rewrite/AvoidInfix2.stat
@@ -71,6 +71,6 @@ behaviour of "..." {
 behavior.of("..." {
   a.shouldBe(b)
 })
-behaviour.of("..." {
+behaviour of "..." {
   a.shouldBe(b)
-})
+}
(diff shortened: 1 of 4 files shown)


scalafmt-tests/src/test/resources/rewrite/AvoidInfix2.stat
```diff
@@ -71,6 +71,6 @@ behaviour of "..." {
 behavior.of("..." {
```
tgodzik, line 71: How is `behavior` actually matched here? The exclude filter has `behaviour`

tgodzik: ach makes sense! I was thinking the other way around

tgodzik (approved):

scalafmt-tests/src/test/resources/rewrite/AvoidInfix2.stat
kitbellew: i know... i am frequently confused myself with the double negative ("never" and "exclude"), plus a little bit with `neverInfix` when used with `AvoidInfix` :)

PR #3707, opened 2023-11-24
Changed example line limit to agree with Intellij default and industr…

samthebest:
The reason why 120 has long been the ubiquitously accepted convention is because Intellij IDE has been the longest ubiquitously accepted IDE for Scala, which came prior to scalafmt and which picks a default of 120 characters.

There are roughly 170 OS repositories that use 120 given by this approximate search: https://github.com/search?q=maxColumn+%3D+120&type=commits

Then about 33 for length 100 https://github.com/search?q=maxColumn+%3D+100&type=commits

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -10,7 +10,7 @@ Here is an example `.scalafmt.conf`:
 
 ```scala config
 align.preset = more    // For pretty alignment.
-maxColumn = 100 // For my wide 30" display.
+maxColumn = 120 // For my wide 30" display.
 ```
 
 ## Most popular


kitbellew:
@samthebest please stop this.

tgodzik:
@samthebest Please respect the maintainers' decision. @kitbellew is the main contributor, who does the majority of work here (like 99%), so I trust him to make a proper decision. I don't think it's worth you time and effort. 

What would probably be better is raising a post on contributors discussing  what would be the best starting configuration, which we would be able to put into recommendation at the start here.

tgodzik:
I raised a PR to make it more explicit that you should not use example configuration https://github.com/scalameta/scalafmt/pull/3710

PR #3710, opened 2023-11-27
improvement: Warn users not to use an example config without discussion

tgodzik:
image

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -10,8 +10,10 @@ Here is an example `.scalafmt.conf`:
 
 ```scala config
 align.preset = more    // For pretty alignment.
-maxColumn = 100 // For my wide 30" display.
+maxColumn = 1234
 ```
+
+> 🚧 Before using specific configuration make sure that your project agrees on the standards. Settings such as `maxColumn` could be a source of issues if different tools such as an IDE uses a different default value.
 
 ## Most popular
 


kitbellew:
it's a good solution. however, do we really need to mention that someone uses a particular value? just below, in the section about `maxColumn`, we explicitly caution against values above 100...

intellij applies 120 as a "hard right margin" (i e., maximum), regardless of language, with the option of adding the so-called "visual guides" at additional stops.

and its original -- and, presumably, primary -- language is java, which has only one formatter (Google Java format) that uses the value of 100 and allows no customization.

not a single formal language style, for scala or Java, which can be found online uses the value of 120 (scala-lang sort of mentions 80, while several others set it to 100).

and, as I mentioned, intellij's global value is really a "nobody should go beyond that since it doesn't fit the screen anymore" rather than "let's pack as many characters as possible before this marker".

in addition to "you should discuss internally" comment, but without a specific value suggested, we can change the example to something ridiculous, like 1234, so that no sane person uses it.

docs/configuration.md
```diff
@@ -10,9 +10,11 @@ Here is an example `.scalafmt.conf`:
 
 ```scala config
 align.preset = more    // For pretty alignment.
-maxColumn = 100 // For my wide 30" display.
+maxColumn = 1234
 ```
 
+> 🚧 Before using specific configuration make sure that your project agrees on the standards. Settings such as `maxColumn` could be a source of issues if different tools such as an IDE uses a different default value.
```
tgodzik, line 16: I changed the example and made the additional comment about the specific problem users might encounter.

kitbellew (approved):

PR #3711, opened 2023-11-27
AvoidInfixSettings: add `behavior.of` to default

### LINKED ISSUE #3699: Allow usage of "behavior of" (used in ScalaTest) by default when AvoidInfix is enabled

kitbellew:
Fixes #3699.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/AvoidInfixSettings.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/AvoidInfixSettings.scala
@@ -74,6 +74,7 @@ object AvoidInfixSettings {
       "in",
       "ignore",
       "be",
+      "behavior\\.of",
       "taggedAs",
       "thrownBy",
       "synchronized",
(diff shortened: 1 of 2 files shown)


piter75:
@kitbellew thanks for fixing this

PR #3729, opened 2023-12-08
bugfix: Show exception from path matcher correctly

tgodzik:
Previously, we would just get a stack trace:
```
Exception in thread "main" java.lang.IllegalArgumentException
	at sun.nio.fs.UnixFileSystem.getPathMatcher(UnixFileSystem.java:286)
	at org.scalafmt.config.ProjectFiles$FileMatcher$Nio.<init>(ProjectFiles.scala:74)
	at org.scalafmt.config.ProjectFiles$FileMatcher$.$anonfun$nio$1(ProjectFiles.scala:70)
	at scala.collection.immutable.List.map(List.scala:250)
	at scala.collection.immutable.List.map(List.scala:79)
	at org.scalafmt.config.ProjectFiles$FileMatcher$.create(ProjectFiles.scala:69)
	at org.scalafmt.config.ProjectFiles$FileMatcher$.nio(ProjectFiles.scala:70)
	at org.scalafmt.config.ProjectFiles$FileMatcher$.apply(ProjectFiles.scala:64)
	at org.scalafmt.cli.ScalafmtCoreRunner$.$anonfun$run$2(ScalafmtCoreRunner.scala:27)
	at metaconfig.Configured$ConfiguredImplicit.fold(Configured.scala:111)
	at org.scalafmt.cli.ScalafmtCoreRunner$.run(ScalafmtCoreRunner.scala:23)
	at org.scalafmt.cli.Cli$.runWithRunner(Cli.scala:140)
	at org.scalafmt.cli.Cli$.run(Cli.scala:91)
	at org.scalafmt.cli.Cli$.mainWithOptions(Cli.scala:61)
	at org.scalafmt.cli.Cli$.main(Cli.scala:43)
	at org.scalafmt.cli.Cli.main(Cli.scala)
```

Reported in https://github.com/VirtusLab/scala-cli/issues/2415

### DIFF
--- a/scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtCoreRunner.scala
+++ b/scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtCoreRunner.scala
@@ -6,6 +6,7 @@ import util.control.Breaks
 import org.scalafmt.Error
 import org.scalafmt.{Formatted, Scalafmt, Versions}
 import org.scalafmt.config.{ProjectFiles, ScalafmtConfig}
+import org.scalafmt.config.ScalafmtConfigException
 import org.scalafmt.CompatCollections.ParConverters._
 
 import scala.meta.parsers.ParseException
@@ -21,10 +22,18 @@ object ScalafmtCoreRunner extends ScalafmtRunner {
       ExitCode.UnexpectedError
     } { scalafmtConf =>
       options.common.debug.println(s"parsed config (v${Versions.version})")
-      val filterMatcher = ProjectFiles.FileMatcher(
-        scalafmtConf.project,
-        options.customExcludes
-      )
+      val filterMatcher =
+        try {
+          ProjectFiles.FileMatcher(
+            scalafmtConf.project,
+            options.customExcludes
+          )
+        } catch {
+          case e: ScalafmtConfigException =>
+            options.common.err.println(e.getMessage())
+            return ExitCode.UnexpectedError // RETURNING EARLY!
+        }
+
       val inputMethods = getInputMethods(options, filterMatcher.matchesPath)
 
       val termDisplay =
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
@@ -70,11 +70,27 @@ object ProjectFiles {
     private def regex(seq: Seq[String]) = create(seq, new Regex(_))
 
     private final class Nio(pattern: String) extends PathMatcher {
-      private val matcher = fs.getPathMatcher(pattern)
+      private val matcher =
+        try { fs.getPathMatcher(pattern) }
+        catch {
+          case _: IllegalArgumentException =>
+            throw new ScalafmtConfigException(
+              s"Illegal pattern in configuration: $pattern"
+            )
+        }
       def matches(path: file.Path): Boolean = matcher.matches(path)
     }
     private final class Regex(regex: String) extends PathMatcher {
-      private val pattern = java.util.regex.Pattern.compile(regex)
+      private val pattern =
+        try { java.util.regex.Pattern.compile(regex) }
+        catch {
+          case e: java.util.regex.PatternSyntaxException =>
+            throw new ScalafmtConfigException(
+              s"""|Illegal regex in configuration: $regex
+                |reason: ${e.getMessage()}""".stripMargin
+            )
+        }
+
       def matches(path: file.Path): Boolean =
         pattern.matcher(path.toString).find()
     }
--- a/scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
+++ b/scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
@@ -798,6 +798,64 @@ class CliTest extends AbstractCliTest with CliTestBehavior {
   testCli("1.6.0-RC4") // test for runDynamic
   testCli(stableVersion) // test for runScalafmt
 
+  test(s"path-error") {
+    val input =
+      s"""|/.scalafmt.conf
+        |version = "${stableVersion}"
+        |project.excludePaths = [
+        |    "glob:**/src/main/scala/besom/rpc/**.scala",
+        |    "foo.scala"
+        |]
+        |/foo.scala
+        |object    A { foo( }
+        |""".stripMargin
+
+    noArgTest(
+      string2dir(input),
+      input,
+      Seq(Array("--test")),
+      ExitCode.UnexpectedError,
+      assertOut = out => {
+        assertContains(
+          out,
+          s"""Illegal pattern in configuration: foo.scala""".stripMargin
+        )
+      },
+      Some(ExitCode.UnexpectedError)
+    )
+
+  }
+
+  test(s"regex-error") {
+    val input =
+      s"""|/.scalafmt.conf
+        |version = "${stableVersion}"
+        |project.excludeFilters = [
+        |    ".*foo("
+        |]
+        |/foo.scala
+        |object    A { foo( }
+        |""".stripMargin
+
+    noArgTest(
+      string2dir(input),
+      input,
+      Seq(Array("--test")),
+      ExitCode.UnexpectedError,
+      assertOut = out => {
+        assertContains(
+          out,
+          """|Illegal regex in configuration: .*foo(
+            |reason: Unclosed group near index 6
+            |.*foo(
+            |""".stripMargin
+        )
+      },
+      Some(ExitCode.UnexpectedError)
+    )
+
+  }
+
   test("Fail if .scalafmt.conf is missing.") {
     val input =
       s"""|/foo.scala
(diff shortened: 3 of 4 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
```diff
@@ -70,7 +70,14 @@ object ProjectFiles {
     private def regex(seq: Seq[String]) = create(seq, new Regex(_))
 
     private final class Nio(pattern: String) extends PathMatcher {
-      private val matcher = fs.getPathMatcher(pattern)
+      private val matcher =
+        try { fs.getPathMatcher(pattern) }
+        catch {
+          case _: IllegalArgumentException =>
+            throw new ScalafmtConfigException(
```
kitbellew, line 77: we should do the same in `Regex` below, and also catch `ReflectionException` in `ScalafmtReflectConfig.projectMatcherField`.

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
```diff
@@ -32,8 +32,9 @@ abstract class AbstractCliTest extends FunSuite {
   ): Unit =
     run(Cli.getConfig(args, init).get, exitCode)
 
-  def run(options: CliOptions, exitCode: ExitCode = ExitCode.Ok) =
+  def run(options: CliOptions, exitCode: ExitCode = ExitCode.Ok) = {
```
kitbellew, line 35: leftover from a debugging session? :)

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
```diff
@@ -798,6 +799,34 @@ class CliTest extends AbstractCliTest with CliTestBehavior {
   testCli("1.6.0-RC4") // test for runDynamic
   testCli(stableVersion) // test for runScalafmt
 
+  test(s"path-error") {
```
kitbellew, line 802: should we add a similar test to the dynamic suite?

kitbellew (changes requested):

scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtCoreRunner.scala
```diff
@@ -21,26 +22,32 @@ object ScalafmtCoreRunner extends ScalafmtRunner {
       ExitCode.UnexpectedError
     } { scalafmtConf =>
       options.common.debug.println(s"parsed config (v${Versions.version})")
-      val filterMatcher = ProjectFiles.FileMatcher(
-        scalafmtConf.project,
-        options.customExcludes
-      )
-      val inputMethods = getInputMethods(options, filterMatcher.matchesPath)
+      try {
```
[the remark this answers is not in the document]
tgodzik: 1. If we don't catch the exception then the stack trace is shown, which is not necessary for the user, but is fine overall.
2. We could, it doesn't make sense to do anything else if filterMatcher failed. We could get the wrong files formatted. I can change it to an option and do nothing on None, but it seemed to use a lot more code. What do you think?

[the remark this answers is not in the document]
tgodzik: This is the alternative, what do you think:
```
override private[cli] def run(
      options: CliOptions,
      termDisplayMessage: String
  ): ExitCode =
    options.scalafmtConfig.fold { e =>
      options.common.err.println(s"${e.msg}")
      ExitCode.UnexpectedError
    } { scalafmtConf =>
      options.common.debug.println(s"parsed config (v${Versions.version})")
      val exitCode = new AtomicReference(ExitCode.Ok)
      val filterMatcher =
        try {
          Some(
            ProjectFiles.FileMatcher(
              scalafmtConf.project,
              options.customExcludes
            )
          )
        } catch {
          case e: ScalafmtConfigException =>
            options.common.err.println(e.getMessage())
            exitCode.set(ExitCode.UnexpectedError)
            None
        }
      filterMatcher match {
        case None =>
        case Some(filterMatcher) =>
          val inputMethods = getInputMethods(options, filterMatcher.matchesPath)
          val termDisplay =
            newTermDisplay(options, inputMethods, termDisplayMessage)
          Breaks.breakable {
            inputMethods.par.foreach { inputMethod =>
              val code = handleFile(inputMethod, options, scalafmtConf)
              exitCode.getAndUpdate(ExitCode.merge(code, _))
              if (options.check && !code.isOk) Breaks.break
              termDisplay.taskProgress(termDisplayMessage)
            }
          }
          termDisplay.completedTask(termDisplayMessage, exitCode.get.isOk)
          termDisplay.stop()
      }
      exitCode.get()
    }
```

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
tgodzik: I think the behaviour will be different in older version, so the test would fail initially.

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
tgodzik: removed!

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
tgodzik: Added with a new test case

scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtCoreRunner.scala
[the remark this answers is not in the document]
kitbellew: what about something as simple as this: replace
```
          val filterMatcher = ProjectFiles.FileMatcher(
            scalafmtConf.project,
            options.customExcludes
          )
```
with
```
          val filterMatcher =
            try {
              ProjectFiles.FileMatcher(
                scalafmtConf.project,
                options.customExcludes
              )
            } catch {
              case e: ScalafmtConfigException =>
                options.common.err.println(e.getMessage())
                return ExitCode.UnexpectedError // RETURNING EARLY!
            }
```

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
kitbellew: for older versions, we could use a check that is appropriate to them, but for this check start with `nightly` (since it actually dynamically loads the current build).

tgodzik:
btw. I think using `regex:` filter will not work currently? `use `excludePaths` with `regex:` prefix` but we only use `nio()` on excludePaths currently.

scalafmt-tests/src/test/scala/org/scalafmt/cli/CliTest.scala
tgodzik: Added two tests to DynamicSuite

scalafmt-cli/src/main/scala/org/scalafmt/cli/ScalafmtCoreRunner.scala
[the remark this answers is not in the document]
tgodzik: Done

kitbellew:
> btw. I think using `regex:` filter will not work currently? `use `excludePaths`with`regex:` prefix` but we only use `nio()` on excludePaths currently.

this i didn't fully understand. `nio()` uses `FileSystem.getPathMatcher` which supports `glob:` and `regex:` prefixes.

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
```diff
@@ -70,11 +70,26 @@ object ProjectFiles {
...
+              s"Illegal pattern in configuration: $pattern"
```
kitbellew, line 78: what happens if we append message from the exception we caught? will it explain the reason better?

tgodzik:
> > btw. I think using `regex:` filter will not work currently? `use `excludePaths`with`regex:` prefix` but we only use `nio()` on excludePaths currently.
> 
> this i didn't fully understand. `nio()` uses `FileSystem.getPathMatcher` which supports `glob:` and `regex:` prefixes.

Ach, ok I didn't know `nio` also supported regexes

kitbellew (changes requested):
lgtm, only that leftover blank line :)

scalafmt-core/shared/src/main/scala/org/scalafmt/config/ProjectFiles.scala
tgodzik: I tried and it seems empty.

tgodzik: Ach, actually for Regex we can add it, it's null for Nio matcher

tgodzik: Added `reason: ...` line

kitbellew (approved):

PR #3732, opened 2023-12-23
[PoC] Extract convert import wildcard to start as independent rewrite

kaatzee:
I would like to know if such change has a chance to be added.
The scenario in which such functionality is needed is when I want to rewrite underscores in imports to stars, but leave possibility to use both old and new syntax for `if` control syntax.
Currently when I set:
- `rewrite.scala3.convertToNewSyntax = true` and `allowStarWildcardImport = true` and `allowSignificantIndentation = true` all old `if` syntax is rewritten to new one
- `rewrite.scala3.convertToNewSyntax = true` and `allowStarWildcardImport = true` and `allowSignificantIndentation = false` it fails on places where new syntax of `if` is used.

Let me know what do you think and what further changes are needed to add this functionality.


kitbellew:
@kaatzee what's the reason for this requirement? are import underscores deprecated whereas "old" `if` syntax is not?

kaatzee:
@kitbellew yes, exactly. I have seen your PR which addresses this approach. Thank you! I am closing this PR then.

PR #3734, opened 2023-12-23
FormatWriter: fix end marker handling

### LINKED ISSUE #3731: Config `insertEndMarkerMinLines` adds end marker for method counting scaladoc
I have a method that has 9 lines but has 11 lines of scaladoc as:

```scala
/**
 * Tries to return a message but might randomly fail. If it fails, retry. A
 * success percent can be passed.
 *
 * @param msg
 *   is the message to be printed
 * @param successPercent
 *   is the percentage of success the effect will have. Defaults to 50%.
 * @return
 *   ZIO[Any, Exception, String]
 */
def eventuallyFail(msg: String, successPercent: Int = 50) =
    require(successPercent >= 0 && successPercent <= 100)
    for
        percent <- Random.nextIntBetween(0, 100)
        _ <- ZIO.when(percent > successPercent):
            ZIO.fail(Exception(msg))
        r <- ZIO.succeed(msg)
    yield r
end eventuallyFail
```

Using the config below, scalafmt adds the `end` marker. I understand the scaladoc lines shouldn't be counted towards the `insertEndMarkerMinLines` config.

```
rewrite.scala3.insertEndMarkerMinLines = 15
rewrite.scala3.removeEndMarkerMaxLines = 14
```

I'm using scalafmt version = 3.7.15.

kitbellew:
The current issue is that the span is computed using `leftLineId` of the token just before and the token at the end. However, `leftLineId` refers to the beginning of the token, thus multiline comments are incorrectly accounted.

Instead, use the token at the beginning of the span and increase by 1. Fixes #3731.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatToken.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatToken.scala
@@ -98,7 +98,15 @@ object FormatToken {
   ) {
     lazy val firstNL = text.indexOf('\n')
     @inline def hasNL: Boolean = firstNL >= 0
-    def countNL: Int = if (hasNL) text.count(_ == '\n') else 0
+    def countNL: Int = {
+      var cnt = 0
+      var idx = firstNL
+      while (idx >= 0) {
+        cnt += 1
+        idx = text.indexOf('\n', idx + 1)
+      }
+      cnt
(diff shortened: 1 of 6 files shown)


tgodzik (approved):
LGTM!

kitbellew:
@tgodzik thank you, happy holidays!

PR #3735, opened 2023-12-24
ConvertToNewScala3Syntax: use the new granular parameter

kitbellew:
Supersedes #3732.


tgodzik:
Do we really need this to be that granular? Maybe it's enough to have separate changes for optional braces and separate for things that are not affected by it?

kitbellew:
@tgodzik i thought about it. looked at braces rewrite which has over time become rather granular, same for indent which had originally been declared inviolate and fixed, and decided to avoid further requests 🙂

kitbellew:
@tgodzik agreed, let's merge `if` and `while`. lump `for` with these two?

kitbellew:
also, since i am not a big expert on backports to scala213 (via `...source3`), would `importAs` and `importStar` make sense as a single flag (say, simply `imports`)?

scalafmt-core/shared/src/main/scala/org/scalafmt/config/RewriteScala3Settings.scala
```diff
@@ -56,4 +60,26 @@ object RewriteScala3Settings {
...
 
+  case class ConvertToNewSyntax(
+      ifCond: Boolean = true,
+      whileCond: Boolean = true,
+      forEnum: Boolean = true,
+      varargStar: Boolean = true,
+      importStar: Boolean = true,
+      importAs: Boolean = true,
+      typeWildcardQuestionMark: Boolean = true,
+      typePlaceholderUnderscore: Boolean = true
```
tgodzik, line 71:
```suggestion
      deprecatedSyntax = true
```
maybe?  This would be controlled by the dialect to see if the new syntax is allowed.

Otherwise if only one thing is set then the user will be quite confused. The separate non optional braces (traditional braced if etc.) will probably not be deprecated ever,.

kitbellew:
@tgodzik all done

tgodzik (approved):
LGTM!

PR #3761, opened 2024-01-11
FormatOps: force NL for non-single-expr in finally

kitbellew:
Fixes a follow-up issue reported in #3653.


mtomko:
Amazing. Thanks! I forked scalafmt last night to try and figure it out myself, but there's no way I could have figured out enough to do this on my own.

tgodzik (approved):

PR #3797, opened 2024-02-23
SortSettings: implement the `styleGuide` preset

### LINKED ISSUE #3788: SortModifiers does not follow Scala style guide
The [Scala Style Guide](https://docs.scala-lang.org/style/declarations.html#modifiers) suggest the following:

> **Modifiers**
>Method modifiers should be given in the following order (when each is applicable):
>
>1. Annotations, each on their own line
>2. Override modifier (override)
>3. Access modifier (protected, private)
>4. Implicit modifier (implicit)
>5. Final modifier (final)
>6. def
> ```scala
>@Transaction
>@throws(classOf[IOException])
>override protected final def foo(): Unit = {
>  ...
>}
>```

Is the current sorting of the modifiers correct and the Scala Style Guide is outdated or is it the other way around?

```
version = 3.8.0 - main branch
...
object SortSettings {
  final case class ModKey(
      name: String,
      matches: Mod => Boolean
  )

  val defaultOrder: List[ModKey] = List(
    ModKey("implicit", _.is[Mod.Implicit]),
    //
    ModKey("final", _.is[Mod.Final]),
    ModKey("sealed", _.is[Mod.Sealed]),
    ModKey("abstract", _.is[Mod.Abstract]),
    //
    ModKey("override", _.is[Mod.Override]),
    //
    ModKey("private", _.is[Mod.Private]),
    ModKey("protected", _.is[Mod.Protected]),
    //
    ModKey("lazy", _.is[Mod.Lazy]),
    ModKey("open", _.is[Mod.Open]),
    ModKey("transparent", _.is[Mod.Transparent]),
    ModKey("inline", _.is[Mod.Inline]),
    ModKey("infix", _.is[Mod.Infix]),
    ModKey("opaque", _.is[Mod.Opaque])
  )

```

When I run scalafmt via CLI like this: `scalafmtAll`

## Steps

Given code like this:
```scala
override protected final def foo(): Unit = {
  ...
}
```

## Problem

```scala
final override protected def foo(): Unit = {
  ...
}
```

## Expectation

```scala
override protected final def foo(): Unit = {
```

kitbellew:
Fixes #3788.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -3241,6 +3241,29 @@ private final case class A2(a: Int)
     extends ADT
 ```
 
+If you prefer to use the order based on the one
+[partially](https://docs.scala-lang.org/style/declarations.html#modifiers)
+specified in the Scala Style Guide, you can use the corresponding preset:
+
+```scala mdoc:scalafmt
+rewrite.rules = [SortModifiers]
+rewrite.sortModifiers.preset = styleGuide
+---
+final lazy private implicit val x = 42
+lazy final implicit private val y = 42
+
+class Test(
+    implicit
+    final private val i1: Int,
+    private final val i2: String
+)
+
+sealed protected[X] trait ADT
+final private case object A1 extends ADT
+private final case class A2(a: Int)
+    extends ADT
+```
+
 If you choose a custom sort order, you can specify some or all modifiers
 in the desired order:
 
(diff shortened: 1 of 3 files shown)


docs/configuration.md
```diff
@@ -3241,6 +3241,29 @@ private final case class A2(a: Int)
     extends ADT
 ```
 
+If you prefer to use the order based on the one
+[partially](https://docs.scala-lang.org/style/declarations.html#modifiers)
+specified in the Scala Style Guide, you can use the corresponding preset:
+
+```scala mdoc:scalafmt
+rewrite.rules = [SortModifiers]
+rewrite.sortModifiers.preset = styleGuide
```
tgodzik, line 3250: Would probably be awesome to have this as a global style that would setup up most needed things. Just random thought

tgodzik (approved):

docs/configuration.md
kitbellew: hmm... i don't know what that entails, to be honest. the style guide hasn't been updated or maintained in a while, but i am happy to let someone who feels strongly about this, to take a stab.

PR #3805, opened 2024-03-01
`:` could be from the single term in ArgClause

kitbellew:
For instance, Term.Block or Term.PartialFunction.


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/FormatOps.scala
```diff
@@ -2093,52 +2093,72 @@ class FormatOps(
               def rightBrace = treeLast(t)
             })
           case t: Term.ArgClause if tokens.getHead(t) eq ft =>
-            def funcSplit(arg: Term.FunctionTerm)(implicit fl: FileLine) = {
```
kitbellew, line 2096: easier to review this with `?w=1`

PR #3815, opened 2024-03-06
RemoveScala3OptionalBraces: handle fewer braces

### LINKED ISSUE #3812: Implement fewer-braces rewrite

kitbellew:
Fixes #3812.


docs/configuration.md
```diff
@@ -3698,6 +3698,13 @@ The section contains the following settings (available since v3.8.1):
     - other flags below might extend rewrites to other cases
 - `oldSyntaxToo`
   - if `true`, applies also to expressions using deprecated syntax
+- `fewerBracesMinStats` and `fewerBracesMaxStats`
```
tgodzik, line 3701: Let's maybe add some info about the defaults? I assume it's 0?

scalafmt-tests/src/test/resources/scala3/FewerBraces.stat
```diff
@@ -1802,3 +1802,312 @@ object a:
...
+  }
+>>>
+foo
+  .mtd1 { x =>
+    x + 1
+  }
+  .mtd2: x =>
+     x + 1
+     x + 2
+  .mtd3 { x =>
```
tgodzik, line 1866: I wouldn't count the parameter, I don't think it's very intuitive

scalafmt-tests/src/test/resources/scala3/FewerBraces.stat
```diff
@@ -1802,3 +1802,312 @@ object a:
...
+      case z => z + 1
+    }
+  }
+>>>
+foo
+  .mtd1 {
+    bar match
+       case x => x + 1
+  }
+  .mtd2:
```
tgodzik, line 2056: I would say it's only one statement and not count the cases inside.

kitbellew: I'll clarify in the codr

scalafmt-tests/src/test/resources/scala3/FewerBraces.stat
kitbellew: i think it counts the invisible block which contains the two statements, not the parameter. checking whether the block is invisible occasionally leads to non idempotent formatting as another rule could also remove the braces. but let me think if I can figure that out.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RemoveScala3OptionalBraces.scala
```diff
@@ -134,7 +154,73 @@ private class RemoveScala3OptionalBraces(implicit val ftoks: FormatTokens)
...
+      else {
+        val next = queue.dequeue()
+        def enqueue(trees: Iterable[Tree]*): Int = {
+          val len = queue.length
+          trees.foreach(queue ++= _)
+          queue.length - len
+        }
+        val delta = next match {
+          case x: Term.Block => enqueue(x.stats)
+          case x: Tree.WithBody => enqueue(x.body :: Nil) - 1
```
kitbellew, line 198: as you can see here, -1 means that the parent expression is not counted and fully replaced by the statements in the body. this applies, for instance, to a function, which doesn't count the parameters.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RemoveScala3OptionalBraces.scala
```diff
@@ -134,7 +154,73 @@ private class RemoveScala3OptionalBraces(implicit val ftoks: FormatTokens)
...
+      else if (queue.isEmpty) cnt >= rob.fewerBracesMinStats
+      else {
+        val next = queue.dequeue()
+        def enqueue(trees: Iterable[Tree]*): Int = {
+          val len = queue.length
+          trees.foreach(queue ++= _)
+          queue.length - len
+        }
+        val delta = next match {
+          case x: Term.Block => enqueue(x.stats)
```
kitbellew, line 197: this is where the block, including one without braces, is counted

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RemoveScala3OptionalBraces.scala
```diff
@@ -134,7 +154,73 @@ private class RemoveScala3OptionalBraces(implicit val ftoks: FormatTokens)
...
+          case x: Tree.WithBody => enqueue(x.body :: Nil) - 1
+          case x: Tree.WithCases => enqueue(x.cases)
```
kitbellew, line 199: you mentioned not counting cases in a match statement. this line applies to partial functions as well, should we not count them either?

kitbellew (commented):
in general, it would have been ideal to count the number of lines and not the statements. however, this is performed before the formatting, and the number of output lines is not known at that point.

counting the number of input lines will lead to non idempotent formatting, as that will change. but the tree won't change, and so the number of statements.

in general, i was trying to avoid rewriting very short blocks, e.g. `foo { bar }`, which might look better on one line, especially since enabled redundant braces actually causes that to be charged to parens _after_ the formatting, hence the minimum statements parameter. but we could also permanently fix it to, say, 2, if we can figure out how to exclude invisible blocks, and remove the parameter.

but the biggest one is the maximum. i think that rewriting a block which is way too large might make the code unreadable. but one can argue that in that case the code was unreadable to begin with, since very large blocks should have been refactored as separate methods.

what do you think? here are the two options, specifically:
- keep it very configurable, with both parameters; in that case, why wouldn't we count cases or other nested blocks?
- keep it simple, replace both parameters with a single boolean, determine minimum size ourselves and do not restrict maximum, if that's how the function argument is written?

tgodzik:
> keep it very configurable, with both parameters; in that case, why wouldn't we count cases or other nested blocks?

I would opt for that, since I can imagine people wanting fewer braces, but not at all possible cases, which without the additional configuration, would be very unfriendly to use.

The only issue I see is that it should be obvious what the parameter does. And currently users would need to understand inner working of the parser to see what is going on. I would try to make it as simple as possible and only count statements inside the body of the block at the toplevel.

kitbellew:
@tgodzik sorry, it's been taking a while, found a few rare bugs and non-idempotent behaviour between rewrite rules. please expect a flurry of changes to fix them, before this one is refreshed. all tracked under #3812.

scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RemoveScala3OptionalBraces.scala
kitbellew: i have decided to go with the same approach that is used for [`runner.optimizer.forceConfigStyleOnOffset` parameter](https://scalameta.org/scalafmt/docs/configuration.html#forcing-config-style), by counting total span of non-whitespace tokens.

that was the only approach that is easy to describe, easy to implement and is not subject to non-idempotence problems (since we know which tokens are being removed during the rewrite session).

docs/configuration.md
```diff
@@ -3698,6 +3698,13 @@ The section contains the following settings (available since v3.8.1):
     - other flags below might extend rewrites to other cases
 - `oldSyntaxToo`
   - if `true`, applies also to expressions using deprecated syntax
+- `fewerBracesMinNonWsSpan` and `fewerBracesMaxNonWsSpan`
```
tgodzik, line 3701:
```suggestion
- `fewerBracesMinSpan` and `fewerBracesMaxSpan`
```
I would try to make it a bit less explicit, if anyone needs to check what is exactly being done they will anyway have to check the docs.

tgodzik (approved):
Just one minor nitpick

docs/configuration.md
kitbellew: thank you, done!

tgodzik: Thanks!

PR #3834, opened 2024-03-16
RedundantBraces: fix "moving" braces around func

kitbellew:
Previously, we'd handle the case of `foo(x => { ... })` by:

1. replacing the `(` with a `{`
2. "pretending" to replace the `{` with a `(`
3. upon encountering the `}` for a fake `(`, removing the `(` and changing on the owner of the `}`
4. then removing the `)`

Instead, let's modify steps 2, 3 and 4 as follows:

2. remove `{`
3. remove `}`
4. replace the closing `)` with a `}` moving it to the position of the `}` in step 3, along with token ownership.


tgodzik (approved):
LGTM!

PR #3840, opened 2024-03-16
RemoveScala3OptionalBraces: allow leading infix op

kitbellew:
Previously, we didn't allow it as the scalameta library didn't provide a way to check when it was actually possible. Helps with #3812.

### DIFF
--- a/scalafmt-tests/src/test/resources/scala3/OptionalBraces_unfold.stat
+++ b/scalafmt-tests/src/test/resources/scala3/OptionalBraces_unfold.stat
@@ -5373,12 +5373,12 @@ object a:
        s"($c) "
      }
    ) +
-     this.match {
-       case MateAdvice(seq, _, _, _) =>
-         seq.desc
-       case CpAdvice(judgment, _, _) =>
-         judgment.toString
-     } + "."
+     this.match
+        case MateAdvice(seq, _, _, _) =>
+          seq.desc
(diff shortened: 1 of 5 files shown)


scalafmt-tests/src/test/resources/scala3/OptionalBraces_unfold.stat
```diff
@@ -5373,12 +5373,12 @@ object a:
...
      }
    ) +
-     this.match {
-       case MateAdvice(seq, _, _, _) =>
-         seq.desc
-       case CpAdvice(judgment, _, _) =>
-         judgment.toString
-     } + "."
+     this.match
+        case MateAdvice(seq, _, _, _) =>
```
tgodzik, line 5377: Why do we have 3 spaces here? It should stay the same, no?

kitbellew: because significant indentation in this file is set to 3:
https://github.com/scalameta/scalafmt/blob/adf6c7941753c030ccd28b3e14fd1903edbf7850/scalafmt-tests/src/test/resources/scala3/OptionalBraces_unfold.stat#L1

tgodzik: ach! Right

tgodzik (approved):

PR #3843, opened 2024-03-22
Replace scalafiddle reference with scastie

sake92:


kitbellew (commented):
@sake92 thank you very much, didn't realize it was as simple as that.

i have a question, though (or, perhaps, @tgodzik knows): i tried scastie, it seems to compile and run the program. how does it display the scalameta trees for parsed code? not sure what worksheet does, or if it's relevant to this.

tgodzik:
> @sake92 thank you very much, didn't realize it was as simple as that.
> 
> i have a question, though (or, perhaps, @tgodzik knows): i tried scastie, it seems to compile and run the program. how does it display the scalameta trees for parsed code? not sure what worksheet does, or if it's relevant to this.

You will need to display either via println or in the worksheets it will display unit statements.

kitbellew:
> > @sake92 thank you very much, didn't realize it was as simple as that.
> > i have a question, though (or, perhaps, @tgodzik knows): i tried scastie, it seems to compile and run the program. how does it display the scalameta trees for parsed code? not sure what worksheet does, or if it's relevant to this.
> 
> You will need to display either via println or in the worksheets it will display unit statements.

thank you, @tgodzik . should we merge this? if there's more to clarify, we could add later.

docs/configuration.md
```diff
@@ -1073,7 +1073,7 @@ node that owns that token), and a list of `parents` (to match the tree containin
 the token).
 
 > To find the `owner` part for a custom tree, look for its type prefix using
-> [ScalaFiddle Playgroud](https://scalameta.org/docs/trees/scalafiddle.html) or
+> [Scastie Playgroud](https://scastie.scala-lang.org/) or
```
tgodzik, line 1076:
```suggestion
> [Scastie Playground](https://scalameta.org/docs/trees/scastie.html) or
```
maybe we should link it to scalameta or directly to the prepared snippet

kitbellew:
@tgodzik do you think this is good now?

PR #3855, opened 2024-03-27
Router: break on `(lambda` to keep trailing comma

kitbellew:
Other cases already implement this. Helps with #3812.

### DIFF
--- a/scalafmt-tests/src/test/resources/rewrite/RedundantBraces.stat
+++ b/scalafmt-tests/src/test/resources/rewrite/RedundantBraces.stat
@@ -1505,13 +1505,12 @@ object a {
 >>>
 Idempotency violated
 => Diff (- obtained, + expected)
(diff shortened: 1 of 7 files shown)


scalafmt-tests/src/test/resources/rewrite/RedundantBraces.stat
```diff
@@ -1505,13 +1505,12 @@ object a {
...
 Idempotency violated
 => Diff (- obtained, + expected)
- object a {
--  mtd1(x => x + 1) + mtd2(x => x + 1 x +2)
-+  mtd1(x => x + 1) + mtd2(x =>
-+    x + 1
-+    x + 2,
-+  )
- }
+   ) + mtd2(
```
kitbellew, line 1508: fyi: this part will be fixed in a subsequent change

PR #3875, opened 2024-03-29
Scalameta: upgrade to v4.9.3

### LINKED ISSUE #3844: Wrong formatting of code blocks with partial functions returning tuples
```
version = 3.8.0
runner.dialect = scala3
```

When I run scalafmt via CLI like this: `scalafmt File.scala`

## Steps

Given code like this:
```scala
class Sfmt {
  val pr: PartialFunction[Int, (String, Int)] = {
    case e: Int =>
      val a = ""
      ("first", 2)
  }

  try {
    ()
  } catch {
    case e: Throwable =>
      println("")
      (1, false)
  }


  1 match {
    case 2 =>
      println("")
      (1, false)
  }
  
  List.empty[Matchable].collect{
    case x:Int =>
      val a = 1
      (a)
  }
}
```

## Problem

```scala
class Sfmt {
  val pr: PartialFunction[Int, (String, Int)] = { case e: Int =>
    val a = "" ("first", 2)
  }

  try {
    ()
  } catch {
    case e: Throwable =>
      println("")(1, false)
  }

  1 match {
    case 2 =>
      println("")(1, false)
  }

  List.empty[Matchable].collect { case x: Int =>
    val a = 1 (a)
  }
}
```

## Expectation

```scala
class Sfmt {
  val pr: PartialFunction[Int, (String, Int)] = { case e: Int =>
    val a = ""
    ("first", 2)
  }

  try {
    ()
  } catch {
    case e: Throwable =>
      println("")
      (1, false)
  }

  1 match {
    case 2 =>
      println("")
      (1, false)
  }

  List.empty[Matchable].collect { case x: Int =>
    val a = 1
    (a)
  }
}
```

## Notes
I gave several examples where the bug occurs.

The pattern is that we have code block with `case` starting from new line, followed by any line (or lines), followed by line containing code in parenthesis (for example tuples).

kitbellew:
Fixes #3844.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -7,7 +7,7 @@ import org.portablescala.sbtplatformdeps.PlatformDepsPlugin.autoImport._
 
 object Dependencies {
   val metaconfigV = "0.12.0"
-  val scalametaV  = "4.9.2"
+  val scalametaV  = "4.9.3"
   val scalacheckV = "1.17.0"
   val coursier    = "2.1.2"
   val munitV      = "0.7.29"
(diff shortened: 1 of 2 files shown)


joan38:
Amazing thanks @kitbellew

PR #3944, opened 2024-04-28
Github actions: explicitly use macOS-12

kitbellew:
Latest version is now 14 but it doesn't include sbt.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,13 +56,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


kitbellew:
fyi @tgodzik 

i found this [article](https://eed3si9n.com/holy-graal-of-gitHub-actions/) but i don't know how to modify our build to work with -14, we don't use `sbtn` that is mentioned there. perhaps @eed3si9n or @keynmol could help.

for now, since this is a check rather than a release step, this should suffice... until -12 disappears, that is.

Issue #3954, opened 2024-05-06
Clean-up config-style and binpacking code


kitbellew:
There's a fair amount of inconsistency between callsite and defnsite handling.

PR #3978, opened 2024-05-11
Router: treat infix argument clause like any other

kitbellew:
Helps with #3954.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
@@ -420,17 +420,16 @@ object TreeOps {
   def isTokenLastOrAfter(token: Token, pos: Position): Boolean = pos.end <=
     token.end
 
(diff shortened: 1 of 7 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/util/TreeOps.scala
```diff
@@ -420,17 +420,16 @@ object TreeOps {
   def isTokenLastOrAfter(token: Token, pos: Position): Boolean = pos.end <=
     token.end
 
-  def isArgClauseSite(tree: Tree)(implicit style: ScalafmtConfig): Boolean =
-    tree match {
-      case t: Member.ArgClause => !t.parent.exists(_.is[Member.Infix]) ||
-        (t.values match {
-          case (_: Term.Assign | _: Lit.Unit) :: Nil => true
-          case Nil | _ :: Nil => false
-          case _ => style.newlines.formatInfix
```
kitbellew, line 429: this is the main change (to `true`). the rest is removing unused implicit.

tgodzik (approved):

PR #3986, opened 2024-05-13
Scalameta: upgrade to v4.9.4

kitbellew:


kitbellew:
@tgodzik any ideas on what to do about this jline error? i don't know where this library is coming from.

also see jline/jline3#954 and sbt/sbt#7500.

should we downgrade our version of sbt, or sbt-native-packager, or something else?

tgodzik:
> should we downgrade our version of sbt, or sbt-native-packager, or something else?

I think we might want to downgrade sbt, though curious why this hasn't happened in a previous PR

kitbellew:
ah, must be something picked up by scalameta, then 😞

PR #4057, opened 2024-06-17
Router: use `indent.{binPackXxx,afterInfix}Site`

kitbellew:
They are needed to implement [scala.js](https://github.com/scala-js/scala-js/blob/main/CODINGSTYLE.md) style.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -315,6 +315,8 @@ object a {
 
 ### `indent.callSite`
 
+This parameter controls indentation of method call arguments.
+
 ```scala mdoc:defaults
 indent.callSite
 ```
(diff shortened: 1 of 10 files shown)


docs/configuration.md
```diff
@@ -376,6 +402,30 @@ parameter1: Type1 // indented by 4
...
+> Since v3.8.3.
+
+Example:
+
+```scala mdoc:scalafmt
+newlines.configStyleDefnSite.prefer = false
+danglingParentheses.defnSite = false
+binPack.defnSite = always
+indent.defnSite = 2
+indent.binPackDefnSite = 4
```
tgodzik, line 420: Why does it need to be separate to `indent.defnSite`, that seems to over complicate things

kitbellew: how would you propose to do it? the have a requirement to use one indent when config style formatting is used, and a different one otherwise. which seems to mean one parameter is not enough...

tgodzik: It's probably not possible any other way, but I don't see how this kind of option is useful for anyone to have some many different indents

tgodzik: Could you point me to the requirement, where it's defined?

tgodzik: Nvm. I found it, but I am still not convinced it's more than just an OCD thing and doesn't really help in readability that much.

kitbellew: i think the reason it's useful to scala.js developers is that it visually distinguishes between clean config style formatting, and more readability-challenged binpack formatting. one of the reasons scala.js has not adopted scalafmt yet.

tgodzik (approved):
Looks good, I just always have doubts if we need to add more options to already an abundance of configuration

PR #4118, opened 2024-07-25
RedundantParens: don't rewrite unary within select

### LINKED ISSUE #4116: Invalid state for bit operations
## Steps

Given code like this:
```scala
(~validByte).toByte
```

## Problem

```scala
~validByte.toByte
```
and makes code broken because for scalac its like `~(validByte.toByte)` and expression become as Int type.

## Expectation

```scala
(~validByte).toByte
```

## Notes

version: 3.8.3

kitbellew:
In reality, it's unclear why ApplyUnary extends Term.Ref, but since Term.Select does, too, let's introduce a special-case handling for it. Fixes #4116.


tgodzik (approved):

armanbilge:
Still running into this, unfortunately I think this fix has not yet been released.

Issue #4133, opened 2024-08-03
Add community build (a-la scalameta), fix formatting bugs


kitbellew:
Test formatting of open-source community code, similar to how it was done in scalameta, and fix any formatting bugs discovered this way.

PR #4137, opened 2024-08-06
Router: apply src=keep around `: return-type`

kitbellew:
Helps with #4133.

### DIFF
--- a/scalafmt-tests/src/test/resources/scala3/OptionalBraces_keep.stat
+++ b/scalafmt-tests/src/test/resources/scala3/OptionalBraces_keep.stat
@@ -3626,8 +3626,8 @@ object a {
   implicit def generate[T](
       value: T
   ): Clue[T] = macro MacroCompatScala2.clueImpl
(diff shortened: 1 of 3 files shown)


scalafmt-tests/src/test/resources/scala3/OptionalBraces_keep.stat
```diff
@@ -3626,8 +3626,8 @@ object a {
   implicit def generate[T](
       value: T
   ): Clue[T] = macro MacroCompatScala2.clueImpl
-  implicit def generate[T](value: T): Clue[T] = macro
```
tgodzik, line 3629: I would expect keep to not change the newlines here,

kitbellew: and now, after the fix, it **does** preserve the newline :)

tgodzik: ach, right my bad. My brain is not working 100% today :sweat_smile:

PR #4183, opened 2024-08-23
Add literal.binPrefix setting to format binary literals.

### LINKED ISSUE #4176: Add a literals.binary setting
Now that Scala supports binary literals (as of 1.5.0) of the form `0b1101`, it'd be nice to add a `literals.binary` setting similar to `literals.float`, `literals.long`, etc.

charpov:
Add formatting to handle binary literals, which were added in Scala 3.5.0.
Those are of the form 0b111 or 0b111L and so are susceptible to both literals.binPrefix and literals.long.

The 3.5.0 release uses 0B111 in its examples, but the original SIP 42 uses 0b111. I chose 0b as the default, as it looks a lot nicer to me.

There is a new test file (`binary.stat`) and the documentation was updated (it's missing a `since` comment).

The changes address my own feature request: fixes #4176.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4873,6 +4873,8 @@ literals.long=Upper
 
 ### `literals.binPrefix`
 
+> Since v3.8.4.
+
 ```scala mdoc:defaults
 literals.binPrefix
 ```
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4792,6 +4792,8 @@ Default formatting:
 0x1Abl
 10E-1
 10e-1D
+0B111
+0b111l
 ```
 
 Each `literals.*` setting has three available options: `Upper`, `Lower`,
@@ -4867,6 +4869,22 @@ literals.long=Upper
 ---
 0xaAaA
 0xaAaAl
+```
+
+### `literals.binPrefix`
+
+```scala mdoc:defaults
+literals.binPrefix
+```
+
+Responsible for the case of binary integer literals prefix `0b`
+
+```scala mdoc:scalafmt
+literals.binPrefix=Lower
+literals.long=Upper
+---
+0B111
+0B111l
 ```
 
 ### `literals.scientific`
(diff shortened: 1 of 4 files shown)


scalafmt-core/shared/src/main/scala/org/scalafmt/util/LiteralOps.scala
```diff
@@ -53,10 +59,12 @@ object LiteralOps {
       suffixCase.process(str.takeRight(1))
     else style.literals.scientific.process(str)
 
-  private def prettyPrintHex(
+  private def prettyPrintHexOrBin(
```
kitbellew, line 62: at this point, i'd rewrite this method as:
```
private def prettyPrintHexOrBin(
      str: String,
  )(implicit style: ScalafmtConfig): String = {
    val (prefix, body) = str.splitAt(2)
    prefix match {
      case "0x" | "0X" =>
        style.literals.hexPrefix.process(prefix) + style.literals.hexDigits.process(body)
      case "0b" | "0B" =>
        style.literals.binPrefix.process(prefix) + body
      case _ => str
  }
}
```

docs/configuration.md
```diff
@@ -4869,6 +4871,22 @@ literals.long=Upper
 0xaAaAl
 ```
 
+### `literals.binPrefix`
+
+```scala mdoc:defaults
+literals.binPrefix
+```
+
+Responsible for the case of binary integer literals prefix `0b`
```
kitbellew, line 4880: could you please prepend `(Since v3.8.4)`, to make it clear when this parameter became available?

kitbellew (changes requested):
thank you, nice and clean!

docs/configuration.md
kitbellew: looks like you mentioned this aspect in your PR comment as well :)

charpov:
I've added the version number and made the suggested code changes.

kitbellew (approved):

PR #4281, opened 2024-09-17
Update coursier-interface to 1.0.21

alexarchambault:


alexarchambault:
Seems that dependency hadn't been updated for a while. Is there a reason for that?

kitbellew:
> Seems that dependency hadn't been updated for a while. Is there a reason for that?

@alexarchambault are you talking about coursier-interface, or some other dependency? i see that native-image builds are failing, i am not exactly sure why but perhaps the new update is somehow incompatible.

kitbellew:
superseded by #4558.

PR #4431, opened 2024-10-11
Router: apply space.inParens consistently

kitbellew:
Helps with #4133.


tgodzik (changes requested):
Looks good, but the tests are failing.

PR #4515, opened 2024-11-07
RemoveScala3OptionalBraces: `()` to fewer braces

kitbellew:


docs/configuration.md
```diff
@@ -3853,6 +3853,11 @@ The section contains the following settings (available since v3.8.1):
   - will only apply the rewrite if the cumulative span of all visible
     (non-whitespace) tokens within the argument is between the two values
   - this rule is disabled if `fewerBracesMaxSpan == 0`
+- (since v3.8.4) `fewerBracesParensToo`
```
tgodzik, line 3856:
```suggestion
- (since v3.8.4) `alsoParens`
```
wouldn't be enough? Since it's already `removeOptionalBraces.fewerBracesParensToo`

tgodzik (approved):
Just a minor nitpick. Otherwise LGTM

docs/configuration.md
kitbellew: will that not be confusing? it only works for fewer braces, not for any other potential case involving optional braces.

Issue #4571, opened 2024-11-20
Can't use different version for native CLI


EnviousSwan:
This issue is observable in the following configuration
- latest version of CLI and previous version in `.scalafmt.conf`

## Configuration (required) ##

```hocon
version = 3.8.3
runner.dialect = scala3
...
rewrite.scala3.convertToNewSyntax = true
rewrite.scala3.removeOptionalBraces = false
```

## Command-line parameters (required) ##

When I run scalafmt via CLI like this: `scalafmt --diff-branch origin/master`

## Steps

Given a GitHub workflow 
```yaml
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v4
    - uses: coursier/setup-action@v1
      with:
        jvm: temurin:21
        apps: scalafmt
    - run: scalafmt --diff-branch origin/master
```

## Problem

When running `scalafmt` when a PR is raised via
```
Run scalafmt --diff-branch origin/master
  scalafmt --diff-branch origin/master
  shell: /usr/bin/bash -e {0}
  env:
    JAVA_HOME: /home/runner/.cache/coursier/arc/https/github.com/adoptium/temurin21-binaries/releases/download/jdk-21.0.5%252B11/OpenJDK21U-jdk_x64_linux_hotspot_21.0.5_11.tar.gz/jdk-21.0.5+11
    COURSIER_BIN_DIR: /home/runner/cs/bin
```

I get the following error
```
Exception in thread "main" java.lang.Exception: Can't use different version for native CLI
	at org.scalafmt.interfaces.Scalafmt$.create(Scalafmt.scala:142)
	at org.scalafmt.interfaces.Scalafmt.create(Scalafmt.scala)
	at org.scalafmt.cli.ScalafmtDynamicRunner$.run(ScalafmtDynamicRunner.scala:21)
	at org.scalafmt.cli.Cli$.runWithRunner(Cli.scala:134)
	at org.scalafmt.cli.Cli$.run(Cli.scala:61)
	at org.scalafmt.cli.Cli$.mainWithOptions(Cli.scala:35)
	at org.scalafmt.cli.Cli$.main(Cli.scala:18)
	at org.scalafmt.cli.Cli.main(Cli.scala)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
[...]

PR #4573, opened 2024-11-20
bugfix: Publish native interfaces with full suffix

tgodzik:
Fixes https://github.com/scalameta/scalafmt/issues/4571

crossVersion actually is set differently if you use `crossProject` so if you disable it you get no suffix in all platforms.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -85,8 +85,6 @@ lazy val interfaces = crossProject(JVMPlatform, NativePlatform)
     moduleName := "scalafmt-interfaces",
     description :=
       "Dependency-free, pure Java public interfaces to integrate with Scalafmt through a build tool or editor plugin.",
-    crossVersion := CrossVersion.disabled,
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -85,8 +85,6 @@ lazy val interfaces = crossProject(JVMPlatform, NativePlatform)
     moduleName := "scalafmt-interfaces",
     description :=
       "Dependency-free, pure Java public interfaces to integrate with Scalafmt through a build tool or editor plugin.",
-    crossVersion := CrossVersion.disabled,
-    autoScalaLibrary := false,
```
kitbellew, line 89: what does this do? not needed for native, then?

kitbellew (approved):

build.sbt
tgodzik: So it doesn't add Scala library for Java artifacts, since it will not be needed.

PR #4574, opened 2024-11-20
SBT: native settings/plugins for native platform

kitbellew:


kitbellew:
@jchyb please let me know if this looks ok.

bin/build-native-image.sh
```diff
@@ -1,2 +1,2 @@
 set -eux
-sbt "cli/nativeImageCopy scalafmt"
+sbt "cliNative/nativeImageCopy scalafmt"
```
tgodzik, line 2: I don't think Scala Native has much to do with NativeImage, that's a totally different thing if I am not mistaken totally? If we switch from native image to pure Scala native than we would need to just do `nativeLinkReleaseFull` and then copy the generated binary from target/scala-2.13.15

Do you want to switch the binaries right now? O after the next release?

kitbellew: goes to show that i am really up to speed here :) so what is native image, then?

let's fix the current state first, then do what you suggest with the switch.

kitbellew: how interesting... then would it be correct to say that nativeImage should only be built from JVM (i.e., this change should accomplish the exact opposite), since there's no reason to build a native image from an already native result?

tgodzik: Yes, native image is only for JVM.

Just checked and the scala native binary is almost 2x smaller than the native image one in case of scalafmt. And it works without adding any special options. We could probably also optimize it better, so that's also more efficient that native image, though I haven't yet compared the two.

jchyb: I should mention that the 2.13 scala native version might likely be slower then 2.13 native image for now, since parallel collections are not used there, as the `scala-parallel-collections` library was not ported to scala native yet they are built into scala stdlib in 2.12, so it's better there).

build.sbt
```diff
@@ -196,6 +196,8 @@ lazy val cli = crossProject(JVMPlatform, NativePlatform)
     ),
     scalacOptions ++= scalacJvmOptions.value,
     Compile / mainClass := Some("org.scalafmt.cli.Cli"),
+  ).nativeSettings(
+    nativeConfig ~= { _.withMode(Mode.releaseFull) },
     nativeImageInstalled := isCI,
     nativeImageOptions ++= {
```
jchyb, line 202: Similar issue as pointed out above, nativeImage settings will be ignored here (basically Native Image still uses classfiles to make a binary file, so it has to use the JVM target, Scala Native compiler outputs something else with the `.nir` extension, which then the linker uses that to make a binary file.

tgodzik (approved):

PR #4616, opened 2024-12-09
SBT: get mdoc version from its plugin

kitbellew:
Otherwise, it was necessary manually to keep `Dependencies.scala` and `plugins.sbt` in sync.


kitbellew:
not sure how to check that it actually pulls in the right version, though.

tgodzik (approved):

tgodzik:
> not sure how to check that it actually pulls in the right version, though.

Which one would be the right one? I guess the BuildInfo file should have the correct one and everything should be fetched correctly.

PR #4618, opened 2024-12-10
Router: for/infix before select like fewer-braces

### LINKED ISSUE #4614: Incorrect formating on for comprehension with Seq operation

configuration parameters that reproduces the problem:
```properties
version = "3.8.4-RC3"
runner.dialect = scala3
```

## Steps

Given code like this:
```scala
@main def main() =
  val first =
    for i <- 0 to 10
    yield i
  .head
  println(first)
```

## Problem

```scala
@main def main() =
  val first =
    for i <- 0 to 10
    yield i.head
  println(first)
```

## Expectation

It shouldn't format the `head` operation

## Workaround

using a bracket on for comprehension can avoid this, code should be like this

```scala
@main def main() =
  val first = {
    for i <- 0 to 10
    yield i
  }.head
  println(first)
```

## Environments

- Operating System: Arch Linux
- Scalafmt version: scalafmt 3.8.4-RC3
- scala version: 3.6.1

kitbellew:
Fixes #4614.

### DIFF
--- a/scalafmt-tests/shared/src/test/resources/scala3/FewerBraces.stat
+++ b/scalafmt-tests/shared/src/test/resources/scala3/FewerBraces.stat
@@ -2329,8 +2329,9 @@ indent.fewerBraces = never
 >>>
 @main def main() =
    val first =
-     for i <- 0 to 10
-     yield i.head
+      for i <- 0 to 10
+      yield i
+   .head
    println(first)
 <<< #4614 for-yield then apply, never
 indent.fewerBraces = never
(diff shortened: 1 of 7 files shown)


scalafmt-tests/shared/src/test/resources/scala3/FewerBraces.stat
```diff
@@ -2317,3 +2317,435 @@ object a:
...
+  .map(_ * 2)
+  .last
+  println(first)
+>>>
+@main def main() =
+   val first =
+      foo :: bar ::
+        baz
+   .map(_ * 2)
+     .last
```
tgodzik, line 2518: This is not a perfect indent, but I think this is such a rare case that we might not want to bother with it

tgodzik (approved):

PR #4697, opened 2024-12-27
Workflow: run CI only native-image upon release

kitbellew:
Also, do not run CI when tagging; when the final commit of a release is merged, currently CI is invoked 3 times:
- upon commit merge into main (triggered by push to branch main)
- upon release (triggered by release published)
- upon tagging along with release (triggered by push to tag `vX.Y.Z`)

Let's remove the last trigger and run only native-image (which uploads release artifacts) on release trigger, as all the rest of the checks would have been run on the same commit when it had been originally pushed to main.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - main
-    tags:
+    tags_ignore:
       - '*'
   pull_request:
   release:
@@ -14,6 +14,7 @@ permissions:
 
 jobs:
   test:
+    if: github.event_name != 'release'
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - main
-    tags:
+    tags_ignore:
```
kitbellew, line 6: perhaps this part should be completely removed, not sure

tgodzik: Alternatively, we could remove the `release` below, no? Since when creating a release a tag is pushed anyway

tgodzik (approved):
I think we can just remove one of those sections, but it's fine either way

.github/workflows/ci.yml
kitbellew: i think release published is needed since that's where it puts the assets. scalameta triggers just on tagging but it doesn't have assets.

PR #4719, opened 2025-01-15
improvement: Add flags to release for JDK 8

tgodzik:


kitbellew:
@tgodzik didn't see you also submitted a fix for this problem, and uploaded my own (#4720). i will abandon mine if yours succeeds :)

tgodzik:
I think this should work, I checked and the class files produced are version 52. I just had a weird issue with javadoc, where the target option would be forwarded to it, which makes no sense, since javadoc doesn't support the same options. I just set the options for it to nil

build.sbt
```diff
@@ -95,6 +95,8 @@ lazy val interfaces = crossProject(JVMPlatform, NativePlatform)
       IO.write(props, "scalafmt properties", out)
       List(out)
     },
+    javacOptions ++= Seq("-source", "1.8", "-target", "1.8"),
+    Compile / doc / javacOptions := Seq(),
```
kitbellew, line 99: chatgpt told me to set it within `.jvmSettings` using `Compile / javacOptions`, and it didn't fail.

what's the difference between `--release 8` and `-source 1.8 -target 1.8`?

build.sbt
```diff
@@ -167,9 +169,13 @@ import sbtassembly.AssemblyPlugin.defaultUniversalScript
...
 val scalacJvmOptions = Def.setting {
   val cross = CrossVersion.partialVersion(scalaVersion.value) match {
-    case Some((2, 13)) =>
-      Seq("-Ymacro-annotations", "-Xfatal-warnings", "-deprecation:false")
-    case _ => Seq.empty
+    case Some((2, 13)) => Seq(
+        "-Ymacro-annotations",
+        "-Xfatal-warnings",
+        "-deprecation:false",
+        "-target:8",
```
kitbellew, line 176: 1. same question: what's the difference between `-release 8` and `-target:8`?
2. also, are you setting it for all projects (including core), and not just dynamic and its two deps?

build.sbt
tgodzik: I think that's the same, though I never remember if that works in scala or java or both.

We can test that and set it to 8

build.sbt
tgodzik: Yes, I thought that's safest and it doesn't cause any issues really

build.sbt
kitbellew: sounds good. this part is not my forte 😔 merge if you're ready.

tgodzik: Switched to 8 to be consistent, still work ok

build.sbt
tgodzik: And looks it actually detected usage of a method added JDK 8 +

tgodzik:
I think we should be good now.

scalafmt-tests-community/common/src/main/scala/org/scalafmt/community/common/TestHelpers.scala
```diff
@@ -58,9 +58,9 @@ object TestHelpers {
         val out1 = x1.formattedCode
         val lines2 = x1.formattedCode.count(_ == '\n')
         val (duration2, result2) = formatCode(out1)
-        def saveFormatted(): Unit = Files.writeString(
+        def saveFormatted(): Unit = Files.write(
```
tgodzik, line 61: This method and the one below was added in JDK 11, but wasn't an issue for released artifacts since it's in the tests only

build.sbt
```diff
@@ -167,9 +172,14 @@ import sbtassembly.AssemblyPlugin.defaultUniversalScript
...
+        "-release:8",
+      )
+    case _ => Seq("-target:8", "-release:8")
```
kitbellew, line 182: rather than adding these twice, to each scala version, should we follow the example of unused below and append separately?

scalafmt-tests-community/common/src/main/scala/org/scalafmt/community/common/TestHelpers.scala
kitbellew: should we instead add new flags to the `Compile /` so it doesn't affect Test?

build.sbt
```diff
@@ -95,7 +95,12 @@ lazy val interfaces = crossProject(JVMPlatform, NativePlatform)
       IO.write(props, "scalafmt properties", out)
       List(out)
     },
-  ).jvmSettings(crossVersion := CrossVersion.disabled, autoScalaLibrary := false)
+  ).jvmSettings(
+    javacOptions ++= Seq("-source", "8", "-target", "8"),
+    Compile / doc / javacOptions := Seq(),
```
kitbellew, line 100: are you sure you still need this for doc? especially if you used the one above with `Compile /`?

tgodzik: yeah, it still didn't work without it

scalafmt-tests-community/common/src/main/scala/org/scalafmt/community/common/TestHelpers.scala
kitbellew: sounds good

kitbellew (approved):

build.sbt
tgodzik: sure, done. I at some point I thought those might differ

PR #4726, opened 2025-01-17
Spaces: add new IfMultipleContextBounds label

kitbellew:
It considers only context bounds, unlike the earlier ifMultipleBounds which looks at all bounds.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4672,6 +4672,12 @@ The parameter also allows the following shortcuts:
 
 ### `spaces.beforeContextBoundColon`
 
+This parameter controls if a space should be used before a colon that precedes
+type context bounds and takes values 'never', 'always', 'ifMultipleBounds', and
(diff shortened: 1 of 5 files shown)


docs/configuration.md
```diff
@@ -4672,6 +4672,12 @@ The parameter also allows the following shortcuts:
 
 ### `spaces.beforeContextBoundColon`
 
+This parameter controls if a space should be used before a colon that precedes
```
tgodzik, line 4675: Do we need a separate setting here?

kitbellew: i am planning to use this same set of values for spaces within braces, and it makes little sense to add those spaces after looking at some other bound which is not within those braces. or to add a completely new set of values which are likely named the same but carry a slightly different meaning. just for consistency.

tgodzik: Ok, thanks for explaining!

tgodzik (approved):

Issue #4729, opened 2025-01-20
Publish images using Scala native instead of Native Image


tgodzik:
I was wondering if it was a good time to switch to using Scala Native instead of Native Image, which seems to have some issues when trying to publish for other architectures or when bumping versions. 

For example we publish some additional images here https://github.com/VirtusLab/scalafmt-native-image/pull/67

I would try to instead publish them here using Scala Native.

Main question is whether anyone would oppose that? If not I can take a look at it.

kitbellew:
i don't know enough (nor use native) to present an opinion, but don't oppose it.

PR #4738, opened 2025-01-24
refactor: Use Scala Native instead of native image

tgodzik:
Fixes https://github.com/scalameta/scalafmt/issues/4729


tgodzik:
Looks like it's simpler than I thought. I also added windows and macos runners.

tgodzik:
I wonder if we should be so strict with the version incompatibility? Maybe we should allow version of the image higher than the version in the config?

.github/workflows/ci.yml
```diff
@@ -98,63 +98,49 @@ jobs:
...
         with:
           java-version: '21'
-          distribution: 'graalvm-community'
-          native-image-musl: ${{ matrix.libc == 'musl' }}
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          distribution: 'temurin'
+          cache: 'sbt'
       - uses: sbt/setup-sbt@v1
-      - run: bin/build-native-image.sh
+      - run: sbt scala-native
```
kitbellew, line 125: should we create a script as well, `bin/build-scala-native.sh` and hide all the overly specific logic pertaining to `scala-2.13` or target location or current architecture (what you have under deploy.XXX)?

kitbellew:
> I wonder if we should be so strict with the version incompatibility? Maybe we should allow version of the image higher than the version in the config?

where does this come into play?

.github/workflows/ci.yml
tgodzik: I can do that for sure, but it's not a lot of logic really and I would have to also deal with windows scripts 😅

tgodzik:
> > I wonder if we should be so strict with the version incompatibility? Maybe we should allow version of the image higher than the version in the config?
> 
> where does this come into play?

We have a check for scala native that if the version of the binary is different than the one in config it will throw an exception.

.github/workflows/ci.yml
kitbellew: let me continue this comment above, under `deploy`.

.github/workflows/ci.yml
```diff
@@ -98,63 +98,49 @@ jobs:
...
-            env:
-              NATIVE_IMAGE_STATIC: nolibc
-          - os: ubuntu-latest
-            libc: musl
-            artifact: scalafmt-linux-musl
-            env:
-              NATIVE_IMAGE_STATIC: musl
-    env: ${{ matrix.env }}
+        deploy: [ 
+          { os : macOS-13, name: scalafmt-x86_64-apple-darwin, path: "scalafmt-cli/native/target/scala-2.13/scalafmt-cli"},
```
kitbellew, line 108: i just thought that perhaps all these OS and architecture values can be easily deduced from the environment and don't have to be hardcoded...

also, should we just call sbt and ask for the location of the target directory, instead of hardcoding with `scala-2.13`?

if that can be done, then i am ok without a script.

kitbellew (changes requested):

kitbellew:
> > > I wonder if we should be so strict with the version incompatibility? Maybe we should allow version of the image higher than the version in the config?
> > 
> > 
> > where does this come into play?
> 
> We have a check for scala native that if the version of the binary is different than the one in config it will throw an exception.

I wouldn't do that. we must use the version specified in the config file, that's the contract. how is "higher version" better than "lower version"? both can produce a slightly different formatting, even if higher likely won't fail on newer configuration parameters.

scalafmt
kitbellew, line 1: are you replacing the formatter? is it a jvm-based build? +36000% :)

build.sbt
```diff
@@ -34,16 +35,25 @@ inThisBuild {
...
 }
 
 name := "scalafmtRoot"
 publish / skip := true
 
-addCommandAlias("native-image", "cli/nativeImage")
-addCommandAlias("scala-native", "cliNative/compile;cliNative/nativeLink")
+lazy val copyScalaNative = taskKey[Unit]("Copy Scala Native output to root")
+
+copyScalaNative := {
```
kitbellew, line 46: thank you for this!

.github/workflows/ci.yml
```diff
@@ -82,63 +82,49 @@ jobs:
...
       - name: Upload release
         if: github.event_name == 'release'
         uses: actions/upload-release-asset@v1.0.2
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ github.event.release.upload_url }}
-          asset_path: scalafmt
-          asset_name: ${{ matrix.artifact }}
+          asset_path: scalafmt-cli/native/target/scala-2.13/scalafmt-cli
```
kitbellew, line 123: is this the result of `copyScalaNative`? if so, shouldn't there be `.exe` for windows?

scalafmt
tgodzik: Ach, I added it by mistake, will remove it soon, but seems something broke somehow

.github/workflows/ci.yml
tgodzik: Fixed, I added a variable.

docs/installation.md
```diff
@@ -295,27 +295,17 @@ If you use sbt to install a pre-release, be sure to add the following setting:
 resolvers += Resolver.sonatypeRepo("snapshots")
 ```
 
-### Native image
+### Native binaries
 
-For macOS and Linux, it's possible to download pre-built GraalVm native binaries
-with instant startup and fast performance for short-lived Scalafmt runs.
+It's possible to download pre-built Scala Native binaries
```
tgodzik, line 300: This might be controversial, but I feel like fixing the `install-scalafmt-native.sh`  is a lot more work and scala cli already properly downloads binaries.

tgodzik: And also on all platforms

docs/installation.md
```diff
@@ -295,27 +295,17 @@ If you use sbt to install a pre-release, be sure to add the following setting:
...
+with instant startup and fast performance for short-lived Scalafmt runs using Scala CLI and 
+[Scala CLI Setup Action](https://github.com/marketplace/actions/scala-cli-setup-action).
 
 ```sh
-VERSION=@STABLE_VERSION@
-INSTALL_LOCATION=/usr/local/bin/scalafmt-native
-curl https://raw.githubusercontent.com/scalameta/scalafmt/master/bin/install-scalafmt-native.sh | \
-  bash -s -- $VERSION $INSTALL_LOCATION
-scalafmt-native --help # should show version @STABLE_VERSION@
+scala-cli format
```
tgodzik, line 305: This will read the scalafmt version if it exists and download the proper binary. We will need to update it and drop the additional repository, but it should be a smooth transition.

docs/installation.md
kitbellew: that sounds good to me, no duplication

kitbellew (dismissed):

build.sbt
```diff
@@ -1,6 +1,7 @@
 import scala.scalanative.build._
 
 import Dependencies._
+import scala.util.Properties
```
kitbellew, line 4: this was formatted to a different line

.github/workflows/ci.yml
```diff
@@ -82,63 +82,51 @@ jobs:
...
+    runs-on: ${{ matrix.deploy.os }}
+    env:
+      BINARY_NAME: ${{ (startsWith(matrix.deploy.os, 'windows') && 'scalafmt.exe') || 'scalafmt' }}
     steps:
       - uses: actions/checkout@v4
-      - name: Set up GraalVM
-        uses: graalvm/setup-graalvm@v1
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-java@v4
```
kitbellew, line 105: a comment for separate consideration: if we are using java 21 for this, should we disable java 8 compatibility in native builds (if they apply, that is)?

kitbellew (approved):

tgodzik:
We can add the script back so that people can migrate, though I would recommend using Scala CLI in that case.

For the script to work with the new setup we would have to support windows etc. Alternatively, I can make the script work with the new setup as well only for those systems that it used to work for.

kitbellew:
apparently, people are saying (on discord) that scala-cli doesn't work. yet, at least. so, in order to avoid disruptions to existing setups, should we restore what we had before, in addition to the new stuff, and remove it later, when we've confirmed everything is working for everyone?

tgodzik:
> apparently, people are saying (on discord) that scala-cli doesn't work. yet, at least. so, in order to avoid disruptions to existing setups, should we restore what we had before, in addition to the new stuff, and remove it later, when we've confirmed everything is working for everyone?

I did a fix for Scala CLI, forgot to publish the latest binaries. It's running now https://github.com/VirtusLab/scalafmt-native-image/releases/tag/v3.8.6

And added the script back, so that should also restore things for now. I can add a warning maybe for newer versions to use SCala CLI? I will make sure the releases are on time and migrate soon to the original binaries here

kitbellew:
> And added the script back, so that should also restore things for now. I can add a warning maybe for newer versions to use SCala CLI? I will make sure the releases are on time and migrate soon to the original binaries here

We could notate that native image will be removed in early 2025.

PR #4743, opened 2025-01-26
ScalafmtRunner: use Future, not `.par`

kitbellew:


scalafmt-cli/shared/src/main/scala/org/scalafmt/cli/ScalafmtCoreRunner.scala
```diff
@@ -31,41 +27,29 @@ object ScalafmtCoreRunner extends ScalafmtRunner {
...
-      val code = handleFile(inputMethod, options, adjustedScalafmtConf)
-      exitCode.getAndUpdate(ExitCode.merge(code, _))
-      if (options.check && !code.isOk) Breaks.break
-      termDisplay.taskProgress(termDisplayMessage)
-    })
-    termDisplay.completedTask(termDisplayMessage, exitCode.get.isOk)
-    termDisplay.stop()
-    exitCode.get()
+      runInputs(options, inputMethods, termDisplayMessage) { inputMethod =>
+        import org.scalafmt.sysops.PlatformCompat.executionContext
```
tgodzik, line 42: Curiously, do we also enable multithreading in Scala Native as well this way? Maybe we don't need the separate executionContext anymore?

Maybe we can remove compatPar now also?

tgodzik: > during building i see messages about multithreading enabled, perhaps we could somehow make it explicit

I think it's automatically disabled when not needed, so no need to make it explicit I guess?

> i can't see much difference in speed in community test running, but it felt to me much faster when i ran the result of sbt cli/assembly with local snapshot version (forcing core runner, instead of dynamic).

Och, interesting! Do you mean the JVM version?

tgodzik (approved):

scalafmt-cli/shared/src/main/scala/org/scalafmt/cli/ScalafmtCoreRunner.scala
kitbellew: yes, the JVM. native was even faster but i didn't check how native performed without this change.

PR #4760, opened 2025-01-28
fix: Add back removed script to allow scalafmt action to work

tgodzik:


tgodzik:
Also, I run the newest release of binaries used by ScalaCLI  https://github.com/VirtusLab/scalafmt-native-image/releases/tag/v3.8.6 so people can use it and migrate

tgodzik:
I can work on that tomorrow, should not be to hard to do

tgodzik:
I added everything back. I don't want to tinker to much with ci.yml and merge both tasks, since that's a bit of a wasted effort if we plan to remove it

kitbellew (approved):

kitbellew:
> I added everything back. I don't want to tinker to much with ci.yml and merge both tasks, since that's a bit of a wasted effort if we plan to remove it

Of course, no need to merge them, they don't even share much in common anyway.

PR #4763, opened 2025-01-29
CI: run `scala-native` only on release

kitbellew:
scala-native doesn't need to run on push since we don't need to upload artifacts, and building is already verified as part of `test`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,7 +127,7 @@ jobs:
         with:
           name: ${{ matrix.artifact }}
           path: scalafmt
-      - name: Upload release
+      - name: Upload release to Github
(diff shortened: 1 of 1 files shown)


tgodzik (approved):
Should we also do that for native image?

kitbellew:
> Should we also do that for native image?

we could. here's what i thought:

1. we will remove it soon, so might as well not touch it :)
2. scala-native is always built as part of the test job (and then run), so the only difference is copying and then uploading which we don't need to test
3. native-image, however, involves some other steps which could potentially broken and only discovered after a release, so might be a reasonable check to keep for now.

let me know what you think.

tgodzik:
Makes sense, let's keep it then

Issue #4761, opened 2025-01-29
Scalafmt Dockerhub has stopped updating


henricook:
# Requirement

I was hoping to pull the latest version of Scalafmt, 3.8.6, to run in a docker container

# Issue

I notice that public images have not been updated on Dockerhub since 3.8.4-RC4 https://hub.docker.com/r/scalameta/scalafmt/tags

# Question

Would it be possible to push the latest images please? :-)

cc @olafurpg as you seem to publish images most regularly

Thanks everyone

henricook:
Thank you @kitbellew

henricook:
@kitbellew Sorry, I don't suppose you could trigger it retrospectively for 3.8.6?

kitbellew:
> [@kitbellew](https://github.com/kitbellew) Sorry, I don't suppose you could trigger it retrospectively for 3.8.6?

Certainly! 

How?

henricook:
Haha @kitbellew  - fair point. I'd be happy to build one by hand and provide it for you to push with the docker creds? But look I don't want to waste your time. If the resolution is that docker images aren't present for the last couple of releases and that they'll be present again from the next release I'm very happy to wait for that.

henricook:
Unfortunately this also didn't succeed with 3.9.0 but I don't have permission to reopen this issue

henricook:
3.9.1 and 3.9.2 wrote images but unfortunately both are faulty

```
docker run -it scalameta/scalafmt:3.9.2 bash
Unable to find image 'scalameta/scalafmt:3.9.2' locally
3.9.2: Pulling from scalameta/scalafmt
f18232174bc9: Already exists 
9a2b35279846: Pull complete 
4f4fb700ef54: Pull complete 
Digest: sha256:0d40b19d5730c4a8e0daa08254e7c2d447ac0a1f61a1108e60c2b64113a99a02
Status: Downloaded newer image for scalameta/scalafmt:3.9.2
exec /bin/scalafmt: no such file or directory
```

kitbellew:
> 3.9.1 and 3.9.2 wrote images but unfortunately both are faulty
> 
> ```
> docker run -it scalameta/scalafmt:3.9.2 bash
> Unable to find image 'scalameta/scalafmt:3.9.2' locally
> 3.9.2: Pulling from scalameta/scalafmt
> f18232174bc9: Already exists 
> 9a2b35279846: Pull complete 
> 4f4fb700ef54: Pull complete 
[...]

PR #4802, opened 2025-02-11
Build and test on arm64 runner

harryzcy:


aairey (approved):

kitbellew:
@aairey should we merge this, despite multiple failures?

kitbellew:
> > @aairey should we merge this, despite multiple failures?
> 
> I don't think so.
> 
> Someone who understands these test failures should take a look.

in that case, what was the purpose of approving this pr?

aairey (commented):
Fix test failures.

.github/workflows/ci.yml
```diff
@@ -21,6 +21,7 @@ jobs:
         os:
           - windows-latest
           - ubuntu-latest
+          - ubuntu-24.04-arm
```
kitbellew, line 24: i am not sure about this... we don't run on macos either.

.github/workflows/ci.yml
```diff
@@ -54,7 +55,7 @@ jobs:
       fail-fast: false
       matrix:
         java: [ '11' ]
-        os: [ windows-latest, ubuntu-latest ]
+        os: [ windows-latest, ubuntu-latest, ubuntu-24.04-arm ]
```
kitbellew, line 58: we definitely don't need this, the whole point was to make sure windows and non-windows are handled correctly.

.github/workflows/release-native.yml
```diff
@@ -61,20 +61,17 @@ jobs:
   dockerize:
     needs: [ scala-native ]
     if: ${{ github.event_name == 'release' == success() }}
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.platform == 'linux/arm64' && 'ubuntu-24.04-arm' || 'ubuntu-latest' }}
```
kitbellew, line 64: not ubuntu-latest anymore, it's a fixed version now.

also, any way to re-use the same deploy matrix (with os and name fields) from above?

.github/workflows/release-native.yml
```diff
@@ -83,12 +80,59 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
```
kitbellew, line 84: what does this do?

.github/workflows/release-native.yml
```diff
@@ -83,12 +80,59 @@ jobs:
...
       - name: Build and push
+        id: build
         uses: docker/build-push-action@v6
         with:
           context: .
+          platforms: ${{ matrix.platform }}
           push: true
-          tags: ${{ steps.meta.outputs.tags }}
+          outputs: type=image,"name=scalameta/scalafmt",push-by-digest=true,name-canonical=true
```
kitbellew, line 92: why do we not need tags here?

.github/workflows/release-native.yml
```diff
@@ -83,12 +80,59 @@ jobs:
...
+      - name: Export digest
+        run: |
+          mkdir -p ${{ runner.temp }}/digests
+          digest="${{ steps.build.outputs.digest }}"
+          touch "${{ runner.temp }}/digests/${digest#sha256:}"
+      - name: Upload digest
+        uses: actions/upload-artifact@v4
+        with:
+          name: digests-${{ matrix.platform == 'linux/arm64' && 'arm64' || 'amd64' }}
```
kitbellew, line 101: this test keeps getting repeated. i'd go with the same, or similar, multi-field matrix strategy (see comment above, about deploy matrix).

kitbellew (changes requested):
separately, since you seem to understand about docker so much, could you help make sense of the claim that docker images no longer work, please?

see context: https://github.com/scalameta/scalafmt/issues/4761#issuecomment-2692170972

.github/workflows/ci.yml
harryzcy: The intention is to make sure it's tested in linux arm environment

.github/workflows/ci.yml
harryzcy: removed

.github/workflows/release-native.yml
harryzcy: switched to 22.04. I'll think about reusing matrix, not implemented yet

.github/workflows/release-native.yml
harryzcy: I believe this is needed for the `platforms` setting in `docker/build-push-action`

.github/workflows/release-native.yml
harryzcy: This is because docker build is split into two steps: first is building and pushing by digest (each digest is only for one platform), and then it will combine the digests and finally create the multi-arch image.

.github/workflows/release-native.yml
harryzcy: okay, will try to do it

harryzcy:
@kitbellew I reworked this PR. Now binaries are pulled within Dockerfile from Release artifacts directly, and then build multi-arch image.

It's now ready for review. I've tested docker build locally and it works out.

harryzcy:
> separately, since you seem to understand about docker so much, could you help make sense of the claim that docker images no longer work, please?
> 
> see context: [#4761 (comment)](https://github.com/scalameta/scalafmt/issues/4761#issuecomment-2692170972)

The reason is very likely due to `musl` vs `glibc` issue. Alpine only contains musl libc, but the binary in Release artifacts is build with glibc.

In this PR, I worked around this with `gcompat` package, which provides a glibc compatibility layer on Alpine. Alternatively, you could also build musl based binaries, and use that instead. That would require changes to `scala-native` job to build for both musl and glibc first though.

Sorry for the late reply

Dockerfile
```diff
@@ -1,6 +1,23 @@
...
+  && VERSION=${VERSION#"v"} \
+  && echo "Installing scalafmt version ${VERSION} for architecture ${TARGETARCH}" \
+  && if [ "${TARGETARCH}" = "amd64" ]; then ARCH=x86_64; \
+  elif [ "${TARGETARCH}" = "arm64" ]; then ARCH=aarch64; \
+  else \
+  echo "Unsupported architecture: ${TARGETARCH}"; \
+  exit 1; \
+  fi \
+  && echo "downloading https://github.com/scalameta/scalafmt/releases/download/${VERSION}/scalafmt-${ARCH}-pc-linux.zip for ${ARCH}" \
+  && curl -L "https://github.com/scalameta/scalafmt/releases/download/v${VERSION}/scalafmt-${ARCH}-pc-linux.zip" -o /tmp/scalafmt.zip \
```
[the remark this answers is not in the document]
harryzcy: To answer your questions:

> download-artifact vs "curl in Dockerfile"

   Both are getting exactly what was built earlier in the pipeline, the only difference is download-artifact prepares the binary ahead of time before docker build start, whereas "curl in Dockerfile" approach prepares the binary within the docker build steps. This allows anyone to rebuild an docker image locally without downloading first or modifying Dockerfile file. But for the purpose of CI pipeline, they should be identical. I can change if you prefer one over the other.

> Have you validated the resulting Docker image running on a native aarch64 host (or via QEMU) actually behaves identically to the amd64 one?

I've tested building natively on arm64 platform and also running the image, and got results successfully:
```shell
$ docker run scalafmt:latest scalafmt --version
scalafmt 3.10.1
```

> Does this result in a single tagged image supporting both architectures? Has an image inspect been done (e.g., docker manifest inspect …) to confirm?

The `docker/build-push-action` should ensure a single tagged image. `docker manifest inspect` only works after pushed to DockerHub, so not really possible ahead of time.

> Is gcompat appropriate?

It should work the same way in x86-64/amd64 and aarch64/arm64. I seems to execute the executable file fine by getting the `--version`.

kitbellew:
> > @tgodzik is this something you could help review?
> 
> I am not an expert in Docker things, but looks ok to me as much as I can verify. We can revert the change if it breaks anything

ok, let's try it, then.

harryzcy:
> @harryzcy thank you for your contribution and your patience. please remove the branch on your end.

Thank you for the review!

PR #4822, opened 2025-02-15
PlatformFileOps: read/write files asynchronously

kitbellew:
Previously, we simply wrapped a Future around a blocking file I/O both on JVM and Native; while we still keep this approach on Native, let's implement actual asynchronous file I/O on JVM, just like on JS.

Also, modify runners to use this async writing and a different thread pool for that, so that it does not clash with formatting tasks.


kitbellew:
> Just curious... Did you notice any performance improvements? Would be great. It takes a while to check formatting in a large code base...

Not really. The build includes community-test checks, and they take about the same time. But they are not slower, and someone might suggest additional tuning. The main purpose is to use the same design for all three platforms, and scala.js, apparently, is inherently asynchronous.

PR #4823, opened 2025-02-16
bugfix: Don't use aggregate in Scala Native

tgodzik:
This causes the nativeLink task to be executed for all the modules and it doesn't make sense to do that for non app modules.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -229,7 +229,7 @@ lazy val cli = crossProject(JVMPlatform, NativePlatform, JSPlatform)
   // TODO: enable NPM publishing
   .jsSettings(scalaJsSettings, scalaJSUseMainModuleInitializer := true)
   .jvmEnablePlugins(NativeImagePlugin)
-  .jvmConfigure(_.dependsOn(dynamic.jvm).aggregate(dynamic.jvm)).aggregate(core)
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -229,7 +229,7 @@ lazy val cli = crossProject(JVMPlatform, NativePlatform, JSPlatform)
   // TODO: enable NPM publishing
   .jsSettings(scalaJsSettings, scalaJSUseMainModuleInitializer := true)
   .jvmEnablePlugins(NativeImagePlugin)
-  .jvmConfigure(_.dependsOn(dynamic.jvm).aggregate(dynamic.jvm)).aggregate(core)
+  .jvmConfigure(_.dependsOn(dynamic.jvm).aggregate(dynamic.jvm).aggregate(core.jvm))
```
kitbellew, line 232: the purpose of this aggregation was only to be able to run `cliXxx/test` and get tests from dependencies (in reality, only sysops).

so, in this case, we should remove this aggregation for all platforms and either:
- add `coreXxx/test` to `ci-test-xxx` commands
- or move `.aggregate(core)` to the `tests` project.

kitbellew (changes requested):

PR #4826, opened 2025-02-18
Remove FAQ entry about Scalariform

froth:
I found this FAQ entry today and thought "oh crazy, is there a new formatting tool that I don't know yet?". I found out that it is indeed a very old and abandoned tool. I think it is fine to not mention it any longer in the FAQ.  Especially as the links don't lead anywhere meaningful.


froth:
I have moved on :)

Issue #4827, opened 2025-02-19
`scalafmt` 3.9.0 doesn't format code with minimal configuration


Gedochao:
## Required: Configuration ##
```hocon
version = "3.9.0"
runner.dialect = scala213
```
```hocon
version = "3.9.0"
runner.dialect = scala212
```
```hocon
version = "3.9.0"
runner.dialect = scala3
```

## Required: Command-line parameters ##

none

## Steps

Given code like this:
```scala
package foo

object Foo       extends       java.lang.Object  {
  def           get()             = 2
}
```

## Problem

Scalafmt 3.9.0 formats code like this (as in, seemingly not at all):
```scala
package foo

object Foo       extends       java.lang.Object  {
  def           get()             = 2
}

```

## Expectation

I would like the formatted output to look like this (3.8.6 and before did so):
```scala
package foo

object Foo extends java.lang.Object {
  def get() = 2
}
```

## Notes
Even when setting the version to 3.8.6 in the configuration, using the newest `scalafmt-cli` launcher (3.9.0) seems to not format the code.
```hocon
version = "3.8.6"
runner.dialect = scala3
```

Discovered while bumping `scalafmt` in `scala-cli`:
- https://github.com/VirtusLab/scala-cli/pull/3507

kitbellew:
> Discovered while bumping `scalafmt` in `scala-cli`:
> 
> * [Update scalafmt-cli_2.13, scalafmt-core to 3.9.0 VirtusLab/scala-cli#3507](https://github.com/VirtusLab/scala-cli/pull/3507)

what exactly in that pr suggests the formatter behaves that way? all i see is it removes trailing whitespace, something that @tgodzik already reported in #4825.

Gedochao:
> > Discovered while bumping `scalafmt` in `scala-cli`:
> > 
> > * [Update scalafmt-cli_2.13, scalafmt-core to 3.9.0 VirtusLab/scala-cli#3507](https://github.com/VirtusLab/scala-cli/pull/3507)
> 
> what exactly in that pr suggests the formatter behaves that way? all i see is it removes trailing whitespace, something that [@tgodzik](https://github.com/tgodzik) already reported in [#4825](https://github.com/scalameta/scalafmt/discussions/4825).

```
integration.test.jvm 9 tests failed: 
  scala.cli.integration.FmtTests.simple scala.cli.integration.FmtTests.simple
[...]

PR #4834, opened 2025-02-19
Cli: ensure input execution context is initialized

kitbellew:


kitbellew:
while i can't repro what's mentioned in #4827, chatgpt thinks it could happen if main ends before thread pools are initialized. this might solve the problem. cc @gedochao.

tgodzik (approved):
Thanks! I will check it out before the next release

PR #4842, opened 2025-02-21
RunnerSettings: rename from ScalafmtRunner

kitbellew:
ScalafmtRunner is also a base class in cli package, let's disambiguate.

### DIFF
--- a/scalafmt-cli/shared/src/main/scala/org/scalafmt/cli/CliOptions.scala
+++ b/scalafmt-cli/shared/src/main/scala/org/scalafmt/cli/CliOptions.scala
@@ -1,16 +1,9 @@
 package org.scalafmt.cli
 
-import org.scalafmt.config.ConfParsed
-import org.scalafmt.config.ScalafmtConfig
-import org.scalafmt.config.ScalafmtConfigException
-import org.scalafmt.sysops.AbsoluteFile
-import org.scalafmt.sysops.GitOps
-import org.scalafmt.sysops.OsSpecific
-import org.scalafmt.sysops.PlatformFileOps
+import org.scalafmt.config._
+import org.scalafmt.sysops._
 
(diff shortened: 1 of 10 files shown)


SethTisue:
In the Scala 2 community build, this made Scastie start failing

We already have Scastie forked for community build purposes, so it was easy to accommodate the renaming in the fork: https://github.com/scalacommunitybuild/scastie/commit/59d57f0f2641003196c3db1fbc2c263e56d4bba1

but that leaves:

```
[scastie] [error] /Users/tisue/cb3/target-0.9.20/project-builds/scastie-0e82b37f219c1b95ff4b30018bd89c932a116726/sbt-runner/src/main/scala/com.olegych.scastie.sbt/FormatActor.scala:29:9: method withDialect in class RunnerSettings cannot be accessed as a member of org.scalafmt.config.RunnerSettings from object FormatActor in package sbt
[scastie] [error]       }.withDialect(dialect)
[scastie] [error]         ^
[scastie] [error] one error found
```

I don't know the Scastie codebase and it isn't obvious to me what to do about `withDialect` becoming package-private. Is there some other way to access it? How sure are you that it should be private?

tgodzik:
I think most of that is internal classes that shouldn't be used. We have interfaces for that and we normally depend on having an actual scalafmt.conf file.

Rewriting it to work similar to metals would probably be the way to go https://github.com/scalameta/metals/blob/main/metals/src/main/scala/scala/meta/internal/metals/FormattingProvider.scala

kitbellew:
@SethTisue i can restore the method, of course, although i didn't realize anyone was using them.

also, it's a simple `.copy(dialect = ...)`.

kitbellew:
@SethTisue fyi #4849

SethTisue:
cool thanks!

PR #4847, opened 2025-02-24
improvement: Upload zip to limit download sizes

tgodzik:


.github/workflows/release-native.yml
```diff
@@ -16,18 +16,17 @@ permissions:
 
 jobs:
   scala-native:
-    if: github.event_name == 'release'
```
kitbellew, line 19: we need to keep this, this workflow supports a dispatch event as well which only runs dockerize.

.github/workflows/release-native.yml
```diff
@@ -45,17 +44,23 @@ jobs:
       - run: sbt scala-native
         env:
           CI: true
+      - name: Zip artifact for deployment
+        if:  ${{ !startsWith(matrix.deploy.os, 'windows') }}
+        run: zip ${{ matrix.deploy.name }} ${{ env.BINARY_NAME }}
+      - name: Zip artifact for deployment
+        if:  ${{ startsWith(matrix.deploy.os, 'windows') }}
+        run: 7z a ${{ matrix.deploy.name }} ${{ env.BINARY_NAME }}
```
kitbellew, line 52: one rule, perhaps:
```
  - name: Zip ...
    run: ${{
      startsWith(matrix.deploy.os, 'windows') &&
      7z matrix.deploy.name 'scalafmt.exe' ||
      zip matrix.deploy.name 'scalafmt'
    }}
```
and remove definition of `BINARY_NAME` under `env`.

.github/workflows/ci.yml
```diff
@@ -11,111 +11,150 @@ permissions:
   contents: read
 
 jobs:
-  test:
```
kitbellew, line 14: i am assuming that there won't really be any changes in this file, and you are just copying release-native for testing purposes.

kitbellew (changes requested):

.github/workflows/release-native.yml
tgodzik: Sure, I am just testing it now 😅

.github/workflows/release-native.yml
tgodzik: I couldn't make it work like this, but I removed `BINARY_NAME ` let me know what you think

tgodzik: I tried that as well and unless there is something else to change I don't think that will work. It actually reported the previous line without any data.

PR #4848, opened 2025-02-26
Router: extract rules to Splits, use class lookup

kitbellew:
A profiler run revealed that `getSplits` is the largest bottleneck of performance, despite being run only once, almost double that of the part which was previously thought to be the more expensive: BestFirstSearch.

Let's refactor away from using a massive `match` to several lookups by the token class.

In community-tests runs, there's a 35% reduction in running times (22 minutes to 14, 15 to 10).


kitbellew:
this is a tricky change :) i tested on scala2+scala3+spark community repos that community-tests runs on, and it produces the same results, but conceivably could be different. if you could check on your repos, would appreciate it.

tgodzik:
> this is a tricky change :) i tested on scala2+scala3+spark community repos that community-tests runs on, and it produces the same results, but conceivably could be different. if you could check on your repos, would appreciate it.

I checked Metals and Scala CLI repos and the results seem the same, but I only changed version in .scalafmt.conf and run using the jar based runner.

kitbellew:
> LGTM though I don't really have time to look at the full changes.

Yes, alas, this change is not easily reviewable. The code is basically the same but in different places; but even with that, there were a few tricky parts:
- now there's a fixed order of rule evaluation (first we look at the left token, then the right, etc), whereas before matching could be in any order
- the location in the source file (line, to be exact) is used as a tiebreaker when comparing formatting states, and since these lines have changed, I had to add an explicit rank field.

PR #4849, opened 2025-02-26
Config: make `withDialect` a public method

kitbellew:
Also:

- remove local copies of dialects: they are completely identical to the originals and they also convey the impression that their type is NamedDialect when it is just Dialect
- replace NamedDialect with sourcecode.Text[Dialect]: this will allow passing a properly named variable containing just a dialect to a method expecting a "named dialect"

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/Scalafmt.scala
@@ -68,7 +68,7 @@ object Scalafmt {
       if (isSbt) {
         val sbtStyle = style.forSbt
         if (sbtStyle.dialect ne baseStyle.dialect) sbtStyle
-        else sbtStyle.withDialect(NamedDialect(dialects.Sbt))
+        else sbtStyle.withDialect(dialects.Sbt, "sbt")
       } else style
     }
     val styleTry =
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/config/NamedDialect.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/config/NamedDialect.scala
@@ -14,20 +14,16 @@ object NamedDialect {
     apply(name.toLowerCase, pair.value)
   }
 
-  val scala212 = Scala212.withAllowTrailingCommas(true) // SIP-27, 2.12.2
-  val scala213 = Scala213.withAllowTrailingCommas(true)
-  val scala3 = Scala3
-
   private[config] val known = Seq[sourcecode.Text[Dialect]](
     Scala211,
-    scala212,
+    Scala212,
     Scala212Source3,
-    scala213,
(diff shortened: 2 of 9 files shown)


SethTisue:
when compiling Scastie against this PR, I still see:

```
[error] /Users/tisue/cb3/newdir/sbt-runner/src/main/scala/com.olegych.scastie.sbt/FormatActor.scala:29:21: type mismatch;
[error]  found   : scala.meta.Dialect
[error]  required: org.scalafmt.config.NamedDialect
[error]       }.withDialect(dialect)
[error]                     ^
[error] one error found
```

so it seems the status quo has not been restored

I was able to get it to compile with

```diff
-      }.withDialect(dialect)
+      }.withDialect(NamedDialect(NamedDialect.getName(dialect).get, dialect))
```

but this seems infelicitous and may actually be incorrect at runtime for all I know; it's just blind fumbling with code I don't understand the context of

kitbellew:
> when compiling Scastie against this PR, I still see:
> 
> ```
> [error] /Users/tisue/cb3/newdir/sbt-runner/src/main/scala/com.olegych.scastie.sbt/FormatActor.scala:29:21: type mismatch;
> [error]  found   : scala.meta.Dialect
> [error]  required: org.scalafmt.config.NamedDialect
> [error]       }.withDialect(dialect)
> [error]                     ^
> [error] one error found
> ```
> 
> so it seems the status quo has not been restored
> 
> I was able to get it to compile with
> 
> ```diff
> -      }.withDialect(dialect)
> +      }.withDialect(NamedDialect(NamedDialect.getName(dialect).get, dialect))
> ```
> 
> but this seems infelicitous and may actually be incorrect at runtime for all I know; it's just blind fumbling with code I don't understand the context of

generally, the expectation was one wouldn't call ` runnerSettings.withDialect` directly...

instead of
```
      val dialect =
        if (scalaTarget.scalaVersion.startsWith("2.12")) NamedDialect.scala212
        else if (scalaTarget.scalaVersion.startsWith("2.13")) NamedDialect.scala213
        else if (scalaTarget.scalaVersion.startsWith("3")) NamedDialect.scala3
        else NamedDialect.scala213

      val runner = {
        if (isWorksheetMode && scalaTarget.hasWorksheetMode)
          RunnerSettings.sbt
        else
          RunnerSettings.default
      }.withDialect(dialect)

      ScalafmtConfig.default.copy(runner = runner)
```

with the new interface one would use
```
      import scala.meta.dialects

      val dialect: NamedDialect =
        if (scalaTarget.scalaVersion.startsWith("2.12.")) dialects.Scala212
        else if (scalaTarget.scalaVersion.startsWith("3.")) dialects.Scala3
        else dialects.Scala213

      ScalafmtConfig.default.withDialect(dialect)
```

Also, `RunnerSettings.sbt` is the same as `RunnerSettings.default`.

tgodzik (approved):

PR #4850, opened 2025-02-27
improvement: Pin image versions

tgodzik:
Without the change it seems it's not possible to use scalafmt on older ubuntu images due to older gcc version. This should not be an issue normally as users can just update. For windows it's just a precaution since that is also the latest image and windows-2019 seems way too old.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,8 +19,8 @@ jobs:
           - '11'
           - '21'
         os:
-          - windows-latest
-          - ubuntu-latest
+          - windows-2022
+          - ubuntu-22.04
         scala:
           - '2.12.20'
           - '2.13.16'
--- a/.github/workflows/release-native.yml
+++ b/.github/workflows/release-native.yml
@@ -25,9 +25,9 @@ jobs:
         deploy: [
           { os: macOS-13, name: scalafmt-x86_64-apple-darwin.zip },
           { os: macOS-14, name: scalafmt-aarch64-apple-darwin.zip },
-          { os: ubuntu-latest, name: scalafmt-x86_64-pc-linux.zip },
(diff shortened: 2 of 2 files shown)


.github/workflows/release-native.yml
```diff
@@ -25,9 +25,9 @@ jobs:
         deploy: [
           { os: macOS-13, name: scalafmt-x86_64-apple-darwin.zip },
           { os: macOS-14, name: scalafmt-aarch64-apple-darwin.zip },
-          { os: ubuntu-latest, name: scalafmt-x86_64-pc-linux.zip },
+          { os: ubuntu-22.04, name: scalafmt-x86_64-pc-linux.zip },
```
kitbellew, line 28: should we do the same in ci.yml? there's a step there which tests that native can be built and that it works.

tgodzik: To be on a super safe side we can. Should I do it in this PR? I don't think anything should really break if we don't do that

kitbellew: sounds reasonable

.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           - '21'
         os:
           - windows-latest
-          - ubuntu-latest
+          - ubuntu-22.04
```
tgodzik, line 23: Changed the version here, since this is the job that also tests Scala Native

kitbellew: no need to do the same for windows-2022?

kitbellew (approved):

.github/workflows/ci.yml
tgodzik: ach yes, done

PR #4913, opened 2025-05-20
native: upgrade to v0.5.7

kitbellew:


project/Extensions.scala
```diff
@@ -0,0 +1,15 @@
...
+
+object Extensions {
+
+  implicit class ImplicitNativeConfig(private val cfg: NativeConfig)
+      extends AnyVal {
+    def withGC: NativeConfig =
+      if (Properties.isWin) cfg.withGC(GC.commix) else cfg.withGC(GC.immix)
+    def forTest: NativeConfig = withGC.withMode(Mode.releaseFast)
+    def forRelease: NativeConfig = withGC.withMode(Mode.releaseFull)
+      .withLTO(LTO.thin)
```
He-Pin, line 12: The `LTO.thin` seems unstable on mac os.

kitbellew: @He-Pin  do you know if something is known to have a problem on windows? it keeps hanging, so i am trying different options.

He-Pin: I only encounter https://github.com/scala-native/scala-native/issues/4341 this problem

kitbellew:
@WojciechMazur @LeeTibbert I need some help with this. After a few iterations, I have finally got ubuntu to work, but windows tests for native keep hanging. What in v0.5.7 could explain this? What can we do?

LeeTibbert:
I'll watch it for a few minutes longer, but on the 2.13 Windows test it appears that the last 
test started is "D:\a\scalafmt\scalafmt\scalafmt-tests\shared\src\test\scala\org\scalafmt\FormatTests:159 Starting: scala3\OptionalBraces_unfold.stat:8762: grouped context bounds: before = always, within = always|
".

So far there is no corresponding "Finished" message, as there are with other tests. (side note: The starting/finished
log messages make it a _lot_ easier for a person dropped into the fray to follow the log. I like that, thank you".


I'm at the end of this session. I'll take another look at the other Windows log files next session and see if they
hang at the same  place or near it.

LeeTibbert:
> MUnitRunNotifier: use concurrent for Native munit#946

A shot in the dark until we get this resolved.  What happens if you build with a scratch munit
which does not have that change?

There have been a number of changes to `java.lang.Process*` in SN 0.5.7 and 0.5.8-SNAPSHOT.
There are also some unresolved SN Issues. In particular #4264 mentions a deadlock in scalafmt.
This problem is in the same neighborhood.

kitbellew:
> > MUnitRunNotifier: use concurrent for Native munit#946
> 
> A shot in the dark until we get this resolved. What happens if you build with a scratch munit which does not have that change, SN only and possibly Windows only?

Nothing positive. What made you think about it? It doesn't make munit multi-threaded, it simply makes it safer with threads.

LeeTibbert:
> Nothing positive. What made you think about it? It doesn't make munit multi-threaded, it simply makes it safer with threads.

Rats!  I was looking at a quick fix. First my Kentucky Derby horse does not come in, then this. 

IIRC, there were some changes in SN 0.5.7 around Windows process shutdown.  I had been hoping 
to not provoke any possible "features" there.  I believe some things got fixed for 0.5.8-SNAPSHOT.

LeeTibbert:
If I clone the kitbellew:4913 branch, any special tricks for building on macOS (or Ubuntu)?

I think building scalafmt on one of those two would be preliminary with my trying to set
up a Windows environment for both of SN & scalafmt.

kitbellew:
> There have been a number of changes to `java.lang.Process*` in SN 0.5.7 and 0.5.8-SNAPSHOT. There are also some unresolved SN Issues. In particular #4264 mentions a deadlock in scalafmt. This problem is in the same neighborhood.

are you talking about a differently-numbered issue, in another repo? scala-native/scala-native#4265?

kitbellew:
@LeeTibbert i updated native dependency to 0.5.8-SNAPSHOT and modified munit-native (also snapshot) to log errors when a suite cannot be instantiated, and here's an error:

```
==> X org.scalafmt.FidelityTest java.io.UncheckedIOException: Cannot open file D:\a\scalafmt\scalafmt\.\scalafmt-config\native\target\scala-2.13\native-test\dependencies\javalib_native0.5_2.13-0.5.8-SNAPSHOT-4\java\util\concurrent\CompletionStage$$anonfun$java$util$concurrent$CompletionStage$$$nestedInanonfun$exceptionallyComposeAsync$1$1.nir - The system cannot find the path specified. (3)
```

not sure what's happening here.

kitbellew:
@LeeTibbert @WojciechMazur I seem to have found the culprit; the hang happens within GitOpsTest which executes a [`sys.process.Process`](https://github.com/scalameta/scalafmt/blob/e87eceb3986513302f49cf7d46ba8309bbdf3b3d/scalafmt-sysops/jvm-native/src/main/scala/org/scalafmt/sysops/PlatformRunOps.scala#L47). Anything you know about running processes under Windows which has changed since v0.5.6?

kitbellew:
@tgodzik just fyi: scala-native v0.5.7 seems to have a [bug with `sys.process.Process` under Windows](https://github.com/scalameta/scalafmt/pull/4913#issuecomment-2932538412), we can't upgrade anything to that version in the dependency tree of this repo (or to what has already been upgraded) until it's fixed.

He-Pin:
Scala native 0.5.8 is released

LeeTibbert:
kitbellew

Congratulations on getting scalafmt working with SN 0.5.9 pre-release.
It certainly has been a long haul.

WojciechMazur:
The stable 0.5.9 should be available next week or maybe during the weekend. It's release would be coordinated with Scala 3.8.0-RC1 cutoff (it might be published before 3.8.0-RC1 would be published)

PR #4917, opened 2025-05-30
bugfix: Update sbt plugin to correctly publish to maven central

tgodzik:


build.sbt
```diff
@@ -43,6 +43,12 @@ inThisBuild {
     testFrameworks += new TestFramework("munit.Framework"),
     // causes native image issues
     dependencyOverrides += "org.jline" % "jline" % "3.29.0",
+    publishTo := {
```
gaeljw, line 46: I think you don't need to set this when using sbt-ci-release, the plugin sets it by default.

project/plugins.sbt
```diff
@@ -1,16 +1,13 @@
 // scalafmt: { maxColumn = 100, align.preset = more, align.allowOverflow = true }
 
-resolvers ++= Resolver.sonatypeOssRepos("releases")
-resolvers ++= Resolver.sonatypeOssRepos("snapshots")
-
```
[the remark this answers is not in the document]
tgodzik: I don't think this was ever needed since we don't use snapshot versions of any of the plugins. We can add maven-snapshots to have the same results, but it doesn't add anything currently.

For releases it's basically the same as maven-central

build.sbt
tgodzik: Thanks! Removed

PR #4962, opened 2025-09-04
CoursierDependencyDownloader: support credentials

kitbellew:
This requires using the full coursier library as it loads credentials from COURSIER_CREDENTIALS env var, just like any default repositories have been loaded from COURSIER_REPOSITORIES.

### DIFF
--- a/scalafmt-dynamic/jvm/src/main/scala/org/scalafmt/dynamic/CoursierDependencyDownloader.scala
+++ b/scalafmt-dynamic/jvm/src/main/scala/org/scalafmt/dynamic/CoursierDependencyDownloader.scala
@@ -1,14 +1,12 @@
 package org.scalafmt.dynamic
 
-import org.scalafmt.CompatCollections.JavaConverters._
-
 import java.io.OutputStreamWriter
 import java.net.URI
 import java.net.URL
 
 import scala.util.Try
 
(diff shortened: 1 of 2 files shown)


scalafmt-dynamic/jvm/src/main/scala/org/scalafmt/dynamic/CoursierDependencyDownloader.scala
```diff
@@ -1,30 +1,25 @@
...
-    val coursierDependencies = dependencies
-      .map(x => CoursierDependency.of(x.group, x.artifact, x.version))
-    // TODO: ttl is unavailable, see https://github.com/coursier/interface/issues/57
-    val cache = Cache.create()
-      .withLogger(Logger.progressBars(downloadProgressWriter))
-    val settings = Fetch.create().withCache(cache)
-      .addRepositories(repositories: _*)
-      .withDependencies(coursierDependencies: _*)
-    settings.fetch().asScala.map(_.toURI.toURL).toList
+    val fileCache = cache.FileCache() // this ctor preserves COURSIER_CREDENTIALS
```
tgodzik, line 17: Och, I wasn't aware of this being an option.

tgodzik (approved):
Looks good! Hopefully this helps or the other fallback I did

PR #4980, opened 2025-09-26
FileOps: implement walkFiles with granular control

kitbellew:
Unfortunately, Files.find has no way to handle exceptions during tree traversal, including those it generates itself (such as when reading the file attributes).

Instead, let's add ability to use Files.walk, with a more fine-grained visitor handler.


kitbellew:
@tgodzik this came about I tried scala-native SNAPSHOT build, which seems to have fixed the problem with Windows processes (thanks @LeeTibbert for the help!) but now revealed an issue with some unreadable build file which is traversed by `Files.find` inside `FidelityTest` while looking for `.scala` files.

tgodzik (approved):
I had a similar issue in other projects, where one race could just break everything. LGTM

PR #5066, opened 2025-12-09
FormatWriter: insert/remove end markers using gaps

### LINKED ISSUE #5064: Feature request: implement "Odersky" rule for end markers
There's some uncertainty in Scala3 regarding "end markers" – where to add them and where not to.

ScalaFmt already provides some options that try to address that (`rewrite.scala3.insertEndMarkerMinLines`/`removeEndMarkerMaxLines`/`countEndMarkerLines`),
But those are quite tricky to get right, because imply tweaking some "magic numbers".

On the other hand, Dr.Ordersky's personal recommendation is the following:
> Use end markers sparingly. My rule is that an end marker is appropriate if the section of code it closes contains blank lines.

Which looks rather straightforward. Moreover, at first glance it shouldn't require any "magic numbers" and could be simply switched on or off in the config. Therefore it could come in handy if ScalaFmt implemented this rule.

Feedback from Dr.Odersky himself:
> It would be great if this rule could be implemented in ScalaFmt. Usage of end markers was the one thing where I was initially unsure what to recommend. Over time, the rule “use an end marker if it closes a scope with blank lines” emerged and it now feels quite natural.

[link](https://contributors.scala-lang.org/t/should-we-gather-usage-statistics-about-sip-44-fewer-braces/7270/108)
<sub>⚠️ WARNING: the thread turned into a flame war about coding styles, but some insights still seem reasonable in there.</sub>

kitbellew:
Instead of counting lines, we could instead count the number of blank gaps within a section. Fixes #5064.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4312,6 +4312,22 @@ is used to calculate the span for the purposes of applying
 
 Prior to v3.10.3, this setting was named `scala3.countEndMarkerLines`.
 
+### `rewrite.scala3.endMarker.spanIs`
+
+```scala mdoc:defaults
+rewrite.scala3.endMarker.spanIs
+```
+
+> Since v3.10.3.
+
+This parameter defines what "span" refers to when applying
+[`insertMinSpan`](#rewritescala3endmarkerinsertminspan) and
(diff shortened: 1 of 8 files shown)


satorg:
Thank you @kitbellew ! Looking forward to give it a try!

PR #5070, opened 2025-12-19
SBT: handle more metadata in assemblyMergeStrategy

jackkoenig:
I tried to run `sbt "cli/assembly"` and it didn't work for me, these rules fix it.

The basic justification for the rules:
* `module-info.class` - discard - we don't need Java Module descriptors in a fat jar
* `javax.inject.Named` - concat - various plexus libraries need their metadata

Applied the rules to both `cli` and `dynamic`. I don't think it's really needed on `dynamic` but SBT seems to run assembly on it even when runing `cli / assembly` so it seems prudent to get rid of that error (better would be not running assembly on it).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,7 @@ jobs:
         run: git config --global user.email "scalafmt@scalameta.org" && git config --global user.name "scalafmt"
       - run: sbt ++${{ matrix.scala }} test-jvm
         if: ${{ (success() || failure()) && steps.gitconfig.outcome == 'success' }}
+        id: testjvm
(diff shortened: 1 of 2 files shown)


jackkoenig:
I don't know if JVM matters but I tried with both GraalVM 17.0.7 and GraalVM 25.0.0 and hit the same merging issues on both.

kitbellew:
would you like to add an `sbt assembly` step in one of these pre-commit checks, to validate that it works?

jackkoenig:
I also added a sanity check of the assembled jar as my first efforts in updating the `assemblyMergeStrategy` involved `MergeStrategy.first` for `javax.inject.Named` which would crash when run.

PR #5282, opened 2026-05-13
BestFirstSearch: recurse on Term.Block siblings in wide arg clauses

gmethvin:
Fixes a `SearchStateExploded` failure on a real-world pattern: a sequence of multi-statement `Term.Block` values inside one arg clause, e.g.

```scala
val testCases = Seq(
  { val nodeData = SomeTrigger(...); TestCase(nodeData = nodeData, ...) },
  { val nodeData = OtherTrigger(...); TestCase(nodeData = nodeData, ...) },
  // ...
)
```

This pattern arises in table-driven tests where each row needs a local `val` to bind a value used twice (typically as both input and expected output). With ≥15 such siblings, BFS exhausts its visit budget and aborts with `SearchStateExploded`.

`BestFirstSearch.getBlockCloseToRecurse` only recurses on blocks wider than `maxColumn * 3`. Each block falls under that threshold, so BFS explores the cross-product of their layouts instead of solving each subproblem once and memoizing.

Update `getBlockCloseToRecurse` to also recurse when the block is one of multiple `Term.Block` siblings inside a wide-enough enclosing `Term.ArgClause`. The "wide enclosing arg clause" guard (`tokens.span(ac) > maxColumn * 3`) keeps compact cases like `f(x, { a; b }, y, { c; d })` unaffected. Existing memoization handles the rest. Sibling blocks share `(indentation, column, depth)` keys, so the recursive subproblem is solved once and reused.

### DIFF
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
@@ -39,9 +39,25 @@ private class BestFirstSearch private (range: Set[Range])(implicit
   private def getBlockCloseToRecurse(ft: FT)(implicit
       style: ScalafmtConfig,
   ): Option[Int] = TokenOps.getEndOfBlock(ft, parens = true).collect {
-    // Block must span at least 3 lines to be worth recursing.
-    case (close, _) if tokens.width(ft, close) > style.maxColumn * 3 =>
-      close.idx
+    // Block must span at least 3 lines to be worth recursing, except for
+    // narrow blocks that are one of several Term.Block siblings inside a
+    // wide arg clause, which would combinatorially explode BFS without
+    // recursion+memoization (e.g., `Seq({ s1; s2 }, { s1; s2 }, ...)` × N).
+    case (close, _)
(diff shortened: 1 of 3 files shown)


kitbellew:
thank for this idea. however, first things first: the test example you've provided does not blow up so it doesn't prove your solution fixes the problem you've described.

in general, we require changes to consist of at least two commits: one adds a test and shows what happens to that test, and the next commit (or commits) modify behaviour and show how that test improves. so i did this with your submission, and the test by itself succeeds.

can we start by figuring out a test which does in fact result in exploded state?

gmethvin:
Sorry, I over-simplified the earlier fixture. I've force-pushed two commits in the order you described: the first adds a test that consistently exhausts the default `runner.maxStateVisits` budget (asserts `stateVisits = 150001`), and the second adds the fix, after which the same test formats cleanly.

kitbellew:
@gmethvin thank you for researching this problem. your solution might require a bit more thought; the problem is that `shortestPathMemo` assumes that the block itself is wide enough and in case of `isOpt = true` only does a lookup, whereas in your case it might be insufficient (it needs to actually compute the path and cache it).

however, it inspired a slightly different approach (see #5283), which seems to handle your example.

gmethvin:
Great, thanks for catching that. I'll close this then.

PR #5283, opened 2026-05-16
BestFirstSearch: recurse on some longer blocks

kitbellew:
Supersedes #5282.


scalafmt-core/shared/src/main/scala/org/scalafmt/internal/BestFirstSearch.scala
```diff
@@ -40,8 +40,11 @@ private class BestFirstSearch private (range: Set[Range])(implicit
       style: ScalafmtConfig,
   ): Option[Int] = TokenOps.getEndOfBlock(ft, parens = true).collect {
     // Block must span at least 3 lines to be worth recursing.
```
gmethvin, line 42: This comment now only applies to the `case _` branch below

gmethvin (approved):

gmethvin:
Thanks for fixing this!

PR #5347, opened 2026-07-17
Fix non-idempotency of rewrite.scala3.preset = "common" for nested objects with blank lines

### LINKED ISSUE #5346: rewrite.scala3.preset = "common" is non-idempotent for nested optional-braces objects with a blank line
## Summary

With `rewrite.scala3.preset = "common"`, scalafmt oscillates (period-2) between two outputs when an optional-braces template contains both a plain member (`val`/`def`) **and** a nested optional-braces template separated by a blank line. Neither state is a fixed point, so `scalafmt --test` can never pass, breaking CI.

## Version

scalafmt 3.11.3

## Minimal `.scalafmt.conf`

```hocon
version = 3.11.3
runner.dialect = scala3
rewrite.scala3.preset = "common"
```

## Minimal input (`Min.scala`)

```scala
object Outer:
  val a = 1

  object Inner:
    val b = 2
```

## Steps to reproduce

```
scalafmt --config .scalafmt.conf Min.scala   # pass 1 → brace/colon
scalafmt --config .scalafmt.conf Min.scala   # pass 2 → back to colon/colon
```

Pass 1 result (outer → braces, inner unchanged):

```scala
object Outer {
  val a = 1

  object Inner:
    val b = 2
}
```

Pass 2 result — **identical to the original input**:

```scala
object Outer:
  val a = 1

  object Inner:
    val b = 2
```

The formatter cycles: `colon/colon → brace/colon → colon/colon → …`

## Trigger conditions

- Outer template uses optional-braces (colon) syntax
- Contains at least one plain member (`val`/`def`) **and** a nested template
- A **blank line** between the member and the nested template

Removing the blank line produces stable output and `--test` passes.

## Root cause analysis

The `common` preset activates two rules simultaneously:

- `convertToNewSyntax = true` — converts brace syntax to colon syntax
- `optionalBraces.insert` with `blankGaps = Between(min = 1)` — inserts braces when the body has ≥1 blank-line gap

These rules run in a single pass. In pass 1, `Outer` is in colon form: `optionalBraces.insert` sees the blank gap and inserts braces. In pass 2, `Outer` is in brace form: `convertToNewSyntax` marks the braces for removal, and `optionalBraces.insert` fails to fire (the `getBlankGaps` counter uses the `{` token as its snapshot reference, and the count comes out differently), so the braces are removed and the input is restored.

The fix should ensure that when `convertToNewSyntax` proposes removing braces from a block that satisfies the `blankGaps` insert condition, the insert rule wins.

## Expected

Formatting is idempotent: `format(format(x)) == format(x)`, and a file that has been formatted passes `--test`.

## Actual

Period-2 oscillation; `--test` never succeeds on either state.

custommonkey:
## Summary

Fixes #5346.

With `rewrite.scala3.preset = "common"`, the formatter oscillated between two states for nested objects separated by a blank line:

- **State A** (colon/colon) → formatter inserts braces on the outer object (blank-gap condition met) → **State B** (brace/colon)
- **State B** (brace/colon) → formatter removes braces from the outer object → **State A** again

Neither state passed `--test`.

### Root cause

`onToken` creates a `{`→`:` Replace for `Template.Body` blocks (brace removal). Because `replaceToken(":")` stores a copy of the FT with `right = T.Colon`, the `onRight` dispatch sends it to `onRightFromBraces`. The colon case in `notOkToRewrite` there only called `shouldRewriteColonOnRight`, which returns `true` for `Template.Body` — unconditionally allowing removal without checking whether the `optionalBraces.insert` blank-gap condition required keeping the braces.

### Fix

Added `|| !skipRightToBraces(left)` to the colon case in `notOkToRewrite` inside `onRightFromBraces`:

```scala
// before
case _: T.Colon => !shouldRewriteColonOnRight(left)

// after
case _: T.Colon =>
  !shouldRewriteColonOnRight(left) || !skipRightToBraces(left)
```

When the insert condition (e.g. blank gaps ≥ 1) is satisfied, `skipRightToBraces` returns `false`, making `notOkToRewrite = true` and preserving the braces. When the condition is not satisfied, `skipRightToBraces` returns `true` and removal proceeds as before.

## Test plan

- [x] Added regression test in `scala3/Issues.stat`: colon/colon input → brace/colon output (only outer object gets braces; inner has no blank-line gap)
- [x] The test framework's built-in idempotency check confirms brace/colon → brace/colon (stable state)
- [x] Full test suite passes (`16228` tests, 0 failures)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scalafmt-tests/shared/src/test/resources/scala3/OptionalBraces.stat
+++ b/scalafmt-tests/shared/src/test/resources/scala3/OptionalBraces.stat
@@ -10213,3 +10213,18 @@ then
    report.errorAndAbort(
      s"Failed to derive a ZLayer: type `${tpeSymbol.fullName}` is not a concrete class."
    )
+<<< preset=common, nested object with blank line is non-idempotent
+rewrite.scala3.preset = "common"
+===
+object Outer:
+  val a = 1
+
+  object Inner:
+    val b = 2
+>>>
+object Outer {
+  val a = 1
+
+  object Inner:
+     val b = 2
+}
--- a/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RemoveScala3OptionalBraces.scala
+++ b/scalafmt-core/shared/src/main/scala/org/scalafmt/rewrite/RemoveScala3OptionalBraces.scala
@@ -185,7 +185,8 @@ private class RemoveScala3OptionalBraces(implicit val ftoks: FormatTokens)
         case _ => false
       }) ||
       (left.ft.right match {
-        case _: T.Colon => !shouldRewriteColonOnRight(left)
+        case _: T.Colon =>
+          !shouldRewriteColonOnRight(left) || !skipRightToBraces(left)
         case _ => false
       })
     ft.right match {
--- a/scalafmt-tests/shared/src/test/resources/scala3/Issues.stat
+++ b/scalafmt-tests/shared/src/test/resources/scala3/Issues.stat
@@ -117,3 +117,18 @@ lazy val versions = new {
 lazy val versions = new {
   val cats = "2.10.0"
 }
+<<< preset=common, nested object with blank line is non-idempotent
+rewrite.scala3.preset = "common"
+===
+object Outer:
+  val a = 1
+
+  object Inner:
+    val b = 2
+>>>
+object Outer {
+  val a = 1
+
+  object Inner:
+    val b = 2
+}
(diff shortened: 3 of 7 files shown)


kitbellew:
@custommonkey thanks for this submission. please ask claude to restructure it this way:
* commit 1: add a copy of the test into each of OptionalBraces*.stat files showing the non-idempotence error
* commit 2: fix the problem and thus adjust the test files showing how the fix improved the result.

kitbellew (changes requested):
see previous comment

scalafmt-tests/shared/src/test/resources/scala3/OptionalBraces.stat
```diff
@@ -10213,3 +10213,18 @@ then
...
+<<< preset=common, nested object with blank line is non-idempotent
+rewrite.scala3.preset = "common"
+===
+object Outer:
+  val a = 1
+
+  object Inner:
+    val b = 2
+>>>
+object Outer {
```
kitbellew, line 10225: i am reviewing this in the first commit. there's no change to the test in the second. does that mean this fix didn't change the test and is thus not needed?

kitbellew (changes requested):
if i check out this tree and use just the first commit (with tests), will the tests succeed? (meaning, asserts on actual vs expected output). if not, then this first commit needs to be revised. if yes, then this entire PR is not needed.

PR #5431, opened 2026-09-20
Fix Windows native CI hangs and support CI runs in forks

alexandrosmagos:
Windows native tests were repeatedly hanging when run through Bash. Switching those jobs to PowerShell lets the full test suite complete on Scala 2.12, 2.13, and 3. The existing retry remains in place, and no tests are skipped.

Running CI in a fresh fork also exposed a separate issue: without upstream release tags, sbt derives version `0.0.0`, which breaks dynamic formatter tests and binary-compatibility checks. This change fetches the upstream tags in fork CI and limits Maven and documentation publishing to the upstream repository.

All five [PR CI jobs](https://github.com/alexandrosmagos/scalafmt/actions/runs/35505581284) passed. All three [Windows native variants](https://github.com/alexandrosmagos/scalafmt/actions/runs/35505602266) also passed on their first attempt, each running 17,498 formatter tests, 22 sysops tests, and 65 CLI tests.


.github/workflows/ci.yml
```diff
@@ -44,6 +44,10 @@ jobs:
         uses: actions/checkout@v7
         with:
           fetch-depth: 0
+      - &upstreamtags
```
kitbellew, line 47: sync tags to your fork, and this won't be necessary. what is the use case for running these in GA from the fork?

.github/workflows/ci.yml
```diff
@@ -85,15 +89,20 @@ jobs:
...
       - *checkout
+      - *upstreamtags
       - *jdk
       - *sbt
       - *gitcfg
       # native run is occasionally flaky (SIGSEGV, scala-native#4917); retry
-      # once so a single crash self-heals instead of red-X'ing main. shell:
-      # bash (not the Windows default pwsh) so `||` behaves on every runner.
-      - if: ${{ !cancelled() && steps.env.outcome == 'success' }}
+      # once so a single crash self-heals instead of red-X'ing main.
```
kitbellew, line 97: this is an unrelated change. your llm states that build hangs in bash and decides to move to powershell; why will that solve the problem?

kitbellew (changes requested):
judging by the pr description and by the level of root-cause analysis depth, you have not followed the instructions published in AGENTS/CONTRIBUTING.md.

.github/workflows/ci.yml
alexandrosmagos: I run the Actions in my fork to reproduce the failures and test changes before opening this PR.
The fork lacked release tags, so sbt derived version 0.0.0 and version-dependent checks failed. Syncing the tags resolved those failures.

You’re right though that this does not require an upstream workflow change. I’ll remove the automatic tag fetching.
The publishing restrictions skipped Maven and documentation publishing jobs in forks, not tests. I added them because those jobs also failed during fork validation.

.github/workflows/ci.yml
alexandrosmagos: The reason for trying PowerShell was to compare how the Windows jobs launch sbt.
The Bash runs stalled after starting native formatter tests. With PowerShell, the complete native suites passed for Scala 2.12, 2.13, and 3.
However, changing shells also changes the sbt launcher path. These results do not establish which difference prevented the hang, but it worked. Dont know the cause

kitbellew: there are two more likely reasons. one is sbt2; it can hang, and it's being addressed. two is native; the reason for running it twice is because it tends to fail... but used to hang as well. shell vs powershell doesn't seem likely at all.

